PR #11, opened 2017-07-12
Remove extra angle-bracket

stuartleeks:


cgillum:
Thanks for the contribution! If you have [DocFX](http://dotnet.github.io/docfx/) installed and run `build.cmd` from the `docfx` directory on your Windows machine, then that will also update the HTML pages which actually show up on https://azure.github.io/azure-functions-durable-extension.

Alternately, you could manually make the same change to the HTML output /docs/articles/topics/checkpointing-and-replay.html.

stuartleeks:
Just mirrored the change in the HTML file. I'm assuming that this additional step is just until the docs are moved over to docs.microsoft.com? :-)

cgillum:
Thanks Stuart. Yeah, it's a little weird, but just until we move to docs.microsoft.com. :)

Issue #13, opened 2017-07-14
Full support for activity functions in languages other than C#


cgillum:
Activity functions should be able to be written in languages other than C# since they don't have a hard dependency on any object model. We close to having this today, but some work is required in the binding extensibility provided by the WebJobs SDK.

The required WebJobs SDK issue can be found here: https://github.com/Azure/azure-webjobs-sdk/issues/1104 

With this fix, it should be possible to implement activity functions in all of our supported languages. Orchestrator functions will still need to be in C#.

cgillum:
This is a breaking change because it fundamentally changes how return values are processed by bindings in Azure Functions.

cgillum:
This is partially fixed in PR https://github.com/Azure/azure-functions-durable-extension/pull/18. However, there are still additional fixes required in the Azure Functions script host to fully enable cross-language support: https://github.com/Azure/azure-webjobs-sdk-script/issues/1422

cgillum:
Katy, this is basically the work that you've already started regarding return values in JavaScript.

nevercast:
Is this blocked by https://github.com/Azure/azure-webjobs-sdk-script/issues/1921 ?

cgillum:
Yes, although I believe this blocking issue has already been fixed. @kashimiz can you confirm?

nevercast:
If that's the case, there could be other things from Sprint 14 also not marked as closed. https://github.com/Azure/azure-webjobs-sdk-script/milestone/38

Will subscribe for updates,
Thanks.

nevercast:
With https://github.com/Azure/azure-webjobs-sdk-script/pull/2049 and https://github.com/Azure/azure-webjobs-sdk/pull/1424 merged. Is this issue in scope for the near future?

cgillum:
Yes, those were the main blockers. We should be able to handle the rest of the work in this repo.

The next step is for us to come up with a protocol which can translate the DTFx orchestration inputs/outputs to/from JSON. This will enable a cross-language story for writing orchestrations.

After that, we need to create the JavaScript-specific programming model for writing orchestrations which implements the JSON protocol. @christopheranderson has started some investigative work into this.

kashimiz:
I'm actively working on this.

cgillum:
And it's done! See here for details: https://github.com/Azure/azure-functions-durable-js

Issue #4, opened 2017-07-03
Activity Binding: Support argument types other than DurableActivityContext


cgillum:
The following argument types should be supported for activity-trigger functions:
* **DurableActivityContext**
* **JObject**
* **Serializable user types**
* **Core primitives** (string, int, DateTime, double, etc.)

Today only `DurableActivityContext` is supported, which is inelegant, especially if the activity function does not need any metadata, like the activity ID.

Most C# developers will want to use core primitive types or POCO types.

JObject binding support will be required for non-.NET languages, like node.js.

PR #18, opened 2017-07-17
Activity binding to use WebJobs return value support

cgillum:
This PR addresses the following issues:

- https://github.com/Azure/azure-functions-durable-extension/issues/4 (binding support for activities) 
- https://github.com/Azure/azure-functions-durable-extension/issues/13 (multi-language support for activities)

It relies on a change in the core WebJobs SDK to support [return values for trigger bindings](https://github.com/Azure/azure-webjobs-sdk/pull/1224).

With this change, activity functions can bind to not only `DurableActivityContext` (the default) but also any JSON-serializeable type, including system types and user types. This also enables activity functions to be implemented in non-.NET languages supported by Azure Functions, like JavaScript.

There is also a breaking change in this PR, which is that `DurableActivityContext.SetOutput` and `DurableOrchestrationContext.SetOutput` are being changed from `public` to `internal`. Functions should use return values instead, now that they are supported by WebJobs.


cgillum:
Before this change can be merged, we need to update to a newer version of the WebJobs SDK NuGet package which contains the dependent changes.

Issue #15, opened 2017-07-15
Sub-Orchestrations


cgillum:
## Summary
This feature allows orchestrator functions to start and wait on sub-orchestrations as part of their logic. Any value returned by the sub-orchestration would be returned to the parent orchestration.

## Motivation
This is useful when you have a library of orchestrations and want to write an orchestration around them.

## Design Notes
The existing `DurableOrchestrationContext.CallFunctionAsync` API would be updated to allow the current orchestrator function to start and wait on other orchestrations, similar to how it calls activity functions.

This is supported by the Durable Task Framework today.  More information can be found [here](https://github.com/Azure/durabletask/wiki/Feature---Sub-orchestrations-%28aka-nested-orchestrations%29).

## Drawbacks
This design does not allow communication between the main orchestration and sub-orchestrations except via inputs and outputs. In that sense, a sub-orchestration is "embedded" into the main orchestration. Actor-style development, on the other hand, would require an ability to start a child-orchestration asynchronously, get back the instance ID, and send messages to it. A separate feature should be designed for this if desired.

mmisztal1980:
+1 I've just encountered this, while implementing a 2-tier fan-out. Having the ability to call orchestration functions from orchestration functions would definitely be cool.

Is there an alternative available at the moment?

cgillum:
There are ways to make this work, but some effort would be involved. Specifically, you could do the following:

1. Create an activity function which starts new orchestrations using `DurableOchestrationClient.StartNewAsync` and returns the instance ID.
2. Create another activity function which queries the status of existing orchestrations using `DurableOrchestrationClient.GetStatusAsync(instanceId)`
3. Have the parent orchestration trigger the child-orchestration via (1) and then wait for it to finish by calling (2) in a loop. This loop code would need to handle null responses and interpret them as "Pending".

So not a great experience right now.

mmisztal1980:
I've encountered a problem while trying to implement 1). I'm getting an exception stating that my sub-orchestration func is expected to have an `[ActivityTrigger]`:

```
System.ArgumentException: 'No activity function named 'os-substations-data-import-orchestrator-func' with version '' is registered.'
```

cgillum:
This is expected. `CallFunctionAsync` can only be used to call activity functions - i.e. those with `[ActivityTrigger]`. Once we implement the sub-orchestration work, you'll be able to call one orchestration from another directly.

[...]

PR #34, opened 2017-08-23
Suborchestration

tohling:
Addressing issue https://github.com/Azure/azure-functions-durable-extension/issues/15

1. Added functionality for orchestrator functions to call other orchestrator functions.
2. Added end-to-end tests for error cases and happy path.


src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -234,9 +234,19 @@ public Task CallFunctionAsync(string functionName, params object[] parameters)
 
             // TODO: Support for versioning
             string version = DefaultVersion;
-            this.config.AssertActivityExists(functionName, version);
+            FunctionType functionType = this.config.GetFunctionType(functionName, version);
```
cgillum: >FunctionType [](start = 12, length = 12)

Rather than introducing a new enum, I would suggest exposing methods like TryGetActivityFunction and TryGetOrchestratorFunction.  It's slightly more performant and introduces fewer new concepts.

test/DurableTaskEndToEndTests.cs
```diff
@@ -343,5 +346,94 @@ public DurableTaskEndToEndTests(ITestOutputHelper output)
...
+            {
+                await host.StartAsync();
+
+                var startArgs = new StartOrchestrationArgs
+                {
+                    FunctionName = nameof(TestOrchestrations.SayHelloInline),
+                    Input = input
+                };
+
+                var client = await host.StartFunctionAsync(nameof(TestOrchestrations.CallActivity), startArgs, this.output);
```
cgillum: >CallActivity [](start = 85, length = 12)

Let's add a comment which shows the call chain. Something like CallActivity --> CallActivity --> SayHelloInline.

test/DurableTaskEndToEndTests.cs
```diff
@@ -343,5 +346,94 @@ public DurableTaskEndToEndTests(ITestOutputHelper output)
...
+            var input = new {Foo = greetingName};
+            var inputJson = JsonConvert.SerializeObject(input);
+
+            using (JobHost host = TestHelpers.GetJobHost(nameof(Orchestration_OnValidOrchestrator)))
+            {
+                await host.StartAsync();
+
+                var startArgs = new StartOrchestrationArgs
+                {
+                    FunctionName = nameof(TestOrchestrations.SayHelloInline),
```
cgillum: >SayHelloInline [](start = 61, length = 14)

Let's use SayHelloWithActivity instead. That's just slightly more interesting/realistic.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -234,9 +235,28 @@ public Task CallFunctionAsync(string functionName, params object[] parameters)
 
             // TODO: Support for versioning
             string version = DefaultVersion;
-            this.config.AssertActivityExists(functionName, version);
+            
+            Task<TResult> callTask = null;
+            string activityFunctions = null;
```
cgillum: > = null [](start = 36, length = 7)

nit: don't initialize to null unless the compiler requires it. By not initializing them, we allow the compiler's flow analysis to validate that these local variables are assigned the way we expect them to be.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -365,16 +365,20 @@ internal void AssertOrchestratorExists(string name, string version)
             }
         }
 
-        internal void AssertActivityExists(string name, string version)
+        internal bool TryGetActivityFunction(string name, string version, out string activityFunctions)
```
cgillum: >TryGetActivityFunction [](start = 22, length = 22)

I know I recommended you use this pattern, but as I look at it implemented I'm realizing that it's problematic because it breaks the convention of existing TryGet* methods which take a key and return a value. This signature is especially strange because you try to get a function and it returns a string list of all functions.

I feel really silly now...but I'm realizing that your original design was more clean and simple, even though it introduced a new type. Really sorry for the back and forth on this, but could we go back to that one (or a variation of it)?

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -234,9 +234,19 @@ public Task CallFunctionAsync(string functionName, params object[] parameters)
 
             // TODO: Support for versioning
             string version = DefaultVersion;
-            this.config.AssertActivityExists(functionName, version);
+            FunctionType functionType = this.config.GetFunctionType(functionName, version);
 
-            Task<TResult> callTask = this.innerContext.ScheduleTask<TResult>(functionName, version, parameters);
+            Task<TResult> callTask = null;
```
cgillum: >null [](start = 37, length = 4)

Remove the null initializer and add a default case to the switch statement which throws if we don't recognize the function type.

src/WebJobs.Extensions.DurableTask/FunctionType.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// The type of a function.
+    /// </summary>
+    public enum FunctionType
```
cgillum: >public [](start = 4, length = 6)

Can this be made internal?

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -234,9 +234,22 @@ public Task CallFunctionAsync(string functionName, params object[] parameters)
...
+            switch (functionType)
+            {
+                case FunctionType.Activity:
+                    callTask = this.innerContext.ScheduleTask<TResult>(functionName, version, parameters);
+                    break;
+                case FunctionType.Orchestrator:
+                    callTask = this.innerContext.CreateSubOrchestrationInstance<TResult>(functionName, version, JsonConvert.SerializeObject(parameters));
+                    break;
+                default:
+                    callTask = null;
```
cgillum: >callTask = null; [](start = 20, length = 16)

This will result in a null-ref below. Please throw an exception instead.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -234,9 +234,23 @@ public Task CallFunctionAsync(string functionName, params object[] parameters)
...
+                case FunctionType.Activity:
+                    callTask = this.innerContext.ScheduleTask<TResult>(functionName, version, parameters);
+                    break;
+                case FunctionType.Orchestrator:
+                    callTask = this.innerContext.CreateSubOrchestrationInstance<TResult>(functionName, version, JsonConvert.SerializeObject(parameters));
+                    break;
+                default:
+                    throw new ArgumentException(
+                        string.Format("The function '{0}' doesn't exist, is disabled, or is not an activity or orchestrator function",
+                            functionName));
```
cgillum: The point of this default clause is to flag the case where we see a function type enum value that we don't expect. Therefore, the correct exception is going to be something like throw new InvalidOperationException($"Unexpected function type {functionType}.");

Issue #30, opened 2017-08-15
Automatic Function Retry


cgillum:
## Summary
This issue tracks adding automatic "retry-on-failure" parameters to the `CallFunctionAsync` API. The behavior is that it will automatically retry a function invocation if it fails with an error up to a specified number of times and using a specified backoff behavior.

## Motivation
Manually implementing retry behavior in code is possible but can add a lot of complexity. Building it into the API could greatly simplify this experience.

tohling:
Addressed in commit https://github.com/Azure/azure-functions-durable-extension/pull/51

PR #51, opened 2017-09-08
Added support for retry options for activity functions.

tohling:
Addressing issue https://github.com/Azure/azure-functions-durable-extension/issues/30

1. Added support for retry options using `DurableTask.Core.OrchestrationContext.ScheduleWithRetry`.

      **Sample client code snippet:**
      ```
     // Given RetryOptions scoped to Extensions.DurableTask.RetryOptions, 
     RetryOptions options = new RetryOptions(TimeSpan.FromSeconds(5), 3); // FirstRetryInterval = 5 seconds, MaxNumberOfAttempts = 3
     await ctx.CallFunctionWithRetryAsync(nameof(TestActivities.Throw), options, message);
     ```
2. Added end-to-end tests.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
@@ -220,6 +220,9 @@ namespace Microsoft.Azure.WebJobs
         /// <param name="retryOptions">The retry option for the activity function.</param>
         /// <param name="parameters">The JSON-serializeable parameters to pass as input to the function.</param>
         /// <returns>A durable task that completes when the called function completes or fails.</returns>
+        /// <exception cref="ArgumentNullException">
+        /// The retry option object is null.
+        /// </exception>
         /// <exception cref="ArgumentException">
         /// The specified function does not exist, is disabled, or is not an orchestrator function.
         /// </exception>
@@ -229,8 +232,13 @@ namespace Microsoft.Azure.WebJobs
         /// <exception cref="DurableTask.Core.Exceptions.TaskFailedException">
         /// The activity function failed with an unhandled exception.
         /// </exception>
-        public Task CallFunctionWithRetryAsync(string functionName, Extensions.DurableTask.RetryOptions retryOptions, params object[] parameters)
+        public Task CallFunctionWithRetryAsync(string functionName, RetryOptions retryOptions, params object[] parameters)
         {
+            if (retryOptions == null)
+            {
+                throw new ArgumentNullException(nameof(retryOptions));    
+            }
+
             return this.CallFunctionWithRetryAsync<string>(functionName, retryOptions, parameters);
         }
 
@@ -263,6 +271,9 @@ namespace Microsoft.Azure.WebJobs
         /// <param name="retryOptions">The retry option for the activity function.</param>
         /// <param name="parameters">The JSON-serializeable parameters to pass as input to the function.</param>
         /// <returns>A durable task that completes when the called function completes or fails.</returns>
+        /// <exception cref="ArgumentNullException">
+        /// The retry option object is null.
+        /// </exception>
         /// <exception cref="ArgumentException">
         /// The specified function does not exist, is disabled, or is not an orchestrator function.
         /// </exception>
@@ -272,8 +283,13 @@ namespace Microsoft.Azure.WebJobs
         /// <exception cref="DurableTask.Core.Exceptions.TaskFailedException">
         /// The activity function failed with an unhandled exception.
         /// </exception>
-        public async Task<TResult> CallFunctionWithRetryAsync<TResult>(string functionName, Extensions.DurableTask.RetryOptions retryOptions, params object[] parameters)
+        public async Task<TResult> CallFunctionWithRetryAsync<TResult>(string functionName, RetryOptions retryOptions, params object[] parameters)
         {
+            if (retryOptions == null)
+            {
+                throw new ArgumentNullException(nameof(retryOptions));
+            }
+
             return await CallDurableTaskFunctionAsync<TResult>(functionName, retryOptions, parameters);
         }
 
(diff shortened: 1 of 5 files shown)


src/WebJobs.Extensions.DurableTask/RetryOptions.cs
```diff
@@ -0,0 +1,90 @@
...
+
+using System;
+using DurableTaskCore = DurableTask.Core;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    /// <summary>
+    ///     Contains retry policies that can be passed as parameters to various operations
+    /// </summary>
+    public class RetryOptions
```
tohling: >RetryOptions [](start = 17, length = 12)

I like the brevity of this name but don't feel too strongly about keeping it. If you feel like the workaround for the conflict of DurableTaskCore.RetryOptions is not a good idea, kindly propose a new name.

src/WebJobs.Extensions.DurableTask/RetryOptions.cs
```diff
@@ -0,0 +1,90 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using DurableTaskCore = DurableTask.Core;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
```
cgillum: Up until this point, all public types that are intended to be used by user code went into the `Microsoft.Azure.WebJobs` namespace. This is what I observed in other extensions as well. I think we should do the same here.

src/WebJobs.Extensions.DurableTask/RetryOptions.cs
```diff
@@ -0,0 +1,90 @@
...
+{
+    /// <summary>
+    ///     Contains retry policies that can be passed as parameters to various operations
+    /// </summary>
+    public class RetryOptions
+    {
+        private DurableTaskCore.RetryOptions retryOptions;
```
cgillum: Make the `retryOptions` field readonly.

src/WebJobs.Extensions.DurableTask/RetryOptions.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        /// <summary>
+        /// Creates a new instance RetryOptions with the supplied first retry and max attempts
+        /// </summary>
+        /// <param name="firstRetryInterval">Timespan to wait for the first retry</param>
+        /// <param name="maxNumberOfAttempts">Max number of attempts to retry</param>
+        /// <exception cref="ArgumentException"></exception>
```
cgillum: Can we add documentation text for the `ArgumentException` tag here?

src/WebJobs.Extensions.DurableTask/RetryOptions.cs
```diff
@@ -0,0 +1,90 @@
...
+        public RetryOptions(TimeSpan firstRetryInterval, int maxNumberOfAttempts)
+        {
+            retryOptions = new DurableTaskCore.RetryOptions(firstRetryInterval, maxNumberOfAttempts);
+        }
+
+        /// <summary>
+        /// Gets or sets the first retry interval
+        /// </summary>
+        public TimeSpan FirstRetryInterval
```
cgillum: I realize this may be a bit of a drag, but the automatic API doc generation tools require that we include <value> tags for each public property we expose. Can you add those to all the properties in this class? Look at the property documentation for other classes, like `DurableOrchestrationContext` for an example of how I've been doing this.

test/DurableTaskEndToEndTests.cs
```diff
@@ -478,6 +478,41 @@ public DurableTaskEndToEndTests(ITestOutputHelper output)
...
+        [Fact]
+        public async Task UnhandledActivityExceptionWithRetry()
+        {
+            string[] orchestratorFunctionNames =
+            {
+                nameof(TestOrchestrations.ThrowWithRetry)
+            };
+            string activityFunctioName = nameof(TestActivities.Throw);
+
+            using (JobHost host = TestHelpers.GetJobHost(loggerFactory, nameof(UnhandledActivityException)))
```
cgillum: This should be `nameof(UnhandledActivityExceptionWithRetry)`.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -398,6 +366,97 @@ public void ContinueAsNew(object input)
...
+                        callTask = this.innerContext.ScheduleTask<TResult>(functionName, version, parameters);
+                    }
+                    else
+                    {
+                        callTask = this.innerContext.ScheduleWithRetry<TResult>(functionName, version,
+                            retryOptions.GetRetryOptions(), parameters);
+                    }
+                    break;
+                case FunctionType.Orchestrator:
+                    callTask = this.innerContext.CreateSubOrchestrationInstance<TResult>(functionName, version,
```
cgillum: What if someone calls `CallFunctionWithRetryAsync` for an orchestrator function and includes a `RetryOptions` parameter? I see that it gets ignored in this implementation, but do you think that behavior is going to be acceptable from a user's perspective? Had you considered throwing an exception? Or do you think we need to revisit reusing CallFunctionAsync for both activity and orchestrator functions since they now have different applicable parameter sets?

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -398,6 +366,97 @@ public void ContinueAsNew(object input)
             this.ContinuedAsNew = true;
         }
 
+        internal async Task<TResult> CallDurableTaskFunctionAsync<TResult>(string functionName,
```
cgillum: Can this method be private? If so, let's make it private. It's good to minimize visibility for both better encapsulation but (IMO more importantly) for better code understandability/maintainability.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
tohling: I've added support for orchestrator functions as well.

---
In reply to: [137793230](https://github.com/Azure/azure-functions-durable-extension/pull/51#discussion_r137793230) [](ancestors = 137793230)

cgillum: Oh, I didn't realize this was already supported! Perfect!

---
In reply to: [137895273](https://github.com/Azure/azure-functions-durable-extension/pull/51#discussion_r137895273) [](ancestors = 137895273,137793230)

PR #60, opened 2017-09-13
Durable Functions Beta Release (.NET Standard 2.0 support, updated docs, etc.)

cgillum:
This change is to add support for .NET Core, which is required for the Functions 2.0 release. I'm also changing the version number to 1.0.0-beta in preparation for the beta release (which contains some significant breaking changes).


cgillum:
@fabiocav, I'm currently targeting both net461 and netstandard2.0, but is there any reason for me to do this? Would it make sense for me to just target netstandard2.0 (which as I understand is implemented by the .NET Framework 4.6.1)?

fabiocav:
Yes, in your guys' case, it would make sense to be .NET Standard 2.0 only.

cgillum:
All tests are passing locally on a semi-clean machine. Not sure why the CI is not happy. Ignoring CI results for now.

PR #80, opened 2017-10-25
Fix typo in sub-orchestrations.md

d-dizhevsky:
Adding missing deviceId parameter to CallSubOrchestratorAsync method


cgillum:
Thanks for catching this. :)

@tdykstra, we'll need to fix this in https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-sub-orchestrations#example as well.

d-dizhevsky:
And I believe in here as well?

cgillum:
@d-dizhevsky Yes, that one will get updated as part of a separate build/publish step that I can take care of on my end. Note, however, that we'll be retiring the docs on github.io and keeping everything on docs.microsoft.com.

PR #90, opened 2017-11-11
Added new HandlesReturnValue attribute to ActivityTrigger.

kashimiz:
Addresses, but does not fully solve, #13 

Reliant on [PR 1424](https://github.com/Azure/azure-webjobs-sdk/pull/1424) in azure-webjobs-sdk.


src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -10,6 +10,10 @@
     <Company>Microsoft Corporation</Company>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <DefineConstants>NETSTANDARD20</DefineConstants>
```
cgillum: No need to define this yourself. When targeting netstandard2.0, the compiler automatically defines a NETSTANDARD2_0 constant.

src/WebJobs.Extensions.DurableTask/ActivityTriggerAttribute.cs
```diff
@@ -14,6 +14,9 @@ namespace Microsoft.Azure.WebJobs
     [AttributeUsage(AttributeTargets.Parameter)]
     [DebuggerDisplay("{Activity} ({Version})")]
     [Binding]
+#if NETSTANDARD20
```
cgillum: Use the built-in NETSTANDARD2_0 constant instead.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -21,7 +25,7 @@
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.0-beta3" />
```
cgillum: Before this can be merged, we'll also need to update this package reference to a version of Microsoft.Azure.WebJobs that contains your new attribute.

src/WebJobs.Extensions.DurableTask/ActivityTriggerAttribute.cs
```diff
@@ -13,7 +13,12 @@ namespace Microsoft.Azure.WebJobs
     /// </summary>
     [AttributeUsage(AttributeTargets.Parameter)]
     [DebuggerDisplay("{Activity} ({Version})")]
+#if NETSTANDARD2_0
+    [Binding(TriggerHandlesReturnValue = true)]
+#else
     [Binding]
+#endif
```
MikeStall: This suggests we're dual compiling for 1.0?  We have a wave of breaking changes coming through (to remove the [Obsolete] attributes), so this may get trickier. You may want to move to a spot where you have 2 separate branches, rather than #ifdef.

cgillum: That's a good point. Right now we have a single NuGet package which targets both .NET 4.6 and .NET Standard 2.0. It's nice because the maintenance is very straightforward. I can certainly see this getting ugly with the new changes you're making. I'll defer that decision to later when I see the scope of the changes. :)

src/WebJobs.Extensions.DurableTask/ActivityTriggerAttribute.cs
```diff
@@ -13,7 +13,12 @@ namespace Microsoft.Azure.WebJobs
...
+
```
cgillum: Can you remove this extra newline?

Issue #100, opened 2017-11-27
Orchestration status values should be enums rather than strings


cgillum:
The `DurableOrchestrationStatus` type exposes the orchestration execution status as a string. This makes writing functions which check for this execution status more painful than they need to be, since they need to reply upon string comparisons. Instead of strings, these should be enum values.

Issue #99, opened 2017-11-25
Allow settable retry-after header to work with Azure 202 patterns


jeffhollan:
Today calling an http function that returns the durable orchestrator status has a 202 header and a location.  When calling from Azure Logic Apps, because there is no `retry-after` header it immediately calls the location header (which returns a 404 as I assume it's still initializing).  Would prefer to return a  `retry-after` header as well to let the polling client know what an appropriate polling interval would be [per Azure 202 patterns](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-async-operations)

Fixes could be either:
1. Allow returning of a retry-after header so the Logic App won't poll until after that duration
2. Make the status endpoint callable immediately as not to receive a 404

cgillum:
Ideally we would implement your second suggestion because getting a 404 response isn't helpful to anyone. :P

In the meantime, try adding the `retry-after` header manually in the HTTP trigger function that is starting the orchestration, like the following:

```csharp
[FunctionName("HttpTrigger_HttpStart")]
public static async Task<HttpResponseMessage> HttpStart(
    [HttpTrigger(AuthorizationLevel.Function, "post")]HttpRequestMessage req,
    [OrchestrationClient]DurableOrchestrationClient starter,
    TraceWriter log)
{
    // Function input comes from the request content.
    ProcessRequest requestData = await req.Content.ReadAsAsync<ProcessRequest>();
    string instanceId = await starter.StartNewAsync("HttpTrigger_Orchestrator", requestData);
 
    log.Info($"Started orchestration with ID = '{instanceId}'.");
 
    HttpResponseMessage response = starter.CreateCheckStatusResponse(req, instanceId);
    response.Headers.RetryAfter = new RetryConditionHeaderValue(TimeSpan.FromSeconds(10));
    return response;
}
```

The resulting response headers will include the `retry-after` header, looking something like this (running locally on Functions v1):

```http
HTTP/1.1 202 Accepted
Content-Type: application/json; charset=utf-8
Date: Mon, 27 Nov 2017 17:25:55 GMT
Location: http://localhost:7071/admin/extensions/DurableTaskExtension/instances/eb49a7c1241840f2a9899fd2a0e84615?taskHub=SampleHubVS&connection=Storage&code=SE2NVBADa7RWYhZnunqMgPURuoJFJIQnMUzMvIQ4QkeRjWe5rvRFzg==
Retry-After: 10
[...]

Issue #97, opened 2017-11-23
Cumbersome to RaiseExternalEvent for Sub-orchestrations.


Kimmen:
**Scenario**
We have function logic that notifies users about an event and awaits responses from those users. The logic notifies the users twice: first via a "standard" notification and if they don't respond within a certain time we notify them via SMS.

We solve this by doing two orchestration: first one determines which users are to be notified and waits for responses. Second orchestration is the "await for response" orchestration, which first sends a notification, waits for a user response or a timeout (similar to the [Hurman Interaction](https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-phone-verification) section in the documentation), using ExternalEvents (Wait and Raise).

The idea was to be able to set a predictable instanceId for the sub-orchestration so it is easy to RaiseExternalEvent once the user do respond. But it seems like you are not able to set an instance id for a sub-orchestration. At first I thought that the sub-orchtestration run in the same context as the parent (somehow) and used the same instance id, but that's not the case. 

I guess you got your reasons for not be able to set instance ids for sub-orchestration, but it would be nice in our case! Now we are persisting the instance id in a store and retrieving it when we get a response from the user.

cgillum:
Actually, there isn't a good reason for this. We simply forgot to include this feature! Another user raised a similar issue here: https://github.com/Azure/azure-functions-durable-extension/issues/91

This is something we plan on adding. Thank you for sharing the details of your scenario. It's a good one that I had not previously thought about.

Kimmen:
Greate! I inspected the code and saw that it was at least possible to provide an instance id to a sub-orchestration. For now we have a working workaround, but we are able to simplify the code once this is in place.

cgillum:
I've added the new APIs which will allow you to specify an instance ID when you start a new sub-orchestration. You can find them in the latest version of the NuGet package. There's also a new automated test which does exactly your scenario, so I feel pretty good about it. :)

Kimmen:
That's great! Now we can remove redundant code, thanks.

Issue #94, opened 2017-11-21
Singleton & eternal orchestrations become slower over time


cgillum:
### Summary
Stress testing has revealed that singleton and eternal orchestrations become slower over time and can eventually become unresponsive.

### Details
The main problem is the way history data is stored in table storage. Row keys include both the execution ID (generation) and the sequence number. Execution IDs are randomly generated, so finding the current generation may require a full partition scan, which is an O(N) I/O operation. This becomes very expensive over time, leading to performance problems and the instance can eventually become unresponsive.

### Proposed Fix
The fix is to change the way the Azure Storage provider stores execution history. Instead of using the execution ID + sequence number as the row key, only the sequence number will be used for the row key. Previous generations' execution history will be overwritten by history events for the new generations. This is nice because the previous generation's history is not useful for anything, so it might as well be discarded. The result is much less data being stored in table storage. It also makes loading instance history much more efficient because only the first N rows in a partition need to be read (where N is the number of history events written for the current generation). The end result is that the load time of an instance will be the same regardless of how many generations it undergoes.

**IMPORTANT**: This is a breaking change. Any in-flight orchestration instances are likely to break when upgrading to a version of the storage provider that contains this change. If preserving data is important, a new function app should be created with either a different storage account or a different task hub name.

PR #103, opened 2017-12-01
Durable Functions Beta 2 Release Payload

cgillum:
This PR contains the following changes:

* Updated README with documentation outline, with links to the [official docs on docs.microsoft.com](https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-overview)
* Several bug fixes and enhancements which are tracked in the [Beta 2 Milestone](https://github.com/Azure/azure-functions-durable-extension/milestone/2)
* Updated reference to [DurableTask.AzureStorage 1.1.1-beta](https://www.nuget.org/packages/DurableTask.AzureStorage/1.1.1-beta), which contains various performance and reliability improvements
* Updated current extension package version to **1.1.0-beta2**
* Updated samples to point to the latest bits

Note that **there are several breaking changes** included in the beta2 payload. This includes breaking API changes as well as breaking Azure Storage schema changes, which will cause any in-flight instances to fail if existing apps are upgraded to this new package version. For that reason, it is highly recommended that any existing projects have their `taskHub` values changed or point to a different storage account.


cgillum:
For tracking purposes, here are the issues addressed by this PR:

- https://github.com/Azure/azure-functions-durable-extension/issues/74
- https://github.com/Azure/azure-functions-durable-extension/issues/75
- https://github.com/Azure/azure-functions-durable-extension/issues/78
- https://github.com/Azure/azure-functions-durable-extension/issues/79
- https://github.com/Azure/azure-functions-durable-extension/issues/86
- https://github.com/Azure/azure-functions-durable-extension/issues/91
- https://github.com/Azure/azure-functions-durable-extension/issues/94
- https://github.com/Azure/azure-functions-durable-extension/issues/97
- https://github.com/Azure/azure-functions-durable-extension/issues/99
- https://github.com/Azure/azure-functions-durable-extension/issues/100

src/WebJobs.Extensions.DurableTask/FunctionFailedException.cs
```diff
@@ -0,0 +1,34 @@
...
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// The exception that is thrown when a sub-orchestrator or activity function fails
+    /// with an error.
+    /// </summary>
+    public class FunctionFailedException : Exception
```
tohling: >public [](start = 4, length = 6)

internal?

tohling:
>     internal static class TestActivities

There are no changes in this file.

---
Refers to: test/TestActivities.cs:10 in 5946c0f. [](commit_id = 5946c0fd7ac041bd8d6dd6029c7bbe77d4aea394, deletion_comment = False)

src/WebJobs.Extensions.DurableTask/FunctionFailedException.cs
cgillum: This is intentionally public. I want developers to be able to explicitly catch this exception in a try/catch block (otherwise they have to catch all exceptions, which isn't always desirable). I've updated the `TryCatchLoop` test orchestration to explicitly catch this exception type to make sure it's being surfaced correctly. The trace validation code should be looking for it as well.

cgillum:
>     internal static class TestActivities

Sorry, there is actually a small whitespace change.

---
In reply to: [348370667](https://github.com/Azure/azure-functions-durable-extension/pull/103#issuecomment-348370667) [](ancestors = 348370667)

---
Refers to: test/TestActivities.cs:10 in 5946c0f. [](commit_id = 5946c0fd7ac041bd8d6dd6029c7bbe77d4aea394, deletion_comment = False)

src/WebJobs.Extensions.DurableTask/FunctionState.cs
```diff
@@ -12,6 +12,7 @@ internal enum FunctionState
         Completed,
         Terminated,
         Failed,
-        ExternalEventRaised
+        ExternalEventRaised,
+        TimerExpired
```
tohling: >TimerExpired [](start = 8, length = 12)

Is there a way to write test for this new state?  Sorry I can't seem to find it.

src/WebJobs.Extensions.DurableTask/FunctionFailedException.cs
tohling: Do we have to worry about this exception crossing app domains? If yes, we should add the [Serializable] attribute.

---
In reply to: [154245498](https://github.com/Azure/azure-functions-durable-extension/pull/103#discussion_r154245498) [](ancestors = 154245498)

cgillum: We don't have to worry about this.

---
In reply to: [154247221](https://github.com/Azure/azure-functions-durable-extension/pull/103#discussion_r154247221) [](ancestors = 154247221,154245498)

src/WebJobs.Extensions.DurableTask/FunctionState.cs
cgillum: Would TestHelpers.GetLogs_TimerExpiration be the right place for this?

---
In reply to: [154245997](https://github.com/Azure/azure-functions-durable-extension/pull/103#discussion_r154245997) [](ancestors = 154245997)

tohling: yes

---
In reply to: [154248091](https://github.com/Azure/azure-functions-durable-extension/pull/103#discussion_r154248091) [](ancestors = 154248091,154245997)

Issue #102, opened 2017-11-28
Allow HTTP triggers to wait synchronously for an orchestration to complete


cgillum:
Kanio Dimitrov made a suggestion about this here: https://azurekan.wordpress.com/2017/11/28/sync-response-for-durable-functions/. 

The basic idea as I understand it is to enable an HTTP trigger to start an instance and wait for it to complete, up to some specified timeout.  The objective is to simplify how HTTP clients can consume durable functions that are expected to complete in a relatively short amount of time (i.e. in less than 10 minutes). If the function completes within the timeout, return the output as part of the HTTP response. Otherwise return the usual HTTP 202 response.

cgillum:
One option would be to create a new API that is a peer to `DurableOrchestrationClient.CreateCheckStatusResponse` that takes a timeout parameter. It can return an `HttpResponseMessage` but the actual response may vary depending on how quickly the instance terminates. Internally it will likely need to rely on polling since there is no reliable way to get a notification when the instance reaches the completed

PR #106, opened 2017-12-05
Adding Sync Response for Orchestration Client

gled4er:
Hello,

The code in this pull request is starting the implementation of the feature described here - https://github.com/Azure/azure-functions-durable-extension/issues/102

Unit tests are to be added but I want to get early feedback if the direction I chose is the correct one.

Thank you!

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
@@ -164,7 +164,8 @@ namespace Microsoft.Azure.WebJobs
         /// </summary>
         /// <param name="instanceId">The ID of the orchestration instance to query.</param>
         /// <returns>Returns a task which completes when the status has been fetched.</returns>
-        public async Task<DurableOrchestrationStatus> GetStatusAsync(string instanceId)
+        // To be discussed if this is an acceptable approach 
+        public virtual async Task<DurableOrchestrationStatus> GetStatusAsync(string instanceId)
         {
             OrchestrationState state = await this.client.GetOrchestrationStateAsync(instanceId);
             if (state == null)
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
@@ -42,26 +42,25 @@ namespace Microsoft.Azure.WebJobs
         /// </summary>
         /// <param name="request">The HTTP request that triggered the current function.</param>
         /// <param name="instanceId">The unique ID of the instance to check.</param>
-        /// <returns>An HTTP response which may include a 202 and locaton header.</returns>
+        /// <returns>An HTTP response which may include a 202 and location header.</returns>
         public HttpResponseMessage CreateCheckStatusResponse(HttpRequestMessage request, string instanceId)
         {
             return this.config.CreateCheckStatusResponse(request, instanceId, this.attribute);
         }
-
 
         /// <summary>
         /// Creates an HTTP response for checking the status of the specified instance supporting synchronous response as well. 
         /// </summary>
         /// <param name="request">The HTTP request that triggered the current function.</param>
         /// <param name="instanceId">The unique ID of the instance to check.</param>
-        /// <param name="totalTimeout">Total allowed timeout for output from the durable function</param>
-        /// <param name="retryTimeout">Timeout between checks for output from the durable function</param>
-        /// <returns>Task of <see cref="HttpResponseMessage"/></returns>
-        public async Task<HttpResponseMessage> CreateCheckStatusResponse(HttpRequestMessage request, string instanceId, int totalTimeout, int retryTimeout)
+        /// <param name="timeout ">Total allowed timeout for output from the durable function.</param>
+        /// <param name="timeout">Total allowed timeout for output from the durable function.</param>
+        /// <param name="retryInterval">Timeout between checks for output from the durable function.</param>
+        /// <returns>An HTTP response which may include a 202 and location header or a 200 with the durable function output in the response body.</returns>
+        public async Task<HttpResponseMessage> CreateCheckStatusResponse(HttpRequestMessage request, string instanceId, TimeSpan timeout, TimeSpan? retryInterval)
         {
-            return await config.CreateCheckStatusResponse(request, instanceId, this.attribute, totalTimeout, retryTimeout);
+            return await this.config.CreateCheckStatusResponse(request, instanceId, this.attribute, timeout, retryInterval ?? TimeSpan.FromSeconds(5));
         }
-
 
         /// <summary>
         /// Starts a new execution of the specified orchestrator function.
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
@@ -55,11 +55,11 @@ namespace Microsoft.Azure.WebJobs
         /// <param name="request">The HTTP request that triggered the current function.</param>
         /// <param name="instanceId">The unique ID of the instance to check.</param>
         /// <param name="totalTimeout">Total timeout that the client should wait for response from the durable function</param>
-        /// <param name="iterationTimeout">Timeout for one iteration between checks for output from the durable function</param>
+        /// <param name="retryTimeout">Timeout for one iteration between checks for output from the durable function</param>
         /// <returns>Task of <see cref="HttpResponseMessage"/></returns>
-        public async Task<HttpResponseMessage> CreateCheckStatusResponse(HttpRequestMessage request, string instanceId, int totalTimeout = 300000, int iterationTimeout = 300)
+        public async Task<HttpResponseMessage> CreateCheckStatusResponse(HttpRequestMessage request, string instanceId, int totalTimeout = 300000, int retryTimeout = 300)
         {
-            return await config.CreateCheckStatusResponse(request, instanceId, this.attribute, totalTimeout, iterationTimeout);
+            return await config.CreateCheckStatusResponse(request, instanceId, this.attribute, totalTimeout, retryTimeout);
         }
 
 
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
@@ -54,10 +54,10 @@ namespace Microsoft.Azure.WebJobs
         /// </summary>
         /// <param name="request">The HTTP request that triggered the current function.</param>
         /// <param name="instanceId">The unique ID of the instance to check.</param>
-        /// <param name="totalTimeout">Total timeout that the client should wait for response from the durable function</param>
-        /// <param name="retryTimeout">Timeout for one iteration between checks for output from the durable function</param>
+        /// <param name="totalTimeout">Total allowed timeout for output from the durable function</param>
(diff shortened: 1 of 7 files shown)


src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -48,6 +48,21 @@ public HttpResponseMessage CreateCheckStatusResponse(HttpRequestMessage request,
             return this.config.CreateCheckStatusResponse(request, instanceId, this.attribute);
         }
 
+
+        /// <summary>
+        /// Creates an HTTP response for checking the status of the specified instance supporting synchronous response as well. 
+        /// </summary>
+        /// <param name="request">The HTTP request that triggered the current function.</param>
+        /// <param name="instanceId">The unique ID of the instance to check.</param>
+        /// <param name="totalTimeout">Total allowed timeout for output from the durable function</param>
```
cgillum: For consistency, each parameter description and summary description should end in a period.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -48,6 +48,21 @@ public HttpResponseMessage CreateCheckStatusResponse(HttpRequestMessage request,
...
+        /// <param name="retryTimeout">Timeout between checks for output from the durable function</param>
+        /// <returns>Task of <see cref="HttpResponseMessage"/></returns>
+        public async Task<HttpResponseMessage> CreateCheckStatusResponse(HttpRequestMessage request, string instanceId, int totalTimeout, int retryTimeout)
```
cgillum: A few comments on the method signature:
* Timeouts should take `TimeSpan` instead of `int`.
* Consider `timeout` instead of `totalTimeout`.
* Consider `retryInterval` instead of `retryTimeout`.
* Developers don't always know how to choose retry intervals, so consider making it optional with a default value of 5 seconds.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -48,6 +48,21 @@ public HttpResponseMessage CreateCheckStatusResponse(HttpRequestMessage request,
...
 
+
+        /// <summary>
+        /// Creates an HTTP response for checking the status of the specified instance supporting synchronous response as well. 
+        /// </summary>
+        /// <param name="request">The HTTP request that triggered the current function.</param>
+        /// <param name="instanceId">The unique ID of the instance to check.</param>
+        /// <param name="totalTimeout">Total allowed timeout for output from the durable function</param>
+        /// <param name="retryTimeout">Timeout between checks for output from the durable function</param>
+        /// <returns>Task of <see cref="HttpResponseMessage"/></returns>
```
cgillum: The `<returns>` comment should be more descriptive. I would expect it explains that either an HTTP 200 or HTTP 202 response is returned. See the `<returns>` description for the original `CreateCheckStatusResponse` for an example.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -48,6 +48,21 @@ public HttpResponseMessage CreateCheckStatusResponse(HttpRequestMessage request,
...
+        public async Task<HttpResponseMessage> CreateCheckStatusResponse(HttpRequestMessage request, string instanceId, int totalTimeout, int retryTimeout)
+        {
+            return await config.CreateCheckStatusResponse(request, instanceId, this.attribute, totalTimeout, retryTimeout);
+        }
+
+
```
cgillum: nit: remove extra newline.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -34,40 +35,17 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
             string instanceId,
             OrchestrationClientAttribute attribute)
         {
-            if (this.config.NotificationUrl == null)
-            {
-                throw new InvalidOperationException("Webhooks are not configured");
-            }
-
-            Uri notificationUri = this.config.NotificationUrl;
+            GetClientResponseLinks(request, instanceId, attribute, out var statusQueryGetUri, out var sendEventPostUri,
```
cgillum: style nit: I prefer all parameters be on the same line or all of them are on their own line. Same for here and below.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -48,6 +48,21 @@ public HttpResponseMessage CreateCheckStatusResponse(HttpRequestMessage request,
...
+        /// Creates an HTTP response for checking the status of the specified instance supporting synchronous response as well. 
+        /// </summary>
+        /// <param name="request">The HTTP request that triggered the current function.</param>
+        /// <param name="instanceId">The unique ID of the instance to check.</param>
+        /// <param name="totalTimeout">Total allowed timeout for output from the durable function</param>
+        /// <param name="retryTimeout">Timeout between checks for output from the durable function</param>
+        /// <returns>Task of <see cref="HttpResponseMessage"/></returns>
+        public async Task<HttpResponseMessage> CreateCheckStatusResponse(HttpRequestMessage request, string instanceId, int totalTimeout, int retryTimeout)
+        {
+            return await config.CreateCheckStatusResponse(request, instanceId, this.attribute, totalTimeout, retryTimeout);
```
cgillum: style nit: to be consistent with the rest of the codebase, please prefix all instance variables, instance properties, and instance methods with `this.`. There are several places in this PR where this needs to be updated.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -76,6 +54,65 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
+            OrchestrationClientAttribute attribute,
+            int totalTimeout,
+            int retryTimeout)
+        {
+            GetClientResponseLinks(request, instanceId, attribute, out var statusQueryGetUri, out var sendEventPostUri,
+                out var terminatePostUri);
+
+            if (totalTimeout < retryTimeout)
+            {
+                throw new InvalidOperationException($"Total timeout {totalTimeout} should be bigger than {retryTimeout}");
```
cgillum: `ArgumentException` or `ArgumentOutOfRangeException` would be more appropriate in this context.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -76,6 +54,65 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
+            }
+
+            var iterationCount = totalTimeout / retryTimeout;
+            JToken durableFunctionOutput = null;
+
+            var client = GetClient(request);
+            var status = await client.GetStatusAsync(instanceId);
+            for (var i = 0; i < iterationCount; i++)
+            {
+                Thread.Sleep(retryTimeout);
```
cgillum: Use `await Task.Delay` instead of `Thread.Sleep` to avoid blocking threads.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -76,6 +54,65 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
+        {
+            GetClientResponseLinks(request, instanceId, attribute, out var statusQueryGetUri, out var sendEventPostUri,
+                out var terminatePostUri);
+
+            if (totalTimeout < retryTimeout)
+            {
+                throw new InvalidOperationException($"Total timeout {totalTimeout} should be bigger than {retryTimeout}");
+            }
+
+            var iterationCount = totalTimeout / retryTimeout;
```
cgillum: I suggest a more robust approach to managing the timeout. The problem with `iterationCount` is that it doesn't work well when `totalTimeout` is not divisible by `retryTimeout`. For example, what should the correct behavior be when `totalTimeout` is `30` but `retryTimeout` is `8`? In the current implementation, the total timeout is actually 24 seconds instead of 30 seconds.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -76,6 +54,65 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
+            JToken durableFunctionOutput = null;
+
+            var client = GetClient(request);
+            var status = await client.GetStatusAsync(instanceId);
+            for (var i = 0; i < iterationCount; i++)
+            {
+                Thread.Sleep(retryTimeout);
+                if (status != null && status.RuntimeStatus == OrchestrationRuntimeStatus.Completed)
```
cgillum: We need to also handle the case where the orchestration status is in other terminal states, like `Failed`, `Terminated`, or `Canceled`. In those cases, it probably makes sense to return an error status value.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -76,6 +54,65 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
+                {
+                    durableFunctionOutput = status.Output;
+                    break;
+                }
+                status = await client.GetStatusAsync(instanceId);
+            }
+
+            if (durableFunctionOutput == null)
```
cgillum: If the orchestration completes but returns a `null` output, this code will return a 202 response, which is not correct. Rather than using the variable for the output, you should use a separate variable for if the operation timed-out.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -76,6 +54,65 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
+                response.Headers.Location = new Uri(statusQueryGetUri);
+                response.Headers.RetryAfter = new RetryConditionHeaderValue(TimeSpan.FromSeconds(10));
+                return response;
+            }
+            else
+            {
+                var response = request.CreateResponse(
+                    HttpStatusCode.OK, durableFunctionOutput);
+                return response;
+
```
cgillum: nit: remove extra newline here and a few lines below.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -76,6 +54,65 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
+                // Implement the async HTTP 202 pattern.
+                response.Headers.Location = new Uri(statusQueryGetUri);
+                response.Headers.RetryAfter = new RetryConditionHeaderValue(TimeSpan.FromSeconds(10));
+                return response;
+            }
+            else
+            {
+                var response = request.CreateResponse(
+                    HttpStatusCode.OK, durableFunctionOutput);
```
cgillum: nit: move these parameters up to be on the same line as the method call.
`var response = request.CreateResponse(HttpStatusCode.OK, durableFunctionOutput);`

cgillum:
Also, code changes should be submitted to the **dev** branch instead of the **master** branch. 
The **master** branch is only updated as part of a documentation update or a release.

gled4er:
@cgillum I addressed the issues that you raised. Let's discuss the right approach for testing. Thank you!

PR #110, opened 2017-12-08
Sync Response feature for Orchestration Client 

gled4er:
Pull request for adding sync response capability to Orchestration Client as described in this issue - https://github.com/Azure/azure-functions-durable-extension/issues/102


gled4er:
@cgillum Please check this pull request. Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -42,12 +42,26 @@ public class DurableOrchestrationClient
...
         }
 
+        /// <summary>
+        /// Creates an HTTP response for checking the status of the specified instance supporting synchronous response as well. 
+        /// </summary>
+        /// <param name="request">The HTTP request that triggered the current function.</param>
+        /// <param name="instanceId">The unique ID of the instance to check.</param>
+        /// <param name="timeout ">Total allowed timeout for output from the durable function.</param>
+        /// <param name="timeout">Total allowed timeout for output from the durable function.</param>
+        /// <param name="retryInterval">Timeout between checks for output from the durable function.</param>
```
cgillum: Let's add a note saying that the default is 5 seconds.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -282,7 +282,8 @@ private static Assembly ResolveAssembly(object sender, ResolveEventArgs args)
             return null;
         }
 
-        internal DurableOrchestrationClient GetClient(OrchestrationClientAttribute attribute)
+        // To be discussed if this is an acceptable approach 
+        internal virtual DurableOrchestrationClient GetClient(OrchestrationClientAttribute attribute)
```
cgillum: I'm fine with making methods virtual. In this case should we make it `protected internal` so others can benefit from it? It feels strange to have an `internal virtual` method since it only works for unit tests that are configured for access to internal members..

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -472,6 +473,23 @@ internal string GetIntputOutputTrace(string rawInputOutputData)
...
+        // Get a response that will wait for response from the durable function for predefined period of time before 
+        // pointing to our webhook handler. 
+        internal async Task<HttpResponseMessage> CreateCheckStatusResponse(
+            HttpRequestMessage request,
+            string instanceId,
+            OrchestrationClientAttribute attribute,
+            TimeSpan timeout,
+            TimeSpan retryInterval)
+        {
+            if (DisableHttpManagementApis)
```
cgillum: Please prefix with `this.` to make it clear to the reader that this is an instance member.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -34,48 +35,65 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
-            if (!string.IsNullOrEmpty(notificationUri.Query))
+        internal async Task<HttpResponseMessage> CreateCheckStatusResponse(
+            HttpRequestMessage request,
+            string instanceId,
+            OrchestrationClientAttribute attribute,
+            TimeSpan timeout,
+            TimeSpan retryInterval)
+        {
+            this.GetClientResponseLinks(request, instanceId, attribute, out var statusQueryGetUri, out var sendEventPostUri, out var terminatePostUri);
+            if (TimeSpan.Compare(timeout, retryInterval) != 1)
```
cgillum: Could this be rewritten as `if (retryInterval > timeout)`? I think that would be more readable.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -295,5 +313,84 @@ private DurableOrchestrationClient GetClient(HttpRequestMessage request)
...
+                // This is expected to include the auto-generated system key for this extension.
+                querySuffix += "&" + notificationUri.Query.TrimStart('?');
+            }
+
+            statusQueryGetUri = instancePrefix + "?" + querySuffix;
+            sendEventPostUri = instancePrefix + "/" + RaiseEventOperation + "/{eventName}?" + querySuffix;
+            terminatePostUri = instancePrefix + "/" + TerminateOperation + "?reason={text}&" + querySuffix;
+        }
+
+        private HttpResponseMessage CreateResponseMessage(HttpRequestMessage request, string instanceId, string statusQueryGetUri, string sendEventPostUri, string terminatePostUri)
```
cgillum: Let's rename this to CreateCheckStatusResponseMessage.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -243,7 +261,7 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
             }
 
             string mediaType = request.Content.Headers.ContentType?.MediaType;
-            if (!string.IsNullOrEmpty(mediaType) && 
+            if (!string.IsNullOrEmpty(mediaType) &&
```
cgillum: Isn't this check redundant? `string.Equals` knows how to handle null values.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -34,48 +35,65 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
+            TimeSpan retryInterval)
+        {
+            this.GetClientResponseLinks(request, instanceId, attribute, out var statusQueryGetUri, out var sendEventPostUri, out var terminatePostUri);
+            if (TimeSpan.Compare(timeout, retryInterval) != 1)
             {
-                // This is expected to include the auto-generated system key for this extension.
-                querySuffix += "&" + notificationUri.Query.TrimStart('?');
+                throw new ArgumentException($"Total timeout {timeout.TotalSeconds} should be bigger than retry timeout {retryInterval.TotalSeconds}");
             }
+            this.GetRetryValues(timeout, retryInterval, out var iterationCount, out var leftOverInterval);
```
cgillum: I'm a little concerned about the complexity of this timeout implementation because of how it involves multiple methods, iteration counts, and division.  Could we refactor the timeout logic to be simpler?  Here is a suggestion:

```csharp
Stopwatch stopwatch = Stopwatch.StartNew();
while (true)
{
    var status = await client.GetStatusAsync(instanceId);

    // ...do checks on the status accordingly...

    TimeSpan elapsed = stopwatch.Elapsed;
    if (elapsed < timeout)
    {
        TimeSpan remainingTime = timeout.Subtract(elapsed);
        await Task.Delay(remainingTime > retryInterval ? retryInterval : remainingTime);
    }
    else
    {
        break;
    }
}
```

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -34,48 +35,65 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
+                else
+                {
+                    return request.CreateErrorResponse(HttpStatusCode.InternalServerError, $"The durable function's runtime status is {status.RuntimeStatus}");
+                }
+            }
+            var response = request.CreateResponse(HttpStatusCode.OK, durableFunctionOutput);
             return response;
         }
 
+
```
cgillum: nit: Please remove extra newlines.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -34,48 +35,65 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
-            response.Headers.RetryAfter = new RetryConditionHeaderValue(TimeSpan.FromSeconds(10));
+            if (durableFunctionOutput == null)
+            {
+                if (status.RuntimeStatus == OrchestrationRuntimeStatus.Running)
+                {
+                    return this.CreateResponseMessage(request, instanceId, statusQueryGetUri, sendEventPostUri, terminatePostUri);
+                }
+                else
+                {
+                    return request.CreateErrorResponse(HttpStatusCode.InternalServerError, $"The durable function's runtime status is {status.RuntimeStatus}");
```
cgillum: I don't think this error message provides enough information. It might be a good idea to return the same JSON payload that `HandleGetStatusRequestAsync` returns, for consistency with the async polling pattern.

gled4er:
@cgillum Thank you for the great feedback. I will work on the changes today in the evening and tomorrow.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
gled4er: Sure.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
gled4er: Great. I like simplifying the code.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
gled4er: Sure.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
gled4er: Thanks

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
gled4er: Great suggestion! I will improve it.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
gled4er: Sure

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
gled4er: What does "nit" stands for?

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
gled4er: Good point

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
cgillum: Sorry, it means I'm being [nit-picky](https://www.thefreedictionary.com/nitpicking). :) In other words, it's not an important thing to fix.

cgillum:
Sorry for the delay. I took a brief look and it's looking much better! I'll try to finish my review in the next few days.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
gled4er: thanks for the clarification.

gled4er:
@cgillum thank you very much!

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -34,48 +36,67 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
-                    statusQueryGetUri = statusQueryGetUri,
-                    sendEventPostUri = sendEventPostUri,
-                    terminatePostUri = terminatePostUri
-                });
-
-            // Implement the async HTTP 202 pattern.
-            response.Headers.Location = new Uri(statusQueryGetUri);
-            response.Headers.RetryAfter = new RetryConditionHeaderValue(TimeSpan.FromSeconds(10));
+                    durableFunctionOutput = status.Output;
+                    break;
```
cgillum: I'm having a hard time reasoning about the expected behavior of the code because there are many conditions that I have to keep track of in my head. I think this should be simplified a bit more to reduce complexity. Can we simplify this by removing the `break` statements and replacing them with `return` statements? For example, can we instead `return request.CreateResponse(HttpStatusCode.OK, status.Output);` here?

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -34,48 +36,67 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
+                }
+                if (status != null && (status.RuntimeStatus == OrchestrationRuntimeStatus.Canceled || status.RuntimeStatus == OrchestrationRuntimeStatus.Failed || status.RuntimeStatus == OrchestrationRuntimeStatus.Terminated))
+                {
+                    break;
```
cgillum: Related to my previous comment, could we simplify this by doing `return await this.HandleGetStatusRequestAsync(request, instanceId);` here instead of `break`?

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -34,48 +36,67 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
+                
+                var elapsed = stopwatch.Elapsed;
+                if (elapsed < timeout)
+                {
+                    var remainingTime = timeout.Subtract(elapsed);
+                    await Task.Delay(remainingTime > retryInterval ? retryInterval : remainingTime);
+                }
+                else
+                {
+                    break;
```
cgillum: Related to my previous comments, can we simplify this by doing `return this.CreateCheckStatusResponseMessage(...);` here instead of `break`?

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -34,48 +36,67 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
+                }
+            }
+            stopwatch.Stop();
```
cgillum: If we implement my previous suggestions, would that allow us to delete all the rest of the code in this method?

test/DurableOrchestrationClientMock.cs
```diff
@@ -0,0 +1,65 @@
...
+using DurableTask.Core;
+using Microsoft.Azure.WebJobs;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+using Newtonsoft.Json;
+
+namespace WebJobs.Extensions.DurableTask.Tests
+{
+    public class DurableOrchestrationClientMock : DurableOrchestrationClient
+    {
+
```
cgillum: nit: can we remove all these extra newlines?

test/DurableOrchestrationClientMock.cs
```diff
@@ -0,0 +1,65 @@
...
+        internal DurableOrchestrationClientMock(IOrchestrationServiceClient serviceClient, DurableTaskExtension config, OrchestrationClientAttribute attribute, EndToEndTraceHelper traceHelper) : base(serviceClient, config, attribute, traceHelper)
+        {
+        }
+
+        public int Counter { get; set; }
+
+
+        public override async Task<DurableOrchestrationStatus> GetStatusAsync(string instanceId)
+        {
+            var runtimeStatus = OrchestrationRuntimeStatus.Running; ;
```
cgillum: nit: extra `;` at the end.

test/HttpApiHandlerTests.cs
```diff
@@ -0,0 +1,123 @@
...
+            Assert.Equal(status["terminatePostUri"], "http://localhost:7071/admin/extensions/DurableTaskExtension/instances/7b59154ae666471993659902ed0ba742/terminate?reason={text}&taskHub=SampleHubVS&connection=Storage&code=mykey");
+        }
+
+        [Fact]
+
+        public async Task CreateCheckStatusResponse_Returns_HTTP_202_Response_After_Timeout()
+        {
+            var httpApiHandler = new HttpApiHandler(new DurableTaskExtensionMock() { NotificationUrl = new Uri(TestConstants.NotificationUrl) }, null);
+            var stopWatch = Stopwatch.StartNew();
+            var httpResponseMessage = await httpApiHandler.CreateCheckStatusResponse(new HttpRequestMessage() { RequestUri = new Uri(TestConstants.RequestUri) }, TestConstants.RandomInstanceId, new OrchestrationClientAttribute() { TaskHub = TestConstants.TaskHub, ConnectionName = TestConstants.ConnectionName }, TimeSpan.FromSeconds(100), TimeSpan.FromSeconds(10));
```
cgillum: Let's break this into multiple lines to avoid so much horizontal scrolling.

gled4er:
@cgillum  Thank you for the great feedback.  I will fix the issues as soon as possible. Thank you!

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
gled4er: Great point!

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
gled4er: Yes, simpler code, easier to understand.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
gled4er: Sure

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
gled4er: Great

test/DurableOrchestrationClientMock.cs
gled4er: Sure

test/DurableOrchestrationClientMock.cs
gled4er: Yes, great catch

test/HttpApiHandlerTests.cs
gled4er: Nice!

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -472,6 +473,23 @@ internal string GetIntputOutputTrace(string rawInputOutputData)
             return this.httpApiHandler.CreateCheckStatusResponse(request, instanceId, attribute);
```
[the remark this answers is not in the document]
gled4er: @cgillum maybe I did not get completely your idea here but this change will require to change the signature of the method to **async Task<...>**:
```csharp
        internal async Task<HttpResponseMessage> CreateCheckStatusResponse(
            HttpRequestMessage request,
            string instanceId,
            OrchestrationClientAttribute attribute)
        {
            return await this.CreateCheckStatusResponse(request, instanceId, attribute, TimeSpan.MaxValue, TimeSpan.MaxValue);
        }
```
 that will require to change the DruableOrchestrationClient's method CreateCheckStatusResponse signature to **async Task<...>** signature. Or in other words we will have a breaking change in the API.

gled4er:
@cgillum I pushed changes for based on your last comments with the exception of your first comment. Let me know what should be the best way to address it and I will make the changed very quickly. Thank you!

gled4er:
@cgillum Thank you very much for your patience! I learnt a lot from the review and the I hope next time I will save you several iterations during the review. Great idea about the documentation and the blog. I will work on that next. It was great working on this feature with you! I am looking forward to the next time we will have chance to work together!

gled4er:
@cgillum I found a repetition in the comments for the method overload and also while I was writing the documentation I realized that we don't have samples for the new overload that I can link to. I added simple extension of the existing sample calling the new method both to the script and precompiled versions. Let me know if it is acceptable to add this changes now or if it is too late I will revert them. Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -42,12 +42,26 @@ public class DurableOrchestrationClient
...
+        /// <summary>
+        /// Creates an HTTP response for checking the status of the specified instance supporting synchronous response as well. 
+        /// </summary>
+        /// <param name="request">The HTTP request that triggered the current function.</param>
+        /// <param name="instanceId">The unique ID of the instance to check.</param>
+        /// <param name="timeout ">Total allowed timeout for output from the durable function.</param>
+        /// <param name="timeout">Total allowed timeout for output from the durable function.</param>
+        /// <param name="retryInterval">Timeout between checks for output from the durable function. The default value is 5 seconds.</param>
+        /// <returns>An HTTP response which may include a 202 and location header or a 200 with the durable function output in the response body.</returns>
+        public async Task<HttpResponseMessage> CreateCheckStatusResponse(HttpRequestMessage request, string instanceId, TimeSpan timeout, TimeSpan? retryInterval)
```
cgillum: After seeing the sample you wrote, I'm just now realizing that this method name is a bit misleading. If the orchestration completes or fails before the timeout expires, then it won't be a "check-status" response, but rather an output response. I feel we should rename this method both here and in the `DurableTaskExtension` class. What about `WaitForCompletionOrCreateCheckStatusResponse`?  The "wait for completion" part is important to communicate that this API may not return immediately.

samples/csx/HttpSyncStart/run.csx
```diff
@@ -0,0 +1,24 @@
...
+    // Function input comes from the request content.
+    dynamic eventData = await req.Content.ReadAsAsync<object>();
+    string instanceId = await starter.StartNewAsync(functionName, eventData);
+    
+    log.Info($"Started orchestration with ID = '{instanceId}'.");
+    
+    return await starter.CreateCheckStatusResponse(
+                req,
+                instanceId,
+                TimeSpan.FromSeconds(2),
```
cgillum: Let's not hardcode the timeout and retry intervals. Instead, let's make them optional query string parameters. That will make this start function more generic. For example, I would imagine someone could do the following:

```http
POST /orchestrators/myfunction/wait?timeout=10&retryInterval=1
```

The `timeout` and `retryInterval` parameters will be interpreted in seconds. If they are not specified, then we can have default values. If we want to support sub-second values, then we could allow them to use floating point numbers, like `retryInterval=0.5`.

For the default values, I would suggest making the timeout at least 10 seconds and the retry interval at least 1 second. Retrying too rapidly could create a lot of I/O overhead on the storage account. I also don't expect that orchestrations will normally complete that quickly (though I could be wrong - feel free to disagree with this one).

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -42,12 +42,25 @@ public class DurableOrchestrationClient
...
 
+        /// <summary>
+        /// Creates an HTTP response for checking the status of the specified instance supporting synchronous response as well. 
+        /// </summary>
+        /// <param name="request">The HTTP request that triggered the current function.</param>
+        /// <param name="instanceId">The unique ID of the instance to check.</param>
+        /// <param name="timeout">Total allowed timeout for output from the durable function.</param>
+        /// <param name="retryInterval">Timeout between checks for output from the durable function. The default value is 5 seconds.</param>
+        /// <returns>An HTTP response which may include a 202 and location header or a 200 with the durable function output in the response body.</returns>
+        public async Task<HttpResponseMessage> CreateCheckStatusResponse(HttpRequestMessage request, string instanceId, TimeSpan timeout, TimeSpan? retryInterval)
```
cgillum: It just occurred to me from reading the samples that this method name can be misleading. I suggest an alternative name that better describes the behavior, like `WaitForCompletionOrCreateCheckStatusResponseAsync`. The important behavior to communicate in the method name is that the method call may not complete immediately (unlike `CreateCheckStatusResponse`, which always completes immediately). I would suggest making that name change both in DurableOrchestrationClient and in DurableTaskExtension.

samples/csx/HttpSyncStart/function.json
```diff
@@ -0,0 +1,23 @@
+{
+  "bindings": [
+    {
+      "authLevel": "anonymous",
+      "name": "req",
+      "type": "httpTrigger",
+      "direction": "in",
+      "route": "syncorchestrators/{functionName}",
```
cgillum: I wouldn't change the base of the route. `syncorchestrators` makes me think that it's a different kind of orchestrator function, which it's not. Instead, I would add a suffix to the existing one. Something like `"route": "orchestrators/{functionName}/wait"`.

gled4er: Good point!

samples/csx/HttpSyncStart/run.csx
gled4er: Actually I was thinking the same yesterday. It will be great if we can define the timeout details with every call instead being driven by the server side. Great change!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
gled4er: I like when the methods and variables are self-descriptive and the dev does not need to go and study the documentation in details. I will work on that! Thank you for the great feedback!

gled4er:
@cgillum I added the requested changes. Please let me know what you think and what should be improved. Thank you!

samples/precompiled/HttpSyncStart.cs
```diff
@@ -24,11 +24,11 @@ public static class HttpSyncStart
 
             log.Info($"Started orchestration with ID = '{instanceId}'.");
 
-            return await starter.CreateCheckStatusResponse(
+            return await starter.WaitForCompletionOrCreateCheckStatusResponseAsync(
                 req,
-                instanceId,
-                TimeSpan.FromSeconds(2),
-                TimeSpan.FromMilliseconds(50));
```
cgillum: In my previous feedback, I suggested having values that come from the query string, but that feedback was only intended to apply to the sample code. This change pulls the timeout values from the request object in _all_ cases, which was not my intention. I feel that is too limiting for users who either want a different way of specifying these values or don't want the caller to specify them at all.

I would prefer if `timeout` and `retryInterval` remained as explicit parameters. Only the `HttpSyncStart` sample should look at the request query string values.

gled4er: Hello @cgillum ,

Thank you very much for the clarification. I will work on the changes and submit them as soon as possible.

Thank you!

gled4er:
@cgillum  I added more changes related to the way we pass the timeout and retryInterval. The samples are also updated. I decided to use more DIY method for the query string parameters in order to have identical samples for the pre-compiled and script sample projects. Let me know what should be optimized. Thank you!

samples/csx/HttpSyncStart/run.csx
```diff
@@ -15,8 +20,33 @@ public static async Task<HttpResponseMessage> Run(
...
-        instanceId);
+
+    //Get query string parameters and add them to a dictionary
+    var queryPairs = req.RequestUri.Query.Replace("?", string.Empty)
+        .Split(new[] {"&"}, StringSplitOptions.None);
+    var parameters = queryPairs.Select(
+        item => item.Split(new[] {"="}, StringSplitOptions.None))
+        .Where(queryPairElementsArray => queryPairElementsArray.Length > 1).ToDictionary(
+        queryPairElementsArray => queryPairElementsArray[0],
+        queryPairElementsArray => queryPairElementsArray[1]);
```
cgillum: I don't feel good about the DIY method of query string parameter parsing. I worry that this complexity distracts from the point of this sample, which is how to start an orchestrator function synchronously. The built-in Azure Functions HTTP trigger samples use `GetQueryNameValuePairs()`, and I'd prefer we do the same here. I think it's okay to diverge the implementation between csx and precompiled if necessary.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -34,48 +41,53 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
+            return this.CreateCheckStatusResponseMessage(request, instanceId, statusQueryGetUri, sendEventPostUri, terminatePostUri);
+        }
 
-            string taskHub = WebUtility.UrlEncode(attribute.TaskHub ?? config.HubName);
-            string connection = WebUtility.UrlEncode(attribute.ConnectionName ?? config.AzureStorageConnectionStringName ?? ConnectionStringNames.Storage);
 
-            string querySuffix = $"{TaskHubParameter}={taskHub}&{ConnectionParameter}={connection}";
-            if (!string.IsNullOrEmpty(notificationUri.Query))
+        internal async Task<HttpResponseMessage> WaitForCompletionOrCreateCheckStatusResponseAsync(
+            HttpRequestMessage request,
```
cgillum: It looks like there are a variety of indent issues in this code. For example, these parameters are only indented by 3 spaces when elsewhere they are indented by 4. Can you fix those?

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -295,5 +306,58 @@ private DurableOrchestrationClient GetClient(HttpRequestMessage request)
 
             return this.config.GetClient(attribute);
         }
+
+        private void GetClientResponseLinks(
+           HttpRequestMessage request,
```
cgillum: More indenting issues here.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -267,25 +278,25 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
         {
-            string taskHub = null;
+            string taskHub = null; 
             string connectionName = null;
 
             var pairs = request.GetQueryNameValuePairs();
             foreach (var key in pairs.AllKeys)
             {
-                if (taskHub == null 
+              if (taskHub == null
```
cgillum: More indenting issues here.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -242,14 +254,13 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
-            if (!string.IsNullOrEmpty(mediaType) && 
-                !string.Equals(mediaType, "application/json", StringComparison.OrdinalIgnoreCase))
+            var mediaType = request.Content.Headers.ContentType?.MediaType;
+            if (!string.Equals(mediaType, "application/json", StringComparison.OrdinalIgnoreCase))
             {
                 return request.CreateErrorResponse(HttpStatusCode.BadRequest, "Only application/json request content is supported");
             }
 
-            string stringData = await request.Content.ReadAsStringAsync();
+            var stringData = await request.Content.ReadAsStringAsync();
```
cgillum: I prefer not using `var` in place of primitive type names. I generally only use it when the type is already specified in the right-hand expression - e.g. `var x = new Foo()`. Using `var` carefully is good to ensure that the code is readable and as easy to understand as possible.

test/DurableOrchestrationClientMock.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System;
```
cgillum: All source files need to include the copyright header.

test/DurableTaskExtensionMock.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
```
cgillum: All source files need to include the copyright header.

test/HttpApiHandlerTests.cs
```diff
@@ -0,0 +1,194 @@
+﻿using System;
```
cgillum: All source files need to include the copyright header.

test/TestConstants.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace WebJobs.Extensions.DurableTask.Tests
```
cgillum: All source files need to include the copyright header.

gled4er:
@cgillum Thank you for the feedback. I will start working on the issues.

samples/csx/HttpSyncStart/run.csx
gled4er: Great!

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
gled4er: I will fix these issues.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
gled4er: Got it. I will follow this approach from now on.

test/DurableOrchestrationClientMock.cs
gled4er: Great point. I missed this element.

test/DurableTaskExtensionMock.cs
gled4er: I will add it shortly.

test/TestConstants.cs
gled4er: I will add it shortly.  Thank you!

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
gled4er: Great catch!

samples/csx/HttpSyncStart/run.csx
gled4er: As we agreed I will set the values explicitly in the source code before the next updates.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
gled4er: I went over the code in this class and I think that most of the parameters in the existing methods are intended by 3 tabs so I followed this approach. Let me know if you prefer another one.

gled4er:
Hello @cgillum,

I pushed new changes to the pull request. 

Please review when you have time and let me know what needs to be fixed.

Thank you!

samples/precompiled/HttpSyncStart.cs
```diff
@@ -2,6 +2,7 @@
 using System.Linq;
 using System.Net.Http;
 using System.Threading.Tasks;
+using ImpromptuInterface.Dynamic;
```
cgillum: Why is ImpromptuInterface.Dynamic needed here?

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -47,11 +41,11 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
 
 
         internal async Task<HttpResponseMessage> WaitForCompletionOrCreateCheckStatusResponseAsync(
-            HttpRequestMessage request,
-            string instanceId,
-            OrchestrationClientAttribute attribute,
-            TimeSpan timeout,
-            TimeSpan retryInterval)
+             HttpRequestMessage request,
```
cgillum: Extra whitespace padding was added to these parameters? It was already 4 spaces and now it is 5.

samples/precompiled/HttpSyncStart.cs
gled4er: Great catch! This was left my testing for another approach. I will remove it shortly.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
gled4er: Thank you for noticing that. I will fix it.

gled4er:
Hello @cgillum,

I pushed changes for the last issues you find in the pull request.

Let me know if I need to fix something else.

Thank you!

PR #116, opened 2017-12-22
Adding abstract classes

gled4er:
Hello,

This pull request is for https://github.com/Azure/azure-functions-durable-extension/issues/64 

This pull request is not ready yet and its goal is to start the discussion for the correct implementation of the code for this issue. 

Thank you!

Issue #64, opened 2017-09-20
Create interfaces for some classes to allow unit testing.


SimonLuckenuik:
Hi,

The following classes would benefit from implementing an interface: Microsoft.Azure.WebJobs.DurableOrchestrationClient, Microsoft.Azure.WebJobs.DurableOrchestrationContext,  Microsoft.Azure.WebJobs.DurableActivityContext

There is no way currently that I can unit test any piece of code using those classes, considering that the ctor is also internal. Creating a simple IDurableOrchestrationClient would enable us to test properly. Currently, I am forced to create a decorator that copies all the public methods  and wrappes the instance in order to do unit testing.

Thanks!
Simon

cgillum:
I'm personally not a big fan of interfaces because of the versioning challenges they create. Would abstract base classes work for you or does your unit testing mechanism require interfaces?

SimonLuckenuik:
I do not understand how they create versioning challenge, considering that you currently expose the full class anyway, and that you would need to prevent breaking changes. Could you elaborate on that? 
In theory, if your interface is aligned with your class public members implementation, it should not be an issue. 

Abstract classes would involve having most your methods virtual, which in my view makes it more complicated. And considering that those classes are currently sealed I believe it would break you design for the wrong reason.

The goal here would be to allow us to choose to rely either on concrete classes or interfaces for dependency injection.

cgillum:
The versioning challenge I was referring to is that if we want to add a new method to an existing interface sometime in the future, that will be a breaking change for any existing implementations (including yours) since all implementations are required to implement all methods. Abstract classes don't have this problem because they allow us to create a default implementations and can therefore be versioned more easily. This is based on my experience working in large .NET framework libraries.

Either way, this is a reasonable request and we'll look into it.

SimonLuckenuik:
Excellent! 
Considering that we are using mocking of interfaces, I guess it won't be an issue in our case. 

You could also consider having different interfaces with less responsibilities which would also minimize the impact of the issue you are referring to.

cgillum:
This has been resolved using abstract base classes. See the linked PR for details. See the new sample test project for an example of how to unit test an orchestrator function using [Moq](https://github.com/Moq/moq4/): https://github.com/Azure/azure-functions-durable-extension/blob/dev/samples/VSSample.Tests/HelloSequenceTests.cs

Thanks to @gled4er for this contribution!

gled4er:
[...]

PR #117, opened 2017-12-29
New approach for enabling unit testing using virtual methods

gled4er:
Hello @cgillum 

I am opening this new pull request based on your feedback provided in previous pull request - https://github.com/Azure/azure-functions-durable-extension/pull/116

This pull request is work in progress for implementing issue - https://github.com/Azure/azure-functions-durable-extension/issues/64

I changed the public methods to virtual and modified the access levels for few constructors and few classes.

**I am very interested in your opinion especially for these changes**:
- enum FunctionType from **internal** to **public**
- class EndToEndTraceHelper from **internal** to **public**
- removing **sealed** from class **DurableOrchestrationContext**

My idea with the changes in this pull request is to enable developers to implement the following mocks:

```csharp

 public class DurableOrchestrationClientMock : DurableOrchestrationClient
    {
        public DurableOrchestrationClientMock(IOrchestrationServiceClient serviceClient, DurableTaskExtension config,
            OrchestrationClientAttribute attribute, EndToEndTraceHelper traceHelper) : base(serviceClient, config,
            attribute, traceHelper)
        {

        }

        public override async Task<DurableOrchestrationStatus> GetStatusAsync(string instanceId)
        {
            await Task.Delay(5);
            return new DurableOrchestrationStatus();
        }
    }


    public class DurableActivityContextMock : DurableActivityContext
    {
        public DurableActivityContextMock(string instanceId, string serializedInput) : base(instanceId, serializedInput)
        {
           
        }

        public override string InstanceId => "123";

        public override string GetRawInput()
        {
            return "Raw Input";
        }
    }

    public class DurableOrchestrationContextMock : DurableOrchestrationContext
    {
        public DurableOrchestrationContextMock(
            DurableTaskExtension config,
            string functionName,
            string functionVersion) : base(config, functionName, functionVersion)
        {
          
        }

        public override DateTime CurrentUtcDateTime => new DateTime();

        public override Task CallActivityAsync(string functionName, object input)
        {
            return Task.Run(() => "Hello World!");
        }
    }

```

Thank you!


src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -105,7 +105,7 @@ internal void AssignToCurrentThread()
         /// <returns>
         /// The raw JSON-formatted orchestrator function input.
         /// </returns>
-        public string GetRawInput()
+        public virtual string GetRawInput()
```
cgillum: Let's avoid exposing `GetRawInput`. I don't think it's a common scenario.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -117,7 +117,7 @@ public string GetRawInput()
         /// <returns>
         /// The parsed <c>JToken</c> representation of the orchestrator function input.
         /// </returns>
-        public JToken GetInputAsJson()
+        public virtual JToken GetInputAsJson()
```
cgillum: Let's avoid exposing `GetInputAsJson`. I don't think it is a common scenario.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -80,7 +80,7 @@ public sealed class DurableOrchestrationContext
         /// <value>
         /// <c>true</c> if the orchestrator function is currently being replayed; otherwise <c>false</c>.
         /// </value>
-        public bool IsReplaying => this.innerContext.IsReplaying;
+        public virtual bool IsReplaying => this.innerContext.IsReplaying;
```
cgillum: Let's move `IsReplaying` into the abstract base class as a virtual property with a default value of `false`.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -57,7 +57,7 @@ public HttpResponseMessage CreateCheckStatusResponse(HttpRequestMessage request,
         /// <exception cref="ArgumentException">
         /// The specified function does not exist, is disabled, or is not an orchestrator function.
         /// </exception>
-        public Task<string> StartNewAsync(string orchestratorFunctionName, object input)
+        public virtual Task<string> StartNewAsync(string orchestratorFunctionName, object input)
```
cgillum: Since this overload just calls the other `StartNewAsync` overload, let's move it into the base class and make it virtual. The more complex overload can be made abstract.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -99,7 +99,7 @@ public Task<string> StartNewAsync(string orchestratorFunctionName, object input)
         /// <param name="eventData">The JSON-serializeable data associated with the event.</param>
         /// <returns>A task that completes when the event notification message has been enqueued.</returns>
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1030:UseEventsWhereAppropriate")]
-        public async Task RaiseEventAsync(string instanceId, string eventName, object eventData)
+        public virtual async Task RaiseEventAsync(string instanceId, string eventName, object eventData)
```
cgillum: Let's keep the implementation of `RaiseEventAsync` here but create an abstract method in a base class.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -131,7 +131,7 @@ public Task<string> StartNewAsync(string orchestratorFunctionName, object input)
         /// <param name="instanceId">The ID of the orchestration instance to terminate.</param>
         /// <param name="reason">The reason for terminating the orchestration instance.</param>
         /// <returns>A task that completes when the terminate message is enqueued.</returns>
-        public async Task TerminateAsync(string instanceId, string reason)
+        public virtual async Task TerminateAsync(string instanceId, string reason)
```
cgillum: Let's keep the implementation of `TerminateAsync` here but create a corresponding abstract method in the abstract base class.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -149,7 +149,7 @@ public Task<string> StartNewAsync(string orchestratorFunctionName, object input)
         /// </summary>
         /// <param name="instanceId">The ID of the orchestration instance to query.</param>
         /// <returns>Returns a task which completes when the status has been fetched.</returns>
-        public async Task<DurableOrchestrationStatus> GetStatusAsync(string instanceId)
+        public virtual async Task<DurableOrchestrationStatus> GetStatusAsync(string instanceId)
```
cgillum: Let's leave the implementation of `GetStatusAsync` here but create a corresponding abstract method in the abstract base class.

gled4er:
Hello @cgillum ,

Thank you for the detailed feedback.

I will work on the comments you gave me.

Thank you!

gled4er:
Hello @cgillum,

I followed your recommendations and I added two new base abstract classes. I wanted to clarify how you prefer to manage the comments for the methods that are included in both the abstract class and the derived class. Currently I have them in both places but I think in order to minimize the risk of going out of sync we should have them in one class only - for example only in the abstract class. What do you think? 

Please let me know if you see other issues that need to be fixed.

Thank you!

cgillum:
Excellent, thank you! For comments, I think the best way to minimize duplication between the parent and child class is to use `/// <inheritdoc />`.  For example:

```csharp
/// <inheritdoc />
public override HttpResponseMessage CreateCheckStatusResponse(HttpRequestMessage request, string instanceId)
{
    return this.config.CreateCheckStatusResponse(request, instanceId, this.attribute);
}
```

The API documentation builder will know to get the documentation from the base class.

gled4er:
Great! I will do that. Thank you for the quick response.

cgillum:
I quickly reviewed these changes and I think they look great. Can you add a few tests that demonstrate how they can be used?

gled4er:
Great idea! 

Sure, I will work on that next.

Thank you!

gled4er:
Hello @cgillum ,

I added a test project and 2 very simple tests. I mocked `DurableOrchestrationContext` for now. I wanted to hear your opinion about the Testing project. Now I am using the default Visual Studio Test Project but in the main project you use xUnit based one. What type of project do you prefer to use for the sample tests - class library or console app that is modified to support xUnit or another option? 

The other question is about the approach I am using. I changed the type of the parameter we use in the sample from `DurableOrchestrationContext` to `DurableOrchestrationContextBase` so I can mock an object and pass it to the function. Let me know if you have another idea for illustrating the use better:

```csharp
            [OrchestrationTrigger] DurableOrchestrationContextBase context)
```

Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClientBase.cs
```diff
@@ -0,0 +1,81 @@
+﻿using System;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
```
cgillum: >Microsoft.Azure.WebJobs.Extensions.DurableTask [](start = 10, length = 46)

Since this is a public type, let's change the namespace to Microsoft.Azure.WebJobs. I caught this when looking at your change to the sample.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContextBase.cs
```diff
@@ -0,0 +1,364 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
```
cgillum: >Microsoft.Azure.WebJobs.Extensions.DurableTask [](start = 10, length = 46)

Let's change this to Microsoft.Azure.WebJobs.

samples/VSSample.Tests/VSSample.Tests.csproj
```diff
@@ -0,0 +1,191 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
cgillum: >Project [](start = 1, length = 7)

It looks like you're using the legacy VS project file format. Can you please use the new .NET Standard project system, like we do with the other projects?

samples/VSSample.Tests/packages.config
```diff
@@ -0,0 +1,68 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
```
cgillum: >packages [](start = 1, length = 8)

This file should not be needed when you convert to the new .NET Standard project system.

samples/VSSample.Tests/VSSample.Tests.csproj
```diff
@@ -0,0 +1,191 @@
...
+  <Import Project="..\precompiled\packages\MSTest.TestAdapter.1.2.0\build\net45\MSTest.TestAdapter.props" Condition="Exists('..\precompiled\packages\MSTest.TestAdapter.1.2.0\build\net45\MSTest.TestAdapter.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{7FD2908F-B6BD-442D-B735-CDA61766D3BD}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>VSSample.Tests</RootNamespace>
+    <AssemblyName>VSSample.Tests</AssemblyName>
+    <TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>
```
cgillum: >TargetFrameworkVersion [](start = 5, length = 22)

Let's target .NET Standard for the target framework version, like we do for the unit test project.

samples/precompiled/HelloSequence.cs
```diff
@@ -4,14 +4,15 @@
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
```
cgillum: This new using statement should not be necessary.

samples/VSSample.Tests/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,20 @@
+using System.Reflection;
```
cgillum: Let's not add AssemblyInfo.cs for this project. I don't believe it is normally included when using the new .NET Standard project system.

cgillum:
> using System;

New source files should include the following header:

---
Refers to: src/WebJobs.Extensions.DurableTask/DurableOrchestrationClientBase.cs:1 in 20439bd. [](commit_id = 20439bd19a5fe1106698542cd72ea69678426ad6, deletion_comment = False)

cgillum:
> using System;

New source files should include the following header:

---
Refers to: src/WebJobs.Extensions.DurableTask/DurableOrchestrationContextBase.cs:1 in 20439bd. [](commit_id = 20439bd19a5fe1106698542cd72ea69678426ad6, deletion_comment = False)

gled4er:
Hello Chris,

Thank you very much for the detailed feedback.

I will fix all issues you commented.

Thank you!


Kanio

Sent from my Windows 10 phone

test/DurableOrchestrationClientBaseTests.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            var instanceId = Guid.NewGuid().ToString();
+            const string functionName = "sampleFunction";
+            var durableOrchestrationClientBaseMock = new Mock<DurableOrchestrationClientBase> {CallBase = true};
+            durableOrchestrationClientBaseMock.Setup(x => x.StartNewAsync(functionName, string.Empty, null)).ReturnsAsync(instanceId);
+
+            var result = await durableOrchestrationClientBaseMock.Object.StartNewAsync(functionName, null);
+            result.Should().Be(instanceId);
+        }
+    }
```
cgillum: Should we add tests for the other abstract methods on `DurableOrchestrationClientBase` as well?

test/DurableOrchestrationClientBaseTests.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
```
cgillum: These new test files will need to copyright headers as well.

gled4er: I will add it. Thank you for catching that.

test/DurableOrchestrationClientBaseTests.cs
gled4er: Yes, I will look into that.

gled4er:
Hello @cgillum,

Great idea to add StyleCop! 

Of course I will fix the warning. 

Thank you for the review remarks.

gled4er:
Hello @cgillum ,

I started to add tests for DurableOrchestrationClient and the first two were ok. However I needed to modify the access to two method to protected:
```csharp
protected internal virtual HttpResponseMessage CreateCheckStatusResponse( ...

protected internal virtual async Task<HttpResponseMessage> WaitForCompletionOrCreateCheckStatusResponseAsync( ...

```
Please let me know if this is acceptable.

I wanted to apply the same approach to the rest methods in the `DurableOrchestrationClient` class but they depend on ` private readonly TaskHubClient client;`  which is initialized in the constructor  ` this.client = new TaskHubClient(serviceClient);`  . 

Currently I am thinking how to mock the  ` TaskHubClient`  variable but due to its private access level and the sealed definition of  ` TaskHubClient` I think this will be difficult. 

Please let me know if you see better approach or workaround. I think it will be great if we can abstract the initialization of  `client` via wrapping class but this will require further changes in the code and I want to align the vision with you first.

Thank you!

samples/VSSample.Tests/HttpSyncStartTests.cs
```diff
@@ -0,0 +1,77 @@
...
+using FluentAssertions;
+using Microsoft.Azure.WebJobs;
+using Microsoft.Azure.WebJobs.Host;
+using Newtonsoft.Json;
+
+namespace VSSample.Tests
+{
+    public class HttpSyncStartTests
+    {
+        private const string FuncitonName = "SampleFunction";
```
cgillum: Typo. `FuncitonName` should be `FunctionName`

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -491,8 +491,9 @@ internal string GetIntputOutputTrace(string rawInputOutputData)
             }
         }
 
-        // Get a response that will point to our webhook handler.
-        internal HttpResponseMessage CreateCheckStatusResponse(
+
+        // Get a response that will point to our webhook handler. 
+        protected internal virtual HttpResponseMessage CreateCheckStatusResponse(
```
cgillum: Why was it necessary to make this `protected` and `virtual`? I can't think of a good reason to expose anything in  `DurableTaskExtension`.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -505,9 +506,9 @@ internal string GetIntputOutputTrace(string rawInputOutputData)
             return this.httpApiHandler.CreateCheckStatusResponse(request, instanceId, attribute);
         }
 
-        // Get a response that will wait for response from the durable function for predefined period of time before
-        // pointing to our webhook handler.
-        internal async Task<HttpResponseMessage> WaitForCompletionOrCreateCheckStatusResponseAsync(
+        // Get a response that will wait for response from the durable function for predefined period of time before 
+        // pointing to our webhook handler. 
+        protected internal virtual async Task<HttpResponseMessage> WaitForCompletionOrCreateCheckStatusResponseAsync(
```
cgillum: Same comment as previous.

test/DurableOrchestrationClientTests.cs
```diff
@@ -0,0 +1,52 @@
...
+using DurableTask.Core;
+using FluentAssertions;
+using Microsoft.Azure.WebJobs;
+using Microsoft.Azure.WebJobs.Host;
+using Moq;
+using Xunit;
+
+namespace WebJobs.Extensions.DurableTask.Tests
+{
+    public class DurableOrchestrationClientTests
```
cgillum: I don't think it's necessary to add unit testing for `DurableOrchestrationClient`. I imagine we already have good coverage elsewhere, and I feel a little uncomfortable about mocking some of the internal implementation details, like `EndToEndTraceHelper` and `DurableTaskExtension`. Can we leave these changes out of the PR for now? The main goal is to enable developers to write mocks for the public surface area, and I think your other changes have already achieved and tested this.

gled4er:
Hello @cgillum ,

Thank you for the detailed feedback. 

I will revert the changes around `DurableOrchestrationClient`. 

I needed to confirm first with you before making the changes. 

When you mentioned in a previous comment - **"Should we add tests for the other abstract methods on `DurableOrchestrationClientBase` as well?"**,  what did you have in mind?  Initially I added unit tests only for the methods in the abstract class that were calling overloads to ensure that we have these changes tested. So it will be great if you give me few more hints how to test the pure abstract methods in the abstract classes inside the main project. 

Thank you very much!

samples/VSSample.Tests/HttpSyncStartTests.cs
gled4er: Thank you for catching that!

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
gled4er: This was done in order to override the behavior in the mock class. I will revert these changes based on your feedback. Thank you!

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
gled4er: I will revert these changes based on your feedback. Thank you!

test/DurableOrchestrationClientTests.cs
gled4er: Thank you for the detailed feedback here. I will revert these changes based on your feedback. Thank you!

cgillum:
I apologize, I think I misunderstood the intent of your tests. I understand now that you were specifically testing abstract methods which were calling other overloads. I was thinking we wanted to mock the pure abstract methods as well to ensure that any breaking changes to these abstract APIs would be caught by the tests. It may be a bit overkill, so we don't have to add any more tests than what you have.

If you can revert the latest changes around `DurableOrchestrationClient`, then I think I can sign off on this PR.

gled4er:
Hello @cgillum ,

Thank you for the quick reply!

I am already working on reverting the changes.

Thank you!

gled4er:
Hello @cgillum ,

I reverted the test changes around `DurableOrchestrationClient'. Style Cop warnings are also addressed.

May I ask you to review again and let me know what should be changed next?

Thank you!

gled4er:
Hello @cgillum ,

Thank you for the great feedback and guidance! 

I will be happy to help with the documentation. Let me know if you have some suggestions in terms of location in the existing documentation pages and the depth of the guidance. 

Maybe we can use the samples' code as foundation and provide explanations around it?

Thank you!

cgillum:
Actually, before we merge, can you verify that the `E1_HelloSequence` sample can still run when using `DurableOrchestrationContextBase` instead of `DurableOrchestrationContext`? I have a feeling it might not and that another small change may be necessary.

gled4er:
Hello @cgillum ,

It works well. I tested if both from `HttpSyncStart` and from `HttpSyncStart`. 

I am not able to trace it exactly at the moment but I think that `DurableOrchestrationClientBase starter` is assigned behind the scenes to instance of `DurableOrchestrationClient` so the code is actually acting as IoC container if that case. 

Let me know if you want me to look deeper into it.

Thank you!

PR #121, opened 2018-01-03
Support cross-language return values for orchestration triggers

### LINKED ISSUE #120: Durable Orchestration Client/Triggers aren't part of generated function.json in VS2017; execution fails
When using the Visual Studio 2017 Functions Tools, creating an Azure Functions v2 Durable Function results in a failure at runtime due to a malformed function.json file which doesn't contain the correct binding information for the `OrchestrationTrigger` or `OrchestrationClient` objects in the function signature
#### Repro steps

Provide the steps required to reproduce the problem
1. Create a new Azure Function v2 Project in VS2017.15.5.2 with Azure Functions and Web Jobs Tools v15.0.31201.0
1. Select HTTPTrigger as the function type
1. Update the Webjobs nuget package to v1.0.7
1. Add the [Microsoft.Azure.WebJobs.extensions.DurableTask v1.1.0-beta2](https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Extensions.DurableTask/1.1.0-beta2) nuget package to the project
1. Change the code for the Function created from step 2 to:
```csharp
[FunctionName("Function1")]
public static async Task<HttpResponseMessage> Run([HttpTrigger(AuthorizationLevel.Function, "post")]HttpRequestMessage req, [OrchestrationClient]DurableOrchestrationClient starter, TraceWriter log)
{
    return new HttpResponseMessage(HttpStatusCode.OK);
}
```
6. Hit F5

#### Expected behavior

The Functions runtime executes in .Net Core (dotnet.exe) and gives me the URL to hit for the created Durable Function.

#### Actual behavior

Functions runtime spits out errors like:
~~~
[1/3/2018 11:21:16 PM] ScriptHost initialization failed
[1/3/2018 11:21:16 PM] System.Private.CoreLib: 'TriggerHandlesReturnValue' property specified was not found.
Listening on http://localhost:7071/
Hit CTRL-C to exit...
Object reference not set to an instance of an object.
~~~

Upon further investigation one can find the generated `function.json` file in bin\debug\netstandard2.0\Function1 and see it's missing the binding for the `OrchestratorClient` object:
```js
{
  "generatedBy": "Microsoft.NET.Sdk.Functions.Generator-1.0.7",
  "configurationSource": "attributes",
  "bindings": [
    {
      "type": "httpTrigger",
      "methods": [
        "post"
      ],
      "authLevel": "function",
      "name": "req"
    }
  ],
  "disabled": false,
  "scriptFile": "../bin/AzureFunctions.Durable.dll",
  "entryPoint": "AzureFunctions.Durable.Orchestrator.Run"
}
```

Changing the Function code to:
```csharp
[FunctionName("Function1")]
public static void Run([OrchestrationTrigger]DurableOrchestrationClient starter, TraceWriter log)
{
    return;
}
```
will generate the exepected values in `function.json` of:
```js
{
  "generatedBy": "Microsoft.NET.Sdk.Functions.Generator-1.0.7",
  "configurationSource": "attributes",
  "bindings": [
    {
      "type": "orchestrationTrigger",
      "name": "starter"
    }
  ],
  "disabled": false,
  "scriptFile": "../bin/FunctionApp1.dll",
  "entryPoint": "FunctionApp1.Function1.Run"
}
```
but the same errors occur during runtime.

brandonh-msft:
This seems like it'll fix #120

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/OrchestrationClientAttribute.cs
+++ b/src/WebJobs.Extensions.DurableTask/OrchestrationClientAttribute.cs
@@ -1,9 +1,9 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-using Microsoft.Azure.WebJobs.Description;
 using System;
 using System.Diagnostics;
+using Microsoft.Azure.WebJobs.Description;
 
 namespace Microsoft.Azure.WebJobs
 {
--- a/src/WebJobs.Extensions.DurableTask/OrchestrationClientAttribute.cs
+++ b/src/WebJobs.Extensions.DurableTask/OrchestrationClientAttribute.cs
@@ -12,11 +12,7 @@ namespace Microsoft.Azure.WebJobs
     /// </summary>
     [AttributeUsage(AttributeTargets.Parameter)]
     [DebuggerDisplay("TaskHub={TaskHub}, ConnectionName={ConnectionName}")]
-#if NETSTANDARD2_0
-    [Binding(TriggerHandlesReturnValue = false)]
-#else
     [Binding]
-#endif
     public sealed class OrchestrationClientAttribute : Attribute, IEquatable<OrchestrationClientAttribute>
     {
         /// <summary>
--- a/src/WebJobs.Extensions.DurableTask/OrchestrationClientAttribute.cs
+++ b/src/WebJobs.Extensions.DurableTask/OrchestrationClientAttribute.cs
@@ -1,10 +1,9 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using Microsoft.Azure.WebJobs.Description;
 using System;
 using System.Diagnostics;
-using Microsoft.Azure.WebJobs.Description;
-using Microsoft.Azure.WebJobs.Extensions.DurableTask;
 
 namespace Microsoft.Azure.WebJobs
 {
@@ -13,7 +12,11 @@ namespace Microsoft.Azure.WebJobs
     /// </summary>
     [AttributeUsage(AttributeTargets.Parameter)]
     [DebuggerDisplay("TaskHub={TaskHub}, ConnectionName={ConnectionName}")]
+#if NETSTANDARD2_0
+    [Binding(TriggerHandlesReturnValue = false)]
+#else
     [Binding]
+#endif
     public sealed class OrchestrationClientAttribute : Attribute, IEquatable<OrchestrationClientAttribute>
     {
         /// <summary>
--- a/src/WebJobs.Extensions.DurableTask/OrchestrationTriggerAttribute.cs
+++ b/src/WebJobs.Extensions.DurableTask/OrchestrationTriggerAttribute.cs
@@ -1,9 +1,9 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using Microsoft.Azure.WebJobs.Description;
 using System;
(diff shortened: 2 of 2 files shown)


src/WebJobs.Extensions.DurableTask/OrchestrationClientAttribute.cs
```diff
@@ -13,7 +12,11 @@ namespace Microsoft.Azure.WebJobs
     /// </summary>
     [AttributeUsage(AttributeTargets.Parameter)]
     [DebuggerDisplay("TaskHub={TaskHub}, ConnectionName={ConnectionName}")]
+#if NETSTANDARD2_0
+    [Binding(TriggerHandlesReturnValue = false)]
```
cgillum: `Binding.TriggerHandlesReturnValue` is only intended for trigger bindings and `OrchestrationClientAttribute` is a non-trigger binding, so I wouldn't expect this change to be necessary.

src/WebJobs.Extensions.DurableTask/OrchestrationTriggerAttribute.cs
```diff
@@ -12,7 +12,11 @@ namespace Microsoft.Azure.WebJobs
     /// </summary>
     [AttributeUsage(AttributeTargets.Parameter)]
     [DebuggerDisplay("{Orchestration} ({Version})")]
+#if NETSTANDARD2_0
+    [Binding(TriggerHandlesReturnValue = false)]
```
cgillum: @kashimiz I recall we added this for `ActivityTriggerAttribute` but was there a reason we didn't originally add this to `OrchestrationTriggerAttribute`?

src/WebJobs.Extensions.DurableTask/OrchestrationClientAttribute.cs
kashimiz: It's not; the runtime won't look for the attribute on a non-trigger binding. This can be safely removed.

src/WebJobs.Extensions.DurableTask/OrchestrationTriggerAttribute.cs
kashimiz: This attribute signals to the function indexer that triggers to which it's applied are meant to handle a return value from non-C# functions. Since we were focused on a regular functions scenario, missing this was probably oversight.

brandonh-msft: Interestingly enough, the associated Issue (#120) happens when I'm using the Client, not the Trigger...

kashimiz: We do in the future want orchestration triggers to handle return values in non-C# scenarios, so this should be true instead of false.

src/WebJobs.Extensions.DurableTask/OrchestrationClientAttribute.cs
```diff
@@ -1,10 +1,9 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using Microsoft.Azure.WebJobs.Description;
```
kashimiz: Minor stylistic quibble: System* using statements should appear before any others.

src/WebJobs.Extensions.DurableTask/OrchestrationTriggerAttribute.cs
```diff
@@ -1,9 +1,9 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using Microsoft.Azure.WebJobs.Description;
```
kashimiz: Same style quibble: keep System* using statements up top.

PR #125, opened 2018-01-05
StyleCop

cgillum:
Because this is an open source project and we have a variety of contributers with different coding styles, I thought it would be a good to introduce StyleCop to help enforce some basic consistency guidelines.

For now, all rules are warnings, but we could escalate them to errors later. Even better if we can simply gate PRs by ensuring they don't contain any warnings so that contributers don't have to fix StyleCop warnings until they are ready to do so.

I'm using the new "Analyzers" feature of Visual Studio, which is apparently the new way to do StyleCop. Hopefully it will work well for us.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
@@ -1,5 +1,5 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the MIT License. See License.txt in the project root for license information.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
 using System.Diagnostics;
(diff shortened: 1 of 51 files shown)


src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -50,6 +52,8 @@ public static string LocalSlotName
             }
         }
 
+#pragma warning disable SA1117 // Parameters should be on same line or separate lines
```
tohling: Why do you need too disable this?

cgillum: It's for the `this.logger.LogInformation(...)` calls. I want to keep the long list of parameters on a single line underneath the template string so that we can easily compare the two and ensure we didn't add a parameter to one and not the other.

Issue #16, opened 2017-07-15
Query Orchestration History


cgillum:
## Summary
The HTTP status query endpoint allows the caller to fetch the current execution status, input, output, and last updated time of a particular orchestration instance. However, there is no way to view the actual execution history of the instance - e.g. which activity functions it called into.

The HTTP status query endpoint should support fetching this history information and formatting it in a useful way. This history can be thought of as the "call stack" of the orchestration function and contain data such as the activities that were called, their relative order, their outputs (or errors), timers, external events received, etc.

## Motivation
There are several use cases for this kind of feature but the primary cases are likely auditing and debugging failed or "stuck" orchestrations. In the latter case, an orchestrator might be "stuck" waiting for some external event that never arrived. It would be useful for an engineer to see which event it is currently waiting on as part of this execution history.

## Design Notes
This should be built into the existing HTTP status query API handler. Because the response could be potentially very large for large orchestrations, it should only be included in the response when explicitly asked for via a query string parameter, e.g. `?showHistory=true`.

Most of the history information is available in the history table and is surfaced up via DTFx APIs. Some aggregation of this history data will be required to create a more nicely formatted output.

SimonLuckenuik:
That would help troubleshooting instead of relying on the Table Storage directly (and create a dependency on internal implementation).

cgillum:
Yes, absolutely.

gled4er:
Hello @cgillum ,

I wanted to get your opinion before I start the implementation. 

I think the following format will be a good starting point for the query orchestration history:

```javascript
{
    "createdTime": "2018-01-23T05:26:37Z",
    "input": null,
    "lastUpdatedTime": "2018-01-23T05:26:58Z",
    "output": [
        "Hello Tokyo!",
        "Hello Seattle!",
        "Hello London!"
    ],
    "runtimeStatus": "Completed",
    "historyEvents": [{
        "name": "E1_SayHello",
        "scheduleId": 0, 
        "type": "TaskCompleted",
        "output": "Hello Tokyo!"
      },
      {
        "name": "E1_SayHello",
        "scheduleId": 1,
        "type": "TaskCompleted",
        "output": "Hello Seattle!"
      },
      {
        "name": "E1_SayHello",
        "scheduleId": 2,
        "type": "TaskCompleted",
        "output":  "Hello London!"
      },
      {
        "type": "TimerCreated",
        "fireAt":  "2018-01-23T06:45:38.724Z"
      },
      {
[...]

PR #142, opened 2018-01-24
Adding first logic for support of Quering Orchestration History descr…

gled4er:
This is an early version for the logic for **Query Orchestration History** feature -https://github.com/Azure/azure-functions-durable-extension/issues/16 

The main processing and aggregation logic is missing and it will be added shortly after confirming the final design. This commit contains the foundation for the feature that enables showing the history.

The aggregation logic will be added once the design in finalized.


gled4er:
@cgillum  This is pull request is very early stage but I want to check quickly if my approach is good enough and then I will add the aggregation logic that we are discussing here - https://github.com/Azure/azure-functions-durable-extension/issues/16

Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -172,5 +163,27 @@ private static JToken ParseToJToken(string value)
...
+            return new DurableOrchestrationStatus
+            {
+                Name = orchestrationState.Name,
+                InstanceId = orchestrationState.OrchestrationInstance.InstanceId,
+                CreatedTime = orchestrationState.CreatedTime,
+                LastUpdatedTime = orchestrationState.LastUpdatedTime,
+                RuntimeStatus = (OrchestrationRuntimeStatus)orchestrationState.OrchestrationStatus,
+                Input = ParseToJToken(orchestrationState.Input),
+                Output = ParseToJToken(orchestrationState.Output),
+                History = ParseToJToken(history),
```
cgillum: It seems inefficient to fetch the history as a string and then format it into JSON here. Can we instead fetch it as JSON and assign it more directly?

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -203,6 +220,17 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
             return response;
         }
 
+        private static bool GetQueryParameterValue(HttpRequestMessage request, string queryParameterName)
+        {
+            if (!request.GetQueryNameValuePairs().AllKeys.Contains(queryParameterName))
```
cgillum: Rather than checking for the query parameter key and then fetching the value, it's more efficient to check the value always and compare it to null to determine whether it exists. One lookup rather than two.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -203,6 +220,17 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
+            {
+                return false;
+            }
+
+            var value = request.GetQueryNameValuePairs()[queryParameterName];
```
cgillum: `GetQueryNameValuePairs` parses the query string every time it is called, right? The `GetQueryParameterValue` method is called several times in a row. Can we refactor this so that we only call it once and cache the result so we can quickly look up all the different query string values we care about without reparsing?

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -203,6 +220,17 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
+            return !string.IsNullOrEmpty(value) && bool.Parse(value);
```
cgillum: `bool.Parse(value)` will throw if `value` is not a properly formatted boolean. I suggest the following line instead:

```csharp
return bool.TryParse(value, out bool parsedValue) && parsedValue;
```

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -203,6 +220,17 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
             return response;
         }
 
+        private static bool GetQueryParameterValue(HttpRequestMessage request, string queryParameterName)
```
cgillum: Since this method is only concerned with boolean query string parameters, I would rename it to `GetBooleanQueryParameterValue`.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -187,7 +193,18 @@ public HttpApiHandler(DurableTaskExtension config, TraceWriter traceWriter)
...
+                request.CreateResponse(
+                statusCode,
+                new
+                {
+                    runtimeStatus = status.RuntimeStatus.ToString(),
+                    input = status.Input,
+                    output = status.Output,
+                    createdTime = status.CreatedTime.ToString("s") + "Z",
+                    lastUpdatedTime = status.LastUpdatedTime.ToString("s") + "Z",
+                 });
```
cgillum: The duplication of logic here makes me a bit nervous. I wonder if we should create a dedicated data contract for this response payload and use an attribute to omit the `historyEvents` property if it is null.  Something like:

```csharp
[DataContract]
class StatusResponsePayload
{
    [DataMember(Name = "runtimeStatus")]
    public string RuntimeStatus { get; set; }

    [DataMember(Name = "input")]
    public JToken Input { get; set; }

    [DataMember(Name = "output")]
    public JToken Output { get; set; }

    [DataMember(Name = "createdTime")]
    public string CreatedTime { get; set; }

    [DataMember(Name = "lastUpdatedTime")]
    public string LastUpdatedTime { get; set; }

    [DataMember(Name = "historyEvents", EmitDefaultValue = false)]
    public JArray HistoryEvents { get; set; }
}
```

Depending on how you put together the history events data, this property could be either a `JArray` or a generic array of a new data contract type that you add.

gled4er:
@cgillum  Thank you for the great feedback!

I will look into all your comments and work on better solutions.

Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
gled4er: @cgillum I think this is limitation of the underlying Durable Task Framework and especially this method:

```csharp
public async Task<string> GetOrchestrationHistoryAsync(OrchestrationInstance instance)
        {
            if (instance == null || string.IsNullOrEmpty(instance.InstanceId) ||
                string.IsNullOrEmpty(instance.ExecutionId))
            {
                throw new ArgumentNullException("instance");
            }

            ThrowIfInstanceStoreNotConfigured();

            IEnumerable<OrchestrationHistoryEventEntity> eventEntities =
                await
                    tableClient.ReadOrchestrationHistoryEventsAsync(instance.InstanceId, instance.ExecutionId)
                        .ConfigureAwait(false);
            var events = new List<HistoryEvent>(eventEntities
                .OrderBy(ee => ee.SequenceNumber)
                .Select(historyEventEntity => historyEventEntity.HistoryEvent));

            string history = JsonConvert.SerializeObject(
                events, Formatting.Indented, new JsonSerializerSettings
                {
                    TypeNameHandling = TypeNameHandling.Objects
                });
            return history;
        }
```
It is a bit unfortunate that the data is retrieved  first as JSON but then the method decides to serialize it. 

Another way is to go directly to the storage table and retrieve the content directly but I prefer not to do it since Azure Storage Table in this case is an implementation detail and and in my opinion it is better to go via the Durable Task Framework for accessing the history. In the same time I am not able to find another method that gives access to the history content. Let me know if you have better suggestion.

Thank you!

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
gled4er: @cgillum  This is great suggestion! I was trying to play a bit and modify the anonymous type yesterday but I decided to leave in this unpleasant way just to push the first commit. I was thinking as well for a dedicated data structure but I wasn't sure if it will be OK to introduce it. Now it is clear. Thank you for the code snippet as well!

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
gled4er: @cgillum  Great point! I will change it.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
gled4er: @cgillum  Nice improvement! Thank you! I will modify accordingly.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
gled4er: @cgillum Perfect! Thank you!

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
gled4er: @cgillum Sure. I like very much this optimization! I will work on it!

gled4er:
@cgillum  I addressed your previous comments with the last commit. And now I will proceed with the parsing and processing part of the history content. Let me know if I need to fix something else before that.

Thank you!

src/WebJobs.Extensions.DurableTask/StatusResponsePayload.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Runtime.Serialization;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    /// <summary>
+    /// Response for Orchestration Status Query.
+    /// </summary>
+    [DataContract]
+    public class StatusResponsePayload
```
cgillum: Can this be internal? I don't see a reason why developers would need to reference this directly.

gled4er: @cgillum  Good point. I will modify it. Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
cgillum: I agree we shouldn't go directly to table storage. What about simply doing [JArray.FromObject(events)](https://www.newtonsoft.com/json/help/html/M_Newtonsoft_Json_Linq_JArray_FromObject.htm)?

gled4er: @cgillum Great! I will change and we will be able to follow the original data structure.

gled4er:
@cgillum Thank you for the quick feedback. I fixed the access to `StatusResponsePayload` and now `History` property is from type `JArray`.

Thank you!

gled4er:
Hello @cgillum ,

I added the first version of the aggregation and I wanted to check what you think about this approach. After inspecting closely the various events I think we can only aggregate (scheduled / created and completed / failed events) for the following types because for them we have straightforward mapping criteria based on EventType and available ID-s  (**EventId and TaskScheduledId)** :

**TaskScheduled	-> TaskCompleted / TaskFailed
SubOrchestrationInstanceCreated -> SubOrchestrationInstanceCompleted / SubOrchestrationInstanceFailed**

For the following couples we don't have ID properties we can use for the mapping. We are missing the **TaskScheduledId** property that we can use to map the initial and the consecutive events:

**TimerCreated	-> TimerFired
OrchestratorStarted -> OrchestratorCompleted
ExecutionStarted	-> ExecutionCompleted / ExecutionFailed / ExecutionTerminated**

I believe that it is better to keep the event we can't aggregate in the history output because they are giving context for the orchestration execution.  Let me know if you think some of them will not be useful and I will add logic to discard them.

**I plan to optimize the approach for removing certain properties and remove the duplication in the method `RemoveProperties` .**

Currently if we request Input / Result to be added to the output, we are getting the following output for E1_HelloSequence:

```javascript

"historyEvents": [
        {
            "EventId": -1,
            "EventType": "OrchestratorStarted",
            "Timestamp": "2018-01-26T13:37:20.5952551Z"
        },
        {
            "EventId": -1,
            "EventType": "ExecutionStarted",
            "Input": "null",
            "Name": "E1_HelloSequence",
            "Tags": null,
            "Timestamp": "2018-01-26T13:37:17.0044271Z"
        },
        {
            "EventId": -1,
            "EventType": "OrchestratorCompleted",
            "Timestamp": "2018-01-26T13:37:20.7577519Z"
        },
        {
            "EventId": -1,
            "EventType": "OrchestratorStarted",
            "Timestamp": "2018-01-26T13:37:21.2858393Z"
        },
        {
            "EventId": -1,
            "EventType": "TaskCompleted",
            "Result": "\"Hello Tokyo!\"",
            "StartTime": "2018-01-26T13:37:20",
            "Timestamp": "2018-01-26T13:37:21.1373369Z"
        },
        {
            "EventId": -1,
            "EventType": "OrchestratorCompleted",
            "Timestamp": "2018-01-26T13:37:21.4428364Z"
        },
        {
            "EventId": -1,
            "EventType": "OrchestratorStarted",
            "Timestamp": "2018-01-26T13:37:21.8228387Z"
        },
        {
            "EventId": -1,
            "EventType": "TaskCompleted",
            "Result": "\"Hello Seattle!\"",
            "StartTime": "2018-01-26T13:37:21",
            "Timestamp": "2018-01-26T13:37:21.6958395Z"
        },
        {
            "EventId": -1,
            "EventType": "OrchestratorCompleted",
            "Timestamp": "2018-01-26T13:37:21.9993407Z"
        },
        {
            "EventId": -1,
            "EventType": "OrchestratorStarted",
            "Timestamp": "2018-01-26T13:37:22.42734Z"
        },
        {
            "EventId": -1,
            "EventType": "TaskCompleted",
            "Result": "\"Hello London!\"",
            "StartTime": "2018-01-26T13:37:21",
            "Timestamp": "2018-01-26T13:37:22.3168391Z"
        },
        {
            "EventId": 3,
            "EventType": "ExecutionCompleted",
            "OrchestrationStatus": 1,
            "Result": "[\"Hello Tokyo!\",\"Hello Seattle!\",\"Hello London!\"]",
            "Timestamp": "2018-01-26T13:37:22.595339Z"
        },
        {
            "EventId": -1,
            "EventType": "OrchestratorCompleted",
            "Timestamp": "2018-01-26T13:37:22.6038435Z"
        }
    ] 
```
**And when we discard the Input / Result we are getting:**
```javascript
    "historyEvents": [
        {
            "EventId": -1,
            "EventType": "OrchestratorStarted",
            "Timestamp": "2018-01-26T13:37:20.5952551Z"
        },
        {
            "EventId": -1,
            "EventType": "ExecutionStarted",
            "Name": "E1_HelloSequence",
            "Tags": null,
            "Timestamp": "2018-01-26T13:37:17.0044271Z"
        },
        {
            "EventId": -1,
            "EventType": "OrchestratorCompleted",
            "Timestamp": "2018-01-26T13:37:20.7577519Z"
        },
        {
            "EventId": -1,
            "EventType": "OrchestratorStarted",
            "Timestamp": "2018-01-26T13:37:21.2858393Z"
        },
        {
            "EventId": -1,
            "EventType": "TaskCompleted",
            "StartTime": "2018-01-26T13:37:20",
            "Timestamp": "2018-01-26T13:37:21.1373369Z"
        },
        {
            "EventId": -1,
            "EventType": "OrchestratorCompleted",
            "Timestamp": "2018-01-26T13:37:21.4428364Z"
        },
        {
            "EventId": -1,
            "EventType": "OrchestratorStarted",
            "Timestamp": "2018-01-26T13:37:21.8228387Z"
        },
        {
            "EventId": -1,
            "EventType": "TaskCompleted",
            "StartTime": "2018-01-26T13:37:21",
            "Timestamp": "2018-01-26T13:37:21.6958395Z"
        },
        {
            "EventId": -1,
            "EventType": "OrchestratorCompleted",
            "Timestamp": "2018-01-26T13:37:21.9993407Z"
        },
        {
            "EventId": -1,
            "EventType": "OrchestratorStarted",
            "Timestamp": "2018-01-26T13:37:22.42734Z"
        },
        {
            "EventId": -1,
            "EventType": "TaskCompleted",
            "StartTime": "2018-01-26T13:37:21",
            "Timestamp": "2018-01-26T13:37:22.3168391Z"
        },
        {
            "EventId": 3,
            "EventType": "ExecutionCompleted",
            "OrchestrationStatus": 1,
            "Timestamp": "2018-01-26T13:37:22.595339Z"
        },
        {
            "EventId": -1,
            "EventType": "OrchestratorCompleted",
            "Timestamp": "2018-01-26T13:37:22.6038435Z"
        }
    ]
```
Now, looking at the output I think we can remove `EventId` and set string value for `OrchestrationStatus`.

I named the property holding the time of the initial event `StartTime` because in the case of `SubOrchestration` is created time but if you think that `ScheduledTime`  is better, I will modify it. I will also add more precision to this variable to match the ` Timestamp` .

Thank you!

gled4er:
Hello @cgillum ,

With the latest commit, I combined the logic for parsing and aggregating events so we have only one pass over the JSON content. The strucutre of the aggregated JSON is a bit changed - the `StartTime` field is changed to `ScheduledTime` as you suggested originally.  I also added `SchedulerName` which shows the `Name` value from the `TaskScheduled` or `SubOrchestrationInstanceCreated` in the `TaskCompleted` / `SubOrchestrationInstanceCompleted`. Let me know if you come up with better name. 

Now the value of   `OrchestrationStatus` is string and  `EventId` and `IsPlayed` are removed. The precision of `ScheduledTime` matches the `Timestamp`'s one. 

When we show the history with default options for **E1_HelloSequence**, we get: 

```javascript
 "historyEvents": [
        {
            "EventType": "OrchestratorStarted",
            "Timestamp": "2018-01-29T09:36:39.8854387Z"
        },
        {
            "EventType": "ExecutionStarted",
            "Name": "E1_HelloSequence",
            "Timestamp": "2018-01-29T09:36:39.3329226Z"
        },
        {
            "EventType": "OrchestratorCompleted",
            "Timestamp": "2018-01-29T09:36:40.2989395Z"
        },
        {
            "EventType": "OrchestratorStarted",
            "Timestamp": "2018-01-29T09:36:41.2524564Z"
        },
        {
            "EventType": "TaskCompleted",
            "ScheduledTime": "2018-01-29T09:36:40.2984385Z",
            "SchedulerName": "E1_SayHello",
            "Timestamp": "2018-01-29T09:36:41.0749547Z"
        },
        {
            "EventType": "OrchestratorCompleted",
            "Timestamp": "2018-01-29T09:36:41.3269597Z"
        },
        {
            "EventType": "OrchestratorStarted",
            "Timestamp": "2018-01-29T09:36:41.7784697Z"
        },
        {
            "EventType": "TaskCompleted",
            "ScheduledTime": "2018-01-29T09:36:41.3269597Z",
            "SchedulerName": "E1_SayHello",
            "Timestamp": "2018-01-29T09:36:41.6614742Z"
        },
        {
            "EventType": "OrchestratorCompleted",
            "Timestamp": "2018-01-29T09:36:41.9329721Z"
        },
        {
            "EventType": "OrchestratorStarted",
            "Timestamp": "2018-01-29T09:36:42.33798Z"
        },
        {
            "EventType": "TaskCompleted",
            "ScheduledTime": "2018-01-29T09:36:41.9329721Z",
            "SchedulerName": "E1_SayHello",
            "Timestamp": "2018-01-29T09:36:42.1904787Z"
        },
        {
            "EventType": "ExecutionCompleted",
            "OrchestrationStatus": "Completed",
            "Timestamp": "2018-01-29T09:36:42.6139849Z"
        },
        {
            "EventType": "OrchestratorCompleted",
            "Timestamp": "2018-01-29T09:36:42.6219918Z"
        }
    ]
```
And if we request the **input and result** values as well, we get:

```javascript
 "historyEvents": [
        {
            "EventType": "OrchestratorStarted",
            "Timestamp": "2018-01-29T09:36:39.8854387Z"
        },
        {
            "EventType": "ExecutionStarted",
            "Input": "null",
            "Name": "E1_HelloSequence",
            "Timestamp": "2018-01-29T09:36:39.3329226Z"
        },
        {
            "EventType": "OrchestratorCompleted",
            "Timestamp": "2018-01-29T09:36:40.2989395Z"
        },
        {
            "EventType": "OrchestratorStarted",
            "Timestamp": "2018-01-29T09:36:41.2524564Z"
        },
        {
            "EventType": "TaskCompleted",
            "Result": "\"Hello Tokyo!\"",
            "ScheduledTime": "2018-01-29T09:36:40.2984385Z",
            "SchedulerName": "E1_SayHello",
            "Timestamp": "2018-01-29T09:36:41.0749547Z"
        },
        {
            "EventType": "OrchestratorCompleted",
            "Timestamp": "2018-01-29T09:36:41.3269597Z"
        },
        {
            "EventType": "OrchestratorStarted",
            "Timestamp": "2018-01-29T09:36:41.7784697Z"
        },
        {
            "EventType": "TaskCompleted",
            "Result": "\"Hello Seattle!\"",
            "ScheduledTime": "2018-01-29T09:36:41.3269597Z",
            "SchedulerName": "E1_SayHello",
            "Timestamp": "2018-01-29T09:36:41.6614742Z"
        },
        {
            "EventType": "OrchestratorCompleted",
            "Timestamp": "2018-01-29T09:36:41.9329721Z"
        },
        {
            "EventType": "OrchestratorStarted",
            "Timestamp": "2018-01-29T09:36:42.33798Z"
        },
        {
            "EventType": "TaskCompleted",
            "Result": "\"Hello London!\"",
            "ScheduledTime": "2018-01-29T09:36:41.9329721Z",
            "SchedulerName": "E1_SayHello",
            "Timestamp": "2018-01-29T09:36:42.1904787Z"
        },
        {
            "EventType": "ExecutionCompleted",
            "OrchestrationStatus": "Completed",
            "Result": "[\"Hello Tokyo!\",\"Hello Seattle!\",\"Hello London!\"]",
            "Timestamp": "2018-01-29T09:36:42.6139849Z"
        },
        {
            "EventType": "OrchestratorCompleted",
            "Timestamp": "2018-01-29T09:36:42.6219918Z"
        }
    ]
```

Do you think that we need to show the input in the aggregated events as well? Now,  looking at the response I think we should add it. 

Let me know what you think I need to improve. 

I plan to work next on your feedback, adding unit tests and working on documentation changes. 

Thank you!

cgillum:
Hey Kanio, sorry for the delay in reviewing this. Things have been quite busy for me. Here is some feedback based on the examples you've provided above. I haven't looked at the code yet, but I'll do so once we agree on the format of the JSON.

- I recall that the inputs are not included in the history given to us by the Durable Task Framework, so it looks like we'll need to leave those out of our history results JSON as well.
- Let's remove the **OrchestratorStarted** and **OrchestratorCompleted** events. I don't think those are particularly useful for the user and there are quite a lot of them.
- Instead of **SchedulerName**, could we use **FunctionName**? And I think should we use the same field name in the **ExecutionStarted** event, which currently uses **Name**.
- I noticed that the **Result** fields are all encoded string values. Can we make those native JSON values instead? For example, instead of `"Result": "[\"Hello Tokyo!\",\"Hello Seattle!\",\"Hello London!\"]"`, it would be nice if it could be `"Result": ["Hello Tokyo!","Hello Seattle!","Hello London!"]`. In the code, this usually means the `Result` property needs to be a `JToken` instead of a `string`. You'll notice I did something similar in the data contract for the existing status query code.

gled4er:
Hello @cgillum ,

I know that you are very busy so no worries about that.

I will write you when I am ready with the changes.

Thank you!

gled4er:
Hello @cgillum ,

I  implemented the changes you requested in the last review. 

When you have time please check the code and let me know what I need to fix next.

Thank you!

gled4er:
Hello @cgillum ,

I forgot to add the current history output for a quick reference:

When we request the history with **{statusQueryGetUri}&showHistory=true**

```javascript
{
    "createdTime": "2018-02-01T04:11:46Z",
    "historyEvents": [
        {
            "EventType": "ExecutionStarted",
            "FunctionName": "E1_HelloSequence",
            "Timestamp": "2018-02-01T04:11:46.3300255Z"
        },
        {
            "EventType": "TaskCompleted",
            "FunctionName": "E1_SayHello",
            "ScheduledTime": "2018-02-01T04:11:49.6427304Z",
            "Timestamp": "2018-02-01T04:11:50.0307362Z"
        },
        {
            "EventType": "TaskCompleted",
            "FunctionName": "E1_SayHello",
            "ScheduledTime": "2018-02-01T04:11:50.2502321Z",
            "Timestamp": "2018-02-01T04:11:50.4837315Z"
        },
        {
            "EventType": "TaskCompleted",
            "FunctionName": "E1_SayHello",
            "ScheduledTime": "2018-02-01T04:11:50.6862485Z",
            "Timestamp": "2018-02-01T04:11:50.8949288Z"
        },
        {
            "EventType": "ExecutionCompleted",
            "OrchestrationStatus": "Completed",
            "Timestamp": "2018-02-01T04:11:51.1764413Z"
        }
    ],
    "input": null,
    "lastUpdatedTime": "2018-02-01T04:11:51Z",
    "output": [
        "Hello Tokyo!",
        "Hello Seattle!",
        "Hello London!"
    ],
    "runtimeStatus": "Completed"
}
```

and  when we request the history and we opt-in for the output with **{statusQueryGetUri}&showHistory=true&showHistoryOutput=true** we get:

```javascript
{
    "createdTime": "2018-02-01T04:11:46Z",
    "historyEvents": [
        {
            "EventType": "ExecutionStarted",
            "FunctionName": "E1_HelloSequence",
            "Timestamp": "2018-02-01T04:11:46.3300255Z"
        },
        {
            "EventType": "TaskCompleted",
            "FunctionName": "E1_SayHello",
            "Result": "Hello Tokyo!",
            "ScheduledTime": "2018-02-01T04:11:49.6427304Z",
            "Timestamp": "2018-02-01T04:11:50.0307362Z"
        },
        {
            "EventType": "TaskCompleted",
            "FunctionName": "E1_SayHello",
            "Result": "Hello Seattle!",
            "ScheduledTime": "2018-02-01T04:11:50.2502321Z",
            "Timestamp": "2018-02-01T04:11:50.4837315Z"
        },
        {
            "EventType": "TaskCompleted",
            "FunctionName": "E1_SayHello",
            "Result": "Hello London!",
            "ScheduledTime": "2018-02-01T04:11:50.6862485Z",
            "Timestamp": "2018-02-01T04:11:50.8949288Z"
        },
        {
            "EventType": "ExecutionCompleted",
            "OrchestrationStatus": "Completed",
            "Result": [
                "Hello Tokyo!",
                "Hello Seattle!",
                "Hello London!"
            ],
            "Timestamp": "2018-02-01T04:11:51.1764413Z"
        }
    ],
    "input": null,
    "lastUpdatedTime": "2018-02-01T04:11:51Z",
    "output": [
        "Hello Tokyo!",
        "Hello Seattle!",
        "Hello London!"
    ],
    "runtimeStatus": "Completed"
}
```

Let me know if you have recommendations for further improving the format of the response.

Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -172,5 +165,190 @@ private static JToken ParseToJToken(string value)
...
+            {
+                return;
+            }
+
+            jsonToken["Result"] = ParseToJToken(jsonToken["Result"]?.ToString());
+        }
+
+        private static void ChangePrpertyName(ref JToken jsonToken, string currentPropertyName, string newPropertyName)
+        {
+            var value = jsonToken?[currentPropertyName];
```
cgillum: The use of `jsonToken?` null-checks seems pretty excessive. I suggest you should leave these null-checks out. If you think null is a possibility for a history item, then I suggest you filter these out in the calling code (e.g. at the top of the `for` loop) so that these helper methods are never called with null inputs.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -172,5 +165,190 @@ private static JToken ParseToJToken(string value)
...
+                jsonToken[properyName]?.Parent?.Remove();
+            }
+        }
+
+        private static void TrackNameAndScheduledTime(JToken historyItem, EventType eventType, int index, ref Dictionary<string, EventIndexDateMapping> eventMapper)
+        {
+            eventMapper.Add($"{eventType}_{historyItem["EventId"]}", new EventIndexDateMapping { Index = index, Name = historyItem["Name"].ToString(), Input = historyItem["Input"]?.ToString(), Date = (DateTime)historyItem["Timestamp"] });
+        }
+
+        private static void AddScheduledEventDataAndAggreagate(IReadOnlyDictionary<string, EventIndexDateMapping> eventMapper, string prefix, JToken historyItem, ref List<int> indexList)
```
cgillum: It seems like `indexList` does not need to be `ref`.

Also, I think there is a typo in the name. I think it should be `AddScheduledEventDataAndAggregate `.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -172,5 +165,190 @@ private static JToken ParseToJToken(string value)
...
+        {
+            var taskScheduledData = eventMapper[$"{prefix}_{historyItem["TaskScheduledId"]}"];
+            historyItem["ScheduledTime"] = taskScheduledData.Date;
+            historyItem["FunctionName"] = taskScheduledData.Name;
+            indexList.Add(taskScheduledData.Index);
+        }
+
+        private static void ConvertOutputToJToken(ref JToken jsonToken, bool showHistoryOutput)
```
cgillum: `jsonToken` should not be `ref` here.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -172,5 +165,190 @@ private static JToken ParseToJToken(string value)
...
+        {
+            if (!showHistoryOutput)
+            {
+                return;
+            }
+
+            if (jsonToken == null)
```
cgillum: I would remove these `jsonToken` null checks from all helper methods per my other feedback.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -172,5 +165,190 @@ private static JToken ParseToJToken(string value)
...
+        }
+
+        private static void TrackNameAndScheduledTime(JToken historyItem, EventType eventType, int index, ref Dictionary<string, EventIndexDateMapping> eventMapper)
+        {
+            eventMapper.Add($"{eventType}_{historyItem["EventId"]}", new EventIndexDateMapping { Index = index, Name = historyItem["Name"].ToString(), Input = historyItem["Input"]?.ToString(), Date = (DateTime)historyItem["Timestamp"] });
+        }
+
+        private static void AddScheduledEventDataAndAggreagate(IReadOnlyDictionary<string, EventIndexDateMapping> eventMapper, string prefix, JToken historyItem, ref List<int> indexList)
+        {
+            var taskScheduledData = eventMapper[$"{prefix}_{historyItem["TaskScheduledId"]}"];
```
cgillum: I wonder if we should have a null-check here in case the event data comes in out of order for some reason.

src/WebJobs.Extensions.DurableTask/EventIndexDateMapping.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    internal class EventIndexDateMapping
```
cgillum: I didn't check, but does any other class than `DurableOrchestrationClient` use `EventIndexDateMapping`? If not, instead of adding it as it's own file, can we make it a private inner class of `DurableOrchestrationClient`.

test/DurableTaskEndToEndTests.cs
```diff
@@ -107,11 +130,48 @@ public DurableTaskEndToEndTests(ITestOutputHelper output)
...
                 Assert.Equal("World", status?.Input);
                 Assert.Equal("Hello, World!", status?.Output);
+                if (!showHistory)
+                {
+                    Assert.Equal(null, status?.History);
+                }
+                else
+                {
+                    Assert.Equal(3, status?.History.Count);
+                    Assert.Equal<string>("ExecutionStarted", status?.History[0]["EventType"].ToString());
```
cgillum: Rather than using `status?` everywhere, we should just `Assert.NotNull(status)` earlier in this method.

gled4er:
Hello @cgillum ,

Thank you for the detailed feedback. I really like the points you made in the review and I will work to fix the issues as soon as possible.

Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -172,5 +165,190 @@ private static JToken ParseToJToken(string value)
 
             return state;
         }
+
+        private async Task<DurableOrchestrationStatus> GetDurableOrchestrationStatusAsync(OrchestrationState orchestrationState, bool showHistory, bool showHistoryOutput)
+        {
+            JArray historyArray = null;
+            if (showHistory)
+            {
+                var history = await this.client.GetOrchestrationHistoryAsync(orchestrationState.OrchestrationInstance);
```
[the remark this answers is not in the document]
gled4er: Sure. I will follow this approach in future. Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -172,5 +165,190 @@ private static JToken ParseToJToken(string value)
...
+                if (!string.IsNullOrEmpty(history))
+                {
+                    historyArray = JArray.Parse(history);
+
+                    var eventMapper = new Dictionary<string, EventIndexDateMapping>();
+                    var indexList = new List<int>();
+
+                    for (var i = 0; i < historyArray.Count; i++)
+                    {
+                        var historyItem = historyArray[i];
```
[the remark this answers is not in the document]
gled4er: Yes, great point. I will change it. Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -172,5 +165,190 @@ private static JToken ParseToJToken(string value)
...
+                        if (Enum.TryParse(historyItem["EventType"].Value<string>(), out EventType eventType))
+                        {
+                            historyItem["EventType"] = eventType.ToString();
```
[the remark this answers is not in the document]
gled4er: Sure. I will add this comment - "Changing the value of EventType from integer to string for better understanding in the history output".

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -172,5 +165,190 @@ private static JToken ParseToJToken(string value)
...
+                            switch (eventType)
+                            {
+                                case EventType.TaskScheduled:
+                                    TrackNameAndScheduledTime(historyItem, eventType, i, ref eventMapper);
+                                    UpdateProperties(ref historyItem, "Version", true);
```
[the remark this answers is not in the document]
gled4er: I like very much the idea and I will update the code. This approach simplifies a lot the code.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -172,5 +165,190 @@ private static JToken ParseToJToken(string value)
...
+                CreatedTime = orchestrationState.CreatedTime,
+                LastUpdatedTime = orchestrationState.LastUpdatedTime,
+                RuntimeStatus = (OrchestrationRuntimeStatus)orchestrationState.OrchestrationStatus,
+                Input = ParseToJToken(orchestrationState.Input),
+                Output = ParseToJToken(orchestrationState.Output),
+                History = historyArray,
+            };
+        }
+
+        private static void UpdateProperties(ref JToken jsonToken, List<string> propertyNames, bool hideInput = false, bool hideResult = false)
```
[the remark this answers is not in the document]
gled4er: Great catch! I will remove the method. Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -172,5 +165,190 @@ private static JToken ParseToJToken(string value)
...
+        {
+            if (jsonToken == null)
```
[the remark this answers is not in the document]
gled4er: Thank you, I will remove the method.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -172,5 +165,190 @@ private static JToken ParseToJToken(string value)
...
+                propertyNames.Add("Result");
+            }
+
+            foreach (var propertyName in propertyNames)
+            {
+                UpdateProperties(ref jsonToken, propertyName);
+            }
+        }
+
+        private static void UpdateProperties(ref JToken jsonToken, string properyName, bool hideInput = false, bool hideResult = false)
```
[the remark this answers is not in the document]
gled4er: Great optimization for removing properties when we use JObject! Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -172,5 +165,190 @@ private static JToken ParseToJToken(string value)
...
+
+            if (!string.IsNullOrEmpty(properyName))
+            {
+                jsonToken[properyName]?.Parent?.Remove();
+            }
+        }
+
+        private static void TrackNameAndScheduledTime(JToken historyItem, EventType eventType, int index, ref Dictionary<string, EventIndexDateMapping> eventMapper)
+        {
+            eventMapper.Add($"{eventType}_{historyItem["EventId"]}", new EventIndexDateMapping { Index = index, Name = historyItem["Name"].ToString(), Input = historyItem["Input"]?.ToString(), Date = (DateTime)historyItem["Timestamp"] });
```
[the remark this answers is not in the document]
gled4er: Yes, I will remove the coded related to `Input`. I change the approach and I will use `(string)` from now on. Thank you for the advice!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
gled4er: I will change it. Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
gled4er: I will change it. Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -172,5 +165,190 @@ private static JToken ParseToJToken(string value)
...
+
+            if (jsonToken == null)
+            {
+                return;
+            }
+
+            jsonToken["Result"] = ParseToJToken(jsonToken["Result"]?.ToString());
+        }
+
+        private static void ChangePrpertyName(ref JToken jsonToken, string currentPropertyName, string newPropertyName)
```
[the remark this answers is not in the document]
gled4er: I decided to remove this method since it is called only once and with the use of `JObject` becames trivial.

src/WebJobs.Extensions.DurableTask/EventIndexDateMapping.cs
gled4er: Yes, great point. I will move it. Thank you very much!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
gled4er: Great point. I  decided to remove this method since it is called only once and with the use of `JObject` becames trivial. Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
gled4er: Got it. I will remove it. Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
gled4er: Yes, great point. I think it is a case we need to check for. I will add it.

Thank you!

gled4er:
Hello @cgillum ,

I implemented the changes that were requested in the last review iteration.

When you have time, may I ask you to check again and let me know what should be fixed next?

Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -165,12 +158,177 @@ private static JToken ParseToJToken(string value)
...
+        }
+
+        private static void ConvertOutputToJToken(JObject jsonObject, bool showHistoryOutput)
+        {
+            if (!showHistoryOutput)
+            {
+                return;
+            }
+
+            if (jsonObject == null)
```
cgillum: I don't think this null-check is necessary. The code would have failed much earlier than this if `jsonObject` was null.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -165,12 +158,177 @@ private static JToken ParseToJToken(string value)
...
+        }
+
+        private static void AddScheduledEventDataAndAggregate(ref Dictionary<string, EventIndexDateMapping> eventMapper, string prefix, JToken historyItem, List<int> indexList)
+        {
+            if (!eventMapper.ContainsKey($"{prefix}_{historyItem["TaskScheduledId"]}"))
+            {
+                return;
+            }
+
+            var taskScheduledData = eventMapper[$"{prefix}_{historyItem["TaskScheduledId"]}"];
```
cgillum: I don't think the null checks you have in this method are necessary. Also, please use `TryGetValue` instead of `ContainsKey` followed by an index lookup. `TryGetValue` is more efficient since you're only querying the dictionary once. Here is an alternate implementation of this method:
```csharp
if (eventMapper.TryGetValue($"{prefix}_{historyItem["TaskScheduledId"]}", out EventIndexDateMapping taskScheduledData))
{
    indexList.Add(taskScheduledData.Index);
}
```

cgillum:
Also, it looks like one of your tests are failing in the CI: HelloWorldOrchestration_Activity_History

gled4er:
Hello @cgillum ,

Thank you very much for the reviews.

I will work on these next and I will let you know when I am done.

Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
gled4er: Great! I need to use this more to be my first option in similar scenarios. Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
gled4er: Great optimization! Thank you!

gled4er:
Hello @cgillum ,

Thank you for pointing me to the failed test. After a bit of investigation, I identified this line as the problematic one:

```csharp

 if (DateTime.TryParse(status.History[1]["Timestamp"].ToString(), out DateTime timestamp) &&
                        DateTime.TryParse(status.History[1]["ScheduledTime"].ToString(), out DateTime scheduledTime))
                    {
                        Assert.True(timestamp > scheduledTime);
                    }

```

After my last logic change commit, I got two test failing pointing to the same Assert. Because of that I decided to relax a bit the condition since I wasn't able to reproduce locally:

```csharp

 if (DateTime.TryParse(status.History[1]["Timestamp"].ToString(), out DateTime timestamp) &&
                        DateTime.TryParse(status.History[1]["ScheduledTime"].ToString(), out DateTime scheduledTime))
                    {
                       Assert.True(timestamp **>=** scheduledTime);
                    }

```
My current theory is that the test is executed very quickly on the CI build infrastructure that results in  `timestamp` and `scheduledTime` being very close in time.

I am interested in your opinion about this situation.

Thank you!

cgillum:
That seems to have worked. If we find it's still flaky, it might be worth adding more logging so we can see what the actual timestamps are and why they are not what we expect.

Issue #141, opened 2018-01-23
WaitForExternalEvent cannot be awaited more than once using the same event name


cgillum:
Consider the following example orchestration:

```csharp
[FunctionName("Loopy")]
public static async Task WaitForFiles([OrchestrationTrigger] DurableOrchestrationContext ctx)
{
    var receivedFiles = new List<string>();
    while (receivedFiles.Count < 3)
    {
        var file = await ctx.WaitForExternalEvent<string>("file");
        receivedFiles.Add(file);
    }

    // continue on...
}
```
It is designed to wait for three files to be delivered via `RaiseEvent` calls with the event name "file" before moving on to the next step.

### Expected
The orchestration should wait until three "file" events are sent to this instance using `RaiseEvent` before it moves to the next step.

### Actual
The orchestrator function goes into an infinite loop after receiving the first "file" event.

The problem is the internal implementation of `WaitForExternalEvent`. After the first event is received, subsequent calls to `WaitForExternalEvent` will return the same completed task instead of generating a new one.

This will be very important to fix for scenarios that rely on processing batches of external events.

### Workaround
If possible, refactor the orchestrator function and the client to ensure that each event has a unique name (this may not be possible in all scenarios).

cgillum:
This has been fixed by @brandonh-msft. A "prerelease" version of this fix is available on myget: https://www.myget.org/feed/azure-appservice/package/nuget/Microsoft.Azure.WebJobs.Extensions.DurableTask/1.1.1-beta2. More officially, this will go out with the Beta 3 release.

stuartleeks:
Very happy to see this fixed and released!
Thanks @brandonh-msft  :-)

PR #144, opened 2018-01-24
Fix infinite WaitForExternalEvent firing (#141)

brandonh-msft:
cc @cgillum 

this makes sense to me - thoughts?

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
@@ -397,6 +397,10 @@ namespace Microsoft.Azure.WebJobs
                     Type tcsType = tcs.GetType();
                     Type genericTypeArgument = tcsType.GetGenericArguments()[0];
 
+                    // If we're going to raise an event we should remove it from the pending collection
+                    // because otherwise WaitForExternal() will always find one with this key and run infinitely. Fixes #141
+                    this.pendingExternalEvents.Remove(name);
+
                     object deserializedObject = MessagePayloadDataConverter.Default.Deserialize(input, genericTypeArgument);
                     MethodInfo trySetResult = tcsType.GetMethod("TrySetResult");
                     trySetResult.Invoke(tcs, new[] { deserializedObject });
--- a/test/DurableTaskEndToEndTests.cs
+++ b/test/DurableTaskEndToEndTests.cs
@@ -248,8 +248,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
             {
                 await host.StartAsync();
 
-                int initialValue = 0;
-                var client = await host.StartOrchestratorAsync(nameof(TestOrchestrations.BatchActor), initialValue, this.output);
+                var client = await host.StartOrchestratorAsync(nameof(TestOrchestrations.BatchActor), null, this.output);
 
                 // Need to wait for the instance to start before sending events to it.
                 // TODO: This requirement may not be ideal and should be revisited.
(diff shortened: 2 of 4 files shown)


src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -397,9 +397,18 @@ internal void RaiseEvent(string name, string input)
...
+                    {
+                        object deserializedObject = MessagePayloadDataConverter.Default.Deserialize(input, genericTypeArgument);
+                        MethodInfo trySetResult = tcsType.GetMethod("TrySetResult");
+                        trySetResult.Invoke(tcs, new[] { deserializedObject });
+                    }
+                    finally
+                    {
+                        // If we even attempt to raise an event, whether or not it failed, we should remove it from the pending collection
+                        // because otherwise WaitForExternal() will always find one with this key and run infinitely. Fixes #141
+                        this.pendingExternalEvents.Remove(name);
```
cgillum: `TaskCompletionSource.TrySetResult` will execute synchronously, so if the continuation wants to wait for another event with the same name, then you'll still get the infinite loop. I think you can fix this by removing it from the `pendingExternalEvents` list *first*, then calling `TrySetResult` on it.

brandonh-msft: k, yeah i wondered about flipping the order as well. I'm in process of including a Unit Test for this alongside the checkin; will submit that change with it.

Thanks!

test/TestOrchestrations.cs
```diff
@@ -80,6 +80,33 @@ public static string SayHelloInline([OrchestrationTrigger] DurableOrchestrationC
             return currentValue;
         }
 
+        public static async Task<int> InfiniteCounter([OrchestrationTrigger] DurableOrchestrationContext ctx)
+        {
+            int currentValue = ctx.GetInput<int>();
+
+            bool done = false;
+            while (!done)
```
cgillum: Can we model the test orchestration as something other than a potentially infinite actor? What we have here is actually a bit of an anti-pattern. In general we advise that infinite loops because of the unbounded CPU and memory overhead they create if the loop is not terminated. Obviously this works fine as a test case because we know that it will never actually be treated like an infinite orchestration, but I worry about others looking at this as a valid use case and going in a bad direction.

cgillum: Maybe we can instead model it after your document batch processing scenario?

test/TestOrchestrations.cs
```diff
@@ -80,6 +80,33 @@ public static string SayHelloInline([OrchestrationTrigger] DurableOrchestrationC
...
         }
 
+        public static async Task<int> InfiniteCounter([OrchestrationTrigger] DurableOrchestrationContext ctx)
+        {
+            int currentValue = ctx.GetInput<int>();
+
+            bool done = false;
+            while (!done)
+            {
+                System.Console.WriteLine($@"Current value is {currentValue}");
```
cgillum: Let's leave out `Console.WriteLine` from the test code.

test/DurableTaskEndToEndTests.cs
```diff
@@ -238,6 +238,54 @@ public DurableTaskEndToEndTests(ITestOutputHelper output)
             }
         }
 
+        /// <summary>
+        /// End-to-end test which validates the ContinueAsNew functionality by implementing a counter actor pattern.
```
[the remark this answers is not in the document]
brandonh-msft: lol that was in the back of my mind to correct but I got excited :)

test/TestOrchestrations.cs
brandonh-msft: Ah, makes sense. Sure thing I'll make this more closely model the "file set" use case.

test/TestOrchestrations.cs
brandonh-msft: Sure thing. I couldn't figure out how to get *any* output to the Test harness from w/in the actual execution :(

brandonh-msft:
@cgillum updated! LMK how it looks. If you want, I can also put the determination of which items to wait for in to an Activity function to be even more "real world" - up to you.

Issue #137, opened 2018-01-19
Passing null parameter into Activity causes exception


MisinformedDNA:
If I add a new Function, choosing the Durable function template, I get an exception if I pass in null for the string.

    [FunctionName("Function1")]
    public static async Task<List<string>> RunOrchestrator(
        [OrchestrationTrigger] DurableOrchestrationContext context)
    {
        var outputs = new List<string>();

        outputs.Add(await context.CallActivityAsync<string>("Function1_Hello", "Tokyo"));
        outputs.Add(await context.CallActivityAsync<string>("Function1_Hello", ""));
        outputs.Add(await context.CallActivityAsync<string>("Function1_Hello", null));  // exception

        return outputs;
    }

    [FunctionName("Function1_Hello")]
    public static string SayHello([ActivityTrigger] string name, TraceWriter log)
    {
        log.Info($"Saying hello to {name}.");
        return $"Hello {name}!";
    }

> DurableTask.Core.Exceptions.TaskFailedException
>   HResult=0x80131500
>   Message=Exception while executing function: Function1_Hello
>   Source=System.Private.CoreLib
>   StackTrace:
>    at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
>    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
>    at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
>    at DurableTask.Core.TaskOrchestrationContext.<ScheduleTaskInternal>d__14.MoveNext()
>    at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
>    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
>    at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
>    at DurableTask.Core.TaskOrchestrationContext.<ScheduleTaskToWorker>d__13`1.MoveNext()
>    at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
>    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
>    at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
>    at DurableTask.Core.TaskOrchestrationContext.<ScheduleTask>d__12`1.MoveNext()
[...]

PR #148, opened 2018-02-03
When parameter is null, convert to empty string.

tohling:
Fix for: https://github.com/Azure/azure-functions-durable-extension/issues/137

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Bindings/ObjectValueProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/ObjectValueProvider.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         public string ToInvokeString()
         {
-            return (this.value != null) ? this.value.ToString() : null;
+            return this.value?.ToString();
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/WebJobs.Extensions.DurableTask/Bindings/ObjectValueProvider.cs
```diff
@@ -38,7 +38,7 @@ public Task<object> GetValueAsync()
 
         public string ToInvokeString()
         {
-            return this.value.ToString();
+            return (this.value != null) ? this.value.ToString() : string.Empty;
```
cgillum: I don't remember what `ToInvokeString()` is used for, though I believe @MikeStall told me at one point. I thought maybe it was just for manual execution in the portal, or for the old monitoring system. Do you know?

Also, had you considered simply returning null instead of an empty string? The problem with converting null into an empty string generally is that there are some cases where `null` and `string.Empty` are not meant to be equivalent, and could in theory cause unexpected behavior.

cgillum: Also, do other bindings in Azure Functions implement the same kind of null handling behavior?

tohling: @cgillum, I am not sure. @MikeStall can probably provide more insights to your questions above.
 
I have updated the PR to return null.

src/WebJobs.Extensions.DurableTask/Bindings/ObjectValueProvider.cs
```diff
@@ -38,7 +38,7 @@ public Task<object> GetValueAsync()
 
         public string ToInvokeString()
         {
-            return this.value.ToString();
+            return (this.value != null) ? this.value.ToString() : null;
```
cgillum: A more concise fix would be to do `return this.value?.ToString();`, but what you have here is fine too. :)

Issue #118, opened 2017-12-29
Replace "Counter" sample with "Monitor" sample


cgillum:
Related to https://github.com/Azure/azure-functions-durable-extension/issues/67, we need to deprecate the Counter sample/pattern and replace it with something that better suits the strengths of Durable Functions. My thought is that we can replace it with a "Monitor" or "Watcher" sample - e.g. something along the lines of this blog post written by @jeffhollan: https://medium.com/@jeffhollan/serverless-and-bitcoin-creating-price-watchers-dynamically-beea36ef194e. I've seen similar monitoring agent-like use cases from our internal security team so I think this is a pattern that is already resonating with people.

I think the basic outline of the sample orchestration is:

1. Orchestrator that runs in a loop and uses a durable timer to delay between iterations.
2. Uses an activity function to check for some status.
3. If the status check comes back with something actionable, call another activity function (otherwise do nothing).
4. Terminate after some period of time.

Note that the idea is to make this like a template which can be modified to suit people's needs. We shouldn't go into too much detail or complexity when building samples lest they lose track of what they should be learning.

This work item tracks the following:

- [x] Update the [code samples](https://github.com/Azure/azure-functions-durable-extension/tree/dev/samples) and replace the Counter sample with the new Monitor sample.
- [x] Update the [Overview](https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-overview) documentation page, replacing the Counter sample with the new Monitor sample. We probably want to rename pattern 4 accordingly and remove all references to "actors".
- [x] Replace the [Stateful singleton](https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-counter) documentation (which describes the sample) based on the above. It probably needs to be an entirely new article. The existing one should probably be deleted and redirect to the new one.
- [x] Update the [README](https://github.com/Azure/azure-functions-durable-extension/blob/dev/README.md) to point to the new Monitor sample (it does NOT link to the Counter sample currently since I knew we were deprecating it).
- [x] Update the [Install](https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-install) instructions to point to an updated samples .zip.

Other notes:

* We'll continue to support `ContinueAsNew`, but let's not include any usage of it in the overview. It's best to think of this as an advanced but not mainstream tool.
[...]

PR #165, opened 2018-02-22
Replace Counter sample with Monitor sample. #118

kashimiz:
Issue #118. Carried over from PR #145, implements requested changes from @tohling I missed before merge - extended setup guidance in comments and logging.

### DIFF
--- a/samples/precompiled/Monitor.cs
+++ b/samples/precompiled/Monitor.cs
@@ -11,7 +11,27 @@ using Microsoft.Azure.WebJobs.Host;
 using Newtonsoft.Json.Linq;
 using Twilio;
 
-// To run this sample, you'll need a Weather Underground API key. Sign-up and documentation here: https://www.wunderground.com/weather/api/d/docs
+/* This sample demonstrates the Monitor workflow. In this pattern, the orchestrator function is
+ * used to periodically check something's status and take action as appropriate. While a
+ * Timer-triggered function can perform similar polling action, the Monitor has additional
+ * capabilities:
+ *
+ *   - manual termination (via request to the orchestrator termination endpoint)
+ *   - termination when some condition is met
+ *   - monitoring of multiple arbitrary subjects
+ *
+ * To run this sample, you'll need to define the following app settings:
+ *
+ *   - TwilioAccountSid: your Twilio account's SID
+ *   - TwilioAuthToken: your Twilio account's auth token
+ *   - TwilioPhoneNumber: an SMS-capable Twilio number
+ *   - WeatherUndergroundApiKey: a WeatherUnderground API key
+ *
+ * For Twilio trial accounts, you also need to verify the phone number in your MonitorRequest.
+ *
+ * Twilio: https://www.twilio.com
+ * WeatherUnderground API: https://www.wunderground.com/weather/api/d/docs
+ */
 namespace VSSample
 {
     public static class Monitor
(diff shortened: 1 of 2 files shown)


samples/precompiled/Monitor.cs
```diff
@@ -107,7 +145,7 @@ public enum WeatherCondition
         Clear,
         Precipitation,
     }
-    
+
     internal class WeatherUnderground
     {
         private static HttpClient httpClient = new HttpClient();
```
cgillum: nit: we should also make this `readonly`.

samples/precompiled/Monitor.cs
```diff
@@ -135,11 +173,21 @@ internal class WeatherUnderground
             var callString = string.Format("http://api.wunderground.com/api/{0}/conditions/q/{1}/{2}.json", Environment.GetEnvironmentVariable("WeatherUndergroundApiKey"), location.State, location.City);
             var response = await httpClient.GetAsync(callString);
             var conditions = await response.Content.ReadAsAsync<JObject>();
+            var toString = conditions.ToString();
 
             JToken currentObservation;
             if (!conditions.TryGetValue("current_observation", out currentObservation))
             {
-                throw new ArgumentException("Could not find weather for this location. Try being more specific.");
+                var error = conditions.SelectToken("response.error"); //type, description
```
cgillum: nit: prefer using the explicit variable type instead of `var` in cases where the right-hand expression doesn't contain the type name, like here.

samples/precompiled/Monitor.cs
```diff
@@ -135,11 +173,21 @@ internal class WeatherUnderground
             var callString = string.Format("http://api.wunderground.com/api/{0}/conditions/q/{1}/{2}.json", Environment.GetEnvironmentVariable("WeatherUndergroundApiKey"), location.State, location.City);
             var response = await httpClient.GetAsync(callString);
             var conditions = await response.Content.ReadAsAsync<JObject>();
+            var toString = conditions.ToString();
```
cgillum: Is this `toString` variable being used anywhere?

samples/precompiled/Monitor.cs
```diff
@@ -11,7 +11,32 @@
...
+ *   - TwilioPhoneNumber: an SMS-capable Twilio number
+ *   - WeatherUndergroundApiKey: a WeatherUnderground API key
+ *
+ * For Twilio trial accounts, you also need to verify the phone number in your MonitorRequest.
+ *
+ * Twilio: https://www.twilio.com
+ * WeatherUnderground API: https://www.wunderground.com/weather/api/d/docs
+ *
+ * View runtime logs:
+ *   - Locally at C:\Users\{user}\AppData\Local\Temp\LogFiles\Application\Functions
```
cgillum: nit: This location only applies for Windows machines. It will be different for macOS/Linux. We should probably clarify that so macOS/Linux customers don't feel left out.

samples/precompiled/Monitor.cs
```diff
@@ -11,7 +11,32 @@
 using Newtonsoft.Json.Linq;
 using Twilio;
 
-// To run this sample, you'll need a Weather Underground API key. Sign-up and documentation here: https://www.wunderground.com/weather/api/d/docs
+/* This sample demonstrates the Monitor workflow. In this pattern, the orchestrator function is
```
cgillum: I really like these detailed overview comments. We should definitely start doing more of this in our other samples.

samples/precompiled/Monitor.cs
```diff
@@ -20,26 +45,37 @@ public static class Monitor
         public static async Task Run([OrchestrationTrigger] DurableOrchestrationContext monitorContext, TraceWriter log)
         {
             MonitorRequest input = monitorContext.GetInput<MonitorRequest>();
+            log.Info($"Received monitor request. Location: {input.Location}. Phone: {input.Phone}.");
```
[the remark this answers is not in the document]
kashimiz: Agreed; adding IsReplaying checks cluttered up the code and are likely to be something people (like me!) forget to do. I did leave off the check on the very last log statement, as it comes at the end of the orchestrator and should only run once.

samples/precompiled/Monitor.cs
kashimiz: Moving this into a PR for the official docs so it's more visible/centralized.

samples/precompiled/Monitor.cs
```diff
@@ -20,26 +40,54 @@ public static class Monitor
         public static async Task Run([OrchestrationTrigger] DurableOrchestrationContext monitorContext, TraceWriter log)
         {
             MonitorRequest input = monitorContext.GetInput<MonitorRequest>();
+            if (!monitorContext.IsReplaying)
```
cgillum: While I normally advise against this, I will make an exception here: the code would be more readable if we made all these conditional log statements a one-liner: 
```csharp
if (!monitorContext.IsReplaying) log.Info(...);
```

Issue #143, opened 2018-01-24
Attempting to poll the result of CreateCheckStatusResponse can return 404


mark-r-cullen:
Hi,

Running beta-2. I have run in to the same issue as #99 . It seems that the Azure Function can return before something, somewhere, is initialized, even with the RetryAfter set (admittedly to "just" 1 second).

e.g

```
var request = await req.Content.ReadAsAsync<...>().ConfigureAwait(false);

var instanceId = await starter.StartNewAsync("Orchestrator", request).ConfigureAwait(false);

var response = starter.CreateCheckStatusResponse(req, instanceId);
response.Headers.RetryAfter = new RetryConditionHeaderValue(TimeSpan.FromSeconds(1.0));

return response;
```

If the caller then polls the response Location header url, even taking in to account the initial 1.0 second delay, sometimes it will still get a 404.

Setting the RetryAfter to something larger, say 10 seconds, just to avoid this seems like a bit of a poor workaround? I would expect to be able to poll the URL in the location header immediately?

cgillum:
Yes, ideally one should be able to query the instance immediately after querying it. Thanks for opening this as a separate issue. I think this will be an important fix for us to include in the GA release.

PR #168, opened 2018-02-26
Make StartNewAsync return only when the instance has started running

gled4er:
Hello @cgillum ,

Please find in this pull request my suggestion for adding  polling to `StartNewAsync`  in order to wait until the orchestration actually starts as described in this issue - https://github.com/Azure/azure-functions-durable-extension/issues/143 

I went ahead and tried to define some safety mechanisms that we did not discuss yet. I set retry period of 500 ms and maximum waiting time of 10 seconds. 

What do you think about these values? I think we can make them configurable but since we think this is temp solution I think we should simplify the experience for the developer. 

I remember that you said that we shouldn't call storage too often so I set the relatively high retry interval of 500 ms. 

One thing that I saw while testing with samples is that the response in terms of `DurableOrchestrationStatus` is coming relatively late. If this is acceptable trade-off we need to explain it well in the documentation. 

Let me know how I can improve the current code and optimize the performance further.

Thank you!

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Diagnostics;
 using System.Net.Http;
 using System.Threading.Tasks;
 using DurableTask.Core;
@@ -50,7 +51,7 @@ namespace Microsoft.Azure.WebJobs
         }
 
         /// <inheritdoc />
-        public override async Task<string> StartNewAsync(string orchestratorFunctionName, string instanceId, object input, bool waitUntilOrchestrationStarts = false)
+        public override async Task<string> StartNewAsync(string orchestratorFunctionName, string instanceId, object input, bool waitUntilOrchestrationStarts = true)
         {
             this.config.AssertOrchestratorExists(orchestratorFunctionName, DefaultVersion);
 
@@ -69,12 +70,11 @@ namespace Microsoft.Azure.WebJobs
             if (waitUntilOrchestrationStarts)
             {
                 DurableOrchestrationStatus status = await this.GetStatusAsync(instance.InstanceId);
-                int counter = 0;
-                while ((status == null || status.RuntimeStatus == OrchestrationRuntimeStatus.Pending) && counter < 20)
+                Stopwatch stopwatch = Stopwatch.StartNew();
+                while ((status == null || status.RuntimeStatus == OrchestrationRuntimeStatus.Pending) && stopwatch.Elapsed < TimeSpan.FromSeconds(10))
                 {
-                    await Task.Delay(500);
+                    await Task.Delay(200);
                     status = await this.GetStatusAsync(instanceId);
-                    counter++;
                 }
             }
 
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
@@ -53,7 +53,7 @@ namespace Microsoft.Azure.WebJobs
         }
 
         /// <inheritdoc />
-        public override async Task<string> StartNewAsync(string orchestratorFunctionName, string instanceId, object input, bool waitUntilOrchestrationStarts = true)
+        public override async Task<string> StartNewAsync(string orchestratorFunctionName, string instanceId, object input)
         {
             this.config.AssertOrchestratorExists(orchestratorFunctionName, DefaultVersion);
 
@@ -69,15 +69,17 @@ namespace Microsoft.Azure.WebJobs
                 functionType: FunctionType.Orchestrator,
                 isReplay: false);
 
-            if (waitUntilOrchestrationStarts)
+            DurableOrchestrationStatus status = await this.GetStatusAsync(instance.InstanceId);
+            Stopwatch stopwatch = Stopwatch.StartNew();
+            while ((status == null || status.RuntimeStatus == OrchestrationRuntimeStatus.Pending) && stopwatch.Elapsed < TimeSpan.FromSeconds(30))
             {
-                DurableOrchestrationStatus status = await this.GetStatusAsync(instance.InstanceId);
-                Stopwatch stopwatch = Stopwatch.StartNew();
-                while ((status == null || status.RuntimeStatus == OrchestrationRuntimeStatus.Pending) && stopwatch.Elapsed < TimeSpan.FromSeconds(10))
-                {
-                    await Task.Delay(200);
-                    status = await this.GetStatusAsync(instanceId);
-                }
+                await Task.Delay(200);
+                status = await this.GetStatusAsync(instance.InstanceId);
+            }
+
+            if (status == null || status.RuntimeStatus == OrchestrationRuntimeStatus.Pending)
+            {
+                throw new TimeoutException($"Timeout expired while waiting for the new instance to start. This can happen if the task hub is overloaded or if the orchestration host failed to process the start message. Please check the orchestration logs to see whether an internal failure may have occurred. Instance ID: {instance.InstanceId}");
             }
 
             return instance.InstanceId;
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
@@ -50,7 +50,7 @@ namespace Microsoft.Azure.WebJobs
         }
 
         /// <inheritdoc />
-        public override async Task<string> StartNewAsync(string orchestratorFunctionName, string instanceId, object input)
+        public override async Task<string> StartNewAsync(string orchestratorFunctionName, string instanceId, object input, bool waitUntilOrchestrationStarts = false)
         {
             this.config.AssertOrchestratorExists(orchestratorFunctionName, DefaultVersion);
 
@@ -65,6 +65,18 @@ namespace Microsoft.Azure.WebJobs
                 reason: "NewInstance",
                 functionType: FunctionType.Orchestrator,
                 isReplay: false);
+
+            if (waitUntilOrchestrationStarts)
+            {
+                DurableOrchestrationStatus status = await this.GetStatusAsync(instance.InstanceId);
+                int counter = 0;
+                while ((status == null || status.RuntimeStatus == OrchestrationRuntimeStatus.Pending) && counter < 20)
+                {
+                    await Task.Delay(500);
+                    status = await this.GetStatusAsync(instanceId);
+                    counter++;
+                }
+            }
 
             return instance.InstanceId;
         }
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
@@ -2,10 +2,12 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Net.Http;
 using System.Threading.Tasks;
 using DurableTask.Core;
+using DurableTask.Core.History;
 using Microsoft.Azure.WebJobs.Extensions.DurableTask;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
@@ -124,7 +126,7 @@ namespace Microsoft.Azure.WebJobs
         }
 
         /// <inheritdoc />
-        public override async Task<DurableOrchestrationStatus> GetStatusAsync(string instanceId)
+        public override async Task<DurableOrchestrationStatus> GetStatusAsync(string instanceId, bool showHistory = false, bool showHistoryOutput = false)
         {
             OrchestrationState state = await this.client.GetOrchestrationStateAsync(instanceId);
             if (state == null)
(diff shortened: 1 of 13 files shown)


src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -50,7 +50,7 @@ public override HttpResponseMessage CreateCheckStatusResponse(HttpRequestMessage
         }
 
         /// <inheritdoc />
-        public override async Task<string> StartNewAsync(string orchestratorFunctionName, string instanceId, object input)
+        public override async Task<string> StartNewAsync(string orchestratorFunctionName, string instanceId, object input, bool waitUntilOrchestrationStarts = false)
```
cgillum: I wonder if we should set `waitUntilOrchestrationStarts = true` by default. The problem we have now is that the behavior does not match peoples' expectations. If we wait by default, then we will meet people's expectations and I suspect it will match the default behavior for the longer-term fix. For those who don't care about waiting, they can set this to `false` explicitly. Thoughts?

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -66,6 +66,18 @@ public override HttpResponseMessage CreateCheckStatusResponse(HttpRequestMessage
                 functionType: FunctionType.Orchestrator,
                 isReplay: false);
 
+            if (waitUntilOrchestrationStarts)
+            {
+                DurableOrchestrationStatus status = await this.GetStatusAsync(instance.InstanceId);
+                int counter = 0;
```
cgillum: The problem with using a loop and a counter is that the timeout cannot be measured accurately. For example, if `GetStatusAsync` is slow for any reason, callers might end up waiting much longer than 10 seconds. Instead, I suggest using a `Stopwatch`. For example:

```csharp
var stopwatch = Stopwatch.StartNew();
while ((status == null || status.RuntimeStatus == OrchestrationRuntimeStatus.Pending) && stopwatch.Elapsed < TimeSpan.FromSeconds(10))
{
    await Task.Delay(200);
    status = await this.GetStatusAsync(instanceId);
}
```

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -66,6 +66,18 @@ public override HttpResponseMessage CreateCheckStatusResponse(HttpRequestMessage
...
+                while ((status == null || status.RuntimeStatus == OrchestrationRuntimeStatus.Pending) && counter < 20)
+                {
+                    await Task.Delay(500);
```
cgillum: I would reduce this a little bit - maybe 200 milliseconds?

gled4er:
Hello @cgillum ,

Thank you for the great feedback. I will work on fixing the issues as soon as possible!

Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
gled4er: Sounds great! Initially I configured 300 ms but 200 ms sounds even better.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
gled4er: Thank you for the code snippet! Great improvement!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
gled4er: Great! I really like this approach. You are right, developers expect this as the default behavior. Thank you!

gled4er:
Hello @cgillum,

I will implement the changes and then I will give you more information about the delay I described earlier. 

Thank you!

gled4er:
Hello @cgillum ,

I fixed the issues you highlighted in the review.

Please let me know what I should fix next.

In terms of the behavior, when I run the sample locally I would see the function complete first (for the basic `E1_HelloSequence` scenario)  outputting:

```bash
[2/27/2018 4:10:55 AM] Function completed (Success, Id=93ab102b-c116-491f-8829-852304667284, Duration=59ms)
[2/27/2018 4:10:55 AM] Executed 'E1_HelloSequence' (Succeeded, Id=93ab102b-c116-491f-8829-852304667284)
[2/27/2018 4:11:04 AM] It took 00:00:10.2877880 to complete StartNewAsync.
[2/27/2018 4:11:04 AM] Started orchestration with ID = '9ce536cf2d8c4da58ff11e516161f227'.
[2/27/2018 4:11:04 AM] Function completed (Success, Id=ee7a9d8a-965a-45e0-bd73-da1e2a6a78ab, Duration=10301ms)
[2/27/2018 4:11:04 AM] Executed 'HttpStart' (Succeeded, Id=ee7a9d8a-965a-45e0-bd73-da1e2a6a78ab)
[2/27/2018 4:11:04 AM] Executed HTTP request: {
[2/27/2018 4:11:04 AM]   "requestId": "f5a339b8-4781-40ae-bba6-5ac27264da0d",
[2/27/2018 4:11:04 AM]   "method": "POST",
[2/27/2018 4:11:04 AM]   "uri": "/orchestrators/E1_HelloSequence",
[2/27/2018 4:11:04 AM]   "authorizationLevel": "Anonymous"
[2/27/2018 4:11:04 AM] }
[2/27/2018 4:11:04 AM] Response details: {
[2/27/2018 4:11:04 AM]   "requestId": "f5a339b8-4781-40ae-bba6-5ac27264da0d",
[2/27/2018 4:11:04 AM]   "status": "Accepted"
[2/27/2018 4:11:04 AM] }
```
and only then I am able to receive the standard output:

```bash
HTTP/1.1 202 Accepted
Content-Type: application/json; charset=utf-8
Date: Tue, 27 Feb 2018 04:11:04 GMT
Location: http://localhost:7071/admin/extensions/DurableTaskExtension/instances/9ce536cf2d8c4da58ff11e516161f227?taskHub=SampleHubVS&connection=Storage&code=MhvDuQhOn/TnAE6IqzXrBbv0OoaadRD5wVMYE7SjhT4oFDwfSltMtQ==
Retry-After: 10
Server: Microsoft-HTTPAPI/2.0
Transfer-Encoding: chunked

{
    "id": "9ce536cf2d8c4da58ff11e516161f227",
    "sendEventPostUri": "http://localhost:7071/admin/extensions/DurableTaskExtension/instances/9ce536cf2d8c4da58ff11e516161f227/raiseEvent/{eventName}?taskHub=SampleHubVS&connection=Storage&code=MhvDuQhOn/TnAE6IqzXrBbv0OoaadRD5wVMYE7SjhT4oFDwfSltMtQ==",
    "statusQueryGetUri": "http://localhost:7071/admin/extensions/DurableTaskExtension/instances/9ce536cf2d8c4da58ff11e516161f227?taskHub=SampleHubVS&connection=Storage&code=MhvDuQhOn/TnAE6IqzXrBbv0OoaadRD5wVMYE7SjhT4oFDwfSltMtQ==",
    "terminatePostUri": "http://localhost:7071/admin/extensions/DurableTaskExtension/instances/9ce536cf2d8c4da58ff11e516161f227/terminate?reason={text}&taskHub=SampleHubVS&connection=Storage&code=MhvDuQhOn/TnAE6IqzXrBbv0OoaadRD5wVMYE7SjhT4oFDwfSltMtQ=="
}
```

So I think this is happening because we need most of the time for starting the orchestration and then it is really fast to complete even before we display the standard output with the 3 HTTP endpoints. 

In general, I think this is OK but we need to communicate it clearly to the developers that got used to getting HTTP endpoints response immediately. 

Most of the time during testing locally I am seeing result around 10 seconds for completing `StartNewAsync`. 

Let me know if you have any concerns.

Thank you!

gled4er:
Hello @cgillum ,

I simulated delay in `E1_SayHello` and the new behavior is working as expected. 

It is taking some time to get the standard HTTP Endpoint response from the **Orchestration Client** but  if the **Activities** have more realistic scenario it will take even more time for the **Orchestrator** to complete. So I think everything is OK.

Thank you!

test/DurableTaskEndToEndTests.cs
```diff
@@ -58,6 +58,36 @@ public DurableTaskEndToEndTests(ITestOutputHelper output)
             }
         }
 
+        /// <summary>
+        /// End-to-end test which validates a simple orchestrator function which doesn't call any activity functions.
+        /// </summary>
+        [Fact]
+        public async Task HelloWorldOrchestration_StartNewAsync_Polling()
```
cgillum: If you agree that we should remove the `waitUntilOrchestrationStarts`, then I think we don't need a new unit test. Instead, we can simply add your Stopwatch logic and Elapsed validation to the existing unit test.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -68,6 +69,17 @@ public override HttpResponseMessage CreateCheckStatusResponse(HttpRequestMessage
                 functionType: FunctionType.Orchestrator,
                 isReplay: false);
 
+            if (waitUntilOrchestrationStarts)
+            {
+                DurableOrchestrationStatus status = await this.GetStatusAsync(instance.InstanceId);
+                Stopwatch stopwatch = Stopwatch.StartNew();
+                while ((status == null || status.RuntimeStatus == OrchestrationRuntimeStatus.Pending) && stopwatch.Elapsed < TimeSpan.FromSeconds(10))
```
cgillum: Lets change the wait time to 30 seconds and throw a timeout exception if it still hasn't started. Realistically, it should never take longer than 30 seconds to have an instance start and reach it's first checkpoint (which is when we would be able to see it's status). I worry that 10 seconds may be too short and 30 seconds is a common value to use for timeouts.

The reason I think a timeout would be better is because after 30 seconds something is obviously wrong and the user should be explicitly notified about this. The error message should say something like `$"Timeout expired while waiting for the new instance to start. This can happen if the task hub is overloaded or if the orchestration host failed to process the start message. Please check the orchestration logs to see whether an internal failure may have occurred. Instance ID: {instanceId}"`

gled4er:
Hello @cgillum ,

I think you decision is the correct and we should not change the public API especially because we will have better internal implementation soon.

I will make the changes soon and I will let you now.

Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -52,7 +53,7 @@ public override HttpResponseMessage CreateCheckStatusResponse(HttpRequestMessage
         }
 
         /// <inheritdoc />
-        public override async Task<string> StartNewAsync(string orchestratorFunctionName, string instanceId, object input)
+        public override async Task<string> StartNewAsync(string orchestratorFunctionName, string instanceId, object input, bool waitUntilOrchestrationStarts = true)
```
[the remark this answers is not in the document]
gled4er: Sure. I will remove it. I am glad that we migrating to this new approach.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
gled4er: I think this change is very good because it simplifies the workflow and also follows best practices as you mentioned. Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClientBase.cs
```diff
@@ -35,12 +35,12 @@ public abstract class DurableOrchestrationClientBase
         /// </summary>
         /// <param name="orchestratorFunctionName">The name of the orchestrator function to start.</param>
         /// <param name="input">JSON-serializeable input value for the orchestrator function.</param>
-        /// <param name="waitUntilOrchestrationStarts">If set to true, the method will wait (up to 10 seconds) until it can confirm that the orchestration instance has started.</param>
+        /// <param name="waitUntilOrchestrationStarts">The method will wait (up to 10 seconds) until it can confirm that the orchestration instance has started.</param>
```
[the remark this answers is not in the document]
gled4er: I will remove this comment as part of eliminating the parameters as we agreed above.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClientBase.cs
```diff
@@ -35,13 +35,14 @@ public abstract class DurableOrchestrationClientBase
         /// </summary>
         /// <param name="orchestratorFunctionName">The name of the orchestrator function to start.</param>
         /// <param name="input">JSON-serializeable input value for the orchestrator function.</param>
+        /// <param name="waitUntilOrchestrationStarts">The method will wait (up to 10 seconds) until it can confirm that the orchestration instance has started.</param>
         /// <returns>A task that completes when the start message is enqueued.</returns>
         /// <exception cref="ArgumentException">
         /// The specified function does not exist, is disabled, or is not an orchestrator function.
         /// </exception>
-        public virtual Task<string> StartNewAsync(string orchestratorFunctionName, object input)
+        public virtual Task<string> StartNewAsync(string orchestratorFunctionName, object input, bool waitUntilOrchestrationStarts = true)
```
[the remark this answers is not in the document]
gled4er: Sure. Thank you for the clarification.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClientBase.cs
```diff
@@ -51,12 +51,12 @@ public virtual Task<string> StartNewAsync(string orchestratorFunctionName, objec
         /// <param name="orchestratorFunctionName">The name of the orchestrator function to start.</param>
         /// <param name="instanceId">A unique ID to use for the new orchestration instance.</param>
         /// <param name="input">JSON-serializeable input value for the orchestrator function.</param>
-        /// <param name="waitUntilOrchestrationStarts">If set to true, the method will wait (up to 10 seconds) until it can confirm that the orchestration instance has started.</param>
+        /// <param name="waitUntilOrchestrationStarts">The method will wait (up to 10 seconds) until it can confirm that the orchestration instance has started.</param>
```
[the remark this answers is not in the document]
gled4er: I will remove this comment as part of eliminating the parameters as we agreed above.

gled4er:
Hello @cgillum ,

Thank you for the quick review. I don't plan to add anything else.

Thank you!

Issue #130, opened 2018-01-11
Orchestration termination reason not shown


markheath:
I tried calling the `terminatePostUri` method passing in a `reason` query string parameter of "oops". It successfully terminated my orchestration, but the reason doesn't appear either in the reponse to the `statusQueryGetUri` or anywhere in the logs (I'm testing locally). 

Here's how I called the terminate uri:

```
Invoke-WebRequest -Method Post "http://localhost:7071/admin/extensions/DurableTaskExtension/instances/00
57dcd4b7be4ebc902716d609e665f4/terminate?reason=oops&taskHub=DurableFunctionsHub&connection=Storage&code=7LsL56Ro1sQHneX
F9Ei1OSAi8MVyIoj5v5CzswPCASlWAMDa3TRElw=="
```

The termination message in the logs just shows a blank reason:

```
[11/01/2018 21:42:18] 0057dcd4b7be4ebc902716d609e665f4: Function 'ProcessVideoOrchestrator (Orchestrator)', version '' was terminated. Reason: . State: Terminated. HubName: DurableFunctionsHub. AppName: . SlotName: . ExtensionVersion: 1.1.0.0.
```

It would be nice for this reason at least to appear in the logs, and also could be useful to be in the response to the `getStatusUri`. Is there something different I should be doing?

cgillum:
That's odd. I would have expected it to show up. Do you at least see it in the Azure Storage table?

markheath:
hmm, struggling to find it in local table storage either in the logs or in the hub history table

cgillum:
Sorry for the delay. I've confirmed that the issue is a query string parsing bug. Thanks for reporting this, it will be fixed in the next release. As a workaround, you can use the `TerminateAsync` API directly, like this:

```csharp
[FunctionName("Terminate")]
public static async Task Terminate(
    [HttpTrigger(AuthorizationLevel.Anonymous, "POST")] TerminatePayload payload,
    [OrchestrationClient] DurableOrchestrationClient client)
{
    await client.TerminateAsync(payload.InstanceId, payload.Reason);
}

public class TerminatePayload
{
    public string InstanceId { get; set; }
    public string Reason { get; set; }
}
```

PR #171, opened 2018-02-27
Fix query string parsing bug and add regression test

cgillum:
This PR fixes https://github.com/Azure/azure-functions-durable-extension/issues/130, which was caused by a query string parsing bug.

For easier mocking, I did some light refactoring in the product code to replace usage of `DurableOrchestrationClient` with `DurableOrchestrationClientBase`.

Apologies for the diff on HttpApiHandlerTests.cs.  It had inconsistent line endings and fixing that with Visual Studio made it look like the whole file changed. Turning off whitespace diffing should make it easier to view the *actual* diff.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/HttpRequestMessageExtensions.cs
+++ b/src/WebJobs.Extensions.DurableTask/HttpRequestMessageExtensions.cs
@@ -68,7 +68,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             var separator = '&';
 
             int l = (s != null) ? s.Length : 0;
-            int i = 0;
+            int i = 1; // skip the leading '?' character
 
             while (i < l)
             {
--- a/test/HttpApiHandlerTests.cs
+++ b/test/HttpApiHandlerTests.cs
@@ -201,7 +201,7 @@ namespace WebJobs.Extensions.DurableTask.Tests
         }
 
         [Fact]
-        public async Task TerminateInstance()
+        public async Task TerminateInstanceWebhook()
         {
             string testInstanceId = Guid.NewGuid().ToString("N");
             string testReason = "TerminationReason" + Guid.NewGuid();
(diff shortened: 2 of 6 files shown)


test/HttpApiHandlerTests.cs
```diff
@@ -1,202 +1,272 @@
...
+                .Returns(Task.FromResult(
+                    new DurableOrchestrationStatus
+                    {
+                        InstanceId = testInstanceId,
+                        RuntimeStatus = OrchestrationRuntimeStatus.Running,
+                    }));
+
+            var terminateRequestUriBuilder = new UriBuilder(TestConstants.NotificationUrl);
+            terminateRequestUriBuilder.Path += $"/Instances/{testInstanceId}/terminate";
+            terminateRequestUriBuilder.Query = $"?reason={testReason}&{terminateRequestUriBuilder.Query.TrimStart('?')}";
```
kashimiz: UriBuilder.Query automatically prepends a ? to whatever is assigned to it. The current code will result in "{path}/??reason={etc}".

src/WebJobs.Extensions.DurableTask/HttpRequestMessageExtensions.cs
```diff
@@ -68,7 +68,7 @@ public static NameValueCollection GetQueryNameValuePairs(this HttpRequestMessage
             var separator = '&';
 
             int l = (s != null) ? s.Length : 0;
-            int i = 0;
+            int i = 1; // skip the leading '?' character
```
kashimiz: nit: s.TrimStart('?') feels more readable to me; is there a reason we don't do this?

cgillum: Yeah, mainly for low level performance. TrimStart('?') will allocate a new string object, and I was hoping to avoid that.

test/HttpApiHandlerTests.cs
cgillum: Hmm...when i looked at it in the debugger I thought it was resolving correctly. I'll look at it again. I'll admit I didn't really understand the behavior of this API.

PR #172, opened 2018-02-28
Support for binding expressions in activity functions

### LINKED ISSUE #123: Support for declarative binding expressions
With normal Azure Functions, I'm able to bind to do some nice declarative work:

        [FunctionName("InjestAndProcessFile")]
        public static async Task Run(
            [BlobTrigger("test/{name}")] Stream input, 
            string name,
            [Blob("archive/0/test/{name}", FileAccess.Write)] Stream output)
        {
            Console.WriteLine($"Archiving {name}...");
            await input.CopyToAsync(output);
        }

However, if I wanted to use Durable Functions, I have to switch an imperative model, which is much noisier:

        [FunctionName("InjestFile")]
        public static async Task Run(
            [BlobTrigger("test/{name}")] Stream input,    // BlobTrigger is required, but the stream is useless here
            string name,
            [OrchestrationClient] DurableOrchestrationClient starter)
        {
            await starter.StartNewAsync("ProcessFile", name);
        }

        [FunctionName("ProcessFile")]
        public static async Task Run(
            [OrchestrationTrigger] DurableOrchestrationContext context,
            IBinder binder)
        {
            var name = context.GetInput<string>();
            using (var input = await binder.BindAsync<Stream>(new BlobAttribute($"test/{name}", FileAccess.Read)))    // For some reason I'm getting an error here: "Cannot access a disposed object"
            using (var output = await binder.BindAsync<Stream>(new BlobAttribute($"archive/0/test/{name}", FileAccess.Write)))
            {
                Console.WriteLine($"Archiving {name}...");
                await input.CopyToAsync(output);
            }
        }

This is just one sample off the top of my head, but you can see how quickly the existing bindings are pushed aside to make way for the Orchestration bindings. Is there anything that can be done to preserve  the existing bindings while still using Durable Functions. The bindings are such a great feature of Azure Functions and I'd like to see them continue to shine.

cgillum:
This PR fixes #123 (enable binding expressions in activity functions. it allows the function author to bind to the name of the parameter or to one of the top-level fields in the activity payload.

### Example 1: Binding to the parameter name
This activity function takes a string argument (the name of a blob) and uses that value to bind to blob input and output bindings.

**Activity Function**
```csharp
public static async Task ArchiveBlob(
    [ActivityTrigger] string name,
    [Blob("test/{name}", FileAccess.Read)] Stream input,
    [Blob("test/{name}-archive", FileAccess.Write)] Stream output)
{
    await input.CopyToAsync(output);
}
```

**Orchestrator Function**
```csharp
public static async Task Orchestrator([OrchestrationTrigger] DurableOrchestrationContext ctx)
{
    string blobName = ctx.GetInput<string>();
    await ctx.CallActivityAsync("ArchiveBlob", blobName);
}
```

### Example 2: Binding to the activity data payload
This activity function takes a custom POCO argument (it could also be a JSON argument) and uses the data contents to bind to blob inputs and outputs. Note the use of `data` in the binding expression, which always acts as a reference to the activity argument data.

**Activity Function**
```csharp
public static async Task CopyBlob(
    [ActivityTrigger] JObject ignored,
    [Blob("test/{data.InputPrefix}-{data.Suffix}", FileAccess.Read)] Stream input,
    [Blob("test/{data.OutputPrefix}-{data.Suffix}", FileAccess.Write)] Stream output)
{
    await input.CopyToAsync(output);
}
```

**Orchestrator Function**
```csharp
public static async Task Orchestrator([OrchestrationTrigger] DurableOrchestrationContext ctx)
{
    var data = new
    {
        InputPrefix = "Input",
        OutputPrefix = "Output",
        Suffix = 42,
    };

    await ctx.CallActivityAsync("CopyBlob", data);
}
```

Both of these examples are verified in the new unit tests that are included in this PR.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
@@ -63,18 +63,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private class ActivityTriggerBinding : ITriggerBinding
         {
-            private static readonly IReadOnlyDictionary<string, Type> StaticBindingContract =
-                new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase)
-                {
-                    // This binding supports return values of any type
-                    { "$return", typeof(object).MakeByRefType() },
-                    { nameof(DurableActivityContext.InstanceId), typeof(string) },
-                };
+            private const string DataBindingPropertyName = "data";
 
             private readonly ActivityTriggerAttributeBindingProvider parent;
             private readonly ParameterInfo parameterInfo;
             private readonly ActivityTriggerAttribute attribute;
             private readonly FunctionName activityName;
+            private readonly IReadOnlyDictionary<string, Type> contract;
 
             public ActivityTriggerBinding(
                 ActivityTriggerAttributeBindingProvider parent,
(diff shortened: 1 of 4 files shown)


src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
```diff
@@ -86,11 +81,47 @@ private class ActivityTriggerBinding : ITriggerBinding
...
+            private static IReadOnlyDictionary<string, Type> GetBindingDataContract(ParameterInfo parameterInfo)
+            {
+                var contract = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase)
+                {
+                    // This binding supports return values of any type
+                    { "$return", typeof(object).MakeByRefType() },
+                    { nameof(DurableActivityContext.InstanceId), typeof(string) },
+                };
+
+                if (IsSimple(parameterInfo.ParameterType))
```
MikeStall: What is this section doing? The reflection code in IsSimple is a bit scary

test/BindingTests.cs
```diff
@@ -100,5 +103,114 @@ public BindingTests(ITestOutputHelper output)
...
         }
+
+        [Fact]
+        public async Task BindToBlobViaParameterName()
+        {
+            using (JobHost host = TestHelpers.GetJobHost(this.loggerFactory, nameof(this.ActivityTriggerAsNumber)))
+            {
+                await host.StartAsync();
+
+                string connectionString = AmbientConnectionStringProvider.Instance.GetConnectionString(ConnectionStringNames.Storage);
```
MikeStall: FYI, we could simplify this test to avoid Blobs and be totally in-memory. 
We have a short binding extension (you could copy from https://github.com/Azure/azure-webjobs-sdk/pull/1513  ) that operates entirely in-memory.

src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
cgillum: The intent is to check and see if the input parameter is a "simple type" that can be used for binding - e.g. a primitive, string, decimal or a nullable of one of these types. The main thing I want to avoid is allowing users to bind to complex types which can't easily be converted to a meaningful string value.  Do you think this check is overkill? I wasn't sure what the underlying binding infrastructure does if a user tries to use a random complex type in a binding expression.

test/BindingTests.cs
cgillum: Good to know. One thing I like about end-to-end tests, though, is that they sometimes make for good mini-samples.

src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
MikeStall: I'd prefer we skip the check here (it's very policy specific); and can push it down to the SDK level (so that it's consistent and all the extensions are doing the same thing).

PR #191, opened 2018-03-14
add comment with copyright & license information

mburleigh:
All the other .cs files had the copyright & license at the top of the file. Please reject if the omission was intentional.


samples/precompiled/HttpSyncStart.cs
```diff
@@ -1,3 +1,6 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
```
cgillum: Thanks for finding this. Actually, the name of the license file is LICENSE, so the header should be:

```
```

mburleigh:
Thanks @cgillum  - I fixed that. I'll check the rest of the files for the same issue.

Issue #192, opened 2018-03-14
Yield values from a long-lived orchestrator function


ethanroday:
**Scenario:**

I have an application, and I want to strategically prompt users to share the application with their friends. In order to increase the chance of success and not annoy my users, I want to target a specific set of users based on their engagement/other behavior over a long period of time (multiple days), and only show the prompt when I think it will not disrupt them.
Modeling this interaction flow using durable functions seems like an intuitive choice. Based on some initial trigger, I can start up an orchestrator function, have the client send interaction events, and complete the orchestrator function with a directive to take an action (show the prompt) or not. The orchestrator function encapsulates all of the logic on the server side, and the client-server interaction is generic, meaning I can try out variations on the scenario as the need arises without having to update client code, and I can create multi-platform scenarios very easily.

**Problem:**

At any given time, the client only needs to notify the orchestrator function of a few specific events (ones that are relevant to the current state of the interaction flow). I'd like to avoid having clients send every possible event, since that would result in a lot of unnecessary network requests.
From the client perspective, after the initial triggering, the ideal would be to call an API that takes an event and associated payload as input and comes back with 1) the next set of events to send and 2) any intermediate actions that the client should take.
The orchestrator function knows what those specific events and actions are, but how can it communicate them to the client? The HTTP API doesn't allow arbitrary data to be passed back in the status response. The best that I can think of is to create an HTTP-triggered activity function that does the following:
1. The activity calls `StartNewAsync` on a short-lived orchestrator `Messenger`.
2. `Messenger` calls `RaiseEventAsync` on the main orchestrator function.
3. The main orchestrator function sends a message back to `Messenger` with `RaiseEventAsync` with information about the next set of events and actions for the client.
4. `Messenger` completes with the payload from the main orchestrator function.
5. The activity, which has been waiting for `Messenger` to finish (#110), returns the response to the client.

Is there a more straightforward way to accomplish this?

cgillum:
I think what you may be looking for is this: https://github.com/Azure/azure-functions-durable-extension/issues/96, but with a little more richness than a simple string.

Essentially, it allows an orchestration to expose arbitrary data to the outside world while it is still running. You can imagine the orchestrator function would have a `context.SetCustomStatus(data)` method where `data` is some arbitrary payload which gets exposed to clients as JSON.  Clients can then read this custom status data by making calls to get the orchestrator status.  When using the HTTP status query API, it would look something like this:

**Request:**
This is the existing status query API:
```http
[...]

Issue #96, opened 2017-11-22
Simple custom progress strings


danielearwicker:
Idea: allow orchestration function to emit strings, which are recorded in the event stream of the execution, e.g.

```csharp
context.EmitProgress("going-nicely");
```

The response from the status end-point would include the most recently emitted string:

```json
{
    "runtimeStatus": "Running",
    "input": {},
    "createdTime": "2017-11-22T15:18:03Z",
    "lastUpdatedTime": "2017-11-22T15:19:07Z",
    "progress": "going-nicely" <---
}
```

This would be useful for clients that want to observe the progress of the overall operation during its execution. So I can launch a function, poll the status end-point and display a progress UI that knows which stage the execution has reached so far. It would be up to the app author to decide on a string format.

cgillum:
Thanks for the suggestion. This is a nice idea that I've been thinking about. The Durable Task Framework does have a "status" concept which allows orchestrator logic to expose it's status. I think I just need to think about how to properly expose it.

SimonLuckenuik:
Very good added value to the framework in my view, currently if I want to provide progression like that I am forced to manage it on the side. I came here to raise that suggestion :-) A simple SetProgress on the DurableOrchestrationContext would do the trick for me.

cgillum:
This feature has been merged into the **dev** branch and will go out with the next release. See the description in the linked PR for more details.

PR #193, opened 2018-03-15
Allow orchestrator functions to set a custom status value

cgillum:
This is a feature which allows users to set a custom status value from their orchestrator function, and have it exposed via the HTTP GetStatus API or the `DurableOrchestrationClient.GetStatusAsync` API.

It is intended to address https://github.com/Azure/azure-functions-durable-extension/issues/96 and https://github.com/Azure/azure-functions-durable-extension/issues/192.

Here is a simple example of how it works, starting with the orchestrator function code:

```csharp
public static async Task SetStatusTest([OrchestrationTrigger] DurableOrchestrationContext ctx)
{
    // ...do work...

    // update the status of the orchestration with some arbitrary data
    var customStatus = new { nextActions = new [] {"A", "B", "C"}, foo = 2, };
    ctx.SetCustomStatus(customStatus);

    // ...do more work...
}
```

While the orchestration is running, external clients can fetch this custom status. For example:

```http
GET /admin/extensions/DurableTaskExtension/instances/instance123
```
...yields the following JSON status payload...

```json
{
  "runtimeStatus": "Running",
  "input": null,
  "customStatus": { "nextActions": ["A", "B", "C"], "foo": 2 },
  "output": null,
  "createdTime": "2017-10-06T18:30:24Z",
  "lastUpdatedTime": "2017-10-06T19:40:30Z"
}
```
Right now, I'm limiting this payload to 16 KB of UTF-16 JSON text because it needs to be able to fit in an Azure Table Storage column. We could store more than this, but it seems that if developers need more data for their status, then they should probably externalize it instead of relying on this feature.

To make this work end-to-end, we need to update the **DurableTask.AzureStorage** provider persist these custom status values (the data already flows to this provider, we just need to add a means of storing it somewhere). This can be done as part of a separate PR that @gled4er is working on: https://github.com/Azure/durabletask/pull/145.


src/WebJobs.Extensions.DurableTask/MessagePayloadDataConverter.cs
```diff
@@ -31,15 +32,31 @@ public MessagePayloadDataConverter()
```
gled4er: Can you add comments for this method?

src/WebJobs.Extensions.DurableTask/MessagePayloadDataConverter.cs
```diff
@@ -31,15 +32,31 @@ public MessagePayloadDataConverter()
 
         public override string Serialize(object value)
         {
-            string serializedJson = base.Serialize(value);
+            return this.Serialize(value, MaxMessagePayloadSizeInKB);
+        }
+
+        public string Serialize(object value, int maxSizeInKB)
```
gled4er: Can you add comments for this method?

gled4er:
Hello @cgillum ,

Thank you for inviting me to review this pull request.

May I ask you to check the tests - currently 3 are failing? 

Thank you!

test/DurableTaskEndToEndTests.cs
```diff
@@ -1149,6 +1149,41 @@ private static void CompareTwoComplexTypeObjects(ComplexType firstObject, Comple
             }
         }
 
+        [Fact]
+        public async Task SetStatusOrchestration()
+        {
+            using (JobHost host = TestHelpers.GetJobHost(this.loggerFactory, nameof(this.SetStatusOrchestration)))
```
gled4er: This test is great and it is showing that we can update custom status. Shall we have a simple one covering the basic case on a single set of the status? I think we can reuse `SetStatus` method but also pass the number of iterations. So the simple case will have only a single iteration. I think this is an optional change.

test/TestOrchestrations.cs
```diff
@@ -256,5 +256,17 @@ public static string BigReturnValue([OrchestrationTrigger] DurableOrchestrationC
             int stringLength = ctx.GetInput<int>();
             return new string('*', stringLength);
         }
+
```
gled4er: I think the main thing is to fix the tests and the other points I made are minor. Thank you!

cgillum:
Thanks for the review. The new test is expected to fail until we make the necessary changes to **DurableTask.AzureStorage**. I'll look into the large message test failures.

gled4er:
@cgillum thank you for the explanation. Let me know when the tests are passing. I will be happy to take a look again.

Thank you!

cgillum:
I rebased my PR on top of @tohling's latest changes and that fixed the failing message size tests.

src/WebJobs.Extensions.DurableTask/MessagePayloadDataConverter.cs
cgillum: Comments added.

test/DurableTaskEndToEndTests.cs
cgillum: Do you think a simple one would cover more than what is already covered here? In general, I prefer to write fewer tests that cover more scenarios to avoid having too many tests.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -156,6 +157,26 @@ internal string GetSerializedOutput()
...
+        /// </summary>
+        /// <remarks>
+        /// The <paramref name="customStatusObject"/> value is serialized to JSON and will be made available
+        /// to the orchestration status query APIs.
+        /// </remarks>
+        /// <param name="customStatusObject">The JSON-serializeable value to use as the orchestrator function's custom status.</param>
+        public void SetCustomStatus(object customStatusObject)
+        {
+            // Limit the custom status payload to 16 KB
+            const int MaxPayloadSizeInKB = 16 * 1024;
```
tohling: >const int MaxPayloadSizeInKB = 16 * 1024; [](start = 12, length = 41)

Could we move this up to where all the const declarations are?

tohling:
>                         "The UTF-16 size of the JSON-serialized payload must not exceed 60 KB. The current payload size is {0:N0} KB.",

This should be formatted to whatever "maxSizeInKB" is.

---
Refers to: src/WebJobs.Extensions.DurableTask/MessagePayloadDataConverter.cs:65 in 15041a8. [](commit_id = 15041a8f07dbfd1434616188a11e23acb2ae1783, deletion_comment = False)

src/WebJobs.Extensions.DurableTask/MessagePayloadDataConverter.cs
gled4er: @cgillum Thank you!

test/DurableTaskEndToEndTests.cs
gled4er: @cgillum Thank you for the clarification. Your are approach is very good. In general I see more value in the integration tests and this is great approach for implementing integration tests.

PR #197, opened 2018-03-17
80-85 char length line formatting for better docs embedding.

kashimiz:
Per request of @tdykstra on [this docs pull request](https://github.com/MicrosoftDocs/azure-docs-pr/pull/34331#pullrequestreview-104381919), code files should have a character line limit of 80-85 to prevent horizontal scrolling when viewed as embeds in documentation.

### DIFF
--- a/samples/csx/E2_BackupSiteContent/run.csx
+++ b/samples/csx/E2_BackupSiteContent/run.csx
@@ -2,7 +2,8 @@
 
 public static async Task<long> Run(DurableOrchestrationContext backupContext)
 {
-    string rootDirectory = Environment.ExpandEnvironmentVariables(backupContext.GetInput<string>() ?? "");
+    string rootDirectory = Environment.ExpandEnvironmentVariables(
+        backupContext.GetInput<string>() ?? "");
     if (string.IsNullOrEmpty(rootDirectory))
     {
         throw new ArgumentException("A directory path is required as an input.");
@@ -10,7 +11,9 @@ public static async Task<long> Run(DurableOrchestrationContext backupContext)
 
     if (!Directory.Exists(rootDirectory))
     {
-        throw new DirectoryNotFoundException($"Could not find a directory named '{rootDirectory}'.");
+        throw new DirectoryNotFoundException(string.Format(
+            "Could not find a directory named '{0}'.", 
+            rootDirectory));
     }
 
     string[] files = await backupContext.CallActivityAsync<string[]>(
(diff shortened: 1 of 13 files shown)


samples/csx/E2_BackupSiteContent/run.csx
```diff
@@ -2,15 +2,18 @@
 
 public static async Task<long> Run(DurableOrchestrationContext backupContext)
 {
-    string rootDirectory = Environment.ExpandEnvironmentVariables(backupContext.GetInput<string>() ?? "");
+    string rootDirectory = Environment.ExpandEnvironmentVariables(
```
tohling: Undo all the auto-correction that was done to the spacing.

cgillum:
80-85 characters feels too restrictive. I understand the need to make this narrower for documentation, but I worry it could make code in the editor harder to read. Typically I try to do no more than 120 characters. Is it possible to compromise on this a bit? What screen layout are we optimizing for?

tdykstra:
@cgillum The shorter line length is only a nice-to-have for readability, not a requirement. Whether you go with 80-85 or 120 characters is up to you and @kashimiz . Should we close this PR and leave the code as-is? Or merge it after restoring some of the long lines where chopping them compromised code readability in VS?

cgillum:
Let me take another pass through this. There are pros and cons for both sides. I think it makes sense to do it on a case-by-case basis. I just wish GitHub had an option to show line lengths. :)

cgillum:
Actually @kashimiz did you automate this using a tool (like Resharper)? If so, going at this case-by-case might be counter-productive. Would it make sense to just compromise at 100 characters?

One general piece of feedback is that I'd like to avoid using `string.Format` over `$"..."`, especially if we can just drop the string onto a new line to save some characters.

kashimiz:
I didn’t (although I wish I had.)

Re: string.Format, I went there because string interpolation broke the line length limit I’d imposed, even when dropping the string to a new line.

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

cgillum:
I assume the motivation for shorter line lengths is primarily for smaller devices. I went through the documentation on my iPhone in landscape mode and horizontal scrolling wasn't a problem for the majority of the code snippets. I think having column limits is a good idea, but 80-85 is too restrictive IMO. I could live with 120.

For now, let's close this PR as it has gotten stale and we have several new samples which have since been added. I'm open to revisiting this at a later time.

Issue #71, opened 2017-09-30
Tracking events in Application Insights cannot be ordered reliably


cgillum:
The problem seems to be that some events contain the same timestamp, so they cannot be ordered reliably. We may want to consider adding a "sequence number" value in the application insights trace to make it easy to order the results for a single instance.

We'll need to update the [Diagnostics](https://azure.github.io/azure-functions-durable-extension/articles/topics/diagnostics.html) topic of the documentation once there is a fix/workaround.

cgillum:
Potential solution here: https://stackoverflow.com/a/49139184/2069

cgillum:
Looks like this is a bug in Application Insights: https://github.com/Microsoft/ApplicationInsights-Home/issues/189

cgillum:
This has been included in the [Beta 3.1 release (v1.2.1-beta3)](https://github.com/Azure/azure-functions-durable-extension/releases/tag/v1.2.1-beta3).

PR #206, opened 2018-03-23
Adding ordering to App Insights Events

gled4er:
Hello @cgillum ,

May I ask you to take look on this first attempt to implement this issue - https://github.com/Azure/azure-functions-durable-extension/issues/71

It is possible that I am missing some of the edge cases or additional logic, but I tested with the samples project and then by running the following query against Application Insights:

```sql
let targetInstanceId = "33681f657c0141288fd1c0712dfe5e4a";
let start = datetime(2018-03-23T08:00:00);
traces
| where timestamp > start and timestamp < start + 50m
| where customDimensions.Category == "Host.Triggers.DurableTask"
| extend functionName = customDimensions["prop__functionName"]
| extend instanceId = customDimensions["prop__instanceId"]
| extend state = customDimensions["prop__state"]
| extend isReplay = tobool(tolower(customDimensions["prop__isReplay"]))
| extend sequenceNumber = tolong(customDimensions["prop__sequenceNumber"]) 
| where isReplay == false
| where instanceId == targetInstanceId
| sort by sequenceNumber asc
| project timestamp, functionName, state, instanceId, sequenceNumber, appName = cloud_RoleName
```

I got this output:

image

Let me know if this is acceptable approach.

I am looking forward for your feedback and review comments.

Thank you!

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
@@ -76,8 +76,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 IsReplay: isReplay);
 
             this.logger.LogInformation(
-                "{instanceId}: Function '{functionName} ({functionType})', version '{version}' scheduled. Reason: {reason}. IsReplay: {isReplay}. State: {state}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}.",
-                instanceId, functionName, functionType, version, reason, isReplay, FunctionState.Scheduled, hubName, LocalAppName, LocalSlotName, ExtensionVersion);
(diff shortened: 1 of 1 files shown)


cgillum:
This theoretically works, but given that orchestration execution is single-threaded, there should be a way for us to use a monotonically incrementing integer. I would prefer that over `DateTime.Utc.TickCount` because it's easier to read and more clearly tells you the exact sequence of events.  This could be as simple as adding an integer field to this class and incrementing it for every log call.

gled4er:
Hello @cgillum ,

I replaced the `DateTime.UtcNow.Ticks` with a new `long` sequence counter. You suggested `int` but I decided to actually use long in order to have more values available. 

Let me know if you prefer to change it to `int`.

Please review the latest changes when you have time and let me know what I should fix next.

Thank you!

gled4er:
Hello @cgillum ,

I see that one of the unit tests is failing.

I will focus on fixing it first.

Thank you!

gled4er:
Hello @cgillum ,

After relaxing one test condition, everything is passing. So I hope this additional change is OK with you.

Let me know if you spotted something else I need to fix.

Thank you!

gled4er:
Hello @cgillum ,

Thank you very much for approving and merging the changes over the weekend.

I did very similar change for the Instances table feature.

Thank you!

Issue #210, opened 2018-03-26
Durable Functions extension broken in latest Azure Functions v2 runtime


cgillum:
### Issue

On 3/26/2018, an Azure Functions v2 runtime update was deployed to West US 2.  See [this announcement](https://github.com/Azure/app-service-announcements/issues/94) for more details.  Unfortunately this update contains several breaking changes for various extensions, including the Durable Functions extension.

### Symptoms

The breaking changes will prevent the Functions host from starting if the Durable Functions extension is loaded. The error message you can expect to see looks something like this:

```
The function runtime is unable to start. Microsoft.Azure.WebJobs.Extensions.DurableTask: Method not found: 'Microsoft.Azure.WebJobs.Host.TraceWriter Microsoft.Azure.WebJobs.Host.Config.ExtensionConfigContext.get_Trace()'.
Session Id: e4250f6d669f479b9535a2b676341502

Timestamp: 2018-03-26T23:18:31.462Z
```

This issue will require an update to the Durable Functions extension before it can be used in the latest Functions v2 runtime.

### Workaround(s)
Use Functions v1 instead of v2 (Durable Functions supports both runtime versions).

GregLevenhagen:
Did this hist North Central US in the last day? Not getting the same error message, but it is an issue with DurableTask.

cgillum:
Yes, I believe North Central US was updated yesterday. Just to confirm, what error message are you seeing?

GregLevenhagen:
```
2018-03-28T14:22:50.888 [Information] Loaded custom extension: DurableTaskExtension from 'referenced by: Method='<assembly namespace>.HttpStart', Parameter='starter'.'
2018-03-28T14:22:50.890 [Trace] Extension loading complete.
2018-03-28T14:22:52.133 [Error] A ScriptHost error has occurred
2018-03-28T14:22:52.224 [Information] Stopping Host
```

Just started with our nightly process early in the AM today. No further details.

cgillum:
This has been fixed in the [Beta 3.1 release (v1.2.1-beta3)](https://github.com/Azure/azure-functions-durable-extension/releases/tag/v1.2.1-beta3).

PR #212, opened 2018-03-29
Functions V2 Compatibility Update

cgillum:
Fixes https://github.com/Azure/azure-functions-durable-extension/issues/210. This is fairly urgent since durable functions on v2 is broken in production due to the recent beta5 breaking change roll-out.

Two main changes in this PR:

1. Replace all usage of `TraceWriter` with `ILogger`.
2. Remove all usage of converters in binding trigger providers.

It turns out that these changes actually make the code cleaner by simply removing unnecessary code.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
@@ -29,8 +29,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.durableTaskConfig = durableTaskConfig;
             this.extensionContext = extensionContext;
             this.traceHelper = traceHelper;
-
-            ActivityTriggerBinding.RegisterBindingRules(extensionContext.Config);
         }
 
         public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext context)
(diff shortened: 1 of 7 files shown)


src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -178,9 +181,10 @@ void IExtensionConfigProvider.Initialize(ExtensionConfigContext context)
 
             // Register the trigger bindings
             JobHostConfiguration hostConfig = context.Config;
+            ILogger logger = context.Config.LoggerFactory.CreateLogger(LoggerCategoryName);
```
brettsam: nit: You can use `LogCategories.CreateTriggerCategory("DurableTask")`, which will generate `Host.Triggers.DurableTask`. I made that function so we could keep things consistent. You're already doing that, but thought I'd point it out.

src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
```diff
@@ -115,39 +113,19 @@ public Task<ITriggerData> BindAsync(object value, ValueBindingContext context)
...
+                {
+                    convertedValue = activityContext;
+                }
+                else if (destinationType == typeof(JObject))
+                {
+                    convertedValue = ActivityContextToJObject(activityContext);
+                }
+                else
+                {
+                    convertedValue = activityContext.GetInput(destinationType);
```
brettsam: Were these changes related to the nuget update? Did some public surface go away?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
cgillum: Cool - I like consistency. :) I'll switch to `LogCategories.CreateTriggerCategory("DurableTask")`

src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
cgillum: Yes, previously I relied on `IConverterManager`, but that surface area was removed in the latest nuget update and I had to come up with something else. I searched but couldn't find any trigger examples that converted inputs into any arbitrary type. What I ended up realizing though was that I didn't really need the `IConverterManager` stuff at all. Instead I could simply implement my conversions directly in my binding logic. This surprisingly simplified the code quite a bit, though I'm not sure whether there are any downsides to opting-out of using rules/converters for my trigger bindings.

@MikeStall let me know if you have any thoughts on this as well.

brettsam: The new way would be to do something like this: https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/WebJobs.Extensions.CosmosDB/Config/CosmosDBConfiguration.cs#L82-L86.

But if this works, let's get it in to unblock.

cgillum: That was the code that I spent the most time looking at, but it only showed how to bind to `string` and `JArray`. My main concern was figuring out how to bind to an unbounded set of arbitrary types.

brettsam: That's what `OpenType` is for. The Cosmos trigger doesn't use it since the types are currently restricted, but the other bindings do -- like here: 
- https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/WebJobs.Extensions.CosmosDB/Config/CosmosDBConfiguration.cs#L65
- https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/WebJobs.Extensions.CosmosDB/Config/CosmosDBConfiguration.cs#L172

But I believe this pattern is to build declarative bindings so tooling can determine the list of supported types. The way you're doing it should still work. If you've tested it, go ahead and get it in. You may want to make a note to come back and refactor to the "new" way in a future update.

cgillum: I've seen that as well, but I've only seen `OpenType` used in the context of collections and not for singular inputs. Anyways, we can follow up on that later. I'll go with what I have for now.

Issue #205, opened 2018-03-23
Race condition: HTTP triggers fail on cold-starts


cgillum:
When running in the consumption plan, I sometimes notice that HTTP requests to start new orchestration instances fail because the target orchestration is not known.  If I wait a few seconds and try again, it succeeds.

I believe this is a race condition in the startup path of the Azure Functions host, where HTTP triggers can start executing before all extensions have finished loading. To fix this, the `DurableOrchestrationClient` and potentially the `DurableOrchestrationContext` API calls should have logic to wait for the listener to finishing starting up before attempting to perform their respective actions.

cgillum:
I think the fix requires coordination between calls to StartNewAsync, CallActivityAsync, CallSubOrchestrationAsync, etc. with listener startup. Specifically, these calls should block until _at least_ [DurableTaskExtension. StartTaskHubWorkerIfNotStartedAsync](https://github.com/Azure/azure-functions-durable-extension/blob/master/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs#L448) is called. A [ManualResetEvent](https://docs.microsoft.com/en-us/dotnet/standard/threading/manualresetevent-and-manualreseteventslim) could be used as a simple coordination primitive to block calls to [AssertOrchestratorExists](https://github.com/Azure/azure-functions-durable-extension/blob/f2b342eb922b766d47312f1b4723b9ac15ff997c/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs#L395) or [ThrowIfInvalidFunctionType](https://github.com/Azure/azure-functions-durable-extension/blob/f2b342eb922b766d47312f1b4723b9ac15ff997c/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs#L408)

But even this is imperfect. It's possible that the extension's internal task hub worker has finished starting up, but not all the orchestrator or activity functions have been registered. I don't know how likely this is, but it seems possible. Unfortunately the only way to truly fix this would be to know when the functions host is done initializing. I don't believe any such hook exists today in the functions host, which is the big challenge. If we can't get such a hook, then we should consider waiting up to 5 seconds after our internal task hub worker finishes completing before throwing these validation exceptions.

gled4er:
Hello @cgillum ,

Thank you very much for assigning the issue for me.

It sounds very exciting! 

May I ask you for recommendation who will be a good contact in Azure Functions team to check about the hook?

Thank you!

[...]

PR #217, opened 2018-03-30
Fix for HTTP Race condition on Cold Start

gled4er:
Hello @cgillum ,

May I ask you to review when you have time this initial commit for HTTP Race Condition on Cold Start as described here - https://github.com/Azure/azure-functions-durable-extension/issues/205 ?

I decided to call  `RegisterOrchestrator(FunctionName orchestratorFunction, ITriggeredFunctionExecutor executor)` inside `CreateListenerAsync` (in  `ActivityTriggerBinding.CreateListenerAsync` and  `OrchestrationTriggerBinding.CreateListenerAsync` ) because there I have access to the `ListenerFactoryContext` which will give me the `ITriggeredFunctionExecutor Executor`. I did not see a way to get access to the Executor in `TryCreateAsync` methods without changing the dependencies for the class. 

Let me know if this is an acceptable approach or I need to change the direction.

Thank you!

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Bindings/OrchestrationTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/OrchestrationTriggerAttributeBindingProvider.cs
@@ -103,6 +103,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     throw new ArgumentNullException(nameof(context));
                 }
 
+               this.config.RegisterOrchestrator(this.orchestratorName, context.Executor);
+
                 var listener = new DurableTaskListener(
                     this.config,
                     this.orchestratorName,
--- a/src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
@@ -159,6 +159,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     throw new ArgumentNullException(nameof(context));
                 }
 
+                this.parent.durableTaskConfig.RegisterActivity(this.activityName, context.Executor);
+
                 var listener = new DurableTaskListener(
                     this.parent.durableTaskConfig,
                     this.activityName,
--- a/src/WebJobs.Extensions.DurableTask/Listener/DurableTaskListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/DurableTaskListener.cs
@@ -36,15 +36,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         public Task StartAsync(CancellationToken cancellationToken)
         {
-            if (this.isOrchestrator)
-            {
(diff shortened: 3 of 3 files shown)


cgillum:
I think the best approach would be to do this in two phases. In `TryCreateAsync` we can simply register the existence of the function (e.g. add the key to the dictionary but leave the value as null). This should be sufficient to fix the race condition issue. Then, in `CreateListenerAsync` we can update the registration value with the executor. Does that make sense?

gled4er:
Hello @cgillum ,

I broke the update process to two steps as you advised in the previous review comment.

May I ask you to review the latest changes when you have time and let me know what should be fixed next?

Thank you!

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -399,9 +399,19 @@ internal void DeregisterOrchestrator(FunctionName orchestratorFunction)
...
+                {
+                    throw new ArgumentException($"The executor of an activity function named '{activityFunction}' cannot be updated from previous null value.");
+                }
+            }
+            else
+            {
+                if (!this.registeredActivities.TryAdd(activityFunction, executor))
+                {
+                    throw new ArgumentException($"The activity function named '{activityFunction}' is already registered.");
+                }
```
cgillum: I like this approach and how it ensures correctness. Maybe we can simplify it slightly?

```csharp
// Allow adding with a null key and subsequently updating with a non-null key.
if (!this.registeredActivities.TryUpdate(activityFunction, executor, null))
{
    if (!this.registeredActivities.TryAdd(activityFunction, executor))
    {
        throw new ArgumentException("The activity function named '{activityFunction}' is already fully registered."};
    }
}
```

src/WebJobs.Extensions.DurableTask/Bindings/OrchestrationTriggerAttributeBindingProvider.cs
```diff
@@ -53,6 +53,7 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
 
             // The orchestration name defaults to the method name.
             var orchestratorName = new FunctionName(name, trigger.Version);
+            this.config.RegisterActivity(orchestratorName, null);
```
cgillum: This needs to be a call to `RegisterOrchestrator`, which will also need the same new logic that you added for `RegisterActivity`.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
gled4er: Great optimization! Thank you very much!

src/WebJobs.Extensions.DurableTask/Bindings/OrchestrationTriggerAttributeBindingProvider.cs
gled4er: Thank you for catching that! I will fix it. Thank you!

gled4er:
Hello @cgillum ,

Thank you!

gled4er:
Hello @cgillum ,

I fixed the issue you identifies in the last review.

May I ask you to check once more and let me know what I should fix next?

Thank you!

Kanio

Issue #216, opened 2018-03-30
Post orchestration lifecycle events to Azure Event Grid


cgillum:
There are various scenarios where it would be useful to have an external system subscribe to orchestration lifecycle events. The most interesting events would probably be **orchestration started** and **orchestration completed** (success or fail).

One idea is to allow configuring an Event Grid SaS key as an application setting, and the extension could then post to the corresponding topic.

TsuyoshiUshio:
I'd love to contribute this as a DevOps guy. :)

cgillum:
よろしくお願いします！:)

TsuyoshiUshio:
Hi @cgillum I have a question.
For sending event to EventGrid Topic, we can use SDK or implement REST by myself. When I see the dependency, Durable doesn't have a lot of dependency. Can I use SDK with new nuget package or prefer to keep less dependency of nuget with direct REST call?

cgillum:
What is the nuget pack for the SDK? Is it [Microsoft.Azure.Management.EventGrid
](https://www.nuget.org/packages/Microsoft.Azure.Management.EventGrid)? I looked at these dependencies and I'm a little worried about them. For this particular case, I would be more comfortable with targeting the REST API.

TsuyoshiUshio:
Sure!

brandonh-msft:
How would we feel about an Event Grid output binding? https://github.com/brandonh-msft/Functions.Bindings.EventGrid

yuka1984:
I'm seeking opinions on the behavior when an Exception occurs when accessing EventGrid's API with HttpClient.

https://github.com/TsuyoshiUshio/azure-functions-durable-extension/blob/feature/lifecycledev/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs#L464

cgillum:
I went ahead and merged this PR. Thanks for this great contribution! I opened up a separate GitHub issue for tracking error handling strategies: https://github.com/Azure/azure-functions-durable-extension/issues/241

CastleArg:
Hi guys, this is great that this feature exists, I just came to request it after a lot of google fu got me nowhere.
Is it due to go G.A and be documented anytime soon?  I'm a bit wary of installing r.c version.
Essentially I want to trigger functions on error AND have realtime web dashboard for some long running task status so pub sub would be ideal.
Cheers!

akourbat:
Is this feature working?  I cannot get it to work at all

[...]

PR #220, opened 2018-04-02
Post orchestration lifecycle events to Azure Event Grid

TsuyoshiUshio:
Hi Chris, 

I re-create a pull request. I create this from clean dev branch.  I try to use this build however, the issue is the same. I open this to start conversation with you.

This pull request is the new version of this pull request.
https://github.com/Azure/azure-functions-durable-extension/pull/219

I open this pull request to start conversation with you. This pull request is created by the several people since we have mob programming. you can refer the https://github.com/Azure/azure-functions-durable-extension/pull/219 We just separate the commit to make it who is contribute this pull request. We had all-time mob-programming all together.

https://github.com/Azure/azure-functions-durable-extension/issues/216

We are implementing the feature. It will be like this. 

If you specify the host.json. Then the customer can receive the event from EventGrid. 

_host.json_

```
{
  "durableTask": {
    "EventGridTopicEndpoint": "YOUR_EVENT_GRID_TPIC_ENDPOINT_HERE",
    "EventGridKey": "YOUR_EVENT_GRID_TOPIC_KEY_HERE"
  }
}
```
You can find the sample program. It is just template code with host.json configuration with New feature.
https://github.com/TsuyoshiUshio/LifeCycleEventSpike

For this sample code, you can copy host.json.example to host.json If  you need EventGrid environment, I'll share with you.

However, when I execute this, you'll get this error. 

```
[02/04/2018 11:07:27] 21a6e8c580da40a89a681db6aff4fec4: Function 'Sample (Orchestrator)', version '' failed with an error. Reason: System.InvalidOperationException: Multithreaded execution was detected. This can happen if the orchestrator function previously resumed from an unsupported async callback.
[02/04/2018 11:07:27]    at Microsoft.Azure.WebJobs.DurableOrchestrationContext.ThrowIfInvalidAccess()
[02/04/2018 11:07:27]    at Microsoft.Azure.WebJobs.DurableOrchestrationContext.<CallDurableTaskFunctionAsync>d__47`1.MoveNext()
[02/04/2018 11:07:27] --- End of stack trace from previous location where exception was thrown ---
[02/04/2018 11:07:27]    at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[02/04/2018 11:07:27]    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[02/04/2018 11:07:27]    at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
[02/04/2018 11:07:27]    at LifeCycleEventSpike.Sample.<RunOrchestrator>d__0.MoveNext() in C:\Users\tsushi\source\repos\LifeCycleEventSpike\LifeCycleEventSpike\Sample.cs:line 19
[02/04/2018 11:07:27] --- End of stack trace from previous location where exception was thrown ---
[02/04/2018 11:07:27]    at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[02/04/2018 11:07:27]    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[02/04/2018 11:07:27]    at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
[02/04/2018 11:07:27]    at Microsoft.Azure.WebJobs.Host.Executors.FunctionInvoker`2.<InvokeAsync>d__9.MoveNext() in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Executors\FunctionInvoker.cs:line 63
[02/04/2018 11:07:27] --- End of stack trace from previous location where exception was thrown ---
[02/04/2018 11:07:27]    at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[02/04/2018 11:07:27]    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[02/04/2018 11:07:27]    at Microsoft.Azure.WebJobs.Extensions.DurableTask.TaskOrchestrationShim.<Execute>d__7.MoveNext(). IsReplay: False. State: Failed. HubName: DurableFunctionsHub. AppName: . SlotName: . ExtensionVersion: 1.2.1.0. SequenceNumber: 3.
[02/04/2018 11:07:28]
```
However, it successfully send message to the event grid for Start. I'm not sure if our code was wrong fix or not. We create the LifeCycleTraceHelper.cs then add it to the DurableTaskExtension.cs and  Listener/TaskOrchestrationShim.cs


cgillum:
Thanks for this PR! I'm excited to have this feature.

The reason for the `InvalidOperationException` is because this PR introduces async operations in the orchestration thread, which is forbidden. We will need to find a different place to call these new methods. Let me try to think of some ideas.

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -79,6 +87,15 @@ public void SetFunctionInvocationCallback(Func<Task> callback)
                     e.ToString(),
                     FunctionType.Orchestrator,
                     this.context.IsReplaying);
+
+                await this.config.LifeCycleTraceHelper.OrchestratorFailedAsync(
```
cgillum: Make sure `this.context.IsReplaying` is false before sending this notification.

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -103,6 +120,15 @@ public void SetFunctionInvocationCallback(Func<Task> callback)
                 FunctionType.Orchestrator,
                 this.context.IsReplaying);
 
+            await this.config.LifeCycleTraceHelper.OrchestratorCompletedAsync(
```
cgillum: Make sure `this.context.IsReplaying` is false before sending this notification.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -158,6 +159,23 @@ public class DurableTaskExtension :
         /// </value>
         public bool DisableHttpManagementApis { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value of Event Grid Topic Endpoint for emitting lifecycle events.
+        /// If this property has been set, it will emit the event to the Event Grid Topic.
+        /// <see cref="LifeCycleTraceHelper"/>
```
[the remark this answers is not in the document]
hamamotsu: Done.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -158,6 +159,23 @@ public class DurableTaskExtension :
...
+        /// </summary>
+        public string EventGridTopicEndpoint { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value of Event Grid Key.
+        /// If this property has been set, it will emit the event to the Event Grid Topic.
+        /// <see cref="LifeCycleTraceHelper"/>
+        /// </summary>
+        public string EventGridKey { get; set; }
```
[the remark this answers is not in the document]
hamamotsu: Done.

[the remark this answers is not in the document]
TsuyoshiUshio: This is coming later , we fix at that time.

src/WebJobs.Extensions.DurableTask/LifeCycleTraceHelper.cs
```diff
@@ -0,0 +1,230 @@
...
+        private readonly ILogger logger;
+        private readonly DurableTaskExtension config;
+
+        private bool UseTrace { get; }
+
+        private static HttpClient httpClient = null;
+
+        public LifeCycleTraceHelper(DurableTaskExtension config, ILogger logger)
+        {
+            this.config = config;
```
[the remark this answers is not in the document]
kingkino: I have done.
Please check it.

src/WebJobs.Extensions.DurableTask/LifeCycleTraceHelper.cs
```diff
@@ -0,0 +1,230 @@
...
+using System.Diagnostics;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    internal class LifeCycleTraceHelper
```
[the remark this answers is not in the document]
rukasakurai: Done

src/WebJobs.Extensions.DurableTask/LifeCycleTraceHelper.cs
```diff
@@ -0,0 +1,230 @@
...
+        {
+            this.config = config;
+
+            this.logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+            if (!string.IsNullOrEmpty(config.EventGridTopicEndpoint) && !string.IsNullOrEmpty(config.EventGridKey))
+            {
+                UseTrace = true;
+                httpClient = new HttpClient();
+                httpClient.DefaultRequestHeaders.Add("aeg-sas-key", config.EventGridKey);
```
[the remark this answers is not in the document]
gled4er: Thank you for the detailed explanation and useful sample!

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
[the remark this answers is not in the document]
kingkino: Change proepry name done.

src/WebJobs.Extensions.DurableTask/LifeCycleTraceHelper.cs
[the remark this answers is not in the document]
rukasakurai: Done by Kanio

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -56,6 +56,14 @@ public void SetFunctionInvocationCallback(Func<Task> callback)
                 FunctionType.Orchestrator,
                 this.context.IsReplaying);
 
+            await this.config.LifeCycleTraceHelper.OrchestratorStartingAsync(
```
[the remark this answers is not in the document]
TsuyoshiUshio: I've done this. :)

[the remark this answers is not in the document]
TsuyoshiUshio: Done. However we also change the this part.  DurableOrchecturatorContext. We used to initialize the deferedFunctions Array in `AddDeferredTask`. We move it to the constructor. We want to know the lifecycle of the DurableOrchestratorContext. Is the instance created on each reply? 

```
        public override void AddDeferredTask(Func<Task> function)
        {
            this.deferedFunctions.Add(function);
        }

        public override async Task RunDeferredTasks()
        {
            if (this.deferedFunctions == null)
            {
                throw new InvalidOperationException(
                    "AddDeferredTask needs to be called first.");
            }
            await Task.WhenAll(this.deferedFunctions.Select(x => x()));
            this.deferedFunctions.Clear();
        }
```

src/WebJobs.Extensions.DurableTask/LifeCycleTraceHelper.cs
```diff
@@ -0,0 +1,230 @@
...
+        public LifeCycleTraceHelper(DurableTaskExtension config, ILogger logger)
+        {
+            this.config = config;
+
+            this.logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+            if (!string.IsNullOrEmpty(config.EventGridTopicEndpoint) && !string.IsNullOrEmpty(config.EventGridKey))
+            {
+                UseTrace = true;
+                httpClient = new HttpClient();
```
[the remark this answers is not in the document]
TsuyoshiUshio: I add some comment. Please review it.

src/WebJobs.Extensions.DurableTask/LifeCycleTraceHelper.cs
```diff
@@ -0,0 +1,230 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    internal class LifeCycleTraceHelper
+    {
+        private readonly ILogger logger;
```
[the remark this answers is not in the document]
TsuyoshiUshio: I've done it. I remove the dependency of the logger and use EndToEndHelper instead. I move the logging detail to the EndToEndHelper. Am I right?

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -35,13 +36,15 @@ public sealed class DurableOrchestrationContext : DurableOrchestrationContextBas
         private string serializedCustomStatus;
         private int owningThreadId;
 
+        private List<Func<Task>> deferedFunctions;
```
cgillum: Please make this `readonly`. Also, I suggest naming this `deferredTasks` or `deferredActions` (note the spelling) instead of `deferedFunctions` because people might confuse "functions" with "Azure Functions".

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -445,5 +448,21 @@ private void ThrowIfInvalidAccess()
...
         }
+
+        public override void AddDeferredTask(Func<Task> function)
+        {
+            this.deferedFunctions.Add(function);
+        }
+
+        public override async Task RunDeferredTasks()
+        {
+            if (this.deferedFunctions == null)
```
cgillum: Please remove the `if` check. `this.deferedFunctions` can never be null.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -445,5 +448,21 @@ private void ThrowIfInvalidAccess()
                     "Multithreaded execution was detected. This can happen if the orchestrator function previously resumed from an unsupported async callback.");
             }
         }
+
+        public override void AddDeferredTask(Func<Task> function)
```
cgillum: Can these two methods be `internal` instead of `public`? I don't think we want to expose this externally.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -56,6 +56,7 @@ public class DurableTaskExtension :
 
         private EndToEndTraceHelper traceHelper;
         private HttpApiHandler httpApiHandler;
+        private LifeCycleNotificationHelper lifeCycleTraceHelper;
```
cgillum: The variable should be renamed to match the new class name: `lifeCycleNotificationHelper`.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -158,6 +159,21 @@ public class DurableTaskExtension :
...
+        /// </summary>
+        public string EventGridTopicEndpoint { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value of Event Grid Key.
+        /// If this property has been set, it will emit the event to the Event Grid Topic.
+        /// </summary>
+        public string EventGridKeySettingName { get; set; }
+
+        internal LifeCycleNotificationHelper LifeCycleTraceHelper => this.lifeCycleTraceHelper;
```
cgillum: This property should be renamed to match the new class name: `LifeCycleNotificationHelper`.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -297,6 +314,10 @@ TaskActivity INameVersionObjectManager<TaskActivity>.GetObject(string name, stri
                     context.InstanceId,
                     context.IsReplaying);
             }
+
+ 
+                await context.RunDeferredTasks();
```
cgillum: Please only use 4 spaces for indenting. Also, please remove the extra newlines above and below.

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -254,6 +254,38 @@ public static string LocalSlotName
...
+            string reason,
+            bool isReplay,
+            FunctionState functionState,
+            string hubName,
+            string appName,
+            string slotName,
+            string extensionVersion
+            )
+        {
+            this.logger.LogError(
```
cgillum: Please add a new `EtwEventSource.Instance.ExternalEventRaised` method which matches this trace.

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -254,6 +254,38 @@ public static string LocalSlotName
                 "{instanceId}: Function '{functionName} ({functionType})', version '{version}' received a '{eventName}' event. State: {state}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
                 instanceId, functionName, functionType, version, eventName, FunctionState.ExternalEventRaised, hubName, LocalAppName, LocalSlotName, ExtensionVersion, this.sequenceNumber++);
         }
+        public void SendMessageFailed(
```
cgillum: Let's name this `SendNotificationFailed`

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -254,6 +254,38 @@ public static string LocalSlotName
...
+            string functionName,
+            FunctionType functionType,
+            string version,
+            string reason,
+            bool isReplay,
+            FunctionState functionState,
+            string hubName,
+            string appName,
+            string slotName,
+            string extensionVersion
```
cgillum: Please remove `functionType` and `isReplay` - we always know that this message is for orchestrator functions and will not be emitted during a replay. Also, please remove `version`, `appName`, `slotName`, and `extensionVersion`. All of these are already defined in this class and do not need to be passed in as parameters.

So to summarize, the only parameters that need to be included are `statusCode`, `instanceId`, `functionName`, and `reason`.

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
```diff
@@ -0,0 +1,245 @@
...
+            }
+        }
+
+        public void SetHttpMessageHandler(HttpMessageHandler handler)
+        {
+            httpClient = new HttpClient(handler);
+            httpClient.DefaultRequestHeaders.Add("aeg-sas-key", eventGridKeyValue);
+        }
+
+        private async Task TraceRequestAsync(
```
cgillum: Let's name this `SendNotificationAsync`.

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
```diff
@@ -0,0 +1,245 @@
...
+            // Details about the Event Grid REST API
+            // https://docs.microsoft.com/en-us/rest/api/eventgrid/
+            var result = await httpClient.PostAsync(this.config.EventGridTopicEndpoint, content);
+
+            if (!result.IsSuccessStatusCode)
+            {
+                var appName = Environment.GetEnvironmentVariable("WEBSITE_SITE_NAME") ?? string.Empty;
+                var slotName = Environment.GetEnvironmentVariable("WEBSITE_SLOT_NAME") ?? string.Empty;
+                var extensionVersion = FileVersionInfo.GetVersionInfo(typeof(DurableTaskExtension).Assembly.Location).FileVersion;
+                this.config.TraceHelper.SendMessageFailed
```
cgillum: See my previous comment about reducing the number of parameters.

test/DurableTaskLifeCycleNotificationTest.cs
```diff
@@ -0,0 +1,361 @@
+﻿using System;
```
cgillum: Please add the copyright header to the top of this new file.

test/DurableTaskLifeCycleNotificationTest.cs
```diff
@@ -0,0 +1,361 @@
...
+                                    }
+                                    else if (callCount == 1)
+                                    {
+                                        Assert.Equal("durable/orchestrator/Completed", o.subject.ToString());
+                                        Assert.Equal("orchestratorEvent", o.eventType.ToString());
+                                        Assert.Equal("1", o.data.EventType.ToString());
+                                    }
+                                    else
+                                    {
+                                        throw new Exception("Call Count is Bad");
```
cgillum: How about `Assert.Fail`?

test/DurableTaskLifeCycleNotificationTest.cs
```diff
@@ -0,0 +1,361 @@
...
+                            return Task.FromResult(new HttpResponseMessage(HttpStatusCode.InternalServerError));
+                        });
+
+                    extension.LifeCycleTraceHelper.SetHttpMessageHandler(mock.Object);
+                }
+
+                var client = await host.StartOrchestratorAsync(orchestratorFunctionNames[0], "World", this.output);
+                var status = await client.WaitForCompletionAsync(TimeSpan.FromSeconds(30), this.output);
+
+                Assert.Equal(OrchestrationRuntimeStatus.Completed, status?.RuntimeStatus);
```
cgillum: Just to make sure I understand: the intent of this test is to ensure that the orchestration completes even if the Event Grid notification fails, correct? Should we also check the trace output to make sure the new trace message is written? Several of the tests in DurableTaskEndToEndTests.cs do similar logging validation.

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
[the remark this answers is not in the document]
cgillum: You should not assume that we create a new `DurableOrchestrationContext` for each reply. Yes, we create one today, but I have another optimization in another branch that causes it to not get recreated in some cases. I think your current implementation is good. However, you can get rid of the `null` check since `this.deferedFunctions` will never be null. :)

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
gled4er: Sure. We will change it. Thank you for catching that.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
gled4er: Thank you! We will change it.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
gled4er: Thank you! We will improve the presentation style here.

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
gled4er: Hello @cgillum ,

I added few more parameters to the function signature and  I ended up with the following definition for the function in ` EndToEndTraceHelper`:

```csharp
 public void SendMessageFailed(
            string hubName,
            string functionName,
            string version,
            string instanceId,
            HttpStatusCode statusCode,
            string httpStatusCodeReasonPhrase,
            string reason)
        {
            FunctionType functionType = FunctionType.Orchestrator;
            bool isReplay = false;

            EtwEventSource.Instance.ExternalEventRaised(
                hubName,
                LocalAppName,
                LocalSlotName,
                functionName,
                version,
                instanceId,
                statusCode,
                httpStatusCodeReasonPhrase,
                reason,
                functionType,
                ExtensionVersion,
                isReplay);

            this.logger.LogError(
                "Error in sending message to the EventGrid. Please check the host.json configuration durableTask.EventGridTopicEndpoint and EventGridKey. LifeCycleNotificationHelper.TraceRequestAsync - Status: {statusCode} Reason Phrase: {httpStatusCodeReasonPhrase} For more detail: {instanceId}: Function '{functionName} ({functionType})', version '{version}' failed with an error. Reason: {reason}. IsReplay: {isReplay}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {ExtensionVersion}.",
                statusCode,
                httpStatusCodeReasonPhrase,
                instanceId,
                functionName,
                functionType,
                version,
                reason,
                isReplay,
                hubName,
                LocalAppName,
                LocalSlotName,
                ExtensionVersion);
        }
```

Let me know if this is acceptable or we need to modify it.

Thank you!

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
gled4er: Thank you! We will modify it.

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
gled4er: Thank you! We will fix it.

test/DurableTaskLifeCycleNotificationTest.cs
gled4er: We will add it. Thank you!

test/DurableTaskLifeCycleNotificationTest.cs
gled4er: Hello @cgillum ,

Since `Assert.Fail` is not available I used the recommended workaround - [xUnit.net alternative: Assert.True(false, "message")](http://xunit.github.io/docs/comparisons.html) and in we will have the following snippet:

```csharp
Assert.True(false, "Call Count is Bad");
```

Let me know if we need to modify it.

Thank you!

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
gled4er: Thank you! We will fix it.

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
gled4er: Sure. We will fix it. Thank you!

gled4er:
Hello @cgillum ,

I increased the timeout from 70 to 80 seconds for `OrchestrationConcurrency` test because we had it continuously failing and after a lot of testing is passing with the new value. We are not able for now to identify the cause of the delay but we will continue to look into it. Let us know if you have any idea. 

Thank you!

gled4er:
Hello @cgillum,

We worked on all issues you found in the previous review iteration.

When you have time may I ask you to check again and let us know what we need to fix next?

Thank you!

gled4er:
Hello @cgillum ,

We worked on the changes you requested and we addressed all of them.

May I ask you to take a look again when you have time in order to fix the next set of issues?

Thank you!

test/DurableTaskLifeCycleNotificationTest.cs
gled4er: Yes, you are right. We added the testing of the trace messages. Thank you!

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -158,6 +159,20 @@ public class DurableTaskExtension :
...
 
+        /// <summary>
+        /// Gets or sets a value of Event Grid Topic Endpoint for emitting lifecycle events.
+        /// If this property has been set, it will emit the event to the Event Grid Topic.
+        /// </summary>
+        public string EventGridTopicEndpoint { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value of Event Grid Key.
+        /// If this property has been set, it will emit the event to the Event Grid Topic.
```
cgillum: I would reword the summary like this:
*Gets or sets the name of the app setting containing the key used for authenticating with the Azure Event Grid custom topic at `<see cref="EventGridTopicEndpoint"/>.`*

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -158,6 +159,20 @@ public class DurableTaskExtension :
         /// </value>
         public bool DisableHttpManagementApis { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value of Event Grid Topic Endpoint for emitting lifecycle events.
+        /// If this property has been set, it will emit the event to the Event Grid Topic.
```
cgillum: I word reword this slightly:
*Gets or sets the URL of an Azure Event Grid custom topic endpoint. When set, orchestration life cycle notification events will be automatically published to this endpoint.*

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
```diff
@@ -0,0 +1,243 @@
...
+                // Post to custom topic for Azure Event Grid
+                // https://docs.microsoft.com/en-us/azure/event-grid/post-to-custom-topic
+                httpClient = new HttpClient();
+                INameResolver nameResolver = extensionConfigContext.Config.GetService<INameResolver>();
+                this.eventGridKeyValue = nameResolver.Resolve(config.EventGridKeySettingName);
+                httpClient.DefaultRequestHeaders.Add("aeg-sas-key", this.eventGridKeyValue);
+            }
+        }
+
+        private bool UseTrace { get; }
```
cgillum: Let's make this a `private readonly bool useTrace` instance variable instead of a property.

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
```diff
@@ -0,0 +1,243 @@
...
+
+            // Details about the Event Grid REST API
+            // https://docs.microsoft.com/en-us/rest/api/eventgrid/
+            var result = await httpClient.PostAsync(this.config.EventGridTopicEndpoint, content);
+
+            if (!result.IsSuccessStatusCode)
+            {
+                var appName = Environment.GetEnvironmentVariable("WEBSITE_SITE_NAME") ?? string.Empty;
+                var slotName = Environment.GetEnvironmentVariable("WEBSITE_SLOT_NAME") ?? string.Empty;
+                var extensionVersion = FileVersionInfo.GetVersionInfo(typeof(DurableTaskExtension).Assembly.Location).FileVersion;
```
cgillum: These three variables can be removed now.

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
```diff
@@ -0,0 +1,243 @@
...
+
+                this.config.TraceHelper.SendMessageFailed(
```
cgillum: I suggest we always trace, regardless of success or failure. That we customers can know if notifications are actually being sent as part of debugging their Event Grid integration. We should also use a `Stopwatch` to track the latency of these notifications in case there might be a performance impact.

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
```diff
@@ -0,0 +1,243 @@
...
+                    DataVersion = "1.0",
+                },
+            };
+        }
+
+        private class EventGridEvent
+        {
+            public EventGridEvent() { }
+
+            public EventGridEvent(string id, string subject, object data, string eventType, DateTime eventTime, string dataVersion, string topic = null, string metadataVersion = null)
```
cgillum: Let's remove this constructor since it is not being used.

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
```diff
@@ -0,0 +1,243 @@
...
+                    EventTime = DateTime.UtcNow,
+                    Data = new
+                    {
+                        HubName = hubName,
+                        FunctionName = functionName,
+                        Version = version,
+                        InstanceId = instanceId,
+                        Reason = reason,
+                        EventType = orchestrationRuntimeStatus,
+                        EventTime = DateTime.UtcNow,
```
cgillum: Isn't this `EventTime` property redundant with the one defined by `EventGridEvent`?

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
```diff
@@ -0,0 +1,243 @@
...
+        private readonly ExtensionConfigContext extensionConfigContext;
+        private readonly string eventGridKeyValue;
+        private static HttpClient httpClient = null;
+
+        public LifeCycleNotificationHelper(DurableTaskExtension config, ExtensionConfigContext extensionConfigContext)
+        {
+            this.config = config ?? throw new ArgumentNullException(nameof(config));
+            this.extensionConfigContext = extensionConfigContext ?? throw new ArgumentNullException(nameof(extensionConfigContext));
+
+            if (!string.IsNullOrEmpty(config.EventGridTopicEndpoint) && !string.IsNullOrEmpty(config.EventGridKeySettingName))
```
cgillum: I suggest we add a bit more validation here in case users make a configuration mistake.
* If `EventGridTopicEndpoint` is set but `EventGridKeySettingName` is not, then we should throw an exception to inform the developer that a key is required.
* Also, further down below, we should throw an exception if `this.eventGridKeyValue` is null since it probably means that the developer had an incorrect app setting name.

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
```diff
@@ -0,0 +1,243 @@
...
+            }
+
+            EventGridEvent[] sendObject = this.CreateEventGridEvent(
+                hubName,
+                functionName,
+                version,
+                instanceId,
+                "",
+                OrchestrationRuntimeStatus.Running);
+            await this.SendNotificationAsync(sendObject, hubName, functionName, version, instanceId, "", FunctionType.Orchestrator, isReplay, FunctionState.Started);
```
cgillum: All the async methods below only use the `await` keyword as the very last statement. Because of that, I suggest we not use async/await compiler feature and instead return the task from `this.SendNotificationAsync` directly. This has the benefit of simplifying exception debugging and should result in a slight performance improvement since the code does not need to go through the compiler-generated async state machine.

For example:
```csharp
public Task OrchestratorCompletedAsync(...)
{
    // ...
    return this.SendNotificationAsync(...);
}
```

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -252,7 +253,53 @@ public static string LocalSlotName
...
+                instanceId,
+                statusCode,
+                httpStatusCodeReasonPhrase,
+                reason,
+                functionType,
+                ExtensionVersion,
+                isReplay);
+
+            this.logger.LogError(
+                "Error in sending message to the EventGrid. Please check the host.json configuration durableTask.EventGridTopicEndpoint and EventGridKey. LifeCycleNotificationHelper.SendNotificationAsync - Status: {statusCode} Reason Phrase: {httpStatusCodeReasonPhrase} For more detail: {instanceId}: Function '{functionName} ({functionType})', function state {functionState} version '{version}' failed with an error. Reason: {reason}. IsReplay: {isReplay}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {ExtensionVersion}.",
```
cgillum: There are several things I think we should change regarding this message. Here is a version that I like better:

```csharp
$"Failed to send the {{functionState}} notification to Azure Event Grid. Please check the configuration values for {nameof(EventGridTopicEndpoint)} and {nameof(EventGridKeySettingName)}. Status code: {{statusCode}}. Reason: {{reason}}. Function '{{functionName}} ({{functionType}})'. InstanceId: {{instanceId}}. Version '{{version}}'. IsReplay: {{isReplay}}. HubName: {{hubName}}. AppName: {{appName}}. SlotName: {{slotName}}. ExtensionVersion: {{ExtensionVersion}}."

File: src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs [hunk 35]
```diff
@@ -252,7 +253,53 @@ public static string LocalSlotName
 
             this.logger.LogInformation(
                 "{instanceId}: Function '{functionName} ({functionType})', version '{version}' received a '{eventName}' event. State: {state}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
-                instanceId, functionName, functionType, version, eventName, FunctionState.ExternalEventRaised, hubName, LocalAppName, LocalSlotName, ExtensionVersion, this.sequenceNumber++);
+                instanceId, functionName, functionType, version, eventName, FunctionState.ExternalEventRaised, hubName,
+                LocalAppName, LocalSlotName, ExtensionVersion, this.sequenceNumber++);
+        }
+
+        public void SendMessageFailed(
```
cgillum: Let's give this a more specific method name, like `EventGridNotificationFailed`.

src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
```diff
@@ -163,6 +164,26 @@ private EtwEventSource()
         {
             this.WriteEvent(209, TaskHub, AppName, SlotName, FunctionName, Version ?? "", InstanceId, Reason, FunctionType, ExtensionVersion, IsReplay);
         }
+
+        [Event(210, Level = EventLevel.Informational)]
+        public void ExternalEventRaised(
```
cgillum: EventGridNotificationFailed

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
```diff
@@ -0,0 +1,243 @@
...
+
+                this.config.TraceHelper.SendMessageFailed(
+                    hubName,
+                    functionName,
+                    functionState,
+                    version,
+                    instanceId,
+                    result.StatusCode,
+                    result.ReasonPhrase,
+                    reason);
```
cgillum: In the case of an error, we should include the failure details that come back from Event Grid. Basically, we should read the string response that comes back and include it in this trace (I assume it is typically simple JSON strings). This will make debugging problems significantly easier.

gled4er:
Hello @cgillum,

Thank you very much for the great feedback.

We will work on fixing the issues.

Thank you!

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
gled4er: Great! We will change it. Thank you!

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
gled4er: Thank you! We will modify it.

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
gled4er: Thank you! We will fix it.

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
gled4er: Great catch. We will fix it. Thank you!

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
gled4er: Thank you! We will fix it.

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
gled4er: Great catch! We will remove it. Thank you!

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
TsuyoshiUshio: This seems fixed by @gled4er  :)

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
TsuyoshiUshio: Thanks! Done!

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
TsuyoshiUshio: Done for it. Also I add a test for this fix. Please check my error message is proper or not.

cgillum: Great! I will suggest some improvements for the error messages.

TsuyoshiUshio:
Hi @cgillum , I have a question. I found my test has flakiness.  Sometimes it works sometimes, fails. Also, When I run all tests, it fails usually, however, when I test only the test case, it works always. This is the code. I suspect that set environment variables could cause a problem since the xunit tests run in  parallel. when I remove this test case it works. I also find set-enviornment variables on the other tests.  Do we have good way to test with environment variable variations ? or Just remove this test?

```
        [Fact]
        public async Task ConfigurationWithoutEventGridKeyValue()
        {
            string eventGridKey = null;
            var eventGridKeySettingName = "eventGridKeySettingName";
            var eventGridEndpoint = "http://dymmy.com/";
            Environment.SetEnvironmentVariable(eventGridKeySettingName, eventGridKey);
            using (JobHost host = TestHelpers.GetJobHost(this.loggerFactory, nameof(this.OrchestrationTerminate), eventGridKeySettingName, eventGridEndpoint))
            {
                var ex = await Assert.ThrowsAsync<ArgumentException>(async () => await host.StartAsync());
                Assert.Equal($"Failed to start lifecycle notification feature. Please check the configuration values for {eventGridKeySettingName} on AppSettings.", ex.Message);
            }
        }
```

cgillum:
@TsuyoshiUshio Let's keep the test, but instead of setting an environment variable (which can have side effects like you mentioned), I suggest you pass these settings into `TestHelpers.CreateJobHost` and then assign them to the `DurableTaskExtension` object directly. That will guarantee that there are no side effects that impact other tests.

TsuyoshiUshio:
@cgillum Very cool, Thank you! I'll have a look and implement that! :)  Also I'll fix the other tests which uses Set Environment variables.

TsuyoshiUshio:
HI @cgillum, I solve the flakiness. Thank you very much! See the last commit. Only one flakiness is the OrchestrationConcurrency() We don't touch it, however, it might be a timeout. 

_DurableTaskEndToEndTests.cs_

```
                // The 100 orchestrations above (which each delay for 10 seconds) should all complete in less than 80 seconds.
                Task parallelOrchestrations = Task.WhenAll(tasks);
                Task timeoutTask = Task.Delay(TimeSpan.FromSeconds(80));

                Task winner = await Task.WhenAny(parallelOrchestrations, timeoutTask);
                Assert.Equal(parallelOrchestrations, winner);
```

Kanio said it happens before we adding the new tests. Our fix overhead for the existing testing is this part. It is just add function to the list. Also, after that, inside the function, it is not execute anything if they don't use new feature. I don't think passing the functions doesn't cost a lot. As I create a sample, 1000 times add to the List<Func<Task>> it cost like this. `Elapsed Time for List = 00:00:00.0000872` 

Do you have good advice for the flakiness?

_TaskOrchestrationShims.cs_

```
            if (!this.context.IsReplaying)
            {
                this.context.AddDeferredTask(() => this.config.LifeCycleNotificationHelper.OrchestratorStartingAsync(
                    this.context.HubName,
                    this.context.Name,
                    this.context.Version,
                    this.context.InstanceId,
                    FunctionType.Orchestrator,
                    this.context.IsReplaying));
            }
  :
```
_LifeCycleNotificationHelper.cs_

```
        public async Task OrchestratorStartingAsync(
            string hubName,
            string functionName,
            string version,
            string instanceId,
            FunctionType functionType,
            bool isReplay)
        {
            if (!this.useTrace)
            {
                return;
            }

            EventGridEvent[] sendObject = this.CreateEventGridEvent(
                hubName,
                functionName,
                version,
                instanceId,
                "",
                OrchestrationRuntimeStatus.Running);
            await this.SendNotificationAsync(sendObject, hubName, functionName, version, instanceId, "", FunctionState.Started);
        }
```

TsuyoshiUshio:
Hi @cgillum, I need your help.  I currently work on add body to the EventGridNotificationFaild / WriteEvent according to your comment. Since it requires getting body, I need to write something like this on LifeCycleNotificationHelper.SendNotificationAsync 

```
            HttpResponseMessage result = await httpClient.PostAsync(this.config.EventGridTopicEndpoint, content);
            var body =  await result.Content.ReadAsStringAsync();   <- this
```

However, when I see the test code, on DurableTaskLifeCycleNotificationTest, I find this code. It try to read Content again. So it starts fail test as timeout.  In this case, How can I avoid this "read twice" problem?  Do you have any good idea?

```
                    var mock = new Mock<HttpMessageHandler>();
                    mock.Protected()
                        .Setup<Task<HttpResponseMessage>>("SendAsync", ItExpr.IsAny<HttpRequestMessage>(), ItExpr.IsAny<CancellationToken>())
                        .Returns((HttpRequestMessage request, CancellationToken cancellationToken) =>
                        {
                            Assert.True(request.Headers.Any(x => x.Key == "aeg-sas-key"));
                            var values = request.Headers.GetValues("aeg-sas-key").ToList();
                            Assert.Single(values);
                            Assert.Equal(eventGridKeyValue, values[0]);
                            Assert.Equal(eventGridEndpoint, request.RequestUri.ToString());
                            var json = request.Content.ReadAsStringAsync().GetAwaiter().GetResult();
                            dynamic content = JsonConvert.DeserializeObject(json);
                            foreach (dynamic o in content)
                            {
```

cgillum:
Interesting. So you see the timeout when trying to read the response? I noticed in the test code that you create an InternalServerError response, but don't include any content. Have you tried updating the InternalServerError response to return some content?

TsuyoshiUshio:
Hi @cgillum , Thank you! You are right! I create a simple spike app and I realize that if I'm not add Content, it cause an no reference exception.  
https://github.com/TsuyoshiUshio/HttpClientMockSpike/blob/master/HttpClientMockSpike/Program.cs

When I change the code at the end of the mock method, it works. 

```
                                var message = new HttpResponseMessage(HttpStatusCode.OK);
                                message.Content = new StringContent("{\"message\":\"OK!\"}");
                                return Task.FromResult(message);
```

However, then I've got new problem. Again flakiness. HandledActivityException, OrchestratorConcurrency , and TimerExpiration starts fail for timeout sometimes. Not always.
When I see the  Task Manager during the execution, it hits CPU 100% a lot of cpu is consumed by Storage Emulator. I close all browse, then all run tests works. However, my machine is Surface book 2 which is very high speed machine. I'll push this fix. I think the code its self have no problem and need to be tested by CI.  Also the code might need some review from others. (Not only my part, but other as well. what makes it slow) So I push this commit to this branch to share. Please try CI.  it is `"log the response body of the EventGrid http call"` commit.  If you have some advice, please let me know.

I just add  `var body = await result.Content.ReadAsStringAsync()` on the SendNortificationAsync on the LifeCycleNotificationHelper. This method is never called from Orchestrator Concurrency and HandledActivityException.

cgillum:
Great! Glad you were able to get that working!

Don't worry too much about the test flakiness. I think the problem is the Storage Emulator as you mentioned. I think the tests are less flaky on the CI because it uses a real storage account. Feel free to do the same when running locally. My desktop is a 12-proc, 16 GB ram machine, so only OrchestratorConcurrency is flaky for me. :)

I will do another full review as soon as I get a chance. Thanks for all your hard work on this!

TsuyoshiUshio:
@cgillum Thank you for helping me a lot. I learned a lot . I appreciate your patience and kindness.

test/TestHelpers.cs
```diff
@@ -182,6 +188,20 @@ private static List<string> GetLogs_HelloWorldOrchestration_Inline(string messag
             return list;
         }
 
+        private static List<string> GetLogs_OrchestrationEventGridApiReturnBadStatus(string messageId, string[] functionNames)
+        {
+            var list = new List<string>()
+            {
+                $"{messageId}: Function '{functionNames[0]} ({FunctionType.Orchestrator})', version '' scheduled. Reason: NewInstance. IsReplay: False. State: Scheduled. HubName: OrchestrationStartAndCompleted. AppName: . SlotName: . ExtensionVersion: 1.2.1.0. SequenceNumber: 0.",
```
cgillum: Let's not check for HubName, AppName, SlotName, ExtensionVersion, or SequenceNumber. Checking for EventVersion is especially problematic because this test will break every time we release a new update.  See the other GetLogs_ implementations for an example of what is okay to check for.

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -252,7 +253,58 @@ public static string LocalSlotName
...
+                statusCode,
+                httpStatusCodeReasonPhrase,
+                reason,
+                functionType,
+                ExtensionVersion,
+                isReplay,
+                latency);
+
+            this.logger.LogError(
+                "LifeCycleNotificationHelper.SendNotificationAsync - Status: {statusCode} Reason Phrase: {httpStatusCodeReasonPhrase}. For more detail: {instanceId}: Function '{functionName} ({functionType})', function state {functionState} version '{version}' failed with an error. Reason: {reason}. IsReplay: {isReplay}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {ExtensionVersion}. Latency: {latency} ms.",
```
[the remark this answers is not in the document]
cgillum: Maybe something like this?

*"{instanceId}: Function '{functionName} ({functionType})', failed to send a '{functionState}' notification event to Azure Event Grid. Status code: {statusCode}. Details: {details}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}."*

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -252,7 +253,58 @@ public static string LocalSlotName
 
             this.logger.LogInformation(
                 "{instanceId}: Function '{functionName} ({functionType})', version '{version}' received a '{eventName}' event. State: {state}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
-                instanceId, functionName, functionType, version, eventName, FunctionState.ExternalEventRaised, hubName, LocalAppName, LocalSlotName, ExtensionVersion, this.sequenceNumber++);
+                instanceId, functionName, functionType, version, eventName, FunctionState.ExternalEventRaised, hubName,
+                LocalAppName, LocalSlotName, ExtensionVersion, this.sequenceNumber++);
+        }
+
+        public void EventGridMessageSent(
```
cgillum: I'm a little confused. The method name is `EventGridMessageSent` but the trace messages are reporting errors. it seems to me that there needs to be two different traces methods: one for successfully sent messages (Informational) and one for errors.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -158,6 +159,24 @@ public class DurableTaskExtension :
...
+        /// <summary>
+        /// Gets or sets the name of the app setting containing the key used for authenticating with the Azure Event Grid custom topic at <see cref="EventGridTopicEndpoint"/>.
+        /// </summary>
+        public string EventGridKeySettingName { get; set; }
+
+        /// <summary>
+        /// Gets or sets the value of the appsetting containing the key used for authenticating with the Azure Event Grid custom topic at <see cref="EventGridKeySettingName"/>
+        /// This value loaded from appsettings. Not from host.json.
+        /// </summary>
+        public string EventGridKeyValue { get; set; }
```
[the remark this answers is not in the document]
TsuyoshiUshio: Done. It works. It is much cooler solution. Thanks!

src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
```diff
@@ -163,6 +164,28 @@ private EtwEventSource()
...
+        [Event(210, Level = EventLevel.Informational)]
+        public void EventGridNotificationFailed(
+            string TaskHub,
+            string AppName,
+            string SlotName,
+            string FunctionName,
+            FunctionState FunctionState,
+            string Version,
+            string InstanceId,
+            string body,
```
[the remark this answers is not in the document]
TsuyoshiUshio: Done

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
```diff
@@ -0,0 +1,240 @@
...
+                        throw new ArgumentException($"Failed to start lifecycle notification feature. Please check the configuration values for {config.EventGridKeySettingName} on AppSettings.");
+                     }
+                }  else
+                {
+                    throw new ArgumentException($"Failed to start lifecycle notification feature. Please check the configuration values for {config.EventGridTopicEndpoint} and {config.EventGridKeySettingName}.");
+                }
+            }
+        }
+
+        private readonly bool useTrace;
```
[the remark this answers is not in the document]
TsuyoshiUshio: Done

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
```diff
@@ -0,0 +1,240 @@
...
+
+                    // Currently, we support Event Grid Custom Topic for notify the lifecycle event of an orchestrator.
+                    // For more detail about the Event Grid, please refer this document.
+                    // Post to custom topic for Azure Event Grid
+                    // https://docs.microsoft.com/en-us/azure/event-grid/post-to-custom-topic
+                    httpClient = new HttpClient();
+                    if (!string.IsNullOrEmpty(config.EventGridKeyValue))
+                    {
+                        httpClient.DefaultRequestHeaders.Add("aeg-sas-key", config.EventGridKeyValue);
+                    }  else
```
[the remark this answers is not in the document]
TsuyoshiUshio: Done

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -252,7 +253,58 @@ public static string LocalSlotName
...
+            string hubName,
+            string functionName,
+            FunctionState functionState,
+            string version,
+            string instanceId,
+            string body, 
+            HttpStatusCode statusCode,
+            string httpStatusCodeReasonPhrase,
```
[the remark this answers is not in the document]
TsuyoshiUshio: Done

src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
```diff
@@ -163,6 +164,28 @@ private EtwEventSource()
...
+            HttpStatusCode StatusCode,
+            string HttpStatusCodeReasonPhrase,
+            string Reason,
+            FunctionType FunctionType,
+            string ExtensionVersion,
+            bool IsReplay,
+            long latency)
```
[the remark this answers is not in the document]
TsuyoshiUshio: Done

src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
```diff
@@ -163,6 +164,28 @@ private EtwEventSource()
         {
             this.WriteEvent(209, TaskHub, AppName, SlotName, FunctionName, Version ?? "", InstanceId, Reason, FunctionType, ExtensionVersion, IsReplay);
         }
+
+        [Event(210, Level = EventLevel.Informational)]
```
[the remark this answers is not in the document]
TsuyoshiUshio: Done

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
```diff
@@ -0,0 +1,240 @@
...
+        {
+            string json = JsonConvert.SerializeObject(eventGridEventArray);
+            StringContent content = new StringContent(json, Encoding.UTF8, "application/json");
+            Stopwatch stopWatch = Stopwatch.StartNew();
+
+            // Details about the Event Grid REST API
+            // https://docs.microsoft.com/en-us/rest/api/eventgrid/
+            HttpResponseMessage result = await httpClient.PostAsync(this.config.EventGridTopicEndpoint, content);
+            var body = await result.Content.ReadAsStringAsync();
+            this.config.TraceHelper.EventGridMessageSent(
```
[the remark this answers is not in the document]
TsuyoshiUshio: Done!

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
```diff
@@ -0,0 +1,240 @@
...
+            string reason,
+            FunctionState functionState)
+        {
+            string json = JsonConvert.SerializeObject(eventGridEventArray);
+            StringContent content = new StringContent(json, Encoding.UTF8, "application/json");
+            Stopwatch stopWatch = Stopwatch.StartNew();
+
+            // Details about the Event Grid REST API
+            // https://docs.microsoft.com/en-us/rest/api/eventgrid/
+            HttpResponseMessage result = await httpClient.PostAsync(this.config.EventGridTopicEndpoint, content);
```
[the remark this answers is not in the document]
TsuyoshiUshio: Done

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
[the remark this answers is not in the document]
TsuyoshiUshio: @cgillum  Do you need sequenceNumber for it? I'm considering to add latencyMs instead. How about that? or do you need both?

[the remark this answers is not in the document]
cgillum: We need both. We use *sequenceNumber* in all traces to help with ordering in Application Insights.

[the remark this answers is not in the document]
TsuyoshiUshio: Sure. I'll do it.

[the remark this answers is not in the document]
TsuyoshiUshio: Hi @cgillum, Done it. Please review it.

test/TestHelpers.cs
TsuyoshiUshio: Done!

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
TsuyoshiUshio: Hi @cgillum, I've done the review. The CI was broken. The error message is related the Storage Account conflict. At least, when I run the test locally, the final commit is all passed.

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -287,21 +287,44 @@ public static string LocalSlotName
                 isReplay,
                 latencyMs);
 
-            this.logger.LogError(
-                "LifeCycleNotificationHelper.SendNotificationAsync - Status: {statusCode}. For more detail: {instanceId}: Function '{functionName} ({functionType})', function state {functionState} version '{version}' failed with an error. Reason: {reason}. IsReplay: {isReplay}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {ExtensionVersion}. Latency: {latencyMs} ms.",
-                statusCode,
-                instanceId,
+            this.logger.LogInformation(
+                "{instanceId}: Function '{functionName} ({functionType})', failed to send a '{functionState}' notification event to Azure Event Grid. Status code: {statusCode}. Details: {details}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}. Latency: {latencyMs} ms.",
```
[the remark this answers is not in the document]
TsuyoshiUshio: Done

src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
```diff
@@ -163,6 +164,47 @@ private EtwEventSource()
...
+            string InstanceId,
+            string Details,
+            HttpStatusCode StatusCode,
+            string Reason,
+            FunctionType FunctionType,
+            string ExtensionVersion,
+            bool IsReplay,
+            long LatencyMs)
+        {
+            this.WriteEvent(210, TaskHub, AppName, SlotName, FunctionName, FunctionState, Version ?? "", InstanceId, Details, StatusCode, Reason, FunctionType, ExtensionVersion, IsReplay, LatencyMs);
```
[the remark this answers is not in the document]
TsuyoshiUshio: Done

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -252,7 +253,78 @@ public static string LocalSlotName
...
+            string instanceId,
+            string details,
+            HttpStatusCode statusCode,
+            string reason,
+            long latencyMs)
+        {
+            FunctionType functionType = FunctionType.Orchestrator;
+            bool isReplay = false;
+
+            EtwEventSource.Instance.EventGridNotificationFailed(
```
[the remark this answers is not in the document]
TsuyoshiUshio: Done

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -252,7 +253,78 @@ public static string LocalSlotName
...
+                details,
+                statusCode,
+                reason,
+                functionType,
+                ExtensionVersion,
+                isReplay,
+                latencyMs);
+
+            this.logger.LogInformation(
+                "{instanceId}: Function '{functionName} ({functionType})', failed to send a '{functionState}' notification event to Azure Event Grid. Status code: {statusCode}. Details: {details}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}. Latency: {latencyMs} ms.",
```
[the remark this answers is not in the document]
TsuyoshiUshio: Done

TsuyoshiUshio:
Hi @cgillum, I finished the fixes. If you find minor mistake or wrong English stuff, feel free to edit it. :) 
However, if you find something takes time, I'll keep on fixing it. Please let me know.

Issue #207, opened 2018-03-25
What is the expected behavior if we are waiting multiple times on the same event name, but with a different shape of data?


SimonLuckenuik:
Ref: [Handling external events in Durable Functions (Azure Functions)](https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-external-events)

Different scenarios we are looking at for multiple systems synchronization. I know there would be workarounds to prevent receiving the same event name multiple times, but my goal here is to clarify the documentation, so let's assume we cannot change the behavior of the external system. Based on the current doc, I am assuming that both examples below are currently supported. Let's assume for both examples that events are raised based on QueueTrigger and that ordering is not guaranteed.

**Example 1**: Wait until a processing is completed or until we receive 3 errors. Number of events is known in advance, but Error value will always be different (or the same). Example of events received: _Error_ event: 503, 503, 404. _Completed_ Event: never received.

```
public static async Task Run([OrchestrationTrigger] DurableOrchestrationContext context)
{
    var completed = context.WaitForExternalEvent<int>("Completed");
    var error1 = context.WaitForExternalEvent<int>("Error");
    var error2 = context.WaitForExternalEvent<int>("Error");
    var error3 = context.WaitForExternalEvent<int>("Error");
    var error = Task.WhenAll(error1, error2, error3);
    var winner = await Task.Any(completed, error);
}
```

**Example 2**: Receive multiple events providing progress on an external system, reflect external system status in the current system status.  Example of events received: _Progress_ event: 0, 5, 10, 30, 70, 85, 90, 100.

```
public static async Task Run([OrchestrationTrigger] DurableOrchestrationContext context)
{
    float progress = 0;
    do
    {
      progress = await context.WaitForExternalEvent<int>("Progress");
      // Reflect external workflow status in the current workflow status.
    }
    while(progress < 100);
}
```
Thank you!

cgillum:
Thanks for posting. I expect Example 2 to succeed but Example 1 to fail. Example 2 was enabled as part of the Beta 3 release (see https://github.com/Azure/azure-functions-durable-extension/issues/141). This fix allowed waiting for events with the same name **sequentially** (which is basically your example 2). However, the fix would not enable listening to events with the same name **in parallel**. A new fix will be needed to support this.

SimonLuckenuik:
Thanks @cgillum, I was expecting the opposite to be supported, since the first example seems to be more predictable.
 
[...]

PR #225, opened 2018-04-04
Support for executing external events with the same name in parallel 

gled4er:
Hello @cgillum ,

May I ask you to check the approach I am taking for implementing this issue - [#207](https://github.com/Azure/azure-functions-durable-extension/issues/207) ?

I added the test you implemented in **origin/multi-external-event** branch.

Let me know if I need to change that. 

Thank you!

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Reflection;
 using System.Threading;
@@ -253,12 +254,29 @@ namespace Microsoft.Azure.WebJobs
 
             lock (this.pendingExternalEvents)
             {
-                object tcsRef;
+                object tcsRefList;
+                List<TaskCompletionSource<T>> tcsList;
                 TaskCompletionSource<T> tcs;
-                if (!this.pendingExternalEvents.TryGetValue(name, out tcsRef) || (tcs = tcsRef as TaskCompletionSource<T>) == null)
+                if (!this.pendingExternalEvents.TryGetValue(name, out tcsRefList) ||
+                    (tcsList = tcsRefList as List<TaskCompletionSource<T>>) == null ||
+                    (tcsList.Count == 0) ||
+                    (tcs = tcsList[0] as TaskCompletionSource<T>) == null)
                 {
                     tcs = new TaskCompletionSource<T>();
-                    this.pendingExternalEvents[name] = tcs;
+                    tcsList = new List<TaskCompletionSource<T>>
+                    {
+                        tcs,
+                    };
+                    this.pendingExternalEvents[name] = tcsList;
+                }
+                else
+                {
+                    if (tcsList != null && tcsList.Count > 0)
+                    {
+                        tcs = new TaskCompletionSource<T>();
+                        tcsList.Add(tcs);
+                        this.pendingExternalEvents[name] = tcsList;
+                    }
                 }
 
                 this.config.TraceHelper.FunctionListening(
@@ -412,15 +430,30 @@ namespace Microsoft.Azure.WebJobs
         {
             lock (this.pendingExternalEvents)
             {
-                object tcs;
-                if (this.pendingExternalEvents.TryGetValue(name, out tcs))
+                object tcsObject;
+                if (this.pendingExternalEvents.TryGetValue(name, out tcsObject))
                 {
+                    var tcsList = tcsObject as IList;
+                    if (tcsList == null || tcsList.Count == 0)
+                    {
+                        return;
+                    }
+
+                    var tcs = tcsList[0];
                     Type tcsType = tcs.GetType();
                     Type genericTypeArgument = tcsType.GetGenericArguments()[0];
 
                     // If we're going to raise an event we should remove it from the pending collection
                     // because otherwise WaitForExternal() will always find one with this key and run infinitely. Fixes #141
-                    this.pendingExternalEvents.Remove(name);
+                    if (tcsList.Count == 1)
+                    {
+                        this.pendingExternalEvents.Remove(name);
+                    }
+                    else
+                    {
+                        tcsList.RemoveAt(0);
+                        this.pendingExternalEvents[name] = tcsList;
+                    }
 
                     object deserializedObject = MessagePayloadDataConverter.Default.Deserialize(input, genericTypeArgument);
                     MethodInfo trySetResult = tcsType.GetMethod("TrySetResult");
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
@@ -23,8 +23,8 @@ namespace Microsoft.Azure.WebJobs
         private const string DefaultVersion = "";
         private const int MaxTimerDurationInDays = 6;
 
-        private readonly Dictionary<string, ICollection> pendingExternalEvents =
-            new Dictionary<string, ICollection>(StringComparer.OrdinalIgnoreCase);
+        private readonly Dictionary<string, Queue> pendingExternalEvents =
+            new Dictionary<string, Queue>(StringComparer.OrdinalIgnoreCase);
 
         private readonly DurableTaskExtension config;
         private readonly string orchestrationName;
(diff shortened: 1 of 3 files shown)


gled4er:
Hello @cgillum ,

Thank you for the comment.

I wanted to check quickly with you if you are thinking of changing to **LinkedList** the private `readonly Dictionary<string, object> pendingExternalEvents =
            new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);` or the new list we added to store multiple event data with same name - `List<TaskCompletionSource<T>>`.

I think the dictionary will be more performant than the **LinkedList**  so I think you are meaning to change the `List<TaskCompletionSource<T>>` to the **LinkedList**, right?

gled4er:
Hello @cgillum ,

Assuming my guess from the previous comment is correct and we are talking about changing **List** to **LinkedList**, I think the code will get more complicated.

Now I am able to cast the object I retrieve in **RaiseEvent** method to **IList** and then relaying on its functionality I can relatively easy get the count and remove the first element:

```csharp
var tcsList = tcsObject as IList;
if (tcsList == null || tcsList.Count == 0)
{
   return;
}

var tcs = tcsList[0];
Type tcsType = tcs.GetType();
Type genericTypeArgument = tcsType.GetGenericArguments()[0];

// If we're going to raise an event we should remove it from the pending collection
// because otherwise WaitForExternal() will always find one with this key and run infinitely. Fixes #141
if (tcsList.Count == 1)
{
    this.pendingExternalEvents.Remove(name);
}
else
{
    tcsList.RemoveAt(0);
    this.pendingExternalEvents[name] = tcsList;
}
```

If we move to **LinkedList**, we will not be able to leverage so useful interface  to achieve the same. This will mean that I will need to rely on Reflection but in the same time I don't know the exact type of `TResult` in `TaskCompletionSource<TResult>` so in my opinion the code will get more difficult to reason.
 
In the same time I think **LinkeList** will give us real advantage only if we want to perform random insertions around element from the list we already have or we want to insert very significant number of elements that will be difficult for the framework to allocate sequential memory for a normal **List**.

One good comparison I found is [here](https://stackoverflow.com/questions/604192/listt-or-linkedlistt)

However, I will be happy to proceed with the changes if you think that they will give us advantage.

Thank you!

cgillum:
I agree with you that my original suggestion might not work well. Here is my updated suggestion: let's continue to use a dictionary but with two slight modifications:

1. In `WaitForExternalEventAsync<T>`, let's add validation that requires multiple events with the same name must also have the same type argument. I think this is a reasonable constraint and help ensure that **RaiseEvent** always behaves in a predictable way.
2. Let's use `Queue<T>` instead of `List<T>`. The reason is that `List<T>.RemoveAt(0)` is an O(N) operation, which is unnecessarily slow. Instead of casting to `IList` you can use `ICollection` to get the count. `Queue<T>` is nice because it implements the exact semantics we want and communicates to the reader that we resume events with the same name in FIFO order.

I think we can make one other small improvement by changing `pendingExternalEvents` into a `Dictionary<string, ICollection>` instead of `Dictionary<string, object>`. That will allow us to remove the cast and makes the intent of this field more clear. Let me know what you think about this alternate proposal.

gled4er:
Hello @cgillum ,

Thank you for the updated suggestion.

I will work on the modifications.

Thank you!

cgillum:
I was thinking about this more after I posted it, and this might still have the reflection problem you mentioned before. I'm not sure how you would cast the ICollection into a Queue<T> when you don't know at compile time what type T is. We could write our own data structure which acts like a non-generic queue, but that might be too much work compared to what you have now.

So in the end, I think the strategy you've chosen might be the best. It's not the most optimal because of IList.RemoveAt(0), but simplicity is critically important when building software and I think the performance bit will very rarely be an issue in real applications. Why don't we go with what you have now and clean it up a little bit. I'll add some feedback to our existing commits.

gled4er:
Hello @cgillum ,

Thank you for your feedback.

I was working on your previous suggestion and I came up with some changes that are using queue.

I will be happy to follow your latest suggestions but before that may I ask you to check my latest changes and see if they are acceptable.

I think the main think we need to consider if it is acceptable to have instead of casting initialization of a queue based on the provided `IColleiton` in `RaiseEvent`:

```csharp
 Queue tcsQueue = new Queue(tcsColleciton);
```
Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -430,29 +430,29 @@ internal void RaiseEvent(string name, string input)
...
                 {
-                    var tcsList = tcsObject as IList;
-                    if (tcsList == null || tcsList.Count == 0)
+                    if (tcsColleciton == null || tcsColleciton.Count == 0)
                     {
                         return;
                     }
 
-                    var tcs = tcsList[0];
+                    Queue tcsQueue = new Queue(tcsColleciton);
```
cgillum: The downside of this approach is that the contents of the queue is being copied every time we receive an event. This is very inefficient, and potentially worse than the `IList.RemoveAt(0)` performance from the previous iteration.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -23,8 +23,8 @@ public sealed class DurableOrchestrationContext : DurableOrchestrationContextBas
         private const string DefaultVersion = "";
         private const int MaxTimerDurationInDays = 6;
 
-        private readonly Dictionary<string, object> pendingExternalEvents =
-            new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
+        private readonly Dictionary<string, ICollection> pendingExternalEvents =
+            new Dictionary<string, ICollection>(StringComparer.OrdinalIgnoreCase);
```
cgillum: Let's switch to `Dictionary<string, Queue>`.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -254,28 +254,28 @@ public override Task<T> WaitForExternalEvent<T>(string name)
 
             lock (this.pendingExternalEvents)
             {
-                object tcsRefList;
-                List<TaskCompletionSource<T>> tcsList;
+                ICollection tcsRefCollection;
+                Queue<TaskCompletionSource<T>> tcsQueue;
```
cgillum: Now that I know there is a non-generic `Queue` class, we should simply use that instead of `ICollection` or `Queue<T>`.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -254,28 +254,28 @@ public override Task<T> WaitForExternalEvent<T>(string name)
...
-                    };
-                    this.pendingExternalEvents[name] = tcsList;
+                    tcsQueue = new Queue<TaskCompletionSource<T>>();
+                    tcsQueue.Enqueue(tcs);
+                    this.pendingExternalEvents[name] = tcsQueue;
                 }
                 else
                 {
-                    if (tcsList != null && tcsList.Count > 0)
+                    if ((tcsQueue = tcsRefCollection as Queue<TaskCompletionSource<T>>) == null)
```
cgillum: If we use `Queue` instead of `Queue<TaskCompletionSource<T>>`, then I think we will need to do a peek of the first element to learn whether `T` is the correct type.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
cgillum: If we use `Queue` instead of `ICollection`, then this inefficient conversion won't be necessary.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -430,29 +430,29 @@ internal void RaiseEvent(string name, string input)
...
-                    if (tcsList.Count == 1)
+                    if (tcsQueue.Count == 0)
                     {
                         this.pendingExternalEvents.Remove(name);
                     }
                     else
                     {
-                        tcsList.RemoveAt(0);
-                        this.pendingExternalEvents[name] = tcsList;
+                        this.pendingExternalEvents[name] = tcsQueue;
```
cgillum: This entire else block can be removed.

gled4er:
Hello @cgillum,

Thank you very much for the great feedback!

I will work on fixing the issues you highlighted as soon as possible.

Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -254,28 +254,28 @@ public override Task<T> WaitForExternalEvent<T>(string name)
...
+                    {
+                        throw new ArgumentException("Events with the same name should have the same type argument.");
+                    }
+                    else
                     {
                         tcs = new TaskCompletionSource<T>();
-                        tcsList.Add(tcs);
-                        this.pendingExternalEvents[name] = tcsList;
+                        tcsQueue.Enqueue(tcs);
+                        this.pendingExternalEvents[name] = tcsQueue;
```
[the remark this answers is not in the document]
gled4er: Thank you! I will remove it.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -430,29 +430,29 @@ internal void RaiseEvent(string name, string input)
...
             lock (this.pendingExternalEvents)
             {
-                object tcsObject;
-                if (this.pendingExternalEvents.TryGetValue(name, out tcsObject))
+                ICollection tcsColleciton;
+                if (this.pendingExternalEvents.TryGetValue(name, out tcsColleciton))
                 {
-                    var tcsList = tcsObject as IList;
-                    if (tcsList == null || tcsList.Count == 0)
+                    if (tcsColleciton == null || tcsColleciton.Count == 0)
```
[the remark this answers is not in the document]
gled4er: Thank you very much! Removing this check will make the code simpler and easier to reason.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
gled4er: Great! I will make the modifications. Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
gled4er: Thank you! I will change it.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
gled4er: Thank you! I will change it.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
gled4er: Yes, this will be great optimization. Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
gled4er: Great! This will simplify the code. Thank you!

gled4er:
Hello @cgillum ,

I optimized the code following your directions.

When you have time may I ask you to check it once more?

Thank you!

Issue #26, opened 2017-08-01
Large Message Support


cgillum:
## Summary
This enhancement allows orchestrator and activity functions to send large argument or event payloads when invoking other functions. By default all function call payloads are serialized into a queue message. If the size of the payload exceeds the queue message limit, then the payload will instead be stored in blob storage. This happens automatically and the function code does not need to be aware of it.

## Motivation
Many users will have scenarios where they need to deliver large payloads to other functions. Having support for this would help save developers a lot of time and effort, especially when they may not be aware of the size limitations of queues (or may not be aware that some payloads will exceed those sizes).

mayoatbm:
Hi,

We're running into this issue (i've read your SO posts etc  on this). 

What patterns do you recommend to work around this problem while still being able to use durable functions? Our scenario involves getting a large (1MB) amount of data and then processing a large amount of data from redis (an action per redis key essentially). Right now, there's no way to get the data back to the orchestrator function (for it to spin up activity functions to process the data).

cgillum:
Do you have the option of compressing the data? For example, could you have the activity function return a string and write code that serializes the payload, compresses it using GZip, and then returning the base64-encoded version.

mayoatbm:
We do, but i'm not sure it's going to help in our particular scenario. We're retrieving a large amount of keys from Redis (tens of thousands) and even compressed, we'll likely be over the 64KB limit.  Unfortunately redis also doesn't seem to have a way for us to get keys in a partitioned manner (so that we could spin up activity functions per partition) .

This is somewhat urgent so i went back to vanilla functions and queues in the meantime.

cgillum:
OK, thanks for letting me know, and sorry that this was a blocker for you. :(

mayoatbm:
Anytime. And no worries; these are early days still for durable functions and i'm excited for what we'll be able to do once the platform matures a bit. 

Thanks to you and the team for all your hard work on it.

JamesInDenver:
This would also be very useful when fanning out.  I've already had to minimize chaining a bit as I can't pass messages around (too big), but now I'm looking at the orchestrator doing more work (loading blobs directly) than it should to manage fanning.

SimonLuckenuik:
Will this be fixed for GA? I currently need to use Blob path outputs from some of my activities just to make sure, removing that overhead and having the framework manage it would be interesting for us.

cgillum:
[...]

PR #242, opened 2018-04-11
Enabling large message support.

tohling:
1. Removed guard max message payload size.
2. Updated tests for large message payloads.


cgillum:
Linking to issue: https://github.com/Azure/azure-functions-durable-extension/issues/26

Issue #84, opened 2017-10-30
Details about activity function failures do not flow to the calling orchestrator function


cgillum:
### Summary
When an activity function throws an exception, the orchestrator function does not see the exception. Instead, it sees a generic `TaskFailedException` (defined in DurableTask.Core.dll) which does not include any details about the original exception.

### Expected Behavior
1. `TaskFailedException` should not be directly exposed in orchestrator functions since it is defined in DurableTask.Core.dll
2. The details of the activity function exception should be preserved. At a minimum, it should be preserved as an inner exception.

### Motivation
It's currently not possible to catch specific exception types when calling activity functions since the original exception information is lost. It's also not possible to fetch the details of the original exception when getting the failed orchestrator status - i.e. all that is shown is "Exception while executing function: {activity function name}". Exception details from orchestrator functions flow fine.

### Other Notes
It's not yet clear how well (if at all) exceptions in sub-orchestrations flow to the parent orchestration. A fix for this issue should also be made to work with sub-orchestrations.

markheath:
Yes, the inner exception is needed when using the `RetryOptions.Handle` func with `CallActivityWithRetryAsync`. Otherwise there's no easy way to make a decision whether to retry or not.

cgillum:
Agreed. FYI @simonporter since I think we will need DTFx support for this.

simonporter:
To stop throwing a TaskFailedException represents a breaking change so this would be a larger change.
To preserve the inner exception is an outstanding ask I can try to address in the short term.

cgillum:
We just checked-in support for this after getting the changes we needed in **DurableTask.Core** (thanks Simon!). It will go out with the next release.

markheath:
I've just been experimenting with this (using local dev environment) and although we have the `InnerException` available now, with the correct message, it does not have the same type as the exception thrown from the activity function - you just get a basic `System.Exception`. Is this intended behavior? It would be nice to do something like

```
new RetryOptions { Handle = ex => ex.InnerException is InvalidOperationException }
```

but that is not currently possible.

vjhbtigoyal:
This is still happening in V2 version. I checked in my local environment. ex.InnerException is coming as System.Exception so my error handling scenario is not able to process it accordingly.
In my case i was raising a custom exception in scenarios where i need retries. Any suggested way to handle that exception in Handle method?

cgillum:
[...]

Issue #146, opened 2018-01-29
Orchestration loops forever when terminated


markheath:
I'm experimenting with eternal orchestrations, where I call an activity, sleep with a timer, and then `ContinueAsNew`. After its run for a few loops I attempt to terminate the orchestration, which returns a 202 accepted. 

Once I have done this, my activity function doesn't get called anymore, but when I query the orchestration status, I still get `runtimeStatus` of `Running`. 

```js
{
    runtimeStatus: "Running",
    input: 5,
    output: null,
    createdTime: "2018-01-29T21:16:01Z",
    lastUpdatedTime: "2018-01-29T21:16:11Z"
}
```

Then every five minutes the functions runtime (I'm testing locally) will run the orchestrator function. I can hit breakpoints and step through, but it doesn't actually call the activity function - it just steps through the whole orchestrator to continueasnew.

```
[29/01/2018 21:32:56] Function started (Id=6aa34ab0-0d59-4d97-8c49-aaf5ab48dbc8)
[29/01/2018 21:32:56] Executing 'O_PeriodicTask' (Reason='', Id=6aa34ab0-0d59-4d97-8c49-aaf5ab48dbc8)
[29/01/2018 21:32:56] c611b9c068754d078d1dc3419bb0c4ef: Function 'O_PeriodicTask (Orchestrator)', version '' started. IsReplay: True. Input: (4 bytes). State: Started. HubName: DurableFunctionsHub. AppName: . SlotName: . ExtensionVersion: 1.1.0.0.
[29/01/2018 21:32:56] c611b9c068754d078d1dc3419bb0c4ef: Function 'A_PeriodicActivity (Activity)', version '' scheduled. Reason: O_PeriodicTask. IsReplay: True. State: Scheduled. HubName: DurableFunctionsHub. AppName: . SlotName: . ExtensionVersion: 1.1.0.0.
[29/01/2018 21:32:56] c611b9c068754d078d1dc3419bb0c4ef: Function 'A_PeriodicActivity (Activity)', version '' completed. ContinuedAsNew: False. IsReplay: True. Output: (replayed). State: Completed. HubName: DurableFunctionsHub. AppName: . SlotName: . ExtensionVersion: 1.1.0.0.
[...]

PR #250, opened 2018-04-13
Flow exception details between orchestrator and activity functions

cgillum:
Fixes https://github.com/Azure/azure-functions-durable-extension/issues/84. The new behavior is that activity and sub-orchestrator functions will continue to throw `FunctionFailedException`, but the `InnerException` will now contain the details of the original exception.

This required taking a new version of **DurableTask.Core**, which has fixes for https://github.com/Azure/azure-functions-durable-extension/issues/146 as well.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Listener/TaskActivityShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/TaskActivityShim.cs
@@ -2,10 +2,12 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
-using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
 using DurableTask.Core;
+using DurableTask.Core.Common;
+using DurableTask.Core.Exceptions;
+using Microsoft.Azure.WebJobs.Host;
 using Microsoft.Azure.WebJobs.Host.Executors;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
@@ -59,20 +61,24 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             FunctionResult result = await this.executor.TryExecuteAsync(triggerInput, CancellationToken.None);
             if (!result.Succeeded)
             {
+                // Flow the original activity function exception to the orchestration
+                // without the outer FunctionInvocationException.
+                Exception exceptionToReport = StripFunctionInvocationException(result.Exception);
+
                 this.config.TraceHelper.FunctionFailed(
                     this.config.HubName,
                     this.activityName,
                     this.activityVersion,
                     instanceId,
-                    result.Exception?.ToString() ?? string.Empty,
+                    exceptionToReport?.ToString() ?? string.Empty,
                     functionType: FunctionType.Activity,
                     isReplay: false);
 
-                if (result.Exception != null)
+                if (exceptionToReport != null)
                 {
-                    // Preserve the original exception context so that the durable task
-                    // framework can report useful failure information.
-                    ExceptionDispatchInfo.Capture(result.Exception).Throw();
+                    throw new TaskFailureException(
+                        $"Activity function '{this.activityName}' failed: {exceptionToReport.Message}",
+                        Utils.SerializeCause(exceptionToReport, MessagePayloadDataConverter.Default));
                 }
             }
 
@@ -94,6 +100,17 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             // This won't get called as long as we've implemented RunAsync.
             throw new NotImplementedException();
+        }
+
+        private static Exception StripFunctionInvocationException(Exception e)
+        {
+            var infrastructureException = e as FunctionInvocationException;
+            if (infrastructureException?.InnerException != null)
+            {
+                return infrastructureException.InnerException;
+            }
+
+            return e;
         }
     }
 }
--- a/src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
@@ -4,6 +4,8 @@
 using System;
 using System.Threading.Tasks;
 using DurableTask.Core;
+using DurableTask.Core.Common;
+using DurableTask.Core.Exceptions;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 {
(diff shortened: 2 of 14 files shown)


src/WebJobs.Extensions.DurableTask/MessagePayloadDataConverter.cs
```diff
@@ -56,6 +53,20 @@ public string Serialize(object value, int maxSizeInKB)
...
+                int payloadSizeInKB = (int)(Encoding.Unicode.GetByteCount(serializedJson) / 1024.0);
+                if (payloadSizeInKB > maxSizeInKB)
+                {
+                    throw new ArgumentException(
+                        string.Format(
+                            "The UTF-16 size of the JSON-serialized payload must not exceed {0:N0} KB. The current payload size is {1:N0} KB.",
+                            maxSizeInKB,
+                            payloadSizeInKB));
+                }
+            }
```
tohling: Why was this added back?

src/WebJobs.Extensions.DurableTask/Listener/TaskActivityShim.cs
```diff
@@ -59,20 +61,24 @@ internal class TaskActivityShim : TaskActivity
...
                     this.activityName,
                     this.activityVersion,
                     instanceId,
-                    result.Exception?.ToString() ?? string.Empty,
+                    exceptionToReport?.ToString() ?? string.Empty,
                     functionType: FunctionType.Activity,
                     isReplay: false);
 
-                if (result.Exception != null)
+                if (exceptionToReport != null)
```
tohling: >null [](start = 41, length = 4)

If for some reason the strip failed, should we log internally what the exception was?

cgillum: I don't think we actually ever expect `exceptionToReport` to be null. I'm not sure what you mean when you say "for some reason the strip failed". Can you elaborate?

---
In reply to: [181521966](https://github.com/Azure/azure-functions-durable-extension/pull/250#discussion_r181521966) [](ancestors = 181521966)

tohling: I was referring to the method StripFunctionInvocationException. If we expect 

e as FunctionInvocationException to never return null, then it's OK.

---
In reply to: [181523045](https://github.com/Azure/azure-functions-durable-extension/pull/250#discussion_r181523045) [](ancestors = 181523045,181521966)

cgillum: Right - I'm saying I don't think that method will ever return null. `e as FunctionInvocationException` might return null, but in that case we return `e` directly, so there shouldn't be any nulls. I can remove the `?` operator if that would make things less confusing.

---
In reply to: [181523423](https://github.com/Azure/azure-functions-durable-extension/pull/250#discussion_r181523423) [](ancestors = 181523423,181523045,181521966)

cgillum: Actually, I'll probably keep the null checks as they are. I don't have any guarantee that `result.Exception` will be non-null when `result.Succeeded` is false since that depends on the behavior of WebJobs. But to your original concern, I don't think there is ever a case where this is an exception and we fail to log it.

---
In reply to: [181523733](https://github.com/Azure/azure-functions-durable-extension/pull/250#discussion_r181523733) [](ancestors = 181523733,181523423,181523045,181521966)

src/WebJobs.Extensions.DurableTask/Listener/TaskActivityShim.cs
```diff
@@ -59,20 +61,24 @@ internal class TaskActivityShim : TaskActivity
...
                 {
-                    // Preserve the original exception context so that the durable task
-                    // framework can report useful failure information.
-                    ExceptionDispatchInfo.Capture(result.Exception).Throw();
+                    throw new TaskFailureException(
```
cgillum: >throw new TaskFailureException( [](start = 20, length = 31)

We throw `TaskFailureException` here because that is the only exception type where the Durable Task Framework will actually try to deserialize the inner exception back in the orchestrator function.

test/TestHelpers.cs
```diff
@@ -298,13 +298,13 @@ private static List<string> GetLogs_UnhandledActivityException(string messageId,
             {
                 $"{messageId}: Function '{orchestratorFunctionNames[0]} ({FunctionType.Orchestrator})', version '' scheduled. Reason: NewInstance. IsReplay: False.",
                 $"{messageId}: Function '{orchestratorFunctionNames[0]} ({FunctionType.Orchestrator})', version '' started. IsReplay: False. Input: \"Kah-BOOOOM!!!\"",
-                $"{messageId}: Function '{activityFunctionName} ({FunctionType.Activity})', version '' scheduled. Reason: Throw. IsReplay: False.",
+                $"{messageId}: Function '{activityFunctionName} ({FunctionType.Activity})', version '' scheduled. Reason: ThrowOrchestrator. IsReplay: False.",
                 $"{messageId}: Function '{orchestratorFunctionNames[0]} ({FunctionType.Orchestrator})', version '' awaited. IsReplay: False.",
                 $"{messageId}: Function '{activityFunctionName} ({FunctionType.Activity})', version '' started. IsReplay: False. Input: [\"Kah-BOOOOM!!!\"]",
-                $"{messageId}: Function '{activityFunctionName} ({FunctionType.Activity})', version '' failed with an error. Reason: Microsoft.Azure.WebJobs.Host.FunctionInvocationException",
```
tohling: >Microsoft.Azure.WebJobs.Host.FunctionInvocationException [](start = 133, length = 56)

The 'Reason: Microsoft.Azure.WebJobs.Host.FunctionInvocationException' entry in all these logs is also giving me a quick clue that the problem is a FunctionInvocationException, vs FunctionTimeoutException. In long and complex execution workflows, it may help quickly identify issues. We lose this visibility with this change. Do you think we really don't need to surface this information?

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -81,27 +84,31 @@ public void SetFunctionInvocationCallback(Func<Task> callback)
...
                         this.context.Version,
                         this.context.InstanceId,
-                        e.ToString(),
+                        exceptionDetails,
                         FunctionType.Orchestrator,
                         this.context.IsReplaying));
                 }
 
-                throw;
+                throw new OrchestrationFailureException(
```
cgillum: >OrchestrationFailureException [](start = 26, length = 29)

`OrchestrationFailureException` is the equivalent of `TaskFailureException` but for orchestrations and sub-orchestrations. We need to throw this exception specifically to ensure that the exception details are surfaced to the calling orchestrator function.

test/TestHelpers.cs
cgillum: I've never heard of `FunctionTimeoutException`. If a function takes to long to execute, WebJobs will recycle the host. But even if we do have other exception types, I'm only ever stripping `FunctionInvocationException`, so you should still be able to figure out what's going on, right? In any case, customers have never needed to deal with these internal WebJobs exceptions, so I didn't want to surface them here.

---
In reply to: [181524809](https://github.com/Azure/azure-functions-durable-extension/pull/250#discussion_r181524809) [](ancestors = 181524809)

Issue #261, opened 2018-04-17
NullReferenceException in v1.3.0-rc when querying an instance which doesn't exist


jeffputz:
Since updating today to v1.3RC, we're getting null ref errors similar to this:

```
System.Exception: [demo]  ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Azure.WebJobs.DurableOrchestrationClient.<GetDurableOrchestrationStatusAsync>d__18.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.Azure.WebJobs.DurableOrchestrationClient.<GetStatusAsync>d__15.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at AMS.Functions.GroupUpdateCalcDispatcher.<Run>d__0.MoveNext()
   --- End of inner exception stack trace ---
```

The signature of these functions looks like this:

```
[FunctionName("GroupUpdateCalcDispatcher")]
public static async Task Run([QueueTrigger(QueueNames.GroupCalculation)]string queueMessage, [OrchestrationClient]DurableOrchestrationClient starter, TraceWriter log)
```

If this is caused by the breaking changes, it's not obvious to me. Looking at the source, I'm assuming that it isn't getting the orchestration state, though it's a little black box for me to understand why. It's obviously triggering on the queue, because otherwise it wouldn't run at all.

SimonLuckenuik:
I have the exact same issue, happens whenever the provided _instanceId_ doesn't exists: 
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Azure.WebJobs.DurableOrchestrationClient.<GetDurableOrchestrationStatusAsync>d__18.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #262, opened 2018-04-17
Fix NullReferenceException when querying nonexistent instance ID.

cgillum:
Fixes https://github.com/Azure/azure-functions-durable-extension/issues/261, which was a regression in the initial RC release.


cgillum:
This was fixed in a different commit.

Issue #248, opened 2018-04-12
Update all samples to use ILogger instead of TraceWriter


cgillum:
`TraceWriter` has been deprecated and everything should move on to using `ILogger`.

cgillum:
Actually, it looks like switching to `ILogger` does not cause output to be written to the console or Azure Portal! https://github.com/Azure/azure-functions-core-tools/issues/130. I would consider that a blocker for this work item.

PR #267, opened 2018-04-19
Replacing TraceWriter with ILogger

gled4er:
Hello @cgillum,

May I ask you when you have time to review these changes addressing [#248](https://github.com/Azure/azure-functions-durable-extension/issues/248)?

For me the logging with `ILogger` only works when I use Functions v1 which is consistent with the comments in [#130](https://github.com/Azure/azure-functions-core-tools/issues/130).

I think this will be confusing for the developers and we need to explicitly clarify it in the documentation.

Thank you!

### DIFF
--- a/samples/VSSample.Tests/HttpStartTests.cs
+++ b/samples/VSSample.Tests/HttpStartTests.cs
@@ -4,13 +4,12 @@
 namespace VSSample.Tests
 {
     using System;
-    using System.Diagnostics;
     using System.Net;
     using System.Net.Http;
     using System.Text;
     using System.Threading.Tasks;
     using Microsoft.Azure.WebJobs;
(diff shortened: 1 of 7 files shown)


cgillum:
My understanding is that `ILogger` only works with v2 and not for v1. Are you seeing something different?

Per the issue you linked to, we have to wait for the next CLI release before the traces will be written to the console in v1, so let's hold off on merging until we can verify this.

Can you also apply this change to the csx examples? Those are expected to run in v2 and should be working.

gled4er:
Hello @cgillum,

Yes, you are right. Thank you for the clarification.

I modified the C# Script samples project.

May I ask you to take a look once more and let me know what should be the next issues that need to be fixed for this PR?

Thank you!

Issue #235, opened 2018-04-09
Add F# samples


mikhailshilkov:
Extend samples with some F# examples similar to C# precompiled project

mikhailshilkov:
Created initial commit at https://github.com/mikhailshilkov/azure-functions-durable-extension/commit/cc9129023ddc298ce700cd8509ebb0d974e97620

`Monitor` sample is still missing.
Only compiles for .NET Standard 2.0.

@cgillum Would that work?

cgillum:
This looks great! The only change I would make is to update the nuget package version to 1.2.**2**-beta3 (makes it compatible with the latest v2.0 runtime in Azure) and name the directory **fsharp** instead of **fs**.

PR #268, opened 2018-04-19
F# samples

mikhailshilkov:
F# samples for .NET Standard 2.0, parity with C# precompiled samples. Addresses https://github.com/Azure/azure-functions-durable-extension/issues/235


mikhailshilkov:
F# is installed with .NET Core SDK, so nothing special should be required.
I build it with `dotnet` CLI and run with `func` CLI. Visual Studio build is OK too.

PR #269, opened 2018-04-19
Nodejs v2

kashimiz:
Enables interpretation/output of JSON schema for out of proc orchestrators.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
@@ -10,6 +10,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using DurableTask.Core;
 using DurableTask.Core.Exceptions;
+using DurableTask.Core.History;
 using Microsoft.Azure.WebJobs.Extensions.DurableTask;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
@@ -74,6 +75,8 @@ namespace Microsoft.Azure.WebJobs
 
         internal bool IsOutputSet => this.serializedOutput != null;
 
+        internal IList<HistoryEvent> History { get; set; }
+
         internal void AssignToCurrentThread()
         {
             this.owningThreadId = Thread.CurrentThread.ManagedThreadId;
@@ -97,9 +100,8 @@ namespace Microsoft.Azure.WebJobs
         /// <returns>
         /// The parsed <c>JToken</c> representation of the orchestrator function input.
         /// </returns>
-        public JToken GetInputAsJson()
+        internal JToken GetInputAsJson()
         {
-            this.ThrowIfInvalidAccess();
             return this.serializedInput != null ? JToken.Parse(this.serializedInput) : null;
         }
 
@@ -118,10 +120,14 @@ namespace Microsoft.Azure.WebJobs
             return MessagePayloadDataConverter.Default.Deserialize<T>(this.serializedInput);
         }
 
-        internal void SetInput(OrchestrationContext frameworkContext, string rawInput)
+        internal void SetInput(string rawInput)
+        {
+            this.serializedInput = rawInput;
+        }
+
+        internal void SetInnerContext(OrchestrationContext frameworkContext)
         {
             this.innerContext = frameworkContext;
-            this.serializedInput = rawInput;
         }
 
         /// <summary>
--- a/src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
@@ -2,10 +2,14 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using System.Threading;
 using System.Threading.Tasks;
 using DurableTask.Core;
 using DurableTask.Core.Common;
 using DurableTask.Core.Exceptions;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 {
@@ -14,6 +18,17 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
     /// </summary>
     internal class TaskOrchestrationShim : TaskOrchestration
     {
+        private enum AsyncActionType
+        {
+            CallActivity = 0,
+            CallActivityWithRetry = 1,
+            CallSubOrchestrator = 2,
+            CallSubOrchestratorWithRetry = 3,
+            ContinueAsNew = 4,
+            CreateTimer = 5,
+            WaitForExternalEvent = 6,
+        }
+
         private readonly DurableTaskExtension config;
         private readonly DurableOrchestrationContext context;
 
@@ -47,7 +62,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             }
 
             this.context.AssignToCurrentThread();
-            this.context.SetInput(innerContext, serializedInput);
+            this.context.SetInnerContext(innerContext);
+            this.context.SetInput(serializedInput);
 
             this.config.TraceHelper.FunctionStarting(
                 this.context.HubName,
@@ -57,6 +73,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 this.config.GetIntputOutputTrace(serializedInput),
                 FunctionType.Orchestrator,
                 this.context.IsReplaying);
+
+            var orchestratorInfo = this.config.GetOrchestratorInfo(new FunctionName(this.context.Name, this.context.Version));
 
             if (!this.context.IsReplaying)
             {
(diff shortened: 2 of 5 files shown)


src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -97,9 +100,8 @@ public string GetRawInput()
         /// <returns>
         /// The parsed <c>JToken</c> representation of the orchestrator function input.
         /// </returns>
-        public JToken GetInputAsJson()
+        internal JToken GetInputAsJson()
```
cgillum: Were you following my GitHub thread with Brandon? :)

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -75,7 +82,7 @@ public void SetFunctionInvocationCallback(Func<Task> callback)
                 Task invokeTask = this.functionInvocationCallback();
                 if (invokeTask is Task<object> resultTask)
                 {
-                    returnValue = await resultTask;
+                    returnValue = orchestratorInfo.IsOutOfProc ? resultTask.Result : await resultTask;
```
cgillum: The call to `result.Task.Result` - that's a blocking I/O operation, right? Let's add a comment calling that out, because it could have performance implications and may need to be revisited in the future. IIRC, the reason we needed to do this was because this is an orchestration thread and orchestrators cannot do async I/O (which is why we have the thread ID check that tries to detect this and throw an exception).

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -117,7 +124,60 @@ public void SetFunctionInvocationCallback(Func<Task> callback)
 
             if (returnValue != null)
             {
-                this.context.SetOutput(returnValue);
+                if (orchestratorInfo.IsOutOfProc && returnValue is JObject)
+                {
+                    var jObj = returnValue as JObject;
+                    if (jObj["actions"] != null)
```
cgillum: In what cases do we expect `jObj["actions"]` to be null? Is this a normal condition or is this a case where it might be useful to throw an exception?

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -117,7 +124,60 @@ public void SetFunctionInvocationCallback(Func<Task> callback)
...
+                    {
+                        var commandSets = JsonConvert.DeserializeObject<AsyncCommand[][]>(jObj["actions"].ToString());
```
cgillum: I see a few places where we're doing dictionary lookups into `jObj`. Would it make sense to just deserialize the entire `jObj` into a strongly-typed class?  I see we're doing that for the `AsyncCommand[][]` but I'm wondering if we should deserialize starting with the outer `jObj`.

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -117,7 +124,60 @@ public void SetFunctionInvocationCallback(Func<Task> callback)
...
+
+                        foreach (AsyncCommand[] commandSet in commandSets)
```
cgillum: I suggest we add some comments describing what the array of arrays represents - e.g. the first dimension represents a particular episode of the orchestration and the second dimension represents all the actions that were scheduled within that episode (I hope I'm understanding it correctly).

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -117,7 +124,60 @@ public void SetFunctionInvocationCallback(Func<Task> callback)
...
+                                    case AsyncActionType.CallActivity:
+                                        tasks.Add(this.context.CallActivityAsync(command.FunctionName, command.Input));
+                                        break;
+                                    case AsyncActionType.CreateTimer:
+                                        using (var cts = new CancellationTokenSource())
+                                        {
+                                            tasks.Add(this.context.CreateTimer(command.FireAt, cts.Token));
+                                            if (command.IsCanceled)
+                                            {
+                                                cts.Cancel();
```
cgillum: I'm not sure how timer cancellation was implemented or tested on the JavaScript side, but it looks to me that this code only handles the case where a timer was created and canceled in the same episode (I'm assuming the `commandSet` is all the commands issued within a particular episode) - i.e. we created a timer, immediately canceled it, and then yielded. Does this work correctly if I schedule a timer, schedule an activity, and then cancel the timer *after the activity completes*?

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -165,4 +225,36 @@ public override void RaiseEvent(OrchestrationContext unused, string eventName, s
             this.context.RaiseEvent(eventName, serializedEventData);
         }
     }
+
+    internal class AsyncCommand
+    {
+        [JsonProperty("ActionType")]
```
cgillum: I worry the JavaScript purists might not be pleased with our Pascal casing here (I normally expect these kinds of JSON objects to be camel-cased). I'm guessing it would only be language worker authors that need to know this schema. Would it make sense to make it camel-case since that seems to be a more commonly accepted standard?

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -117,7 +124,60 @@ public void SetFunctionInvocationCallback(Func<Task> callback)
...
+                    var isDone = bool.Parse(jObj["isDone"].ToString());
+                    if (isDone)
+                    {
+                        this.Context.SetOutput(jObj["output"]);
+                    }
+                }
+                else
+                {
+                    this.context.SetOutput(returnValue);
+                }
```
cgillum: This is a big block of new logic specific to out-of-proc . Would it make sense to factor it into a separate method for improved readability?

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -165,4 +225,36 @@ public override void RaiseEvent(OrchestrationContext unused, string eventName, s
             this.context.RaiseEvent(eventName, serializedEventData);
         }
     }
+
+    internal class AsyncCommand
```
cgillum: Does this (and `AsyncActionType`) need to be an `internal` class? Could we instead make it a private inner class? The C# purist in me who studied the old books on .NET Framework guidelines wants to put internal classes into their own files. But if these classes are only used here, then it makes sense to just make them private inner classes.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
kashimiz: Caught!

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
kashimiz: It did get rather large. I'll move it into a method.

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
kashimiz: I tested this scenario. It does.

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -164,5 +198,91 @@ public override void RaiseEvent(OrchestrationContext unused, string eventName, s
...
+                // An actionSet represents all actions that were scheduled within that execution.
+                foreach (AsyncAction action in actionSet)
+                {
+                    switch (action.ActionType)
+                    {
+                        case AsyncActionType.CallActivity:
+                            tasks.Add(this.context.CallActivityAsync(action.FunctionName, action.Input));
+                            break;
+                        case AsyncActionType.CreateTimer:
+                            tasks.Add(this.context.CreateTimer(action.FireAt, CancellationToken.None));
```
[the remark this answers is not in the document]
kashimiz: I did. That's weird; I'll put that back in.

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -117,7 +137,22 @@ public void SetFunctionInvocationCallback(Func<Task> callback)
 
             if (returnValue != null)
             {
-                this.context.SetOutput(returnValue);
+                if (orchestratorInfo.IsOutOfProc)
+                {
+                    var jObj = returnValue as JObject;
+                    if (returnValue != null)
```
cgillum: This should be `if (jObj != null)`.

Issue #82, opened 2017-10-30
Use instance history caching to improve message processing throughput


cgillum:
Currently we are slow to process orchestrator function messages because we fetch the full history from Azure Storage every time we process a batch of messages for that instance. Some internal MRU caching should help speed things up, especially for highly parallel fan-out/fan-in scenarios.

SimonLuckenuik:
What about adding a snapshot concept to prevent rebuilding from the whole history? See this [article](https://msdn.microsoft.com/en-us/library/jj591577.aspx) for details, this is common in Event Sourcing scenarios.

cgillum:
Yes, something like this would be better. I'm having some discussions with the Durable Task Framework maintainers to see if we can build that into the core framework so that all providers (including the Azure Storage provider used by Durable Functions) can participate. Thanks for the link.

cgillum:
I added a new feature known as "extended sessions" to the Durable Task Framework to address this. It works by keeping the orchestration state in memory for a period of time to avoid replaying the orchestration when new messages continually arrive. The DurableTask.AzureStorage provider starting with version 1.2.0 will include these optimizations. We've merged in support for this in the latest Durable Functions code, and it will be available in the next release.

PR #289, opened 2018-04-30
Extended session support

cgillum:
This PR contains changes for extended sessions. It is intended primarily to address https://github.com/Azure/azure-functions-durable-extension/issues/82 by removing the need for replay except when an orchestration instance actually goes idle.

As part of enabling extended sessions, we now have two different runtime configurations, depending on whether extended sessions is enabled or disabled. For that reason, most of the end-to-end tests are now configured to run twice, once with each configuration.

Bundled with this is also some revamping of our automated tests to make them more robust.


src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -155,6 +154,36 @@ public class DurableTaskExtension :
         /// </summary>
         public string EventGridKeySettingName { get; set; }
 
+        /// <summary>
+        /// Gets or sets a flag indicating whether to enable extended sessions.
+        /// </summary>
+        /// <remarks>
+        /// <para>Extended sessions can improve the performance of orchestrator functions by allowing them to skip
+        /// replays when new messages and received within short periods of time.</para>
```
tohling: >and [](start = 38, length = 3)

are

Issue #282, opened 2018-04-28
Remove "version" information from traces


cgillum:
The DTFx concept of versions is not exposed in durable functions, so we should remove it from the traces we emit.

gled4er:
Hello @cgillum,

If you agree, I want to help with this issue.

Thank you!

cgillum:
Sure, that would be very helpful!

Also, do you think it would be possible to remove AppName and SlotName entries if they are null? For local development, these trace values aren't meaningful, so i was wondering if we could filter them out when during local development.

gled4er:
Hello Chris,

Sure.

Thank you very much for assigning the issue!

Sent from my Windows 10 phone

gled4er:
Hello @cgillum,

I have few questions for clarification about the changes required for this issue. 

Shall we remove the version also from the logs going to App Insights?

How about the app and slot names in terms of sending it to App Insights? Shall we remove them if they are null? 

The criteria to removing app and slot name will be that they are null, right? Or we need to detect when the app is run in local environment?

Thank you!

cgillum:
Version should be removed for everything since we don't use it at all.

Regarding AppName and SlotName, my expectation is that these values will always be null in local environments. In Azure environments, my expectation is that these values will never be null. I don't think any special detection is needed.

gled4er:
Hello Chris,

Thank you very much for the clarification!

Sent from my Windows 10 phone

gled4er:
Hello @cgillum,

Shall we remove references to version in the documentation?

Thank you!

PR #290, opened 2018-05-01
Remove from traces version and appname and slotname when null

gled4er:
Hello @cgillum,

May I ask you to review this pull request for [#282](https://github.com/Azure/azure-functions-durable-extension/issues/282)?

Thank you!

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
+++ b/src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
@@ -32,7 +32,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             string ExtensionVersion,
             bool IsReplay)
         {
-            this.WriteEvent(201, TaskHub, AppName, SlotName, FunctionName, InstanceId, Reason, FunctionType, ExtensionVersion, IsReplay);
+            if (!string.IsNullOrEmpty(AppName) && !string.IsNullOrEmpty(SlotName))
+            {
+                this.WriteEvent(201, TaskHub, AppName, SlotName, FunctionName, InstanceId, Reason, FunctionType, ExtensionVersion, IsReplay);
+            }
+            else
+            {
+                this.WriteEvent(201, TaskHub, FunctionName, InstanceId, Reason, FunctionType, ExtensionVersion, IsReplay);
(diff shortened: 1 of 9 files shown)


src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
```diff
@@ -26,14 +26,20 @@ private EtwEventSource()
...
         {
-            this.WriteEvent(201, TaskHub, AppName, SlotName, FunctionName, Version ?? "", InstanceId, Reason, FunctionType, ExtensionVersion, IsReplay);
+            if (!string.IsNullOrEmpty(AppName) && !string.IsNullOrEmpty(SlotName))
+            {
+                this.WriteEvent(201, TaskHub, AppName, SlotName, FunctionName, InstanceId, Reason, FunctionType, ExtensionVersion, IsReplay);
+            }
+            else
+            {
+                this.WriteEvent(201, TaskHub, FunctionName, InstanceId, Reason, FunctionType, ExtensionVersion, IsReplay);
+            }
```
cgillum: This change will actually break our ETW telemetry because the `WriteEvent` parameters must match the method signature exactly. Instead, let's leave it as it was, but simply remove the `Version` parameter. To be safe, we should also increment the version of this ETW event by updating the `[Event]` attribute as follows:

```csharp
[Event(201, Level = EventLevel.Informational, Version = 2)]
```
Please apply similar changes to the other methods in this class.

gled4er:
Hello @cgillum,

Thank you for the clarification. I think is it better that we are keeping **AppName** and **SlotName**. I wasn't very happy with the code for filtering them out. 

I will modify the methods as you requested. 

Thank you!

cgillum:
We just need to increment the existing version numbers. If they don't have one, then let's set it to `2`. If they are already at `2`, then let's increment it to `3`. No need to make them all consistent.

gled4er:
Hello @cgillum,

Great! 

Thank you for the clarification!

gled4er:
Hello @cgillum,

May I ask you to verify the latest changes related to reveritng the code for stripping **AppName** / **SlotName**  and increasing the version of `EtwEventSource` methods?

Thank you!

PR #291, opened 2018-05-01
JavaScript samples, 2 of 4.

kashimiz:
Includes HelloSequence and BackupSiteContent samples.

### DIFF
--- a/samples/javascript/HttpStart/run.csx
+++ b/samples/javascript/HttpStart/run.csx
@@ -1,0 +1,22 @@
+#r "Microsoft.Azure.WebJobs.Extensions.DurableTask"
+#r "Newtonsoft.Json"
+
+using System.Net;
+using System.Net.Http.Headers;
+
+public static async Task<HttpResponseMessage> Run(
+    HttpRequestMessage req,
+    DurableOrchestrationClient starter,
+    string functionName,
+    TraceWriter log)
+{
+    // Function input comes from the request content.
+    dynamic eventData = await req.Content.ReadAsAsync<object>();
+    string instanceId = await starter.StartNewAsync(functionName, eventData);
+    
+    log.Info($"Started orchestration with ID = '{instanceId}'.");
+    
+    var res = starter.CreateCheckStatusResponse(req, instanceId);
+    res.Headers.RetryAfter = new RetryConditionHeaderValue(TimeSpan.FromSeconds(10));
+    return res;
+}
--- a/samples/javascript/host.json
+++ b/samples/javascript/host.json
@@ -1,0 +1,13 @@
+{
+    "http": {
+      "routePrefix": ""
+    },
+    "applicationInsights": {
+      "sampling": {
+        "isEnabled": false
+      }
+    },
+    "durableTask": {
+      "HubName": "SampleHubJs"
+    }
+}
--- a/samples/javascript/local.settings.json
+++ b/samples/javascript/local.settings.json
@@ -1,0 +1,12 @@
+{
+  "IsEncrypted": false,
+  "Values": {
+    "AzureWebJobsStorage": "<Put your Azure Storage connection string here>",
+    "AzureWebJobsDashboard": "<Put your Azure Storage connection string here>",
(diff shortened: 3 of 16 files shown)


samples/javascript/HttpStart/run.csx
```diff
@@ -0,0 +1,22 @@
+#r "Microsoft.Azure.WebJobs.Extensions.DurableTask"
+#r "Newtonsoft.Json"
+
+using System.Net;
+using System.Net.Http.Headers;
+
+public static async Task<HttpResponseMessage> Run(
```
cgillum: Hmm...it's really unfortunate that we have to have a c# function for this. It's actually possible to start orchestrations using JavaScript, as we demonstrate here: https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-instance-management#starting-instances.

```javascript
module.exports = function (context, input) {
    var id = generateSomeUniqueId();
    context.bindings.starter = [{
        FunctionName: "HelloWorld",
        Input: input,
        InstanceId: id
    }];

    context.done(null);
};
```
The problem I anticipate is that we don't have a way for JavaScript to generate the HTTP response that contains all the useful management links. An unfortunate oversight :(. Can you open an issue to track support for this?

I still think we should put the JavaScript starter function in the checked-in samples. It can live side-by-side with the C# one. For demos, we can use the C# version.

samples/javascript/host.json
```diff
@@ -0,0 +1,13 @@
...
+    "http": {
+      "routePrefix": ""
+    },
+    "applicationInsights": {
+      "sampling": {
+        "isEnabled": false
+      }
+    },
+    "durableTask": {
+      "HubName": "SampleHubCsx"
```
cgillum: Let's name the hub something else since this name implies C# scripting (csx).

samples/javascript/local.settings.json
```diff
@@ -0,0 +1,12 @@
+{
+  "IsEncrypted": false,
+  "Values": {
+    "AzureWebJobsStorage": "UseDevelopmentStorage=true",
+    "AzureWebJobsDashboard": "UseDevelopmentStorage=true",
+    "AZURE_STORAGE_CONNECTION_STRING": "<Put your Azure Storage connection string here>",
```
cgillum: What is the `AZURE_STORAGE_CONNECTION_STRING` setting?

samples/javascript/package-lock.json
```diff
@@ -0,0 +1,679 @@
+{
```
cgillum: I don't know anything about JavaScript packages or npm, but are these lock files supposed to be checked-in?

samples/javascript/HttpStart/run.csx
kashimiz: I did create a JavaScript HttpStart function using that sample as a starting point. Unfortunately, the runtime failed at indexing because DurableOrchestrationClient isn't able to bind to string. We'd need to make that change in order for this to occur.

samples/javascript/host.json
kashimiz: Good catch, thanks.

samples/javascript/local.settings.json
kashimiz: In order to write to blob storage I'm using the [Azure Storage SDK for Node](https://github.com/Azure/azure-storage-node). AZURE_STORAGE_CONNECTION_STRING is used by that module.

samples/javascript/package-lock.json
kashimiz: Convention is that we do check in package-lock.json files. Here's a [Medium](https://medium.com/@Quigley_Ja/everything-you-wanted-to-know-about-package-lock-json-b81911aa8ab8) article on the subject.

samples/javascript/local.settings.json
cgillum: Cool. My suggestion then is that we use the same connection string placeholder text for all three storage app settings. Using a non-emulator connection string is probably better since that's what non-Windows developers are most likely to require.

Issue #285, opened 2018-04-29
Change custom Event Grid eventType property from an int to a string


cgillum:
The data/eventType field in the custom Event Grid notification is currently an int value, representing the orchestration runtime status. This is confusing because:

1. There is already an **eventType** field in the parent JSON object which has a different value.
2. It's not really an "event type" but rather the runtime status of the orchestration
3. Integers are hard to understand when used for status

Here is an example of the current paylod:
```json
[
  {
    "id": "aeafd489-ee98-4b2e-b77f-2aac2594b4b2",
    "topic": null,
    "subject": "durable/orchestrator/Running",
    "data": {
      "hubName": "DFHub",
      "functionName": "Counter",
      "instanceId": "13b602a9d2914d7c981016acbee0b941",
      "reason": "",
      "eventType": 0
    },
    "eventType": "orchestratorEvent",
    "eventTime": "2018-04-29T04:41:43.5299974Z",
    "metadataVersion": null,
    "dataVersion": "1.0"
  }
]
```

My proposal is to change **data/eventType** to **data/runtimeStatus** and make it a `string` instead of an `int`. Here is the same example, but translated:

```json
[
  {
    "id": "aeafd489-ee98-4b2e-b77f-2aac2594b4b2",
    "topic": null,
    "subject": "durable/orchestrator/Running",
    "data": {
      "hubName": "DFHub",
      "functionName": "Counter",
      "instanceId": "13b602a9d2914d7c981016acbee0b941",
      "reason": "",
      "runtimeStatus": "Running"
    },
    "eventType": "orchestratorEvent",
    "eventTime": "2018-04-29T04:41:43.5299974Z",
    "metadataVersion": null,
    "dataVersion": "1.0"
  }
]
```

This will be a breaking change for any consumers of this field. @TsuyoshiUshio, @yuka1984, please let me know if you have any concerns with this change. I can take care of implementing it for GA if you are agreeable to this change.

cgillum:
Another option is to remove this second `eventType` field since it is redundant with the `subject` field.

TsuyoshiUshio:
@cgillum  OK. I'll ask yuka and the guys. I'm ok for both solution, and I can help. Also, we need to care the documentation change. I also send pull request to the documentation as well.

TsuyoshiUshio:
[...]

PR #294, opened 2018-05-03
Change custom Event Grid eventType property from an int to a string

TsuyoshiUshio:
This pull request is for the https://github.com/Azure/azure-functions-durable-extension/issues/285 . 

I change EventGridPayload eventType -> runtimeStatus. 

Here is an example of the current paylod:

```
[
  {
    "id": "aeafd489-ee98-4b2e-b77f-2aac2594b4b2",
    "topic": null,
    "subject": "durable/orchestrator/Running",
    "data": {
      "hubName": "DFHub",
      "functionName": "Counter",
      "instanceId": "13b602a9d2914d7c981016acbee0b941",
      "reason": "",
      "eventType": 0
    },
    "eventType": "orchestratorEvent",
    "eventTime": "2018-04-29T04:41:43.5299974Z",
    "metadataVersion": null,
    "dataVersion": "1.0"
  }
]
```

This pull request change it like this. 

```
[
  {
    "id": "aeafd489-ee98-4b2e-b77f-2aac2594b4b2",
    "topic": null,
    "subject": "durable/orchestrator/Running",
    "data": {
      "hubName": "DFHub",
      "functionName": "Counter",
      "instanceId": "13b602a9d2914d7c981016acbee0b941",
      "reason": "",
      "runtimeStatus": "Running"
    },
    "eventType": "orchestratorEvent",
    "eventTime": "2018-04-29T04:41:43.5299974Z",
    "metadataVersion": null,
    "dataVersion": "1.0"
  }
]
```
@cgillum , However, I have local unit testing problem. Please Full CI for this pull request. 
I tested in my environment, the all tests related this change is all pass. However, some test fails for Timeout.  (Obviously, it is not related this change). 

Failed Tests are

* DurableTaskEndToEndTests
* HandledActivityException
* OrchestrationConcurrency
* TimerCancellation

Please have a look.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
@@ -210,7 +210,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                         FunctionName = functionName,
                         InstanceId = instanceId,
                         Reason = reason,
-                        EventType = orchestrationRuntimeStatus,
+                        RuntimeStatus = $"{orchestrationRuntimeStatus}",
                     },
                     DataVersion = "1.0",
                 },
@@ -233,8 +233,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             [JsonProperty(PropertyName = "reason")]
             public string Reason { get; set; }
 
-            [JsonProperty(PropertyName = "eventType")]
-            public OrchestrationRuntimeStatus EventType { get; set; }
+            [JsonProperty(PropertyName = "runtimeStatus")]
+            public string RuntimeStatus { get; set; }
         }
 
         private class EventGridEvent
--- a/src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
@@ -210,7 +210,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                         FunctionName = functionName,
                         InstanceId = instanceId,
                         Reason = reason,
-                        RuntimeStatus = $"{orchestrationRuntimeStatus}",
+                        RuntimeStatus = orchestrationRuntimeStatus.ToString(),
                     },
                     DataVersion = "1.0",
                 },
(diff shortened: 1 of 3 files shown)


src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
```diff
@@ -210,7 +210,7 @@ public void SetHttpMessageHandler(HttpMessageHandler handler)
                         FunctionName = functionName,
                         InstanceId = instanceId,
                         Reason = reason,
-                        EventType = orchestrationRuntimeStatus,
+                        RuntimeStatus = $"{orchestrationRuntimeStatus}",
```
cgillum: `orchestrationRuntimeStatus.ToString()` is better from a performance perspective.

TsuyoshiUshio: Thanks, I've done.

cgillum:
@TsuyoshiUshio it seems your PRs are very unlucky - I don't normally see as many flaky test failures. :) I'm re-running the CI to see if everything passes this time.

TsuyoshiUshio:
Sure. Or  you can change the Timeout of these tests. It  is just 10 sec.  I locally test it like this. as you said the performance should much better. 

```
            var s = new Stopwatch();
            var status01 = OrchestrationRuntimeStatus.Running;
            var status02 = OrchestrationRuntimeStatus.Completed;
            s.Start();
            for (int i = 0; i < 1000000; i++)
            {
                string c = $"{status01}";
                string d = $"{status02}";
            }
            Console.WriteLine("{xxx} : " + s.ElapsedMilliseconds);
            s.Stop();
            s.Restart();
            for (int i = 0; i < 1000000; i++)
            {
                string b = status01.ToString();
                string d = status02.ToString();
            }

            Console.WriteLine("ToString(): " + s.ElapsedMilliseconds);
            Console.ReadLine();
```
Result

```
{xxx} : 1190
ToString(): 385
```

TsuyoshiUshio:
Other idea for the build time and flakiness, Bing team did a very good job for that. We might talk with them. :) Their build took a couple of hours in local machine. Now they have a bunch of tests, however, they can build test everyhing within 10 - 15 min. They said that their solution becomes available. 
https://stories.visualstudio.com/bing-continuous-delivery/

TsuyoshiUshio:
Also, I'm gonna read the source related EndToEndTest related one.

cgillum:
One thing I noticed when trying to improve robustness a few days ago is that there can be a lot of messages leftover in the queues from previous CI runs. Because task hub names are reused, the test code will often try to execute these old messages, causing unplanned overhead. Deleting these messages often makes the tests more reliable. That's what I did before the last CI restart and now it looks green.

So I think one thing we need to do to improve this situation is find a way to reliably clean up resources after test runs. However, profiling may still be useful because even if there are old messages, I would expect the tests to complete successfully.

Issue #241, opened 2018-04-11
Error handling for Azure Event Grid publishing


cgillum:
This is a followup for https://github.com/Azure/azure-functions-durable-extension/issues/216. Current Event Grid publishing does not have a defined error handling behavior. Protocol errors will be ignored, but connection failures will cause the orchestration instance to fail.

There are ramifications for both failure and non-failure policies, so we should probably allow the user to decide which option is better. For transient errors (connectivity failures, HTTP 503, etc.) we should implement automatic retry for improved reliability. More details on the REST API can be found here: https://docs.microsoft.com/en-us/azure/event-grid/post-to-custom-topic

TsuyoshiUshio:
Do you mean 

* Make customer can add some attribute on `host.json`  for failure/non-failure policies
* Make customer can add some attribute for retry policy
* Implement these

I'm not a good programmer and also, I can't work this in a couple of days and no one want to do it, I'd like to do it!  I only care about your review work load for me like last time. I felt sorry about that.  If these are OK, I'd love to do it.

TsuyoshiUshio:
Other thing is, how about the document contribution as you mentioned last time? Maybe we need to update this page. I can add some example and explanation on this page. e.g. host.json configration.  https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-versioning  Is this what you did intend to?

cgillum:
For this item, I was thinking this:

* Customer adds an attribute on host.json for failure/non-failure policies
* We add automatic retry for HTTP 503, timeouts, and connectivity errors. We can skip configuration for this for now. I expect there are some best practices available that we can follow.

I would love to have your contribution if you are willing. :) There is no rush on getting this done quickly since it will not be commonly needed (hopefully Azure Event Grid is always highly available). However, I do think we will want this in order for it to truly be "production ready" and give customers greater confidence in the feature.

Here is an interesting post on best practices for implementing automatic retry with HttpClient: https://stackoverflow.com/questions/19260060/retrying-httpclient-unsuccessful-requests. This answer in particular looks like a good, clean approach with minimal dependencies: https://stackoverflow.com/a/35183487/2069 (uses [Polly](https://github.com/App-vNext/Polly)).

TsuyoshiUshio:
Thank you @cgillum ! I'd happy to do it! :) To contribute this repo is my pleasure and fun. Also working with you is totally awesome experience. I'll do it carefully.

[...]

PR #307, opened 2018-05-12
Feature event grid error handling

yuka1984:
Hi @cgillum .
I have created Event grid error handling.
Please code review.


src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
```diff
@@ -272,5 +314,76 @@ private class EventGridEvent
...
+                    return response;
+                }
+                else
+                {
+                    throw lastException;
+                }
+            }
+        }
+
+        public class LifeCyclePublishingException : Exception
```
cgillum: Not used?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -169,6 +169,22 @@ public class DurableTaskExtension :
         /// </summary>
         public string EventGridKeySettingName { get; set; }
 
+        /// <summary>
+        /// Gets or sets the Event Grid publish request retry count.
+        /// </summary>
+        public int EventGridPublishRetryCount { get; set; }
```
cgillum: Do you have any tests which try to assign these new properties?

src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
```diff
@@ -205,6 +205,26 @@ private EtwEventSource()
             this.WriteEvent(211, TaskHub, AppName, SlotName, FunctionName, FunctionState, Version ?? "", InstanceId, Details, StatusCode, Reason, FunctionType, ExtensionVersion, IsReplay, LatencyMs);
         }
 
+        [Event(212, Level = EventLevel.Error)]
+        public void EventGridNotificationException(
```
cgillum: This seems very similar to `EventGridNotificationFailed` (Event ID 211). Do we really need a new event or can we use the existing event?

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
```diff
@@ -272,5 +314,76 @@ private class EventGridEvent
...
             public string DataVersion { get; set; }
         }
+
+        internal class HttpRetryMessageHandler : DelegatingHandler
+        {
+            private readonly int maxRetryCount;
+            private readonly Func<int, TimeSpan> retryWaitSpanFunc;
+            private int[] retryTargetStatus;
+
+            public HttpRetryMessageHandler(HttpMessageHandler messageHandler, int maxRetryCount, Func<int, TimeSpan> retryWaitSpanFunc, int[] retryTargetStatusCode)
```
cgillum: It looks like all uses of `Func<int, TimeSpan> retryWaitSpanFunc` specify a constant timeout. The public properties require a static number, so I think we should make this a simple `TimeSpan` parameter instead of a `Func`.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
yuka1984: It is difficult to do this test. Is there any good idea?

cgillum: Right now there is a `SetHttpMessageHandler` which allows you to set a preconfigured message handler. Maybe you could change this into a get/set property? Then you could potentially write a simple unit test which sets `EventGridPublishRetryCount`, etc. and checks to see if the `HttpMessageHandler` was correctly configured. Would this work? I apologize if my suggestion is unclear.

yuka1984: Thank you. I have added a new test for your review.

src/WebJobs.Extensions.DurableTask/LifeCycleNotificationHelper.cs
```diff
@@ -70,10 +66,15 @@ public LifeCycleNotificationHelper(DurableTaskExtension config, ExtensionConfigC
...
-        public void SetHttpMessageHandler(HttpMessageHandler handler)
+        public HttpMessageHandler HttpMessageHandler
         {
-            httpClient = new HttpClient(handler);
-            httpClient.DefaultRequestHeaders.Add("aeg-sas-key", this.eventGridKeyValue);
+            get => httpMessageHandler;
+            set
+            {
+                httpMessageHandler = value;
+                httpClient = new HttpClient(httpMessageHandler);
```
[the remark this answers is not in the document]
yuka1984: Yes. It is better to call the Dispose method.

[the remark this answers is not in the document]
cgillum: Ah, that's a nice use of the `?.` operator...

cgillum:
Linking back to https://github.com/Azure/azure-functions-durable-extension/issues/241

PR #309, opened 2018-05-15
Enable DurableOrchestrationClient output binding to string.

kashimiz:


src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
```diff
@@ -29,6 +30,11 @@ public StartOrchestrationArgs JObjectToStartOrchestrationArgs(JObject input, Orc
             return input?.ToObject<StartOrchestrationArgs>();
         }
 
+        public StartOrchestrationArgs StringToStartOrchestrationArgs(string input, OrchestrationClientAttribute attr)
+        {
+            return JsonConvert.DeserializeObject<StartOrchestrationArgs>(input);
```
cgillum: Should we consider handling null input?

kashimiz: We should. Making that change and merging.

PR #311, opened 2018-05-16
Fix: MessagePayloadDataConverter can't deserialize to string

kashimiz:
Due to Json.NET's JsonSerializer class throwing an exception when presented with stringified JSON and a target type of string.

Knock-on effect of bindings for out of proc functions being strings, not JObjects. (Did we ever figure that out?) This was keeping JavaScript activity functions from successfully binding to JSON data.

### DIFF
--- a/samples/javascript/E3_GetIsClear/function.json
+++ b/samples/javascript/E3_GetIsClear/function.json
@@ -1,0 +1,10 @@
+{
+  "bindings": [
+    {
+      "name": "location",
+      "type": "activityTrigger",
+      "direction": "in"
+    }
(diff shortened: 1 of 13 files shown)


cgillum:
I don't have the full context on this change, so I'm a little hesitant. My fear is that it could negatively impact regular C# scenarios (maybe that's related to the CI failures) since this change doesn't distinguish in-proc (which works with string data just fine) from out-of-proc.

kashimiz:
Nope, you're right. This was too general an approach (as further testing confirmed.) Going for something more specific.

PR #315, opened 2018-05-18
Handle object inputs for out-of-proc activity contexts.

kashimiz:
A less-naïve, more specific implementation to prevent the runtime from throwing an exception on binding an activity trigger for an out-of-proc function whose input is some sort of object.

Ideally, there'd be a simpler way to do this - say, an existing IsValidJson method or a TryParse, but neither exists for Json.NET in 2018. This is a best-effort.


src/WebJobs.Extensions.DurableTask/DurableActivityContext.cs
```diff
@@ -102,6 +103,31 @@ internal object GetInput(Type destinationType)
...
+                {
+                    var deserializedValue = MessagePayloadDataConverter.Default.Deserialize(serializedValue, typeof(object));
+                    if (deserializedValue != null)
+                    {
+                        return serializedValue;
+                    }
+                }
+                catch (JsonReaderException ex)
+                {
+                    throw ex;
```
cgillum: There is no need for a `catch` block which simply rethrows the exception. In fact, there is a problem with this as-written because the original call stack information will be lost when you rethrow the exception this way (normally one is advised to use `throw` without any arguments to rethrow an exception). But anyways, I would remove the try/catch entirely.

src/WebJobs.Extensions.DurableTask/DurableActivityContext.cs
```diff
@@ -102,6 +103,31 @@ internal object GetInput(Type destinationType)
...
+            // workaround for that case.
+            if (destinationType.Equals(typeof(string)))
+            {
+                // Json.NET has no TryParse or IsValidJson method, so we'll optimistically parse
+                // and catch any exceptions (which shouldn't occur: all (in-proc; JSON and not JSON)
+                // or all other (out-of-proc: not-JSON) inputs with destination System.String should
+                // cast to valid JValues and be handled above.)
+                try
+                {
+                    var deserializedValue = MessagePayloadDataConverter.Default.Deserialize(serializedValue, typeof(object));
```
cgillum: Interesting. Technically, a `string` is also an `object`, so does this deserialization work as intended for all cases?

src/WebJobs.Extensions.DurableTask/DurableActivityContext.cs
```diff
@@ -102,6 +103,31 @@ internal object GetInput(Type destinationType)
...
+                    if (deserializedValue != null)
```
cgillum: What does it mean if `deserializedValue` is null?

src/WebJobs.Extensions.DurableTask/DurableActivityContext.cs
```diff
@@ -102,6 +103,31 @@ internal object GetInput(Type destinationType)
...
+            // with a destination type of System.String. Unfortunately, deserializing a JSON string
+            // to a string causes MessagePayloadDataConverter to throw an exception. This is a
+            // workaround for that case.
+            if (destinationType.Equals(typeof(string)))
+            {
+                // Json.NET has no TryParse or IsValidJson method, so we'll optimistically parse
+                // and catch any exceptions (which shouldn't occur: all (in-proc; JSON and not JSON)
+                // or all other (out-of-proc: not-JSON) inputs with destination System.String should
+                // cast to valid JValues and be handled above.)
+                try
```
cgillum: I'm curious - is there a simple and fast test we can apply to the string, such as checking to see if the first character is a `{` or a `[`? I'm not sure if this would be useful or not since I don't yet fully understand how the current iteration works though...

src/WebJobs.Extensions.DurableTask/DurableActivityContext.cs
kashimiz: Thanks for catching this; that's a doc read error on my end. JsonSerializer will throw an exception rather than return null if it can't parse a string, but I read it the other way around. Removing as it makes no sense with that correction.

src/WebJobs.Extensions.DurableTask/DurableActivityContext.cs
kashimiz: It does - a valid JSON string will parse to a JObject or JArray when the Deserialize() call's objectType is object, but will cause an exception to be thrown when the objectType is string.

src/WebJobs.Extensions.DurableTask/DurableActivityContext.cs
kashimiz: That might be better - or not necessary at all. My concern is how understandable our thought process would be, but I suppose that's what comments are for.

Currently, by my testing, here's how the possible cases behave:

In/Out of Proc | JSON?<sub>*</sub> | Destination Type 
--------------- | ----------------------- | -------------------
In-Proc | Yes | System.String 
In-Proc | No | System.String
Out-of-Proc | No | System.String

cgillum: I don't quite understand the meaning of the **JSON?** column. `serializedInput` is always JSON.  Do you actually mean **JObject** (as in a JSON object and not a JSON primitive like int or string)?

Following up on our in-person discussion, I'm starting to think that your proposal to do the following might just work:
```csharp
if (destinationType == typeof(string))
{
    return serializedValue;
}
```

The part I mentioned about string encoding actually doesn't apply because the actual JSON value we're trying to deserialize is not actually a string (otherwise we would have returned at ln 102 like you mentioned), but a JObject.

kashimiz: JObject, yes.

src/WebJobs.Extensions.DurableTask/DurableActivityContext.cs
```diff
@@ -102,6 +103,17 @@ internal object GetInput(Type destinationType)
             }
 
             string serializedValue = jToken.ToString(Formatting.None);
+
+            // Object inputs for out-of-proc activities are passed in their JSON-stringified form with a destination
+            // type of System.String. Unfortunately, deserializing a JSON string to a string causes
+            // MessagePayloadDataConverter to throw an exception. This is a workaround for that case. All other
+            // inputs with destination System.String (in-proc: JSON and not JSON; out-of-proc: not-JSON) inputs with
+            // destination System.String should cast to JValues and be handled above.)
```
cgillum: I think you have an unbalanced `)` at the end of this comment.

src/WebJobs.Extensions.DurableTask/DurableActivityContext.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.IO;
```
cgillum: Is this still needed?

Issue #170, opened 2018-02-27
Log only on non-replay executions


kashimiz:
Currently, log statements are output on every execution, non-replay and replay alike. To log only on non-replay executions, code must look something like:

`if (!ctx.IsReplaying) log.Info("Calling F1.");`

This is tedious to write, clutters up code, and is easily missed or forgotten. it would make for a better developer experience and cleaner code to simplify logging for our users and make logging only on non-replay executions the default behavior.

gled4er:
Hello @cgillum,

Shall this change apply to the logs we send to Application Insights as well? 

Thank you!

cgillum:
That's a great question. I had not thought about it deeply yet, but it probably makes a lot of sense since that is potentially a huge amount of data going to Application Insights that will cost the customer money.

Here is my thinking for the implementation design of this work item:

* I don't think we can implement this for `TraceWriter` because I don't think it is extensible. `TraceWriter` is being deprecated though so maybe that's not an issue.
* Let's create a Boolean flag that controls whether to log replays or not. By default, let's not log replays. Developers can turn replay logging on if they really want it for deeper debugging purposes.
* Let's find a way to wrap the `ILogger` we use with one that can filter out replay logs. I'm not sure yet exactly how to do this, but I feel reasonably confident that we can control this somehow using the host extensibility.

We should continue to write ETW analytics for replays regardless of any user settings - that's because we want that data for our own internal diagnostics. If we focus our changes on `ILogger` then this should not be an issue.

gled4er:
Hello @cgillum,

I just saw your comment only after submitting the initial pull request. 

Thank you very much for the insights! 

I will improve my pull request based on your suggestions.

Thank you!

PR #319, opened 2018-05-24
Skip logs on replay

gled4er:
Hello @cgillum,

I wanted to share my idea for solving issue [#170](https://github.com/Azure/azure-functions-durable-extension/issues/170).

I added very basic method but the idea is to allow the developers to be able to call freely any `ILogger` methods.

In the same time the developer experience doesn't look very smooth so I am looking forward your recommendations:

```csharp
            [OrchestrationTrigger] DurableOrchestrationContextBase context,
            ILogger logger)
        {

            context.LogMessage(new Action(() => logger.LogInformation("Calling E1_SayHello with Tokyo")));
            context.LogMessage(new Action(() => logger.LogInformation("Calling E1_SayHello with Seattle")));
            outputs.Add(await context.CallActivityAsync<string>("E1_SayHello", "London"));
            context.LogMessage(new Action(() => logger.LogInformation("Calling E1_SayHello with London")));

            return outputs;
        }
```

And as output we are getting only once the lines:
- Calling E1_SayHello with Tokyo
- Calling E1_SayHello with Seattle
- Calling E1_SayHello with London

Thank you!


cgillum:
It's an interesting idea, and I've considered it myself. However, I have two pieces of feedback:

* This solution is not specific to logging. I would actually prefer a method name like `RunOnce(...)`.
* This is useful for simplifying the code, but I was hoping to create a solution that didn't require users to learn a new API. What I would love to see is a solution that allows developers to use the `ILogger` interface have it "just work" without needing to do anything else.

Is there a solution we should be considering that modifies the `ILogger` instead that gets passed as the orchestrator function parameter?

gled4er:
Hello @cgillum,

I created this PR before I read your design guidance so I will be researching more how to implement it in a native way.

If you prefer I will close this PR for now and open it when I am ready with better solution.


Ideally I want to use approach leveraging [LogFilterRules](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loggerfilterrule.filter?view=aspnetcore-2.0#Microsoft_Extensions_Logging_LoggerFilterRule_Filter)

Thank you!

gled4er:
Hello @cgillum,

I have another suggestion about avoiding the duplication of the logs. It is not so smooth as you want but it is better than creating a delegate. Do you think we can create `ILogger` extension methods that have the same signature as the original ones but accept `DurableOrchestrationContext` as a first parameter:

```csharp
    /// <summary>
    /// ILogger extension methods.
    /// </summary>
    public static class ILoggerExtensions
    {
        /// <summary>
        /// Log Information only once when the IsReplaying is set to false.
        /// </summary>
        /// <param name="logger">Instance of <see cref="ILogger"/></param>
        /// <param name="contextBase">Instance of <see cref="DurableOrchestrationContextBase"/></param>
        /// <param name="message">Information message</param>
        public static void LogInformationOnce(this ILogger logger, DurableOrchestrationContextBase contextBase, string message)
        {
            if (!contextBase.IsReplaying)
            {
                logger.LogInformation(message);
            }
        }
    }
```
And then the client code will be:

```csharp
            [OrchestrationTrigger] DurableOrchestrationContextBase context)
        {

            logger.LogInformationOnce(context, "E1_SayHello called with Tokyo");
            logger.LogInformationOnce(context, "E1_SayHello called with Seattle");
            outputs.Add(await context.CallActivityAsync<string>("E1_SayHello", "London"));
            logger.LogInformationOnce(context, "E1_SayHello called with London");

            return outputs;
        }
```
The downside of this approach is that we are introducing new API which you want to avoid but on the other hand the developers have the a lot more consistent experience seeing LogOnce methods next to the original ILogger methods.

Let me know what you think?

Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -299,6 +299,15 @@ public override void ContinueAsNew(object input)
             this.ContinuedAsNew = true;
         }
 
+        /// <inheritdoc />
+        public override void RunOnce(Delegate d)
```
cgillum: While very useful for some scenarios, I'm a little worried that developers could abuse this method and accidentally create non-deterministic orchestrations. For example, I imagine someone could do something like the following:

```csharp
Guid id = Guid.Empty;
ctx.RunOnce(() => id = Guid.NewGuid());
// ...do something with id
```

In this scenario, the value of `id` will be different between the first execution and the replay. I think there are many other scenarios like this which could happen.

For this reason, I suggest we not implement `RunOnce` unless we have a really strong reason to do so. I just want to be really careful and thoughtful about adding features which could be useful in some cases but could cause hard-to-diagnose problems for others.

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -329,10 +369,19 @@ public static string LocalSlotName
...
+            }
+        }
 
-            this.logger.LogInformation(
-                "{instanceId}: Function '{functionName} ({functionType})' was resumed by a timer scheduled for '{expirationTime}'. IsReplay: {isReplay}. State: {state}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
-                instanceId, functionName, functionType, expirationTimeString, isReplay, FunctionState.TimerExpired, hubName, LocalAppName, LocalSlotName, ExtensionVersion, this.sequenceNumber++);
+        private bool ShouldReplayEventsBeLogged(bool isReplay)
+        {
+            return this.durableTaskExtension.LogReplayEvents ||
+                   (!this.durableTaskExtension.LogReplayEvents && !isReplay);
```
cgillum: This condition seems a bit redundant. I think it's more concise to say `this.durableTaskExtension.LogReplayEvents || !isReplay`.

test/DurableTaskEndToEndTests.cs
```diff
@@ -232,7 +244,7 @@ public DurableTaskEndToEndTests(ITestOutputHelper output)
                     TestHelpers.AssertLogMessageSequence(
                         this.output,
                         this.loggerProvider,
-                        "HelloWorldOrchestration_Activity",
+                        logReplayEvents ? "HelloWorldOrchestration_Activity" : "HelloWorldOrchestration_Activity_Skip_Replay_Events",
```
cgillum: The `extendedSessions` boolean parameter already filters out replay events. Is there some common logic we can use so that we don't need to create multiple sets of expected log outputs for the same basic tests?

src/WebJobs.Extensions.DurableTask/ILoggerExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// </summary>
+    public static class ILoggerExtensions
+    {
+        /// <summary>
+        /// Log Information only once when the IsReplaying is set to false.
+        /// </summary>
+        /// <param name="logger">Instance of <see cref="ILogger"/></param>
+        /// <param name="contextBase">Instance of <see cref="DurableOrchestrationContextBase"/></param>
+        /// <param name="message">Information message</param>
+        public static void LogInformationOnce(this ILogger logger, DurableOrchestrationContextBase contextBase, string message)
```
cgillum: I'm not comfortable adding new public APIs if we think they might be temporary. I also share the other concerns you mentioned. Let's discuss a little bit more to see if we can find a way to change the behavior of the existing `ILogger`.

gled4er:
Hello @cgillum,

Thank you very much for the great feedback.

I will remove the code related to `RunOnce` and the extension methods.

Thank you for starting the email thread for `ILogger`!

Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
gled4er: Sure. I completely agree. I will remove it from the PR. Thank you!

src/WebJobs.Extensions.DurableTask/ILoggerExtensions.cs
gled4er: Thank you! I will remove it.

test/DurableTaskEndToEndTests.cs
gled4er: Great point. I will work on it. Thank you!

gled4er:
Hello @cgillum,

May I ask you to check my latest changes based on your feedback in the previous review iteration?

I generalized and reused the existing test code and I am interested in your feedback.

Thank you!

gled4er:
Hello @cgillum,

May I ask you to enable the full CI for this pull request?

Thank you!

cgillum:
I've enabled the full CI. However, it looks like there is now a merge conflict - possibly with Yuka's recent change, so that will block the CI from running. Please resolve at your convenience.

gled4er:
Hello @cgillum,

I work on fixing the conflicts.

Thank you!

gled4er:
Hello @cgillum,

I merged the latest changes from dev and the tests passed.

May I ask you to have one more check and let me know what should be fixed next?

Thank you!

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -363,10 +403,18 @@ public static string LocalSlotName
                 functionType.ToString(),
                 ExtensionVersion,
                 IsReplay: isReplay);
+            if (this.ShouldReplayEventsBeLogged(false))
```
cgillum: This is a little misleading to read. `ShouldReplayEventsBeLogged(false)` intuitively means "don't log replay events", but that's not what it means at all. I suggest that you call it as `ShouldReplayEventsBeLogged(isReplay: false)` - basically using named parameters - to make the meaning more clear.

test/TestHelpers.cs
```diff
@@ -81,7 +83,8 @@ internal static class TestHelpers
             string instanceId,
             bool extendedSessions,
             string[] orchestratorFunctionNames,
-            string activityFunctionName = null)
+            string activityFunctionName = null,
+            bool logReplayEvents = true)
```
cgillum: I think I mentioned this in another iteration, but I don't think it's necessary to create this new `logReplayEvents` parameter. The existing `extendedSessions` parameter basically doesn't the same thing already. Can we just rename that parameter to something like `filterOutReplayLogs`?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -230,6 +230,17 @@ public class DurableTaskExtension :
         /// </value>
         public int ExtendedSessionIdleTimeoutInSeconds { get; set; } = 30;
 
+        /// <summary>
+        /// Gets or sets if logs for replay events need to be recorded.
+        /// </summary>
+        /// <remarks>
+        /// The default value is false that prevents logging the replay events.
```
cgillum: "The default value is false , which disables the logging of replay events."

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -16,12 +16,14 @@ internal class EndToEndTraceHelper
         private static string slotName;
 
         private readonly ILogger logger;
+        private readonly DurableTaskExtension durableTaskExtension;
 
         private long sequenceNumber;
 
-        public EndToEndTraceHelper(JobHostConfiguration config, ILogger logger)
+        public EndToEndTraceHelper(JobHostConfiguration config, ILogger logger, DurableTaskExtension durableTaskExtension)
```
cgillum: Rather than passing a reference to the extension, can we simply pass the value of `LogReplayEvents`? I feel like that would make it easier to add unit testing for this class.

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -363,10 +403,18 @@ public static string LocalSlotName
...
+                    "{instanceId}: Function '{functionName} ({functionType})' was resumed by a timer scheduled for '{expirationTime}'. IsReplay: {isReplay}. State: {state}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
+                    instanceId, functionName, functionType, expirationTimeString, isReplay, FunctionState.TimerExpired,
+                    hubName, LocalAppName, LocalSlotName, ExtensionVersion, this.sequenceNumber++);
+            }
+        }
 
-            this.logger.LogInformation(
-                "{instanceId}: Function '{functionName} ({functionType})' was resumed by a timer scheduled for '{expirationTime}'. IsReplay: {isReplay}. State: {state}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
-                instanceId, functionName, functionType, expirationTimeString, isReplay, FunctionState.TimerExpired, hubName, LocalAppName, LocalSlotName, ExtensionVersion, this.sequenceNumber++);
+        private bool ShouldReplayEventsBeLogged(bool isReplay)
```
cgillum: The method name doesn't quite represent what this method is used for. I think you should simply name it `ShouldLogEvent`.

gled4er:
Hello @cgillum,

Thank you very much for the great feedback.

Thank you!

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
gled4er: This a great improvement. Thank you!

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
gled4er: Thank you! I will fix it.

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
gled4er: Great! I will change it. Thank you!

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
gled4er: Great! Thank you very much! Shorter, simpler, better!

test/TestHelpers.cs
gled4er: Hello @cgillum,

Maybe I am missing something but my current understanding is that we have in general 3 options:
- `extendedSessions` is `true`, we don't log the replay events
- `extendedSessions` is `false`, here we will have the replay events available for logging but then based on the value of `logReplayEvents` we can change the behavior -> `logReplayEvents`  is `true`, we will log the replay events and `logReplayEvents` is `false` and then we will skip logging the replay even though `extendedSessions` is `false`.

So if this is true, I think we need to keep both parameters to cover all 3 cases.

Please let me what you think.

I will double check the logic.

Thank you!

gled4er:
Hello @cgillum,

Thank you for approving the pull request.

May I ask you to check my latest comment for skipping the `logReplayEvents` parameter and renaming the `extendedSessions` to `filterOutReplayLogs`?

My current understanding it that we should keep `logReplayEvents`.

Thank you!

test/TestHelpers.cs
cgillum: The only thing this parameter controls is whether to filter out the replay logs. The actual logging behavior is configured separately in another method. I was thinking we rename `extendedSessions` *in this context only* to `filterOutReplayLogs`. The value we use is essentially `filterOutReplayLogs = extendedSessions || !logReplayEvents`. Would that work?

gled4er: Great! Thank you very much for the clarification!

gled4er:
Hello @cgillum,

I added the changes related to renaming and simplifying the use of `extendedSessions`.

Thank you for your support on this PR!

gled4er:
Hello @cgillum,

Thank you very much for the support!

chinarosesz:
I'm also interested in disabling logs going to Application Insights when replay is set to true. From a user control, I understand how to do that with ILogger, but logs that I can't control, I would like to turn it off. I'm trying to figure out which flag should I turn on do accomplish this?

chinarosesz:
I think I found it. Sorry, still learning and navigating myself around Durable Functions. My first try was looking for the property `logReplayEvents` context from https://github.com/MicrosoftDocs/azure-docs/blob/master/includes/functions-host-json-durabletask.md

chinarosesz:
I think I found it. `logReplayEvents`  can be set in host.json https://github.com/MicrosoftDocs/azure-docs/blob/master/includes/functions-host-json-durabletask.md

Issue #155, opened 2018-02-15
Simple API to get status/terminate/raise-event URLs


cgillum:
### Problem
HTTP triggers have a way to return an HTTP response with the get-status, terminate, and raise-event webhook URLs embedded in a JSON payload. However, there is no simple way to get this information for non-HTTP triggered functions. Even when using HTTP, there is no simple way to share these URLs with other systems.

### Suggestion
Create a `GetWebHookUrls` API which takes the instance ID as a parameter and returns a data structure containing the webhook URLs. This API could exist on `DurableOrchestrationClient`, `DurableOrchestrationContext`, and `DurableActivityContext` classes.

For non-.NET functions, we can consider using binding parameters to flow this information.

cgillum:
However, thinking about this more, it's often simple enough to create new HTTP-trigger functions and call the corresponding `DurableOrchestrationClient` methods. We should weigh the benefit of this feature over the ability to implement this manually.

gled4er:
Hello @cgillum,

I think this feature will be useful in cases we need to pass send event, status or terminate url-s from the orchestrator external systems. 

If you remember we had similar scenario when we needed to pass the send custom event url to a logic app that did few steps and then post back to the orchestrator. The issue in this case was that we couldn't retrieve the key from the orchestration context and we passed the url from the orchestration client back to the orchestrator. 

I am interested in your guidance on this feature.

Thank you!

cgillum:
Great! Here is my thinking for this: Let's update [HttpApiHandler.GetClientResponseLinks](https://github.com/Azure/azure-functions-durable-extension/blob/aa4180cf457c2bf7e5cbdc3e971164e539d906a8/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs#L335) to accept a null `HttpRequestMessage` value. If it is null, then we can instead use the value in the `notificationUri` variable. This `Uri` should be used to replace our usage of `request.RequestUri` in this method.  Something like this:

```csharp
Uri notificationUri = this.config.NotificationUrl;
Uri baseUri = request?.RequestUri ?? notificationUri;

// e.g. http://{host}/admin/extensions/DurableTaskExtension?code={systemKey}
string hostUrl = baseUri.GetLeftPart(UriPartial.Authority);

// ...rest of the method...
```

[...]

PR #321, opened 2018-05-25
Adding simple API to get status/terminate/raise-event URLs 

gled4er:
Hello @cgillum ,

May I ask you to review the direction I took for implementing issue [#155](https://github.com/Azure/azure-functions-durable-extension/issues/155#issuecomment-391624661)?

I tried to reduce the duplication as much as possible.

But let me know if you prefer to have dedicated method for the new API.

I also did not add any new code to the `DurableActivityContext` because we don't have dependency on `DurableTaskExtension` where currently the main method I use is implemented. Also I think that it is important to have the method available in `DurableOrchestrationContext` and if we need it in `DurableActivityContext` we can easily pass it.  

Please let me know if this is acceptable approach or it is recommended to add logic to `DurableActivityContext`  as well.

Thank you!


test/HttpApiHandlerTests.cs
```diff
@@ -73,6 +73,78 @@ private void CreateCheckStatusResponse_Throws_Exception_When_NotificationUrl_Mis
...
 
+        [Fact]
+        public void CreateCheckStatus_Returns_Corrent_CheckStatus_based_on_default_values()
+        {
+            var httpApiHandler = new HttpApiHandler(new DurableTaskExtension() { NotificationUrl = new Uri(TestConstants.NotificationUrl) }, null);
+            CheckStatus checkStatus = httpApiHandler.CreateCheckStatus(TestConstants.InstanceId, null, null);
+            Assert.NotNull(checkStatus);
+            Assert.Equal(checkStatus.Id, TestConstants.InstanceId);
+            Assert.Equal(
+                "http://localhost:7071/admin/extensions/DurableTaskExtension/instances/7b59154ae666471993659902ed0ba742?taskHub=DurableFunctionsHub&connection=Storage&code=mykey",
```
cgillum: Is the `http://localhost:7071/admin/extensions/DurableTaskExtension` based on `TestConstants.NotificationUrl`? If so, this might be a good place to reduce string duplication. I mention this mainly because Functions v2 will have a different URL prefix, and it might be nice to be able to update the constant value in one place and have these tests continue to work.

test/HttpApiHandlerTests.cs
```diff
@@ -73,6 +73,78 @@ private void CreateCheckStatusResponse_Throws_Exception_When_NotificationUrl_Mis
...
+                "http://localhost:7071/admin/extensions/DurableTaskExtension/instances/7b59154ae666471993659902ed0ba742?taskHub=SampleHubVS&connection=TestConnection&code=mykey",
+                checkStatus.StatusQueryGetUri);
+            Assert.Equal(
+                "http://localhost:7071/admin/extensions/DurableTaskExtension/instances/7b59154ae666471993659902ed0ba742/raiseEvent/{eventName}?taskHub=SampleHubVS&connection=TestConnection&code=mykey",
+                checkStatus.SendEventPostUri);
+            Assert.Equal(
+                "http://localhost:7071/admin/extensions/DurableTaskExtension/instances/7b59154ae666471993659902ed0ba742/terminate?reason={text}&taskHub=SampleHubVS&connection=TestConnection&code=mykey",
+                checkStatus.TerminatePostUri);
+        }
+
```
cgillum: It's great to have these unit tests to help ensure we don't accidentally break the URL formats - thanks for adding them!

gled4er:
Hello @cgillum,

Thank you very much for the comments.

I will work on fixing the issues next.

Thank you!

src/WebJobs.Extensions.DurableTask/CheckStatus.cs
```diff
@@ -0,0 +1,37 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    /// <summary>
+    /// Data structure containing status, terminate and send external event HTTP endpoints
+    /// </summary>
+    public class CheckStatus
```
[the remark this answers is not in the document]
gled4er: Great! I will change it! Thank you!

test/HttpApiHandlerTests.cs
gled4er: Sure. Tests are super useful to validate all expected behaviors.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContextBase.cs
```diff
@@ -390,10 +391,17 @@ public virtual Task CreateTimer(DateTime fireAt, CancellationToken cancelToken)
...
         /// The <paramref name="customStatusObject"/> value is serialized to JSON and will
-        /// be made available to the orchestration status query APIs. The serialized JSON 
+        /// be made available to the orchestration status query APIs. The serialized JSON
         /// value must not exceed 16 KB of UTF-16 encoded text.
         /// </remarks>
         /// <param name="customStatusObject">The JSON-serializeable value to use as the orchestrator function's custom status.</param>
         public abstract void SetCustomStatus(object customStatusObject);
+
+        /// <summary>
+        /// Creates a <see cref="CheckStatus"/> object that is useful for checking the status of the specified instance.
```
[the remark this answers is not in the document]
gled4er: Thank you! I will update them.

test/HttpApiHandlerTests.cs
gled4er: Thank you! I will change it.

gled4er:
Hello @cgillum,

May I ask you to check the current state of the pull request?

I worked on all issues you raised.

Please let me know what should be the next items for fixing.

Thank you!

test/TestHelpers.cs
```diff
@@ -236,6 +244,61 @@ private static List<string> GetLogs_HelloWorldOrchestration_Inline(string instan
             return list;
         }
 
+        private static List<string> GetLogs_Orchestrator_Retreives_HttpManagementPayload(string instanceId, string[] functionNames, bool extendedSessions)
```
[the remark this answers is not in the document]
gled4er: Sure. I will be careful not to add trace validation. Do you prefer to remove it from the current PR?

test/TestOrchestrations.cs
```diff
@@ -19,6 +19,12 @@ public static string SayHelloInline([OrchestrationTrigger] DurableOrchestrationC
             return $"Hello, {input}!";
         }
 
+        public static HttpManagementPayload GetHttpManagementPayload([OrchestrationTrigger] DurableOrchestrationContext ctx)
+        {
+            HttpManagementPayload httpManagementPayload = ctx.CreateHttpManagementPayload(ctx.InstanceId);
```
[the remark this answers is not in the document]
gled4er: I explained a bit in the comment below. My thinking is that passing the `HttpManagementPayload`  information from the client to the orchestrator and the activity is currently not very useful. I think the most useful use cases is getting `HttpManagementPayload` from the orchestrator and then providing it to any activity that needs it. Maybe I missing part of your idea for design? 

The problem I see with having access to `HttpManagementPayload` only in the client is that first we need to start the orchestration and then to pass `HttpManagementPayload`. I see this working well only when the orchestrator is waiting for external event which make the whole flow a bit heavy in my opinion.

Please feel free to correct my understanding.

test/DurableTaskEndToEndTests.cs
```diff
@@ -1580,6 +1580,162 @@ private static void CompareTwoComplexTypeObjects(ComplexType firstObject, Comple
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// End-to-end test which validates HttpManagementPayload retrieved from Orchestration client when executing a simple orchestrator function which doesn't call any activity functions.
+        /// </summary>
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
```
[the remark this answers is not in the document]
gled4er: This is great clarification! Do you think it make sense to start maintaining a wiki with all similar tips and guidance in order to reduce your review time?

test/TestOrchestrations.cs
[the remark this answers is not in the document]
cgillum: I agree trying to pass this data from the client to the orchestrator to the activity is not very useful. However, my idea was a little different, so I apologize for not being clear. My thought was to simply have an activity function *in the middle of an orchestration* which uses the `[OrchestrationClient]` binding to get an instance of `HttpManagementPayload`. For example:

```csharp
public static void SendInstanceInfo(
    [ActivityTrigger] DurableActivityContext ctx,
    [OrchestrationClient] DurableOrchestrationClient client)
{
    var payload = client.CreateHttpManagementPayload(ctx.InstanceId);
    // send the payload to some external destination
}
```
Let me know what you think.

cgillum:
Sorry, I think I didn't not communicate clearly enough. I mentioned this comment directly in the code, but I'll paste it here as well so it doesn't get lost:

I agree trying to pass this data from the client to the orchestrator to the activity is not very useful. However, my idea was a little different, so I apologize for not being clear. My thought was to simply have an activity function in the middle of an orchestration which uses the `[OrchestrationClient]` binding to get an instance of `HttpManagementPayload`. For example:

```csharp
public static void SendInstanceInfo(
    [ActivityTrigger] DurableActivityContext ctx,
    [OrchestrationClient] DurableOrchestrationClient client)
{
    var payload = client.CreateHttpManagementPayload(ctx.InstanceId);
    // send the payload to some external destination
}
```
Let me know what you think.

test/TestOrchestrations.cs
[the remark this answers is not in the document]
gled4er: Hello @cgillum,

This is great! Thank you very much for the sample! This made it clear for me. I will remove the code related to `DurableOrchestrationContext`. 

I agree with you that we should not include API methods to `DurableOrchestrationContext` if we don't need them. We need to access `HttpManagementPayload` in the activity functions. So support in `DurableOrchestrationClient` will be enough!

Thank you!

gled4er:
Hello @cgillum,

I removed the code relegated to `DurableOrchestrationContext` and I modified one of the existing integration tests to follow the sample you gave me.

Please when you have time take a look at the current state of the code and let me know what should be fixed next.

Thank you!

test/TestActivities.cs
```diff
@@ -86,9 +86,11 @@ public static double BindToDouble([ActivityTrigger] double value)
             await input.CopyToAsync(output);
         }
 
-        public static HttpManagementPayload GetAndPassHttpManagementPayload([ActivityTrigger] DurableActivityContext ctx)
+        public static HttpManagementPayload GetAndPassHttpManagementPayload(
```
cgillum: Let's rename this to `GetAndReturnHttpManagementPayload`.

test/TestHelpers.cs
```diff
@@ -275,7 +257,7 @@ private static List<string> GetLogs_Activity_Gets_HttpManagementPayload(string i
...
-                    $"{instanceId}: Function '{functionNames[1]} ({FunctionType.Activity})' started. IsReplay: False. Input: [{{\"$type\":\"Microsoft.Azure.WebJobs.Extensions.DurableTask.HttpManagementPayload, Microsoft.Azure.WebJobs.Extensions.DurableTask\",\"id\":\"{instanceId}\",\"statusQueryGetUri\":\"{TestConstants.NotificationUrlBase}/instances/{instanceId}?taskHub={taskHubName}&connection=Storage&code=mykey\",\"sendEventPostUri\":\"{TestConstants.NotificationUrlBase}/instances/{instanceId}/raiseEvent/{{eventName}}?taskHub={taskHubName}&connection=Storage&code=mykey\",\"terminatePostUri\":\"{TestConstants.NotificationUrlBase}/instances/{instanceId}/terminate?reason={{text}}&taskHub={taskHubName}&connection=Storage&code=mykey\"}}]. State: Started. HubName: {taskHubName}. AppName: . SlotName: . ExtensionVersion: 1.4.1.0. SequenceNumber: 4.",
+                    $"{instanceId}: Function '{functionNames[1]} ({FunctionType.Activity})' started. IsReplay: False. Input: [null]. State: Started. HubName: {taskHubName}. AppName: . SlotName: . ExtensionVersion: 1.4.1.0. SequenceNumber: 4.",
```
cgillum: Let's remove all the text at `HubName` and beyond. We don't want to test for this because that will cause test failures when, for example, we update the extension version. The test code does a prefix comparison, so the full trace doesn't need to match.

test/TestHelpers.cs
```diff
@@ -236,6 +241,46 @@ private static List<string> GetLogs_HelloWorldOrchestration_Inline(string instan
             return list;
         }
 
+        private static List<string> GetLogs_Activity_Gets_HttpManagementPayload(string instanceId, string[] functionNames, bool extendedSessions)
```
cgillum: Can we actually remove these log checks too? I think the other components of your end-to-end test are already sufficient.

gled4er:
Hello @cgillum,

Thank you!

test/TestActivities.cs
gled4er: Thank you! I will change it.

gled4er:
Hello @cgillum,

I modified the code according your recommendations.

Please check once more when you have time and let me know what should be fixed next.

Thank you!

gled4er:
Hello @cgillum,

Thank you very much for the support and especially for the sample that clarify the design for me!

Thank you!

gled4er:
Hello @cgillum,

Shall we update the documentation with this scenario? If so where do you want to add the new feature?

Thank you!

cgillum:
Yes, we should update the documentation.  I think the existing **Instance Management** topic is the right place for this. Thanks!

gled4er:
Hello @cgillum,

I will work on that.

Thank you!

Issue #161, opened 2018-02-20
Support creating DurableOrchestrationClient outside of Azure Functions


cgillum:
This issue tracks the ability to create an instance of [DurableOrchestrationClient](https://azure.github.io/azure-functions-durable-extension/api/Microsoft.Azure.WebJobs.DurableOrchestrationClient.html) outside the Azure Functions context. It looks like we don't have a public constructor for this type, so it's not currently possible. 

Here is the relevant source code: https://github.com/Azure/azure-functions-durable-extension/blob/master/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs#L27.

The challenge will be creating a very simple public constructor that only requires an Azure Storage connection string and a task hub name.

ConnorMcMahon:
This is getting pushed to October release.

cgillum:
The PR for this is merged and will go out with v2.4.0.

PR #327, opened 2018-06-03
Add support for DurableOrchestrationClient outside Azure Functions

gled4er:
Hello @cgillum,

May I ask you to give me early feedback on my approach for implementing [#161](https://github.com/Azure/azure-functions-durable-extension/issues/161)?

Thank you!

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
@@ -5,9 +5,13 @@ using System;
 using System.Collections.Generic;
 using System.Net.Http;
 using System.Threading.Tasks;
+using DurableTask.AzureStorage;
 using DurableTask.Core;
 using DurableTask.Core.History;
 using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+using Microsoft.Azure.WebJobs.Host.Config;
+using Microsoft.Azure.WebJobs.Logging;
+using Microsoft.Extensions.Logging;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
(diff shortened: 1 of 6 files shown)


src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -42,6 +46,38 @@ public class DurableOrchestrationClient : DurableOrchestrationClientBase
...
+        {
+            this.config = new DurableTaskExtension();
+            JobHostConfiguration jobHostConfiguration =
+                new JobHostConfiguration(connectionName)
+                {
+                    LoggerFactory = new LoggerFactory(),
+                };
+            ExtensionConfigContext extensionConfigContext = new ExtensionConfigContext
+            {
+                Config = new JobHostConfiguration(connectionName),
```
cgillum: You've already created a `JobHostConfiguration` a few lines above this. Why not use that one instead of creating another?

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -42,6 +46,38 @@ public class DurableOrchestrationClient : DurableOrchestrationClientBase
...
+            this.attribute = new OrchestrationClientAttribute
+            {
+                TaskHub = taskHubName ?? this.config.HubName,
+                ConnectionName = connectionName ?? this.config.AzureStorageConnectionStringName ?? ConnectionStringNames.Storage,
+            };
+            AzureStorageOrchestrationServiceSettings settings = this.config.GetOrchestrationServiceSettings(this.attribute);
+            var serviceClient = new AzureStorageOrchestrationService(settings);
+            this.client = new TaskHubClient(serviceClient);
+
+            ILoggerFactory loggerFactory = new LoggerFactory();
```
cgillum: You created another `LoggerFactory` previous to this. Why not reuse that one?

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -42,6 +46,38 @@ public class DurableOrchestrationClient : DurableOrchestrationClientBase
...
+            JobHostConfiguration jobHostConfiguration =
+                new JobHostConfiguration(connectionName)
+                {
+                    LoggerFactory = new LoggerFactory(),
+                };
+            ExtensionConfigContext extensionConfigContext = new ExtensionConfigContext
+            {
+                Config = new JobHostConfiguration(connectionName),
+            };
+            ((IExtensionConfigProvider)this.config).Initialize(extensionConfigContext);
```
cgillum: I worry about using types like `ExtensionConfigContext` in our extension code because they're really only "supposed" to be used by the WebJobs host. My fear is that some behavior could be changed by the functions team in a way we don't expect and break customers who depend on this.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
gled4er: Thank you! I will fix it.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
gled4er: Thank you for catching this! I will improve it.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
gled4er: Yes, I agree with you. When I was calling this method the feeling was that we are hacking the API-s and trying to find workarounds.

Thank you!

ConnorMcMahon:
A similar work item to this has been merged as a part of v2.4.0, but thank you for your contribution @gled4er .

PR #349, opened 2018-06-19
Update to v1.5.0 and additional fixes

cgillum:
This PR:
* Bumps up our package version to **1.5.0** (for the new release)
* Fixes an exception serialization regression caused by our recent move to Newtonsoft.Json when targeting netstandard2.0.
* Fixes a variety of StyleCop warnings and documentation issues that were introduced by other contributors

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClientBase.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClientBase.cs
@@ -48,9 +48,9 @@ namespace Microsoft.Azure.WebJobs
         /// or contains the payload containing the output of the completed orchestration.
         /// </summary>
         /// <remarks>
-        /// If the orchestration instance completes within the specified timeout, then the HTTP response payload will
+        /// If the orchestration instance completes within the default 10 second timeout, then the HTTP response payload will
         /// contain the output of the orchestration instance formatted as JSON. However, if the orchestration does not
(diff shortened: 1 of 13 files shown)


src/WebJobs.Extensions.DurableTask/DurableOrchestrationClientBase.cs
```diff
@@ -40,10 +40,22 @@ public abstract class DurableOrchestrationClientBase
...
 
-        /// <inheritdoc />
+        /// <summary>
+        /// Creates an HTTP response which either contains a payload of management URLs for a non-completed instance
+        /// or contains the payload containing the output of the completed orchestration.
+        /// </summary>
+        /// <remarks>
+        /// If the orchestration instance completes within the specified timeout, then the HTTP response payload will
+        /// contain the output of the orchestration instance formatted as JSON. However, if the orchestration does not
+        /// complete within the specified timeout, then the HTTP response will be identical to that of the
```
kashimiz: This overload of WaitForCompletionOrCreateCheckStatusResponseAsync doesn't take a timeout parameter, so should we change instances of "specified timeout" here to "30 seconds"?

cgillum: Good catch. I had a feeling that a blind copy/paste might get me into trouble...

Issue #365, opened 2018-06-27
Wait for external event with timeout


markheath:
Inspired by [Mikhail Shilkov's tweet](https://twitter.com/MikhailShilkov/status/1011705530648334337) I wonder if we could simplify the syntax for waiting for an external event with a timeout, as there are multiple lines of boilerplate code required to set this up, and it requires implementers remembering to adhere to the correct patterns (remembering to cancel the `CancellationTokenSource` and remembering to use `DurableOrchestrationContext.CurrentUtcDateTime` instead of `DateTime.Now`).

What would be nice is an overload of `WaitForExternalEvent` that simply takes a `TimeSpan` timeout and returns either the received event or something else to indicate a timeout.

I wondered whether this could be simply implemented as an extension method on `DurableOrchestrationContext` like this:

    static class DurableOrchestrationContextExtensions
    {
        public static async Task<T> WaitForExternalEvent<T>(this DurableOrchestrationContext ctx, string eventName, TimeSpan timeoutDuration)
        {
            using (var cts = new CancellationTokenSource())
            {
                var timeoutAt = ctx.CurrentUtcDateTime + timeoutDuration;
                var timeoutTask = ctx.CreateTimer(timeoutAt, cts.Token);
                var waitForEventTask = ctx.WaitForExternalEvent<T>(eventName);

                var winner = await Task.WhenAny(waitForEventTask, timeoutTask);
                if (winner == waitForEventTask)
                {
                    cts.Cancel();
                    return waitForEventTask.Result;
                }
                else
                {
                    return default(T);
                }
            }
        }
    }

This would greatly simplify the orchestrator code to something like:

    var approvalResult = await ctx.WaitForExternalEvent<string>("OrderApprovalResult", TimeSpan.FromMinutes(30));
    if (approvalResult == null)
    {
       // we timed out
    }

However, this doesn't work because of the restrictions on what the orchestrator can do with the `await` keyword (it doesn't even work if I eliminate the `await` keyword from the extension method and use `ContinueWith` instead).

Would it be possible to implement a method like this in the `DurableOrchestrationContext` class itself? I had a quick look at the code, but it wasn't obvious to me what would need to be done to provide this directly on `DurableOrchestrationContext`.

My example implementation uses `default(T)` to indicate timeout, which I suspect would be fine in most cases. It's shame C# doesn't have an idiomatic representation of discriminated unions like F# does, which would be ideal.

[...]

Issue #14, opened 2017-07-15
Allow timers to last longer than 7 days


cgillum:
## Issue
The Azure Storage provider uses invisible queue messages to implement durable timers. Queue messages have a maximum lifetime of 7 days, so durable timer messages cannot live longer than this.

## Proposal
Allow durable timers to have an arbitrary lifetime, which could go as high as weeks or months.

## Motivation
Real world scenarios often involve extended timeouts. For example, a late fee might be applied to a bill if it is not paid 30 days after it is due. A durable timer is a natural way to track this kind of expiration, and a 7 day limitation reduces the usefulness of the feature.

## Design Notes
The fix would be specific to the Azure Storage provider extension in the Durable Task Framework (the Service Bus provider has no such limits). Azure Storage cannot support a message living longer than 7 days in the queue. To work around this, if an orchestrator creates a timer for 30 days, internally a message should be tagged with metadata indicating the intended receive time (e.g. 30 days from now) but it should be enqueued for only **72 hours** (using a number small than 7 days is preferred to avoid potential data loss if the message cannot be immediately dequeued for some reason).

Once a message is received, the provider checks to see if a `ScheduledReceiveTimeUtc` metadata property exists and if it does, compares that time to the current time. If the `ScheduledReceiveTimeUtc` is within 5 minutes from the current time, the message can be processed. Otherwise, a new copy of the message should be enqueued for as many as another 72 hours (or less if the scheduled receive time is less than 72 hours away).

## Drawbacks
* If an orchestrator function is waiting on a timer for extended periods of time, then it will be periodically woken up (consuming CPU and memory) without having any real work to do.
* A single timer event can result in multiple messages being enqueued and received in Azure Storage, which means the message ID cannot be used for end-to-end correlation. The extension already defines its own correlation ID for messages and end-to-end tracing relies on this, so it's not a significant problem, but it is something to be aware of in case someone attempts to change this correlation behavior or try to do correlation purely using Storage logs.

SimonLuckenuik:
@cgillum is there any target date for this issue? Is there a durable task framework related issue I can monitor? I guess this is low priority considering that I can reschedule the message myself.

cgillum:
No specific date - we just need to get around to doing it. We'll definitely include it for GA, which is targeting Spring 2018, but ideally this fix will come much sooner than that.

[...]

PR #385, opened 2018-07-10
WaitForExternalEvent with timeout support

markheath:
This is a first attempt at implementing overloads of `WaitForExternalEvent` with timeout support as discussed in issue #365. There are two overloads - one that throws a `TimeoutException`, the other returns a user-specified default value.

I'd like to get some end to end tests going for this. At the moment, I'm still getting my machine configured to run the tests (ActorOrchestration is currently failing for me).


markheath:
Have added some end to end tests.

markheath:
Awesome! No worries about the delay, I've been pretty busy myself. I'm hoping to revisit #14 at some point to see if a similar approach can be used to solve that.

PR #406, opened 2018-08-19
Add TaskHubName parameter to RaiseEventAsync method

yuka1984:
Hi.
I created this PullRequest because I wanted to specify TaskHubName as a parameter of RiaseEventAsync function of DurableTaskClient.

There are two kinds of means to activate Orchestrator waiting in WaitForExternal: a method of making a Post request to Api and a method of using DurableOrchestrationClient.RaiseEventAsync function.

The way to make a Post request to Api is to specify InstanceId, event name and TaskHubName, but with the method using DurableOrchestrationClient.RaiseEventAsync function you can only specify event name and InstanceId.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
@@ -132,6 +132,54 @@ namespace Microsoft.Azure.WebJobs
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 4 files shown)


cgillum:
That's fine. Other users also will also forget about `Binder`, so this PR is still valuable. If you can can fix the code issues that I mentioned, then I will accept this PR.

yuka1984:
I fix the code.

PR #408, opened 2018-08-23
Updates for Azure Functions v2 breaking changes

cgillum:
This is a pretty sizeable PR which handles the breaking change updates for Functions v2.  It is more complex than other extensions because our single package targets both v1 (.NET 4.6.1) and v2 (.NET Standard 2.0),

The main changes are:

* Split test project into two test projects, all which share the same test code but have different setup code. Previously we only had one project which ran on .NET Core. The differences between our v1 and v2 are now substantial enough to warrant two test projects with shared code.
* Added new cross-version abstractions to minimize the number of places we need to use `#if`.
* Also added improved tracing to make it easier to debug issues.
* Update to the latest version of DurableTask.AzureStorage, which contains a bunch of perf and reliability improvements.


src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -131,8 +134,12 @@ void IExtensionConfigProvider.Initialize(ExtensionConfigContext context)
         {
             ConfigureLoaderHooks();
 
-            // This is a no-op when targetting .NET Standard 2.0 (Functions v2)
-            this.InitializeForFunctionsV1(context);
+            // Functions V1 has it's configuration initialized at startup time (now).
```
kashimiz: Minor grammatical nit: its, not it's.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -106,21 +106,19 @@ public override async Task<string> StartNewAsync(string orchestratorFunctionName
 
         /// <inheritdoc />
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1030:UseEventsWhereAppropriate", Justification = "This method does not work with the .NET Framework event model.")]
-        public override async Task RaiseEventAsync(string instanceId, string eventName, object eventData)
+        public override Task RaiseEventAsync(string instanceId, string eventName, object eventData)
```
kashimiz: Why are we changing this from async? Also, how does this commit fix the infinite loop bug / what was it? (Was it ever filed as an issue?)

src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
```diff
@@ -0,0 +1,284 @@
...
+        /// <summary>
+        /// Gets or sets the amount of time in seconds before an idle session times out. The default value is 30 seconds.
+        /// </summary>
+        /// <remarks>
+        /// This setting is applicable when <see cref="ExtendedSessionsEnabled"/> is set to <c>true</c>.
+        /// </remarks>
+        /// <value>
+        /// The number of seconds before an idle session times out.
+        /// </value>
+        public int ExtendedSessionIdleTimeoutInSeconds { get; set; } = 30;
```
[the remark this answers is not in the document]
cgillum: I can't think of a particular reason to enforce one.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
cgillum: Regarding `async`, it's more efficient and is easier on the debugging experience to have a method return a task directly rather than awaiting a task.  I'm pretty consistent about coding this way.

Regarding the infinite loop: this was caused by a PR I recently accepted, so it was limited to the dev branch. It went uncaught because I forgot to tag the PR will "full-ci" (I'm not sure how the PR author didn't notice it when testing locally) and assumed all tests ran when I saw the green check-mark in GitHub. I wonder whether there is some process we can look into which ensures that tests are actually run before PRs are accepted, or at least makes it more clear that a full CI has not yet been run.

Issue #433, opened 2018-09-10
Strong name sign latest assemblies


mhoeger:
Details and prior work: https://github.com/Azure/azure-webjobs-sdk/issues/1883

cgillum:
This was done as part of our v1.6.1 release, which went out at the same time as Functions v2 RC1.

PR #434, opened 2018-09-11
Strong-name signing for Durable Functions

cgillum:
We have received several asks to have all Azure Functions assemblies to be strong-name signed. This PR creates a new strong-name signing key pair and uses it for assembly signing. This resolves https://github.com/Azure/azure-functions-durable-extension/issues/433.

With this change the new strong name of the extension assembly is:
```
Microsoft.Azure.WebJobs.Extensions.DurableTask, Version=1.0.0.0, Culture=neutral, PublicKeyToken=014045d636e89289, processorArchitecture=MSIL
```

Note that the assembly version is now `1.0.0.0`. The file version will continue to use semver, e.g. `v1.6.1`. The plan moving forward is to only update the assembly version for major updates - i.e. we will always use `1.0.0.0` for all `v1.X.X` releases. Similarly we will use `2.0.0.0` etc. for all `v2.X.X` releases. This design is consistent with many other existing .NET projects.

Note that for open source projects, Microsoft recommends checking in the private key into the repository. This is safe because strong name keys are used for identity, not security. See here for reference: https://docs.microsoft.com/en-us/dotnet/framework/app-domains/strong-named-assemblies

> Do not rely on strong names for security. They provide a unique identity only.

They also specifically address open source projects:

> If you are an open-source developer and you want the identity benefits of a strong-named assembly, consider checking in the private key associated with an assembly into your source control system.


cgillum:
I forgot to update the WebJobs dependencies to 3.0.0-rc1. Will add that in the next commit.

PR #438, opened 2018-09-17
Changed SayHello sample to look more chained

totosan:
**Reason for PR**
According to the Chaining-Sample in MS docs (https://docs.microsoft.com/en-gb/azure/azure-functions/durable-functions-sequence), this modified sample code supports better understanding of what chaining is. It means output of one function is input for another function.

**What's changed**
- VSSample > HelloSequence
  instead calling one function after another with independent out- and input, there are different functions, using the output of it's predecessor.
- it will sequence the create "Hello Tokyo and Seattle and London!".   Start -> Func1("Tokyo") >> "Hello Tokyo" -> Func2(..) >> "Hello Tokyo and Seattle" -> Func3(..) >> "Hello Tokyo and Seattle and London!" -> End

It would be nice, if you could take such an example, instead of the current one.

### DIFF
--- a/samples/VSSample.Tests/HelloSequenceActivityTests.cs
+++ b/samples/VSSample.Tests/HelloSequenceActivityTests.cs
@@ -10,8 +10,22 @@ namespace VSSample.Tests
         [Fact]
         public void SayHello_returns_greeting()
         {
-            var result = HelloSequence.SayHello("John");
-            Assert.Equal("Hello John!", result);
+            var result = HelloSequence.SayHello("there!");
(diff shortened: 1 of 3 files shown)


totosan:
this relates to [Doc-Issue #15070](https://github.com/MicrosoftDocs/azure-docs/issues/15070#issuecomment-422127209)

cgillum:
Before accepting this, we need to see what impact it has on documentation (since the documentation references the sample code directly).

totosan:
So, I could also adjust the sample, if you like.

PR #442, opened 2018-09-18
Durable JS GA changes.

kashimiz:
This PR contains changes to enable the following:

- flow parentInstanceId, isReplaying and instanceId through to DurableOrchestrationContext during out-of-proc binding
- handle out-of-proc requests to CallActivityWithRetry, CallSubOrchestrator, CallSubOrchestratorWithRetry, ContinueAsNew


src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -45,13 +47,43 @@ internal DurableOrchestrationContext(DurableTaskExtension config, string functio
...
+                    return this.innerContext.IsReplaying;
+                }
+                else
+                {
+                    return this.isReplaying;
+                }
+            }
+
+            internal set => this.isReplaying = value;
+        }
```
cgillum: With this change, it's no longer clear what is intended to be the source of truth - the inner context object or the new private members you're adding. I think we have to choose one and go with it or else our future selves (or future contributors) will get confused. There is precedent for mutable properties, such as `ParentOrchestrationId`, so I suggest you follow that model.

src/WebJobs.Extensions.DurableTask/RetryOptionsConverter.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        public override bool CanConvert(Type objectType) => objectType.IsAssignableFrom(typeof(RetryOptions));
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            JObject obj = JObject.Load(reader);
+
+            if (!this.FieldExists(FirstRetryIntervalField, obj))
+            {
+                throw new ArgumentException("Invalid JSON. Must contain field firstRetryIntervalInMilliseconds", "reader");
```
cgillum: I suggest using the constants you defined in these error messages:
```csharp
$"Invalid JSON. Must contain field {FirstRetryIntervalField}"
```

src/WebJobs.Extensions.DurableTask/RetryOptionsConverter.cs
```diff
@@ -0,0 +1,71 @@
...
+            }
+
+            if (!this.FieldExists(MaxAttemptsField, obj))
+            {
+                throw new ArgumentException("Invalid JSON. Must contain field maxNumberOfAttempts", "reader");
```
cgillum: Same comment about reusing your constants.
```csharp
$"Invalid JSON. Must contain field {MaxAttemptsField}."
```

src/WebJobs.Extensions.DurableTask/RetryOptionsConverter.cs
```diff
@@ -0,0 +1,71 @@
...
+            }
+
+            var firstRetryIntervalInMilliseconds = obj[FirstRetryIntervalField].Value<int>();
+            var maxNumberOfAttempts = obj[MaxAttemptsField].Value<int>();
+
+            var target = new RetryOptions(
+                new TimeSpan(firstRetryIntervalInMilliseconds * TimeSpan.TicksPerMillisecond),
```
cgillum: I would prefer `TimeSpan.FromMilliseconds(firstRetryIntervalInMilliseconds)`.

src/WebJobs.Extensions.DurableTask/RetryOptionsConverter.cs
```diff
@@ -0,0 +1,71 @@
...
+                maxNumberOfAttempts);
+
+            if (this.FieldExists(BackoffCoefficientField, obj))
+            {
+                target.BackoffCoefficient = obj[BackoffCoefficientField].Value<double>();
+            }
+
+            if (this.FieldExists(MaxRetryIntervalField, obj))
+            {
+                target.MaxRetryInterval = new TimeSpan(obj[MaxRetryIntervalField].Value<int>() * TimeSpan.TicksPerMillisecond);
```
cgillum: I would prefer:
```csharp
int maxRetryIntervalMilliseconds = obj[MaxRetryIntervalField].Value<int>();
target.MaxRetryInterval = TimeSpan.FromMilliseconds(maxRetryIntervalMilliseconds );
```

src/WebJobs.Extensions.DurableTask/RetryOptionsConverter.cs
```diff
@@ -0,0 +1,71 @@
...
+            }
+
+            if (this.FieldExists(RetryTimeoutField, obj))
+            {
+                target.RetryTimeout = new TimeSpan(obj[RetryTimeoutField].Value<int>() * TimeSpan.TicksPerMillisecond);
```
cgillum: Same idea here:
```csharp
int retryTimeoutMilliseconds = obj[RetryTimeoutField].Value<int>();
target.RetryTimeout = TimeSpan.FromMilliseconds(retryTimeoutMilliseconds);
```

src/WebJobs.Extensions.DurableTask/RetryOptionsConverter.cs
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            throw new NotImplementedException("Unnecessary because CanWrite is false. The type will skip the converter.");
+        }
+
+        private bool FieldExists(string fieldName, JObject jObject)
+        {
+            return jObject[fieldName] != null;
```
cgillum: I'm a little squeamish about relying on null values to determine whether a field exists. You get a more future-proof contract with this:
```csharp
return jObject.TryGetValue(fieldName, out _);
```

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContextBase.cs
```diff
@@ -59,7 +59,11 @@ public abstract class DurableOrchestrationContextBase
         /// <value>
         /// <c>true</c> if the orchestrator function is currently being replayed; otherwise <c>false</c>.
         /// </value>
-        public virtual bool IsReplaying => false;
+        public virtual bool IsReplaying
+        {
+            get => false;
+            internal set => this.IsReplaying = value;
+        }
```
cgillum: This looks like a potential `StackOverflowException`. Can we just do this?
```csharp
public virtual bool IsReplaying { get; internal set; }
```
I don't remember why it was previously hard-coded to `false` but I can't imagine it was necessary.

src/WebJobs.Extensions.DurableTask/RetryOptionsConverter.cs
```diff
@@ -0,0 +1,71 @@
...
+                throw new ArgumentException("Invalid JSON. Must contain field firstRetryIntervalInMilliseconds", "reader");
+            }
+
+            if (!this.FieldExists(MaxAttemptsField, obj))
+            {
+                throw new ArgumentException("Invalid JSON. Must contain field maxNumberOfAttempts", "reader");
+            }
+
+            var firstRetryIntervalInMilliseconds = obj[FirstRetryIntervalField].Value<int>();
+            var maxNumberOfAttempts = obj[MaxAttemptsField].Value<int>();
```
cgillum: It just occurred to me that you're doing double reads on these fields. Rather than having a `FieldExists` method, it would be more efficient to just read once using TryGetValue, like I suggested in my other comment (except you would need to keep the value instead of discarding it).

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -36,6 +36,7 @@ public sealed class DurableOrchestrationContext : DurableOrchestrationContextBas
         private string serializedInput;
         private string serializedOutput;
         private string serializedCustomStatus;
+        private string instanceId;
```
cgillum: Can we remove the `instanceId` field here, remove the `InstanceId` property override below, and make `InstanceId` on the base class virtual instead of abstract?

PR #453, opened 2018-09-28
removing redundant setting

jeffhollan:
@kashimiz I don't think we need this value? I don't see it being accessed from any samples and adds some confusion especially when running without storage emulator

per feedback from @sdras

### DIFF
--- a/samples/javascript/local.settings.json
+++ b/samples/javascript/local.settings.json
@@ -3,7 +3,6 @@
   "Values": {
     "FUNCTIONS_WORKER_RUNTIME": "node",
     "AzureWebJobsStorage": "UseDevelopmentStorage=true",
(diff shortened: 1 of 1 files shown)


sdras:
Thank you!

cgillum:
I asked @kashimiz about this setting and apparently it was needed by one of the samples which [uses the Azure Storage SDK directly](https://github.com/Azure/azure-functions-durable-extension/blob/bd887fce280023445b112d82555a667d4d7b9a60/samples/javascript/E2_CopyFileToBlob/index.js#L3). Is that not the case?

cgillum:
Yeah, I'd be cool with that change. The fewer config settings, the better. :)

jeffhollan:
Closing in favor of this PR:

https://github.com/Azure/azure-functions-durable-extension/pull/480

Issue #296, opened 2018-05-03
Enable out-of-proc orchestration starter functions.


kashimiz:
Currently, the [JavaScript orchestration starter function](https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-instance-management#starting-instances) shown in the docs does not work, throwing an exception when the host runtime indexes it due to DurableOrchestrationClient being unable to bind to string.

We should fix this so our docs are accurate and JavaScript folks can do what the C# folks do.

Work will also be needed so that we can return the status/event/terminate endpoints JSON that a C# starter provides.

cgillum:
@kashimiz is this issue fixed with your recent binding PR?

kashimiz:
Fixed in that we can do out-of-proc starter functions per output bindings as instructed in our docs, but this part hasn't been implemented or designed:

> Work will also be needed so that we can return the status/event/terminate endpoints JSON that a C# starter provides.

kashimiz:
As of the 1.1.2 release of Azure/azure-functions-durable-js, this has been fixed.

PR #479, opened 2018-10-17
Added OrchestrationClient binding output converter.

kashimiz:
This commit adds a converter for OrchestrationClient when used as output. This is necessary to enable an [improved orchestration client experience](https://github.com/Azure/azure-functions-durable-js/issues/12) in Durable JS and other (upcoming) out-of-proc languages, given current limitations on out-of-proc bindings in Azure Functions.

Partially enables #296.


cgillum:
Looks like there is a failing test. The same test fails in both V1 and V2. Please make sure it is passing before merging.

PR #480, opened 2018-10-19
[js] clean sample connection strings

jeffhollan:


tohling:
So proud of you @jeffhollan!

Issue #389, opened 2018-07-13
Durable Functions  (Fan Out) + Azure Storage  = Host Threshold Exceeded (Connections)


mayoatte:
Hey folks,

Thanks for all your hard work on durable functions. I've used them on multiple projects and i'm a big fan.

On a recent project, we ran into an issue that occurs with durable functions  that interact with azure storage. For example: 

- Durable orchestrator receives a batch of event data to save.
- Durable orchestrator fans out activity functions to save event data as small blobs in storage 
- At an interval, another Durable orchestrator fans out activity functions to combine the small blobs into one large blob for upload to another data store e.g. data lake.

This setup causes the "Host Threshold Exceeded (Connections)" error very quickly.  I was able to monitor the outbound connections while running the function app locally and there are too many connections to storage (queues, tables, blobs). I even commented out a bunch of our code and there were 120+ outbound connections to storage within 60 seconds (I have the IPs available) and it only kept increasing.

I think there is an issue here.  Either the outbound connections from the durable framework shouldn't count against the sandbox limits or the durable framework should minimize the number of outbound connections it makes. Taking up so many of the available connections really limits what is left over for the actual logic of the function.

Please note that we followed all the connection management recommendations (lazy init static clients etc). 

Some details about the function app:

- Function Runtime: 2.0.11888.0
- Durable Extension Version: 1.5.0 
- Region: East US 2
- Invocation Id: 4c91d7e6-a206-4997-919d-b9555ce9b2c8

cgillum:
Very interesting. That's definitely not expected and I've not seen this kind of issue with Durable Functions running in our Consumption plan yet.

Just to make sure I'm understanding correctly, you're pretty confident that nothing in your code could be causing these connections to build-up, right?

I'll need to see if I can reproduce this locally - but I agree that using so many connections seems excessive. Under the hood we're just using the Azure Storage client SDK and caching client objects, so I wouldn't expect there to be any connection leaks.

cgillum:
I ran a quick load test on my local box and sure enough, I see the number of established TCP connections spike up to almost 200. In my case, it wasn't a fan-out/fan-in, but rather a large number of sequential orchestrations running concurrently.

As a potential workaround, could you try limiting the concurrency using the [concurrency throttles](https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-perf-and-scale#concurrency-throttles)?

mayoatbm:
[...]

PR #486, opened 2018-10-26
temporary workaround for setting MaxConnectionsPerServer in Storage

brettsam:


cgillum:
@brettsam can you target the **dev** branch instead of **master**? That's our preferred workflow.

brettsam:
oh whoops yep. let me re-push the change.

src/WebJobs.Extensions.DurableTask/MaxConnectionHelper.cs
```diff
@@ -0,0 +1,56 @@
...
+                        handler.MaxConnectionsPerServer = 50;
+                    }
+                    catch (Exception ex)
+                    {
+                        logger.LogDebug($"Azure Storage HttpClientHandler was not updated: '{ex}'. Current value: {handler.MaxConnectionsPerServer}");
+                    }
+                }
+                else
+                {
+                    logger.LogDebug($"Azure Storage HttpClientHandler was not found in assembly: '{storageAssembly.FullName}'.");
```
cgillum: Can we use `LogWarning` instead of `LogDebug`? I would like it if we could see this warning more clearly if the best-effort workaround isn't working.

cgillum:
Strangely the PR is showing your change in addition to a bunch of other changes in this branch since August. Is your fork up-to-date with the latest?

brettsam:
I thought I could re-target the dev branch via github... but apparently there were ~10 commits difference between master and dev. So I had re-target from my local machine. It should be good now.

cgillum:
For the record, this resolves https://github.com/Azure/azure-functions-durable-extension/issues/389.

Issue #476, opened 2018-10-15
Ability to omit the Input field from the GetStatus HTTP API response


andylholloway:
I would like to have the option to not retrieve the Input.Entities[] in the json response when calling the getStatusUri. The reason for this is because we have durable functions that receive large payloads, which is distributed to activity functions using the Fan-Out, Fan-In pattern, and the large Input.Entities[] is rarely required when requesting a status update.

cgillum:
I'm not sure what `Input.Entities` refers to. Can you elaborate?  Is `Entities` the name of a list payload being used as the input to your orchestrator function?

andylholloway:
Hi @cgillum thanks for the reply. 
Yes that's right. `input.Entities[]` is a payload we are sending to the orchestrator function. When we query the `statusQueryGetUri` the response can be quite large. An example of the response is (entities not included for brevity):

```
{
    "instanceId": "1652390b26e247c394b6f632514b451b",
    "runtimeStatus": "Completed",
    "input": {
        "$type": "Common.Models.Payload, Common.Models",
        "Entities": [  ... may contain hundreds of entities ...  ]
    },
    "customStatus": "All Entities were processed successfully",
    "output": {
        "ActivityResults": [ ]
    },
    "createdTime": "2018-10-16T07:44:55Z",
    "lastUpdatedTime": "2018-10-16T07:45:01Z"
}
```
We have total control over the `customStatus` and the `output` which only displays a list of failed activity results. From what I can see the same control isn't possible with the `input` nor can you prevent the `input` data being returned in the json response. Could the API allow the caller to provide a query string param to prevent the `input` data being returned e.g. `showInput`, in the same way `showHistory` and `showHistoryOutput` works?

Currently I think the only option I have is to create a `GetStatus` function which can call `context.GetStatusAsync(instanceId)` and from there I could remove the `input` from the response.

cgillum:
Thanks for clarifying. I think we could solve this by adding a `showInput` flag as you suggested. To avoid a breaking change, we'd set it to `true` by default but callers like you could set it to `false` to avoid the cost of retrieving this value.

andylholloway:
Thanks for adding this issue as an enhancement. Would you have an ETA for this change?

cgillum:
ETA is a bit hard for features like this because it largely depends on how much help we can get. However, I might have someone who can help relatively soon...

cgillum:
This functionality has been checked-in and will be available in the next release.

[...]

PR #490, opened 2018-11-01
Add option to avoid retrieving orchestration instance inputs

glennamanns:
addresses #476 & adds the ability to avoid the cost of retrieving the 'Input' column of the DurableFunctionsHubInstances table.

**Options:** 

1. Directly call GetStatusAsync on DurableOrchestrationClient with the boolean hideInput set to true (`await starter.GetStatusAsync(instanceId, showHistory: false, showHistoryOutput: false, showInput: false`)) 

2. Hit the getStatus URI with showInput=false in the query string (GET http://localhost:7071/runtime/webhooks/durabletask/instances/123456?taskHub=DurableFunctionsHub&connection=Storage&code=blahblah&showInput=false)

**Some things to note:**
- Default value of `showInput` is true.
- For the Durable Functions Extension, the HTTP query parameter name is `showInput`. 
- In DTFx, the name is `fetchInput`, rather than `showInput`, because the name `showInput` implies that the input is being retrieved but not shown. I opted to not retrieve the input at all because there's no point in wasting bandwidth on it. 
- `DurableOrchestrationClient` directly calls `GetOrchestrationStateAsync` of `AzureStoreOrchestrationService.` This was done to avoid making a change to TaskHubClient and the DurableTask.Core assembly.

Corresponding DTFx pull request: https://github.com/Azure/durabletask/pull/226


gled4er:
Hello @cgillum @glennamanns,

Do you think it will be useful to provide also an option for hiding the output as well? 

Have you seen use case scenarios that produce large output?

Thank you!

cgillum:
Outputs are hidden by default. So basically with this change, you opt-in to seeing outputs but opt-out of seeing inputs. It's unfortunately not consistent, but this is the way we have to go to avoid breaking changes.

glennamanns:
I merged with dev and changed the base branch. This means the PR is now in a pretty weird state, because dev is 9 commits ahead of and 10 commits behind master..

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -262,7 +263,11 @@ public async Task<HttpResponseMessage> HandleRequestAsync(HttpRequestMessage req
             var queryNameValuePairs = request.GetQueryNameValuePairs();
             var showHistory = GetBooleanQueryParameterValue(queryNameValuePairs, ShowHistoryParameter);
             var showHistoryOutput = GetBooleanQueryParameterValue(queryNameValuePairs, ShowHistoryOutputParameter);
-            var status = await client.GetStatusAsync(instanceId, showHistory, showHistoryOutput);
+
+            bool? nullableFetchInput = GetNullableBooleanQueryParameterValue(queryNameValuePairs, FetchInputParameter);
+            bool fetchInput = (nullableFetchInput == null) ? true : (bool)nullableFetchInput; // Default to true
```
cgillum: I simpler way to express this is:
```csharp
bool fetchInput = nullableFetchInput.GetValueOrDefault(true);
```

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -363,6 +368,12 @@ private static bool GetBooleanQueryParameterValue(NameValueCollection queryStrin
             return bool.TryParse(value, out bool parsedValue) && parsedValue;
         }
 
+        private static bool? GetNullableBooleanQueryParameterValue(NameValueCollection queryStringNameValueCollection, string queryParameterName)
```
cgillum: This would be a slightly more natural and simpler way to design this API.
```csharp
private static bool GetBooleanQueryParameterValue(NameValueCollection queryStringNameValueCollection, string queryParameterName, bool defaultValue)
{
    var value = queryStringNameValueCollection[queryParameterName];
    return bool.TryParse(value, out bool parsedValue) ? parsedValue : defaultValue;
}
```

cgillum:
BTW, since this adds new public surface area, we need to document it in two places:

* https://docs.microsoft.com/azure/azure-functions/durable-functions-instance-management
* https://docs.microsoft.com/azure/azure-functions/durable-functions-http-api

However, @ggailey777 is currently moving our docs to their own dedicated section, so we'll need to wait for that refactoring to happen first before we can update the content.

ggailey777:
@cgillum we can make updates in the release-functions-durable-migration branch, if we need them ready to go when we complete the migration.

Issue #499, opened 2018-07-10
Orchestrationclient TaskHub configurable option.


praveena-m:
Instead of hardcoding a taskhub name, need a way to define it in an app setting and refer to it using % signs. 

For instance, if you have a setting called CRON_EXPRESSION, then you can refer that as below (Reference : [https://github.com/Azure/azure-functions-host/issues/1934](url) )
`
public static void Run([TimerTrigger("%CRON_EXPRESSION%")]TimerInfo myTimer, TraceWriter log)
`
Currently below code throws exception.

```csharp
[FunctionName("HttpStart")]
public static async Task<HttpResponseMessage> Run([HttpTrigger(AuthorizationLevel.Function, methods: "post", Route = "orchestrators/{functionName}")]
HttpRequestMessage req, [OrchestrationClient(TaskHub ="%MyHubName%")] DurableOrchestrationClient starter, string functionName, TraceWriter log)
{
	// Function input comes from the request content.
	dynamic eventData = await req.Content.ReadAsAsync<object>();
	string instanceId = await starter.StartNewAsync(functionName, eventData);

	log.Info($"Started orchestration with ID = '{instanceId}'.");

	var res = starter.CreateCheckStatusResponse(req, instanceId);
	res.Headers.RetryAfter = new RetryConditionHeaderValue(TimeSpan.FromSeconds(10));
	return res;
}
```

```
### Exception while executing function: HttpStart -> Exception binding parameter 'starter' -> Invalid queue name. Check MSDN for more information about valid queue naming.",
  "errorDetails": "Microsoft.Azure.WebJobs.Host.FunctionInvocationException : Exception while executing function: HttpStart ---> System.InvalidOperationException : Exception binding parameter 'starter' ---> System.ArgumentException : Invalid queue name. Check MSDN for more information about valid queue naming.
   at Microsoft.WindowsAzure.Storage.NameValidator.ValidateShareContainerQueueHelper(String resourceName,String resourceType)
   at Microsoft.WindowsAzure.Storage.NameValidator.ValidateQueueName(String queueName)
   at DurableTask.AzureStorage.AzureStorageOrchestrationService.GetQueueInternal(CloudQueueClient queueClient,String taskHub,String suffix)
   at DurableTask.AzureStorage.AzureStorageOrchestrationService.GetWorkItemQueue(CloudStorageAccount account,String taskHub)
[...]

PR #503, opened 2018-11-08
Adding support for configuring task hub name from application settings

gled4er:
Hello @cgillum,

I wanted to share my idea for implementing https://github.com/Azure/azure-functions-durable-extension/issues/499.

My idea is to configure it via helper `host.json` setting that will point to the application setting that holds the name of the task hub. 

I think this will be even better design because set once we will not need to decorate every function with the app setting name.

I decided to go this way because we use `DurableTaskOptions.HubName` is many places and even if we set the value from the attribute we end up with task hub with name coming from `DurableTaskOptions.HubName`. 

Let me know what you think about this approach and if it is not acceptable and I will continue to research the original one.

Thank you!


cgillum:
Thanks @gled4er for jumping on this so quickly! I was actually looking for a simpler solution, such as updating the value in the `DurableTaskExtension.Initialize()` method, around line 148 or so (before the call to `this.Options.Validate()`. Basically, something like this:

```csharp
if (this.nameResolver.TryResolveWholeString(this.Options.TaskHubName, out string taskHubName))
{
    // use the resolved task hub name
    this.Options.TaskHubName = taskHubName;
}
```

Does something like this seem possible?

gled4er:
Hello @cgillum,

Thank you very much for the suggestion.

I will try it and let you know.

Thank you!

gled4er:
Hello @cgillum,

Thank you for the great optimization!

After studying the code in depth it was clear that your approach is what we needed.

However, I have difficulties making it work just using your snippet so I need to change it a bit:

```csharp
string taskHubName = this.nameResolver.Resolve(this.Options.HubName);

if (!string.IsNullOrEmpty(taskHubName))
{
      this.Options.HubName = taskHubName;
}
```

During my testing `INameResolver.TryResolveWholeString` and `INameResolver.ResolveWholeString` weren't able to retrieve the value of the App Setting variable but continue to use the value that is set to `DurableTaskOptions.HubName`. 

Do you think that this is issue with my environment?

I saw both approaches using  `INameResolver.TryResolveWholeString`  and  `INameResolver.Resolve` methods in the Durable Functions code base.

Is there a documentation showing the differences between the two methods?

My setup is as follows:

`host.json:`

```javascript
{ 
"durableTask": {
      "HubName": "TaskHubName" }
}
```

`local.settings.json:`

```javascript
{
  "IsEncrypted": false,
  "Values": {
    "TaskHubName": "sampletaskhub"
  }
}
```

And when I use `INameResolver.TryResolveWholeString` or `INameResolver.ResolveWholeString`  the task hub name will be **TaskHubName** and when I use `INameResolver.Resolve`  the task hub name will be **sampletaskhub** when applying the code snipped I shared above.

Let me know if you have an advice how to get the expected output from `INameResolver.TryResolveWholeString`.

Thank you!

cgillum:
Based on your findings, I think `INameResolver.TryResolveWholeString` is probably correct because (I think) it allows you to use either an app setting or a hardcoded string. Here is an example setup:

**host.json:**

```javascript
{ 
"durableTask": {
      "HubName": "%TaskHubName%" }
}
```

**local.settings.json:**

```javascript
{
  "IsEncrypted": false,
  "Values": {
    "TaskHubName": "sampletaskhub"
  }
}
```

Basically, it's my understanding that `TryResolveWholeString` allows you to do variable expansion using the `%` characters.

gled4er:
Hello @cgillum,

Thank you for your guidance on this PR. 

It worked as you expected.

I updated the code.

When you have time, may I ask you to take a look?

Thank you!

gled4er:
Hello @cgillum,

Thank you for the great help on this PR!

cgillum:
One thing I just realized is that the original issue was asking about configuring the `[OrchestrationClient]` _binding_ with a task hub name from application settings (in other words, it's the value that goes into the attribute or to function.json). I think that is slightly different from what you implemented, which applies to the task hub name specified in host.json.

If you have cycles, would you be interested in doing another PR which targets the `[OrchestrationClient]` binding? It might be as simple as adding an attribute to the `TaskHub` property, but I haven't experimented with this feature much yet.

gled4er:
Hello @cgillum,

Thank you!

Yes, I will look into it.

Thank you!

gled4er:
Hello @cgillum,

I started to work on the implementation of the original issue for the `TaskHubName`. Based on my testing the creation of the Hub resources is happening immediately after we complete the `DutableTaskExtention.Initialize`. If that is true I think we will need to get the `TaskHub` property value of the `OrchestrationClientAttribute` in this method or otherwise it will be too late. However, I don't see currently a way to do this. Do you have some idea how we can approach it? 

Thank you!

cgillum:
I think the answer might be as simple as adding an `[AutoResolve]` attribute to the attribute property, like [here](https://github.com/Azure/azure-webjobs-sdk-extensions/blob/6f510a2f04b301f2966d84472a7d1ba7eda1047e/src/WebJobs.Extensions.CosmosDB/CosmosDBAttribute.cs#L49).

PR #513, opened 2018-11-13
Adding support for configuring Task Hub Name via OrchestrationClientAttribute

gled4er:
Hello @cgillum,

I wanted to open this PR and start the discussion for resolving https://github.com/Azure/azure-functions-durable-extension/issues/499.

Thank you for the guidance for the `AutoResolve` attribute. It works as expected and it is able to resolve the value of the app setting variable. However, when using it the orchestration gets stacked in the pending state. We don't get any entries to the History table and the initial control queue message stays unprocessed. I think the runtime for some reason is trying to pull from the original task hub (configured in `host.json`) control queues. My guess is it is because the `TaskHubWorker` is initialized with the original `TaskHubName` set in `host.json` in `DurableTaskExtension.Initialize` :

```csharp
this.orchestrationServiceSettings = this.GetOrchestrationServiceSettings();
            this.orchestrationService = new AzureStorageOrchestrationService(this.orchestrationServiceSettings);
            this.taskHubWorker = new TaskHubWorker(this.orchestrationService, this, this);
            this.taskHubWorker.AddOrchestrationDispatcherMiddleware(this.OrchestrationMiddleware);
```

And when we use the new `TaskHubName` coming from the attribute we only execute the following code in `DurableTaskExtension`:

```csharp
protected internal virtual DurableOrchestrationClient GetClient(OrchestrationClientAttribute attribute)
        {
            DurableOrchestrationClient client = this.cachedClients.GetOrAdd(
                attribute,
                attr =>
                {
                    AzureStorageOrchestrationServiceSettings settings = this.GetOrchestrationServiceSettings(attr);
                    AzureStorageOrchestrationService innerClient;
                    if (this.orchestrationServiceSettings != null &&
                        this.orchestrationService != null &&
                        string.Equals(this.orchestrationServiceSettings.TaskHubName, settings.TaskHubName, StringComparison.OrdinalIgnoreCase) &&
                        string.Equals(this.orchestrationServiceSettings.StorageConnectionString, settings.StorageConnectionString, StringComparison.OrdinalIgnoreCase))
                    {
                        // It's important that clients use the same AzureStorageOrchestrationService instance
                        // as the host when possible to ensure we any send operations can be picked up
                        // immediately instead of waiting for the next queue polling interval.
                        innerClient = this.orchestrationService;
                    }
                    else
                    {
                        innerClient = new AzureStorageOrchestrationService(settings);
                    }
                    return new DurableOrchestrationClient(innerClient, this, attr);
                });
            return client;
        }

```
But we continue to use the previous instances of the `TaskHubWorker`.

I tried re-initializing the `TaskHubWorker` in the `GetClient` method but it did not help.

May I ask you to confirm my current understanding?

Thank you!

### DIFF
--- a/test/Common/ClientFunctions.cs
+++ b/test/Common/ClientFunctions.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Threading.Tasks;
-using DurableTask.Core;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
 {
--- a/test/Common/ClientFunctions.cs
+++ b/test/Common/ClientFunctions.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Threading.Tasks;
+using DurableTask.Core;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
 {
@@ -11,6 +12,24 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
         [NoAutomaticTrigger]
         public static async Task StartFunction(
             [OrchestrationClient] DurableOrchestrationClient client,
+            string functionName,
+            string instanceId,
+            object input,
+            TestOrchestratorClient[] clientRef)
+        {
+            DateTime instanceCreationTime = DateTime.UtcNow;
+
+            instanceId = await client.StartNewAsync(functionName, instanceId, input);
+            clientRef[0] = new TestOrchestratorClient(
+                client,
+                functionName,
+                instanceId,
+                instanceCreationTime);
+        }
+
+        [NoAutomaticTrigger]
+        public static async Task StartFunctionWithTaskHub(
+            [OrchestrationClient(TaskHub = "testtaskhub")] DurableOrchestrationClient client,
             string functionName,
             string instanceId,
             object input,
--- a/test/Common/ClientFunctions.cs
+++ b/test/Common/ClientFunctions.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
 
         [NoAutomaticTrigger]
         public static async Task StartFunctionWithTaskHub(
-            [OrchestrationClient(TaskHub = "testtaskhubV2")] DurableOrchestrationClient client,
+            [OrchestrationClient(TaskHub = "%TaskHubName%")] DurableOrchestrationClient client,
             string functionName,
             string instanceId,
             object input,
--- a/test/Common/ClientFunctions.cs
+++ b/test/Common/ClientFunctions.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
 
         [NoAutomaticTrigger]
         public static async Task StartFunctionWithTaskHub(
-            [OrchestrationClient(TaskHub = "%TaskHubName%")] DurableOrchestrationClient client,
+            [OrchestrationClient(TaskHub = "%TestTaskHub%")] DurableOrchestrationClient client,
             string functionName,
             string instanceId,
             object input,
--- a/test/Common/ClientFunctions.cs
+++ b/test/Common/ClientFunctions.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
 
         [NoAutomaticTrigger]
         public static async Task StartFunctionWithTaskHub(
-            [OrchestrationClient(TaskHub = "testtaskhub")] DurableOrchestrationClient client,
+            [OrchestrationClient(TaskHub = "testtaskhubV2")] DurableOrchestrationClient client,
             string functionName,
             string instanceId,
             object input,
--- a/test/Common/DurableTaskLifeCycleNotificationTest.cs
+++ b/test/Common/DurableTaskLifeCycleNotificationTest.cs
@@ -933,7 +933,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
         private static Mock<INameResolver> GetNameResolverMock((string Key, string Value)[] settings)
         {
             var mock = new Mock<INameResolver>();
-            foreach (var setting in settings)
+            var settingList = settings.ToList();
+            settingList.Add(("TestTaskHub", string.Empty));
+            foreach (var setting in settingList)
             {
                 mock.Setup(x => x.Resolve(setting.Key)).Returns(setting.Value);
             }
--- a/test/Common/TestHelpers.cs
+++ b/test/Common/TestHelpers.cs
@@ -643,7 +643,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
     }
 
     /// <summary>
-    /// Helper name resolver class. 
+    /// Test implementation of INameResolver interface.
     /// </summary>
     public class SimpleNameResolver : INameResolver
     {
--- a/test/Common/TestHelpers.cs
+++ b/test/Common/TestHelpers.cs
@@ -642,6 +642,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
         }
     }
 
+    /// <summary>
+    /// Helper name resolver class. 
+    /// </summary>
     public class SimpleNameResolver : INameResolver
     {
         private readonly Dictionary<string, string> values = null;
--- a/test/Common/TestHelpers.cs
+++ b/test/Common/TestHelpers.cs
@@ -643,7 +643,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
     }
 
     /// <summary>
-    /// Helper implementation of INameResolver interface.
+    /// Test helper implementation of INameResolver interface.
     /// </summary>
     public class SimpleNameResolver : INameResolver
     {
--- a/test/Common/TestHelpers.cs
+++ b/test/Common/TestHelpers.cs
@@ -650,7 +650,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
         {
             this.values = new Dictionary<string, string>
             {
-                { "TestTaskHub", "" },
+                { "TestTaskHub", string.Empty },
             };
         }
 
--- a/test/Common/TestHelpers.cs
+++ b/test/Common/TestHelpers.cs
@@ -31,6 +31,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
             Uri notificationUrl = null,
             HttpMessageHandler eventGridNotificationHandler = null)
         {
+            if (nameResolver == null) { nameResolver = new SimpleNameResolver(); }
             var durableTaskOptions = new DurableTaskOptions
             {
                 HubName = GetTaskHubNameFromTestName(testName, enableExtendedSessions),
@@ -638,6 +639,34 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
 
                 return value;
             }
+        }
+    }
+
+    public class SimpleNameResolver : INameResolver
+    {
+        private readonly Dictionary<string, string> values = null;
+
+        public SimpleNameResolver()
+        {
+            this.values = new Dictionary<string, string>
+            {
+                { "TestTaskHub", "" },
+            };
+        }
+
+        public SimpleNameResolver(Dictionary<string, string> values)
+        {
+            this.values = values;
+        }
+
+        public string Resolve(string name)
+        {
+            if (this.values == null)
+            {
+                return null;
+            }
+
+            return !this.values.TryGetValue(name, out string result) ? null : result;
         }
     }
 }
--- a/test/Common/TestHelpers.cs
+++ b/test/Common/TestHelpers.cs
@@ -643,7 +643,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
     }
 
     /// <summary>
-    /// Test implementation of INameResolver interface.
+    /// Test helper implementation of INameResolver interface.
     /// </summary>
     public class SimpleNameResolver : INameResolver
     {
--- a/test/Common/TestHelpers.cs
+++ b/test/Common/TestHelpers.cs
@@ -643,7 +643,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
     }
 
     /// <summary>
-    /// Test helper implementation of INameResolver interface.
+    /// Helper implementation of INameResolver interface.
     /// </summary>
     public class SimpleNameResolver : INameResolver
     {
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
@@ -41,7 +41,7 @@ namespace Microsoft.Azure.WebJobs
 
             this.client = new TaskHubClient(serviceClient);
             this.traceHelper = config.TraceHelper;
-            this.hubName = config.Options.HubName;
+            this.hubName = attribute.TaskHub ?? config.Options.HubName;
             this.attribute = attribute;
         }
 
--- a/src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask.xml
+++ b/src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask.xml
@@ -22,6 +22,11 @@
             Configuration for the Durable Functions extension.
             </summary>
         </member>
+        <member name="M:Microsoft.Azure.WebJobs.Extensions.DurableTask.DurableTaskExtension.#ctor">
+            <summary>
+            Obsolete. Please use an alternate constructor overload.
+            </summary>
+        </member>
         <member name="M:Microsoft.Azure.WebJobs.Extensions.DurableTask.DurableTaskExtension.#ctor(Microsoft.Extensions.Options.IOptions{Microsoft.Azure.WebJobs.Extensions.DurableTask.DurableTaskOptions},Microsoft.Extensions.Logging.ILoggerFactory,Microsoft.Azure.WebJobs.INameResolver,Microsoft.Azure.WebJobs.Extensions.DurableTask.IConnectionStringResolver)">
             <summary>
             Initializes a new instance of the <see cref="T:Microsoft.Azure.WebJobs.Extensions.DurableTask.DurableTaskExtension"/>.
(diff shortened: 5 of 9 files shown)


cgillum:
Wow, it looks like you may have fixed an important bug in your adfb217 commit. Can you write a test which exercises this scenario to we can observe more closely what might be happening? I wonder if one of Yuka's tests might also be impacted.

gled4er:
Hello @cgillum,

Thank you for the great idea!

I just added a new test that will demonstrate the issues I found with the `TaskHubName` configuration. The test is called `HelloWorld_OrchestrationClientTaskHub` and it uses the `Hello World Inline` orchestration. But the actual orchestration is not so important. Let me clarify that all the changes in https://github.com/Azure/azure-functions-durable-extension/pull/513/commits/9a12535fc4c7ce2610e25a9ba7b44b9f1821b014 are temporary and just to explain better the problem I want to discuss. I plan to remove them later if the solution we implement does not need them. That said I will work on a following commit to check why other integration tests are failing as well. Let me explain more about https://github.com/Azure/azure-functions-durable-extension/pull/513/commits/9a12535fc4c7ce2610e25a9ba7b44b9f1821b014 

When we start the execution of the `HelloWorld_OrchestrationClientTaskHub` the test setup creates `TaskHub` from the test name with the help of `TestHelpers.GetJobHost`. In our case the name of the `TaskHub` will called **HelloWorldOrchestrationClientTaskHubV2** when run from the `V2` tests.  


```csharp
        public static JobHost GetJobHost(...)
        {
            var durableTaskOptions = new DurableTaskOptions
            {
                HubName = GetTaskHubNameFromTestName(testName, enableExtendedSessions),
                ....
            };
          ....
        }
```
Then we call the the updated version of the `StartOrchestratorAsync`  from the Test project. I added a new parameter that allow us to call modified `ClientFunctions` method as follows:

```csharp
public static async Task<TestOrchestratorClient> StartOrchestratorAsync(
            this JobHost host,
            string functionName,
            object input,
            ITestOutputHelper output,
            string instanceId = null,
            bool isTaskHubConfiguredViaOrchestrationClient = false)
        {
            var startFunction = isTaskHubConfiguredViaOrchestrationClient ?
                typeof(ClientFunctions).GetMethod(nameof(ClientFunctions.StartFunctionWithTaskHub)) :
                typeof(ClientFunctions).GetMethod(nameof(ClientFunctions.StartFunction));
            var clientRef = new TestOrchestratorClient[1];
            var args = new Dictionary<string, object>
            {
                { "functionName", functionName },
                { "instanceId", instanceId },
                { "input", input },
                { "clientRef", clientRef },
            };

            await host.CallAsync(startFunction, args);
            TestOrchestratorClient client = clientRef[0];
            output.WriteLine($"Started {functionName}, Instance ID = {client.InstanceId}");
            return client;
        }
```
The only difference is which `ClientFunction` method we call. For this sceanrio I want to call the new `StartFunctionWithTaskHub` from `ClientFunctions` class:

```csharp
        [NoAutomaticTrigger]
        public static async Task StartFunctionWithTaskHub(
            [OrchestrationClient(TaskHub = "testtaskhub")] DurableOrchestrationClient client,
            string functionName,
            string instanceId,
            object input,
            TestOrchestratorClient[] clientRef)
        {
            DateTime instanceCreationTime = DateTime.UtcNow;

            instanceId = await client.StartNewAsync(functionName, instanceId, input);
            clientRef[0] = new TestOrchestratorClient(
                client,
                functionName,
                instanceId,
                instanceCreationTime);
        }
```

In this method I hardcode the value for `TaskHubName` and avoid the resolving from `%{value}%` notation since we don't use settings file in the test project and it is not convenient to configure environment variable on the CI server we can use for testing. The setup with the exact value is good enough in my opinion for now.  

I created the `StartFunctionWithTaskHub` method because I did not find quick way to parametrize the `OrchestarionClientAttribute` that will allow me to use the existing `StartFunction` method. 

So when we call `StartFunctionWithTaskHub`, this will call `DurableTaskExtension.GetClient` method that will create new `OrchestrationCclient` since we don't have match for the `OrchestrationClientAttribute` value in the cache:

```csharp
protected internal virtual DurableOrchestrationClient GetClient(OrchestrationClientAttribute attribute)
        {
            DurableOrchestrationClient client = this.cachedClients.GetOrAdd(
                attribute,
                attr =>
                {
                    AzureStorageOrchestrationServiceSettings settings = this.GetOrchestrationServiceSettings(attr);

                    AzureStorageOrchestrationService innerClient;
                    if (this.orchestrationServiceSettings != null &&
                        this.orchestrationService != null &&
                        string.Equals(this.orchestrationServiceSettings.TaskHubName, settings.TaskHubName, StringComparison.OrdinalIgnoreCase) &&
                        string.Equals(this.orchestrationServiceSettings.StorageConnectionString, settings.StorageConnectionString, StringComparison.OrdinalIgnoreCase))
                    {
                        // It's important that clients use the same AzureStorageOrchestrationService instance
                        // as the host when possible to ensure we any send operations can be picked up
                        // immediately instead of waiting for the next queue polling interval.
                        innerClient = this.orchestrationService;
                    }
                    else
                    {
                        innerClient = new AzureStorageOrchestrationService(settings);
                    }

                    return new DurableOrchestrationClient(innerClient, this, attr);
                });

            return client;
        }
```
After that execution and maybe few more internal calls we create new `TaskHub`  with name **testtaskhub** we configured via the `OrchestrationClientAttribute`. So now we have two `TaskHub`-s - **HelloWorldOrchestrationClientTaskHubV2** and **testtaskhub**.

We proceed with the execution of our very simple inline orchestration and I can easily see that in one of the control queue - `testtaskhub-control-02` we have one message:

```javascript

{"$type":"DurableTask.AzureStorage.MessageData, DurableTask.AzureStorage","ActivityId":"37b1d4fc-867f-4f4f-92a8-6eb87aa040ab","TaskMessage":{"$type":"DurableTask.Core.TaskMessage, DurableTask.Core","Event":{"$type":"DurableTask.Core.History.ExecutionStartedEvent, DurableTask.Core","OrchestrationInstance":{"$type":"DurableTask.Core.OrchestrationInstance, DurableTask.Core","InstanceId":"f578f41b4eec44ea847e41c49d42b08d","ExecutionId":"613ddfb164544267a534a18b5746e21d"},"EventType":0,"ParentInstance":null,"Name":"SayHelloInline","Version":"","Input":"\"World\"","Tags":null,"EventId":-1,"IsPlayed":false,"Timestamp":"2018-11-14T06:11:07.0000855Z"},"SequenceNumber":0,"OrchestrationInstance":{"$type":"DurableTask.Core.OrchestrationInstance, DurableTask.Core","InstanceId":"f578f41b4eec44ea847e41c49d42b08d","ExecutionId":"613ddfb164544267a534a18b5746e21d"}},"CompressedBlobName":null,"SequenceNumber":1}

```
And also in the `testtaskhubInstnace` Table we have only one row staying in `Pending` state. We have no rows in `testtaskhubHistory` Table. And the test `HelloWorld_OrchestrationClientTaskHub` is failing with: **"Message: System.TimeoutException : Durable function 'SayHelloInline' with instance ID '12eaa76c4dea44dfb5dcdf3e648d473c' failed to complete."** after the timeout expires. All the resources for **HelloWorldOrchestrationClientTaskHubV2** TaskHub are empty.

My guess continue to be is that the **runtime is checking the control queue belonging to **HelloWorldOrchestrationClientTaskHubV2** and we are not able acutally to start the orchestration at all since the starting control message queue is in a control queue belonging to **testtaskhub**** `TaskHub`.

I hope this makes it a little bit more clear.

Please let me know if I am missing something.

I am very interested in your opinion!

Thank you!

gled4er:
Hello @cgillum,

I found the issue with the failing tests and I fixed it. It was bug in the unit test code.

Currently we have only the new `HelloWorld_OrchestrationClientTaskHub` test failing and this is expected based on the details in the previous comment.

Thank you!

gled4er:
Hello @cgillum,

I continued to debug the `HelloWorld_OrchestrationClientTaskHub` test and looking what is going on Durable Task project.

All the control queues that are monitored are related to the default `TaskHub` that is created based on the integration test name - **HelloWorldOrchestrationClientTaskHubV2.**

I checked [OrchestrationSessionManager.DequeueLoop](https://github.com/Azure/durabletask/blob/4aea0a0d5c8edc4ae1c02ef12529a656c731da7a/src/DurableTask.AzureStorage/OrchestrationSessionManager.cs#L92) and [AzureStorageOrchestrationService.OnPartitionAcquiredAsync](https://github.com/Azure/durabletask/blob/4aea0a0d5c8edc4ae1c02ef12529a656c731da7a/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs#L509). There are not any control queues related to the new `TaskHubName` we configure via the `OrchestrationClientAttribute` - **testtaskhub**.

Do you think it is possible that the assignment of the new `TaskHubName` is happening to late and if so do you think we can reconfigure `DurableTask` to work with the new `TaskHubName`?

Thank you!

cgillum:
@gled4er Everything you're describing sounds like expected behavior to me. If you configure a `DurableOrchestrationClient` with a non-default task hub name, it will try to start orchestrations in a different task hub from the one that the `JobHost` is listening to. The fact that you get the timeout is therefore correct behavior because there is no `JobHost` listening for messages on **testtaskhub**, thus nobody will actually run the orchestration.

The way to test properly this is to do something similar that Yuka did in the new tests she wrote: create two different `JobHost` objects - a default one for **HelloWorldOrchestrationClientTaskHubV2** and one that corresponds to your test task hub **testtaskhub**. The thing to confirm is that the correct task hub (**testtaskhub**) picks up and runs the orchestration, and that it is NOT picked up by **HelloWorldOrchestrationClientTaskHubV2**. Does that make sense?

> In this method I hardcode the value for TaskHubName in the OrchestrationClientAttribute to avoid the resolving from `%{value}%` notation since we don't use settings file in the test project and it is not convenient to configure environment variable on the CI server we use for running the tests. The setup with the exact value is good enough in my opinion for now.

You don't need to use environment variables for this. I think you can instead configure a test `INameResolver` implementation which contains the values of any app settings you want to inject (the default implementation of `INameResolver` looks for environment variables). For example, see you Yuka did this in `DurableTaskLifecycleNotificationTest.OrchestrationStartAndComplete`.

gled4er:
Hello @cgillum,

Thank you for your explanation and advice!

I will work on them.

Thank you!

gled4er:
Hello @cgillum,

I think I have now better understanding of the feature. **My first impression was when specifying `TaskHubName` via the `TaskHub` property of the `OrchestrationClientAttribute` we were aiming to create new task hub and configure the current host to use it instead of the default one configure via `host.json`.** After you gave me the idea to inspect some of the existing integration tests especially the one that starts two hosts I realized that **we should use the attribute's `TaskHub` property for selecting an existing and running host that will use the `TaskHub` we configured with the attribute**. Is my understanding correct? If so, what should be the behavior when we specify `TaskHub` name that does not have already host listening its control queues? In my opinion it is easy developer to get confused by configuring the `TaskHubName` via the attribute expecting that this configuration will override the `host.json` and set the current host to start working (listening its control and worker queue) with the new `TaskHub` instead with the default configured via `host.json`. If my understanding is not correct, how a single Durable Functions will start two hosts at the same time to serve the `TaskHub`-s? 

I have the core part of the test committed and passing. I need more work to fix the behavior of the `AutoResolve` attribute. Yuka was using the `INameResolver.Resolve` and I am not completely sure yet what `AutoResolve` is using behind the scenes. I will continue to investigate but in the meantime it will be great if you confirm my understanding from the previous paragraph.

Thank you!

cgillum:
Your understanding sounds correct, but I don't think it will be hard for people to understand. Let's consider HTTP as an example metaphor. Think of a task hub as a URL. The client is configured to send messages to a particular task hub (e.g. a URL). That task hub can be in the current function app (like `localhost` in HTTP), which is the default behavior, or it could be in a completely different function app (e.g. xyz.azurewebsites.net), in which case you specify a non-default `TaskHubName`. Those are the two scenarios.

But considering that it was confusing for you, we should consider explaining this more clearly in our documentation.

test/Common/ClientFunctions.cs
```diff
@@ -25,5 +26,23 @@ internal static class ClientFunctions
                 instanceId,
                 instanceCreationTime);
         }
+
+        [NoAutomaticTrigger]
+        public static async Task StartFunctionWithTaskHub(
+            [OrchestrationClient(TaskHub = "testtaskhubV2")] DurableOrchestrationClient client,
```
cgillum: If you want to specify a dynamic value for the task hub name, consider binding it to some input data or using a `Binder` object to do it imperatively in the code.

gled4er: Great!

I will work on it!

Thank you!

gled4er: Hello @cgillum,

I know that you are super busy with the release but if you have time it will be great to hear your opinion.

Do you know if there a sample I can check about passing dynamic value to the  `OrchestrtrationClientAttribute`?

I tried adding the `%{value}%` notation for the `TaskHub` in the `OrchestrtrationClientAttribute` after setting the mocked `INameResolver` as Yuka did but I don't think this approach works well with `AutoResolve` attribute. I am getting the following error when I use the `%{value}%` notation for the `TaskHub` in the `OrchestrtrationClientAttribute` :

**"Message: Microsoft.Azure.WebJobs.Host.Indexers.FunctionIndexingException : Error indexing method 'TestOrchestrations.SayHelloInlineTaskHubName'
---- System.InvalidOperationException : '%TaskHubName%' does not resolve to a value.
"**

Thank you!

gled4er: Hello @cgillum,

Thank you very much!

I will try it and let you know.

My approach was to use moq to mock the interface.

Thank you!

gled4er:
Hello @cgillum,

With your great help I think we are very close to solving this issue!

May I ask you to check once more the latest modifications and let me know if we need to change something.

Integrtion tests that did not have injected `INameResolver` were not able to resolve `%TestHubName%` and were failing on build time even though they did not use it in their execution path. That is why I initialized an instances of new class I defined `SimpleNameResolver` in `TestHelpers.GetJobHost` method to avoid this build failures:

```csharp
public static JobHost GetJobHost(
            ILoggerProvider loggerProvider,
            string testName,
            bool enableExtendedSessions,
            string eventGridKeySettingName = null,
            INameResolver nameResolver = null,
            string eventGridTopicEndpoint = null,
            int? eventGridRetryCount = null,
            TimeSpan? eventGridRetryInterval = null,
            int[] eventGridRetryHttpStatus = null,
            bool logReplayEvents = true,
            Uri notificationUrl = null,
            HttpMessageHandler eventGridNotificationHandler = null)
        {
            if (nameResolver == null) { nameResolver = new SimpleNameResolver(); }
            ...
         }

````

Where SimpleNameResolver is the newly added class:

```csharp
public class SimpleNameResolver : INameResolver
    {
        private readonly Dictionary<string, string> values = null;

        public SimpleNameResolver()
        {
            this.values = new Dictionary<string, string>
            {
                { "TestTaskHub", "" },
            };
        }

        public SimpleNameResolver(Dictionary<string, string> values)
        {
            this.values = values;
        }

        public string Resolve(string name)
        {
            if (this.values == null)
            {
                return null;
            }

            return !this.values.TryGetValue(name, out string result) ? null : result;
        }
    }
```
Let me know if this is an acceptable approach.

Thank you!

gled4er:
Hello @cgillum,

May I ask you to check the current state of the CI server?

I think it is not running at the moment.

All the tests were passing after my last main commit https://github.com/Azure/azure-functions-durable-extension/pull/513/commits/6cc24de6075356c5946f533fcaee8fb7abcb35ef. Then I decided to remove unused namespace in https://github.com/Azure/azure-functions-durable-extension/pull/513/commits/eb33117130f6feeb95a7d5710008de569f32d9cf an 3 test failed. All of them are passing for me locally.

I pushed another commit to trigger the CI again but this time 2 tests failed that are passing for me locally - https://github.com/Azure/azure-functions-durable-extension/pull/513/commits/179ce5e55689299dea01ec59260a04d354b77127. 

I added comment in the next commit - https://github.com/Azure/azure-functions-durable-extension/pull/513/commits/74eb353fbdc466675b66474205f275de2136d6c1 again to trigger the CI but from this moment the tests are staying in queued state. 

Let me know if you see something problematic that can disturb the normal execution of the failing tests.

Thank you!

gled4er:
Hello @cgillum,

The tests passed after slight delay.

I need to push one more commit and hopefully all the tests will pass again.

I will monitor and investigate otherwise.

Thank you!

gled4er:
Hello @cgillum,

After several runs I was able to get all the integration tests passing.

Please let me know what should be fixed next.

Thank you!

test/Common/DurableTaskLifeCycleNotificationTest.cs
```diff
@@ -933,7 +933,9 @@ public void OrchestrationEventGridApiConfigureCheck()
         private static Mock<INameResolver> GetNameResolverMock((string Key, string Value)[] settings)
         {
             var mock = new Mock<INameResolver>();
-            foreach (var setting in settings)
+            var settingList = settings.ToList();
+            settingList.Add(("TestTaskHub", string.Empty));
```
cgillum: Why do the lifecycle notification tests need to be updated like this? If it's required to make these tests pass, then I suspect there is some other bug that we're covering up with this change.

test/Common/TestHelpers.cs
```diff
@@ -640,4 +641,35 @@ public string Resolve(string name)
             }
         }
     }
+
+    /// <summary>
+    /// Test helper implementation of INameResolver interface.
+    /// </summary>
+    public class SimpleNameResolver : INameResolver
```
cgillum: Public classes should be placed in their own file. The only exception for this should be inner classes.

test/Common/TestHelpers.cs
```diff
@@ -31,6 +31,7 @@ internal static class TestHelpers
             Uri notificationUrl = null,
             HttpMessageHandler eventGridNotificationHandler = null)
         {
+            if (nameResolver == null) { nameResolver = new SimpleNameResolver(); }
```
cgillum: Why do we need to add the `SimpleNameResolver` to all end-to-end tests? We didn't require this before so I don't understand why your change would require it now.

test/Common/TestHelpers.cs
gled4er: Great point! I will be careful in future.

test/Common/DurableTaskLifeCycleNotificationTest.cs
gled4er: It was beacuase of the build issues I was facing from the` %{value}%` notation for test that does not use injected `INameResolver`.


Thank you for the comment!

test/Common/TestHelpers.cs
gled4er: Thank you for improving the code in the other PR!

gled4er:
Hello @cgillum,

Thank you for the support for this PR!

pksorensen:
I have the following function

```
  [FunctionName(nameof(MirrorStarter))]
        public async Task<HttpResponseMessage> MirrorStarter(
            [HttpTrigger(AuthorizationLevel.Anonymous, Route = "mirror/{instanceId}")]HttpRequestMessage req, string instanceId,
            [OrchestrationClient(TaskHub = "%MirrorHubName%")] DurableOrchestrationClient starter,
            ILogger log)
        {
            var orchestrationInput = await req.Content.ReadAsAsync<MirrorQueueFunctionInput>();

            var queue = radarContext.StorageAccount.CreateCloudQueueClient().GetQueueReference(orchestrationInput.Target);
            await queue.CreateIfNotExistsAsync();

            var existingInstance = await starter.GetStatusAsync(instanceId);
            if (existingInstance == null || existingInstance.RuntimeStatus == OrchestrationRuntimeStatus.Terminated)
            {
                await starter.StartNewAsync(nameof(MirrorQueueFunction), instanceId, orchestrationInput);
                log.LogInformation($"Started orchestration with ID = '{instanceId}'.");
                return starter.CreateCheckStatusResponse(req, instanceId);
            }
            else
            {
                await starter.PurgeInstanceHistoryAsync(instanceId);
                return req.CreateResponse("cleaned");
            }
        }
```

but it dont work on 1.7.1 - it has created both the default name and my custom name.

image

When starting the orchestration, it is stuck in pending mode.
```
{"instanceId":"rjbbr02m","runtimeStatus":"Pending","input":{"$type":"Ascend.Wammo.RadarMirrorFunctions.MirrorQueueFunctionInput, Ascend.Wammo.RadarMirrorFunctions","Source":"ascendradar2","Target":"rjbbr02m","SwipeTime":null,"LoopStart":"2018-11-01T00:00:00+00:00","LoopEnd":"2018-11-14T00:00:00+00:00","TargetPosition":[135.23278,34.43417],"LoopStartMonitor":null},"customStatus":null,"output":null,"createdTime":"2019-05-31T15:30:47Z","lastUpdatedTime":"2019-05-31T15:30:48Z"}
```

but i see the msg in the custom named queue. So my quess is that it listens on the default hub name put correctly put the msg in the custom named hub.

I need to test some more before knowing for sure.

pksorensen:
Same experience on 1.8.1. All orc stuck in pending.

cgillum:
@pksorensen I'm afraid I don't understand what the issue is that you're describing. Can you please open a separate GitHub issue and describe the problem there? That will be much better than asking questions on an old, closed PR.

pksorensen:
Ye, sorry. I will create one soon. Just need to get this out the door first (finding an different path to solve my problem atm).

In short, when using `[OrchestrationClient(TaskHub = "%MirrorHubName%")]` it outs messages on the queues that matches this custom hub name. But they never start, because it listens on the default durablehub name.

Issue #377, opened 2018-07-03
[Feature Request] Add NewGuid() to DurableOrchestrationContext in order to enable creating unique replay friendly values to provide to activities from the orchestrator directly


SimonLuckenuik:
Something similar to DurableOrchestrationContext.CurrentUtcDateTime, which would persist the GUID and return same value on replay. This would allow to create new GUID without calling an Activity to obtain it and persist it for replay, GUID can then be shared to multiple activities.

cgillum:
I like this idea. I think there are two ways we could do this:

1. Create some kind of "built-in" activity function which returns `Guid.NewGuid()`. This is simple, but involves a lot of Azure Storage overhead.
2. Generate deterministic GUIDs that are safe for replay ([example](https://github.com/LogosBible/Logos.Utility/blob/master/src/Logos.Utility/GuidUtility.cs)).

Any thoughts/concerns/preferences?

SimonLuckenuik:
What I had in mind is something like 2. Today, I could create an activity to generate and persist my GUID, but it does involve too much overhead for a simple GUID generation.

gled4er:
Hello @cgillum,

I would like to help with this feature.

Where will be the best place for the logic in the Durable Functions or Durable Task code base?

Thank you!

gled4er:
@cgillum thank you for the clarification!

gled4er:
Hello @cgillum,

I think we should add information for this feature in the documentation page. Shall we continue working from the release-functions-durable-migration branch?

Thank you!

cgillum:
Yes, that is the right branch to target.

gled4er:
Thank you for the confirmation!

I will work on it.

Thank you!

gled4er:
Hello @cgillum,

What documentation page will you advice me to use for the `NewGuid` method?

I checked all the pages and I don't see one dedicated to the `DurableOcrchestrationContext` that we can use for adding more info for `NewGuid` method.

Thank you!

cgillum:
I think the pages which talk about orchestrator function limitations around random data. We should call out that the workaround for the "no random GUIDs" rule is to use your new API, just like how we mention how to get the current DateTime.

PR #514, opened 2018-11-13
Adding support for deterministic GUID

gled4er:
Hello @cgillum,

May I ask you to review the code in this PR that is addressing the issue for generating deterministic `Guid` from the `DurableOrchestrationContext` as described here https://github.com/Azure/azure-functions-durable-extension/issues/377.

Thank you!

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
@@ -6,6 +6,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using DurableTask.Core;
@@ -25,6 +26,7 @@ namespace Microsoft.Azure.WebJobs
         private const string DefaultVersion = "";
         private const int MaxTimerDurationInDays = 6;
 
+        private int newGuidCounter = 0;
         private readonly Dictionary<string, Stack> pendingExternalEvents =
             new Dictionary<string, Stack>(StringComparer.OrdinalIgnoreCase);
 
@@ -103,7 +105,10 @@ namespace Microsoft.Azure.WebJobs
         /// <inheritdoc />
         public override Guid NewGuid()
         {
-            return GuidManager.CreateDeterministicGuid(GuidManager.UrlNamespaceValue, this.InstanceId);
+            this.newGuidCounter++;
+            StringBuilder newGuidNameStringBuilder = new StringBuilder(this.InstanceId);
+            newGuidNameStringBuilder.Append(this.newGuidCounter);
+            return GuidManager.CreateDeterministicGuid(GuidManager.UrlNamespaceValue, newGuidNameStringBuilder.ToString());
         }
 
         internal void SetInput(string rawInput)
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
@@ -105,10 +105,9 @@ namespace Microsoft.Azure.WebJobs
         /// <inheritdoc />
         public override Guid NewGuid()
         {
+            string guidNameValue = this.InstanceId + this.newGuidCounter.ToString();
             this.newGuidCounter++;
-            StringBuilder newGuidNameStringBuilder = new StringBuilder(this.InstanceId);
-            newGuidNameStringBuilder.Append(this.newGuidCounter);
-            return GuidManager.CreateDeterministicGuid(GuidManager.UrlNamespaceValue, newGuidNameStringBuilder.ToString());
+            return GuidManager.CreateDeterministicGuid(GuidManager.UrlNamespaceValue, guidNameValue);
         }
 
         internal void SetInput(string rawInput)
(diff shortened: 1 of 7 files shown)


SimonLuckenuik:
Hi @gled4er,

I just had a quick look at your code and I believe that the following behavior is not what I would expect as a user of the API:
```
            Guid firstGuid = ctx.NewGuid();
            string output = await ctx.CallActivityAsync<string>(nameof(TestActivities.Hello), input);
            Guid secondGuid = ctx.NewGuid();
            return firstGuid == secondGuid;
```
And in your test you check for `Assert.Equal("True", status.Output.ToString());`. 
I would expect firstGuid and secondGuid to be different, not equal.

The goal of NewGuid would be to enable us to create multiple GUIDs prior passing them to different activities. For example, I could create 10 GUIDs, do a fan-out and provide a different GUID to each activity part of the fan-out. The name also implies that it's creating a new GUID at each call, like `Guid.NewGuid()`, which is not the case in your unit test.

Simon

gled4er:
Hello Simon,

Will you mind passing a name that,can be any string to the NewGuid method? If this is possible the behavior will be fixed easily. I can make the name optional parameter but when provided will generate different Guid-s per different names.

Let me know if this is acceptable for you.

Thank you!

cgillum:
Just like `Guid.NewGuid()` doesn't take a parameter, `ctx.NewGuid()` also shouldn't have a parameter.

Would it be possible to simply include a monotonically incrementing counter variable as part of the name? It seems you could increment this counter variable every time `NewGuid` gets called.

gled4er:
Hello @cgillum,

Great idea!

I will modify the code.

Thank you!

gled4er:
Hello @SimonLuckenuik  @cgillum,

I modified the code according our discussion.

May I ask you to take a look and let me know what should be fixed next?

Thank you!

gled4er:
Hello @SimonLuckenuik,

I was wondering if you had time to review the changes after your last comment.

I will be happy to hear if we are getting closer to support your scenario.

Thank you!

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -100,6 +102,15 @@ public override T GetInput<T>()
             return MessagePayloadDataConverter.Default.Deserialize<T>(this.serializedInput);
         }
 
+        /// <inheritdoc />
+        public override Guid NewGuid()
+        {
+            this.newGuidCounter++;
+            StringBuilder newGuidNameStringBuilder = new StringBuilder(this.InstanceId);
+            newGuidNameStringBuilder.Append(this.newGuidCounter);
```
cgillum: Using a `StringBuilder` to concatenate two strings is actually more overhead than doing string concatenation directly. I suggest simply:
```csharp
string guidNameValue = this.InstanceId + this.newGuidCounter.ToString();
this.newGuidCounter++;
```

test/Common/TestOrchestrations.cs
```diff
@@ -28,6 +28,16 @@ public static async Task<string> SayHelloWithActivity([OrchestrationTrigger] Dur
...
         }
 
+        public static async Task<bool> SayHelloWithActivityWithDeterministicGuid([OrchestrationTrigger] DurableOrchestrationContext ctx)
+        {
+            string input = ctx.GetInput<string>();
+            Guid firstGuid = ctx.NewGuid();
+            Guid secondGuid = ctx.NewGuid();
+            string output = await ctx.CallActivityAsync<string>(nameof(TestActivities.Hello), input);
+            Guid thirdGuid = ctx.NewGuid();
+            return firstGuid != secondGuid && firstGuid != thirdGuid && secondGuid != thirdGuid;
```
cgillum: This is a good basic test. Consider two more possible tests:
1. A test that ensures if you create 10K+ GUIDs in your orchestration they are all unique.
2. A test that ensures the same GUID values are generated during replays.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
gled4er: Hello @cgillum,

Great!

I will modify it!

Thank you!

test/Common/TestOrchestrations.cs
gled4er: Hello @cgillum,

I will add the tests.

Thank you!

gled4er:
Hello @cgillum,

I worked on the requested changes during the last review iteration.

May I ask you to take another look when you have time?

Thank you!

test/Common/TestOrchestrations.cs
```diff
@@ -28,6 +28,55 @@ public static async Task<string> SayHelloWithActivity([OrchestrationTrigger] Dur
...
+        }
+
+        public static bool VerifyUniqueGuids([OrchestrationTrigger] DurableOrchestrationContext ctx)
+        {
+            List<Guid> guids = new List<Guid>(10000);
```
cgillum: A `HashSet<Guid>` would be much more natural and efficient than `List<Guid>` for a test like this.

gled4er:
Hello @cgillum,

Thank you very much for the support on this PR!

Thank you!

test/Common/TestOrchestrations.cs
gled4er: Great!

I will change it!

Thank you!

PR #517, opened 2018-11-14
Feature/511 waitforexternalevent nongeneric

### LINKED ISSUE #511: WaitForExternalEvent has no overloaded version with no type
**Summary**
A durable function eventName callback url expects application/json.  The fact that that is forced makes can lead to some confusion since context.WaitForExternalEvent<T>.  That function should be overloaded not to take a type.  Since if you specify <string> and pass no body, it doesn't get called, with no exception.   It appears to be attempting to coerce json to whatever type is given, but that appears to silently fail with no logging or anything and the post to the callback uri returns 202 which is expected.​

**Problem Statement**: Currently, it is not possible for partners to create a Durable Function that uses "WaitForExternalEvent" without giving a type. The partner wants to be able to be able to use this method to wait for an event with no body.

**Desired Experience**: Durable Functions should provide a version of the "WaitForExternalEvent" method without giving a type.

brandonh-msft:
Resolves #511 .

It's worth noting that the existing API does make this change a bit clunky due to the use of `object` for `eventData` and a default value for `connectionName` however all V2 Unit Tests pass.

It's also important to note that I wasn't able to fully run the V1 unit tests, so I'm relying on the CI to execute these. 🤞

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,9 +34,10 @@ The general flow for making a change to the script host is:
 ## Running the tests (Visual Studio) 
 
 1. Build the project and Visual Studio will identify all the tests in the solution.
-2. Set an environment variable named **AzureWebJobsStorage** set to a connection string e.g. to the storage emulator (`UseDevelopmentStorage=true`).
-3. Run Azure Storage Emulator 5.6.
-4. Run the unit tests via Visual Studio Test Explorer by selecting "Run All"
+2. Set an environment variable named **AzureWebJobsStorage** set to an Azure General Purpose Storage Account connection string.
+> Note: While it is possible to use the local storage emulator (`UseDevelopmentStorage=true`), this is not recommended as performance and reliability are severely impacted while running unit tests; using a real storage account in Azure will yield much better results.
+
+3. Run the unit tests via Visual Studio Test Explorer by selecting "Run All"
 
 
 ## Testing code changes locally (Visual Studio) 
--- a/test/Common/DurableTaskEndToEndTests.cs
+++ b/test/Common/DurableTaskEndToEndTests.cs
@@ -644,17 +644,17 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
                 var client = await host.StartOrchestratorAsync(nameof(TestOrchestrations.BatchActorRemoveLast), null, this.output);
 
                 // Perform some operations
-                await client.RaiseEventAsync("deleteItem", null); // deletes last item in the list: item5
-                await client.RaiseEventAsync("deleteItem", null); // deletes last item in the list: item4
-                await client.RaiseEventAsync("deleteItem", null); // deletes last item in the list: item3
-                await client.RaiseEventAsync("deleteItem", null); // deletes last item in the list: item2
+                await client.RaiseEventAsync("deleteItem"); // deletes last item in the list: item5
+                await client.RaiseEventAsync("deleteItem"); // deletes last item in the list: item4
+                await client.RaiseEventAsync("deleteItem"); // deletes last item in the list: item3
+                await client.RaiseEventAsync("deleteItem"); // deletes last item in the list: item2
 
                 // Make sure it's still running and didn't complete early (or fail).
                 var status = await client.WaitForStartupAsync(TimeSpan.FromSeconds(10), this.output);
                 Assert.Equal(OrchestrationRuntimeStatus.Running, status?.RuntimeStatus);
 
                 // Sending this last event will cause the actor to complete itself.
-                await client.RaiseEventAsync("deleteItem", null); // deletes last item in the list: item1
+                await client.RaiseEventAsync("deleteItem"); // deletes last item in the list: item1
 
                 status = await client.WaitForCompletionAsync(TimeSpan.FromSeconds(10), this.output);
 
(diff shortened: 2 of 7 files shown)


test/Common/TestOrchestratorClient.cs
```diff
@@ -54,11 +54,15 @@ public async Task<DurableOrchestrationStatus> GetStatusAsync(bool showHistory =
...
         }
 
+        public async Task RaiseEventAsync(string eventName) => await innerClient.RaiseEventAsync(instanceId: instanceId, eventName: eventName);
+
         public async Task RaiseEventAsync(string eventName, object eventData)
         {
             await this.innerClient.RaiseEventAsync(this.instanceId, eventName, eventData);
         }
 
+        public async Task RaiseEventAsync(string taskHubName, string instanceid, string eventName, string connectionName = null) => await innerClient.RaiseEventAsync(taskHubName: taskHubName, instanceId: instanceid, eventName: eventName);
```
cgillum: Would this be a breaking change? For example, if customers previously had code like the following:
```csharp
client.RaiseEventAsync(instanceId, eventName, "string event data");
```
Will that now conflict with the new overload? Your use of parameter names scares me. If so, we should consider reducing the scope of your change - i.e. delete the new `RaiseEventAsync(string, string, string, string = null)`

CONTRIBUTING.md
```diff
@@ -34,9 +34,10 @@ The general flow for making a change to the script host is:
 ## Running the tests (Visual Studio) 
 
 1. Build the project and Visual Studio will identify all the tests in the solution.
-2. Set an environment variable named **AzureWebJobsStorage** set to a connection string e.g. to the storage emulator (`UseDevelopmentStorage=true`).
-3. Run Azure Storage Emulator 5.6.
-4. Run the unit tests via Visual Studio Test Explorer by selecting "Run All"
+1. Set an environment variable named **AzureWebJobsStorage** set to an Azure General Purpose Storage Account connection string.
```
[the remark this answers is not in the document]
brandonh-msft: Actually markdown's smart like this :) It'll auto-sequence an ordered list while the numbers are the same. [check out the output](https://github.com/Azure/azure-functions-durable-extension/blob/205585f9d5355b560bebf977964bc4c7f6ba5752/CONTRIBUTING.md)

test/Common/TestOrchestratorClient.cs
brandonh-msft: Yeah like i put in the PR comment, one has to now get extra-specific when they want to use either of these methods in order to disambiguate between what we have today and the proposed change here.

I think you might be right that the requirement for people to do this is indeed a breaking change. Shoot.

brandonh-msft:
I won't be able to get to the requested changes until Friday at the earliest. Standby!

brandonh-msft:
⏲ 
@cgillum I noticed that the AppVeyor CI doesn't have any test output... is it running the UTs?

gled4er:
A special label full-ci needs to be added by @cgillum to run the integration tests.

cgillum:
Done. The CI has restarted and you should start seeing test output.

brandonh-msft:
hmmm... investigating failure now. doesn't seem to be in any code related to what I changed, but I'll check it out.

brandonh-msft:
rebased on top of `/dev` and am letting it run again. was all green on my machine 🤷‍♂️

CONTRIBUTING.md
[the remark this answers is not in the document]
cgillum: But you have a `3` below this. Isn't that kind of inconsistent?

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -569,6 +569,45 @@ public async Task BatchedActorOrchestration(bool extendedSessions)
...
+            {
+                await host.StartAsync();
+
+                var client = await host.StartOrchestratorAsync(nameof(TestOrchestrations.BatchActorRemoveLast), null, this.output);
+
+                // Perform some operations
+                await client.RaiseEventAsync("deleteItem", null); // deletes last item in the list: item5
+                await client.RaiseEventAsync("deleteItem", null); // deletes last item in the list: item4
+                await client.RaiseEventAsync("deleteItem", null); // deletes last item in the list: item3
+                await client.RaiseEventAsync("deleteItem", null); // deletes last item in the list: item2
```
cgillum: Why pass `null` for the eventData? I think you should add a `RaiseEventAsync` method to the test client which calls the new API you're introducing so that we have proper code coverage.

brandonh-msft: The only new API we're introducing now is `WaitForExternalEvent()` (non-generic). We removed the `RaiseEventAsync()` that didn't take `eventData` because that would've been a breaking change due to the default parameter signature already in place. So, this usage is actually consistent with how it'll be used in the wild and _does_ hit the new `WaitForExternalEvent()` I've added. So we've got coverage on the change... unless you're seeing something I'm not?

brandonh-msft:
ha, not sure why the CI failed here; all I pushed was a change to CONTRIBUTING.md

test/Common/DurableTaskEndToEndTests.cs
brandonh-msft: ah, gotchya. I'll revisit and push a new ver in a few.

cgillum:
CI is still a bit flaky. I have a couple PRs outstanding that I think will improve the reliability. I'm going to focus on those right now and then re-trigger this PR.

Issue #526, opened 2018-11-22
Support for source-linking in Durable Functions extension


cgillum:
The Azure Functions core DLLs support this already. The Durable Functions extension should support this as well.

More info: https://github.com/dotnet/sourcelink/blob/master/README.md

cgillum:
Ideally we would at it to all three assemblies used by Durable Functions, but the priority order would be:

1. Microsoft.Azure.WebJobs.Extensions.DurableTask
2. DurableTask.AzureStorage
3. DurableTask.Core

gled4er:
Great!

Thank you for the clarification!

PR #528, opened 2018-11-22
Adding support for source-link

gled4er:
Hello @cgillum,

May I ask you to take a look on this first commit for adding support for source-link as described in https://github.com/Azure/azure-functions-durable-extension/issues/526?

Thank you!

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -50,9 +50,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private readonly AsyncLock taskHubLock = new AsyncLock();
 
-        private readonly INameResolver nameResolver;
         private readonly bool isOptionsConfigured;
 
+        private INameResolver nameResolver;
         private AzureStorageOrchestrationService orchestrationService;
         private AzureStorageOrchestrationServiceSettings orchestrationServiceSettings;
         private TaskHubWorker taskHubWorker;
@@ -147,8 +147,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 this.InitializeForFunctionsV1(context);
             }
 
-            if (this.nameResolver != null &&
-                this.nameResolver.TryResolveWholeString(this.Options.HubName, out string taskHubName))
+            if (this.nameResolver.TryResolveWholeString(this.Options.HubName, out string taskHubName))
             {
                 // use the resolved task hub name
                 this.Options.HubName = taskHubName;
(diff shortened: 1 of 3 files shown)


cgillum:
Is this the only change that is required? Would you be able to somehow demonstrate that it is working as expected, maybe with screenshots? I'm not actually familiar with how to set it up, but I assumed there was more to it.

gled4er:
Hello @cgillum,

Thank you for the comment.

Great idea about the testing!

I will continue to work on it and update you about the progress.

Thank you!

gled4er:
Hello @cgillum,

I was able to achieve the expected result with integrating source-link.

Please check the following images:









May I ask you to review the changes and let me know what should be fixed next?

Thank you!

gled4er:
Hello @cgillum,

Yes, it makes perfect sense.

The best way to test it against .Net Framework is to modify the host.json or you suggest other option?

Thank you!

cgillum:
I think all you need to do is create a function app which targets Functions 1.0, clear your nuget cache, and make sure that customers can still manually set breakpoints and break into Durable Functions code. If source-link works, that's great to, but not required.

I think targeting 1.0 is as simple as updating the .csproj file (you may also need to make updates to host.json). When you reference your local build of the nuget package, it will target the right framework automatically.

cgillum:
According to this Twitter thread, it might be expected to work on all versions of .NET: https://twitter.com/kirillosenkov/status/1007052524946255872?lang=en

I'll try to test it as well.

gled4er:
Great!

Thank you!

I will try it soon.

Thank you!

gled4er:
Hello @cgillum,

I was able to test and verify that source-link works for .NET Framework as well as you can see from the images below:









In the last commit you will see a small change in `DurableTaskExtensions.Initialize` method. The problems is that `this.nameResolver` is null when we execute against **.Net Framework** and the host is not able to start. The regression was introduced from the change we made for setting `TaskHub` name from the `local.settings.json` in https://github.com/Azure/azure-functions-durable-extension/pull/503/files#diff-8e71038a95b35db07c2b23ecbc95b728R150 And the fix I added is:

```csharp
            if (this.nameResolver != null &&
                this.nameResolver.TryResolveWholeString(this.Options.HubName, out string taskHubName))
            {
                // use the resolved task hub name
                this.Options.HubName = taskHubName;
            }
```
 I was able to detect the error partially thanks to the new source-link feature.

Let me know if you prefer to make the correction in the way presented above or you prefer to have 

```csharp
#if NETSTANDARD2_0
           if (this.nameResolver.TryResolveWholeString(this.Options.HubName, out string taskHubName))
            {
                // use the resolved task hub name
                this.Options.HubName = taskHubName;
            }
#endif
```

Thank you!

cgillum:
Thanks Kanio for making that additional fix. It's good that we did some manual testing to detect this regression.

gled4er:
Hello @cgillum,

Thank you  for the quick response that allowed us add the correct fix immediately!

Thank you!

Issue #516, opened 2018-11-14
Non-Deterministic workflow detected: TaskScheduledEvent: 0 TaskScheduled CopyIModelToProject


SpicySyntax:
I have recently run into an issue with the "Non-Deterministic workflow detected" message while using Durable functions with Node.js + Typescript. (Note: I am on a consumption plan). The workflow that leads to this exception involves a few components:
- A Http Endpoint function 'SampleIModelProjects' with GET, POST, PATCH, & DELETE
- A Http Orchestration Starter (Starts Orchestration Client -> Orchestration Trigger)
- An Orchestrator (Activated by Orchestration Trigger -> Activity Trigger)
- An Activity Function 'CopyIModelToProject' (Long running, Memory/CPU intensive, with async I/O)(Activated by Activity Trigger) (Usually takes 1-5 minutes)
- An Http Orchestration Status Endpoint

When POST is called on the 'SampleIModelProjects' it calls a few APIs, posts to start the Http Orchestration Starter, and returns some data regarding the created project and the status URI for the Orchestration that was started. This works fine in most cases when the load scales slowly, but when the load scales to quickly (lets say 5 workflows in 10 seconds) is when I get the exception "Non-Deterministic workflow detected" and the CopyIModelToProject fails. See below for more details, any suggestions are welcome.

**This is what the Status endpoint then returns with 500 when this happens:**
Error: StatusCodeError: 500 - {"instanceId":"2ed8b845-b2fc-42c4-9a54-f27c46e4eae3","runtimeStatus":"Failed","input":{"targetProjectName":"IMODEL_DEV_SERVICES_LOAD_TEST","srcProjectName":"Test Project","srcIModelName":"Test IModel","userEmails":["Regular.IModelJsTestUser@mailinator.com"],"targetProjectId":"f7c35bdd-eb24-4450-8273-e2594162c1e5"},"customStatus":null,"output":"Non-Deterministic workflow detected: TaskScheduledEvent: 0 TaskScheduled CopyIModelToProject","createdTime":"2018-11-14T17:45:51Z","lastUpdatedTime":"2018-11-14T17:48:11Z"}

**This is what is logged in App Insights:**
```
Microsoft.Azure.WebJobs.Host.FunctionInvocationException: Exception while executing function: Functions.CopyIModelToProject ---> Microsoft.Azure.WebJobs.Script.Rpc.RpcException: Result: Failure
Exception: Error Copying IModel
Stack: 
[...]

Issue #527, opened 2018-11-22
Bug: No error if activity not found..


ackava:
I made mistake in naming my function, let's say function was "F2_Process",

```javascript
const df = require("durable-functions");

module.exports = df.orchestrator(function*(context){

    const sites = yield context.df.callActivity("F1_GetConstructionSites", "ny");

    const tasks = [];

    if (sites.length === 0) {
        console.log(`No sites to process`);
    }

    for (const site of sites) {
        console.log(`Queued: ${site.url}`);
        tasks.push(context.df.callActivity("F1_Process", JSON.stringify(site)));
    }

    yield context.df.Task.all(tasks);

    console.log("Success");

    return 0;
});
```

Notice that in my `callActivity` to process site, I wrongly named function name as "F1_Process" instead of "F2_Process", I never received any error saying that no such function exist, instead it just hung up and I never see "Success". After investigating Live Metrics Stream in Application Insights, I found that call to start orchestration function completed with result as null;

Error to reproduce is to simply make a call to activity which is not present and see what happens.

cgillum:
This is likely the same root cause as https://github.com/Azure/azure-functions-durable-extension/issues/516, which is that orchestrator failures and not being properly handled in JavaScript. We're working on a fix for this.

PR #530, opened 2018-11-27
Added out-of-proc orchestration state error field/handling.

kashimiz:
Enables resolution of #516.


cgillum:
Also, do you know whether this fix would address https://github.com/Azure/azure-functions-durable-extension/issues/527?

kashimiz:
Re: #527, sadly not on its own, no. I haven't determined exactly what's going on but it looks like the [exception](https://github.com/Azure/azure-functions-durable-extension/blob/c6d3ce6dd427f2c1b9659d0ef6776922da062f44/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs#L489) thrown by the `GetIfFunctionDoesNotExist` check is getting swallowed somewhere.

PR #538, opened 2018-12-05
Updated JavaScript samples for durable-functions 1.1.2 release:

kashimiz:
* Updated activity functions to use preferred async pattern.
* Added HttpStartSync client function.
* Updated HttpStart client function to use DurableOrchestrationClient API.


kashimiz:
Adding @mhoeger to verify samples follow best practices for JavaScript Azure Functions all-up.

samples/javascript/E1_HelloSequence/function.json
```diff
@@ -7,4 +7,4 @@
     }
   ],
   "disabled": false
```
mhoeger: The `disabled` property is only for V1, could you remove references to it?

samples/javascript/E2_CopyFileToBlob/index.js
```diff
@@ -2,7 +2,7 @@ const fs = require("fs");
 const path = require("path");
 const storage = require("azure-storage");
 
-module.exports = function (context, filePath) {
+module.exports = async function (context, filePath) {
     const container = "backups";
     const root = path.parse(filePath).root;
     const blobPath = filePath
```
mhoeger: Out of scope, but probably would suggest making const blobService in the scope of this file and not of the function, want to persist the blob service for as long as possible (and not create on each function invocation)

samples/javascript/E2_GetFileList/index.js
```diff
@@ -1,6 +1,6 @@
 const readdirp = require("readdirp");
```
mhoeger: looks like readdirp is also sync only :(

samples/javascript/E1_HelloSequence/function.json
kashimiz: Done!

samples/javascript/E2_CopyFileToBlob/index.js
```diff
@@ -2,7 +2,7 @@ const fs = require("fs");
 const path = require("path");
 const storage = require("azure-storage");
 
-module.exports = function (context, filePath) {
+module.exports = async function (context, filePath) {
```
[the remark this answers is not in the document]
kashimiz: Oh shoot, that's right! Thank you for catching/reminding me of that. Reversing autopilot.

samples/javascript/E2_GetFileList/index.js
kashimiz: 😞 This is why we can't have nice syntax.

samples/javascript/E3_GetIsClear/index.js
```diff
@@ -2,15 +2,15 @@ const request = require("request");
 
 const clearWeatherConditions = ['Overcast', 'Clear', 'Partly Cloudy', 'Mostly Cloudy', 'Scattered Clouds'];
 
-module.exports = function (context, location) {
+module.exports = async function (context, location) {
     getCurrentConditions(location)
```
mhoeger:
```
try {
 const isClear = await getCurrentConditions(location);
 return isClear;
} catch (err) {
 context.log(error);
 throw err;
}
```

samples/javascript/E3_GetIsClear/index.js
```diff
@@ -34,4 +34,4 @@ function getCurrentConditions(location) {
             resolve(body.current_observation);
         });
     });
-}
```
mhoeger: you could use [request-promise-native](https://github.com/request/request-promise-native) (it looks like there are a [few flavors too](https://www.npmjs.com/package/request#promises--asyncawait)) to make your getCurrentConditions an async/await too!

samples/javascript/E2_CopyFileToBlob/index.js
kashimiz: Not out of scope; good reminder!

samples/javascript/E3_GetIsClear/index.js
kashimiz: Ooh, nice!

samples/javascript/E2_CopyFileToBlob/function.json
```diff
@@ -12,6 +12,5 @@
       "connection": "AzureWebJobsStorage",
```
mhoeger: Other people might have different opinions, but I don't think that we should encourage people to use the same storage account as the function app (AzureWebJobsStorage). It's easier because it comes pre-provisioned but I think it's an easy way for users to shoot themselves in the foot (accidentally delete function execution-related objects in storage or do so many things in the account that they get throttled).

samples/javascript/E3_GetIsClear/index.js
```diff
@@ -1,37 +1,32 @@
...
             if (body.error) {
                 reject(body.error);
-            }
-            if (body.response.error) {
+            } else if (body.response && body.response.error) {
                 reject(body.response.error);
+            } else {
+                resolve(body.current_observation);
             }
-            resolve(body.current_observation);
```
mhoeger: You could make this async too, I generally think it's a bit cleaner!

Something like (although would recommend testing):
```javascript
let body = await request(options);
if (body.error) {
    throw body.error;
} else if (body.response && body.response.error) {
    throw body.response.error;
} else {
    return body.current_observation;
}
```

samples/javascript/E1_SayHello/function.json
```diff
@@ -5,6 +5,5 @@
       "type": "activityTrigger",
       "direction": "in"
     }
-  ],
-  "disabled": false
```
cgillum: Does this mean we don't advertise `"disabled": false` in function.json anymore?  Not a problem, I'm just curious. :)

samples/javascript/E2_CopyFileToBlob/function.json
cgillum: I actually don't even really like this sample since it doesn't make a lot of sense for someone to do this anyways (backup your function content to blobs). We should consider creating a different sample for fan-out/fan-in - not for this PR, but at some later point.

samples/javascript/E4_SendSmsChallenge/index.js
```diff
@@ -1,17 +1,19 @@
...
     const challengeCode = Math.floor(rand() * 10000);
 
     context.log(`Sending verification code ${challengeCode} to ${phoneNumber}.`);
 
+    /*
     context.bindings.message = {
         body: `Your verification code is ${challengeCode.toPrecision(4)}`,
         to: phoneNumber
     };
+    */
```
cgillum: I assume this is commented out temporarily, and that this change shouldn't be pushed out to the samples?

kashimiz: Whoops! That was for local testing. Thanks for catching that.

samples/javascript/E4_SendSmsChallenge/index.js
```diff
@@ -1,17 +1,19 @@
 const seedrandom = require("seedrandom");
 const uuidv1 = require("uuid/v1");
 
-module.exports = function (context, phoneNumber) {
+module.exports = async function (context, phoneNumber) {
     // Get a random number generator with a random seed (not time-based)
     const rand = seedrandom(uuidv1());
```
mhoeger: Could `const rand = seedrandom(uuidv1());` be pulled outside of the function scope? not sure if that generates bad behavior, but if possible may be (slightly, this is a nit) better

kashimiz: This should be fine. Thanks for catching that.

samples/javascript/E1_SayHello/function.json
cgillum: @mhoeger oh wow, I did not know that! Thanks for the update. :)

samples/javascript/E3_GetIsClear/index.js
kashimiz: Tested, and it works.

PR #547, opened 2018-12-16
Correlation with Application Insights

TsuyoshiUshio:
Hi @cgillum 

I open this PR for discussing and reviewing the initial design of distributed tracing implementation.
I'm not implementing the test classes. Once the design is OK. I'll create these.
Let's start discussing the design.

This PR should be releaseed the same as 
https://github.com/Azure/durabletask/pull/239

Also, We need to wait the  web jobs host side fix is released.  I'll also create a PR for azure-webjobs-sdk. It will be based on this branch.
https://github.com/TsuyoshiUshio/azure-webjobs-sdk/tree/correlation/plana

Currently, I refer local csproj since I need to orchestrate three repos. Once finish the discussion and change, I'll switch back to the nuget package then test it.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
@@ -93,6 +93,10 @@ namespace Microsoft.Azure.WebJobs
 
             // Correlation 
             var current = Activity.Current;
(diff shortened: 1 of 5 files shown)


TsuyoshiUshio:
Since I implement the new version the separate the responsibility between DurableTask and Extension, I close this.

PR #556, opened 2018-12-25
Http Method Fix

daronyondem:
The article here suggests copy-pasting the URL into a browser. To make that work the method allowed needs to be a GET instead of a POST. https://docs.microsoft.com/en-us/azure/azure-functions/durable/quickstart-js-vscode


cgillum:
Closing due to inactivity.

PR #562, opened 2019-01-04
Orchestrator should throw exception when activity fails

### LINKED ISSUE #411: Status "Succeeded" on Application Insights when orchestrator failed
If the Activity fails, runtimeStatus will be Failed, but on Application Insights it will succeed.

**Status query result**

```json
{
  "instanceId": "5ad6d773dc3a42a1b74dc89bc5f1097a",
  "runtimeStatus": "Failed",
  "input": {},
  "customStatus": null,
  "output": "***",
  "createdTime": "2018-08-27T07:25:33Z",
  "lastUpdatedTime": "2018-08-27T07:25:34Z"
}
```

**Application Insights**

image

**Runtime**

- Azure Functions v2
- Microsoft.Azure.WebJobs.Extensions.DurableTask = 1.5.0

glennamanns:
Fixes #411 

In the case where an activity function fails, that execution of the orchestrator function should also fail. The Functions host determines an execution's success/failure from the presence of an exception. The orchestration function now throws an `OrchestrationFailureException` that the host can pick up. 

For each activity function, the orchestrator logs have two entries: one for the scheduling of the activity, and one for the execution. Only the log representing the execution of the activity will show a failure.
image

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -11,6 +11,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using DurableTask.AzureStorage;
 using DurableTask.Core;
+using DurableTask.Core.Exceptions;
 using DurableTask.Core.Middleware;
 using Microsoft.Azure.WebJobs.Description;
 using Microsoft.Azure.WebJobs.Host;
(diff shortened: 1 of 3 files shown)


src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -11,6 +11,7 @@
 using System.Threading.Tasks;
 using DurableTask.AzureStorage;
 using DurableTask.Core;
+using DurableTask.Core.Exceptions;
```
glennamanns: extra; I'll remove this in the next iteration/commit

glennamanns:
@cgillum I changed the OrchestrationFailureException to include the full exception details instead of just the exception message. This means the test for that particular exception is going to fail. I can either change it such that the details are only included for this particular code path, rather than both, or update the test

cgillum:
@glennamanns Ah, I missed that when looking at the latest iteration. As you mentioned in person, the call stack isn't actually relevant since we throw that exception outside of user code - so `e.Message` is probably better. I would revert it back to that.

PR #563, opened 2019-01-05
Added bug issue template

kashimiz:
Added issue template for reporting bugs.

PR #564, opened 2019-01-07
Corrected terminology in bug report issue template.

kashimiz:
Fixed per #563 comment.


cgillum:
I'm going to ignore the test failure since this change obviously wouldn't regress any tests.

Issue #573, opened 2019-01-15
Possible race condition causing events to be dropped


olitomlinson:
**Describe the bug**
I can't reliably reproduce this but I feel there is a race condition when awaiting on an activity, and then awaiting on an external event.

```cs
var submitToRouteResult = await context.CallActivityAsync<SubmitToRouteResult>("submit_to_route", @object);
 
// other orchestration code here (no activities)

var result = await context.WaitForExternalEvent<CoreDispatchAttemptResult>($"ChannelSubmissionResult_{route.Name}");
```

I have a scenario where the 'External Event' is being dropped due to the prior Activity not completing.

When I observe the History I see the following events in this order:

PartitionKey | RowKey | Timestamp | EventId | EventType | ExecutionId | IsPlayed | _Timestamp
-- | -- | -- | -- | -- | -- | -- | --
67d8b855-bd13-4cb5-b937-c330eb88b767 | 0000000000000000 | 2019-01-15T14:46:32.856Z | -1 | OrchestratorStarted | 0ccf5cd3f5ee4c84aaf93db34479e430 | false | 2019-01-15T14:46:32.640Z
67d8b855-bd13-4cb5-b937-c330eb88b767 | 0000000000000001 | 2019-01-15T14:46:32.857Z | -1 | ExecutionStarted | 0ccf5cd3f5ee4c84aaf93db34479e430 | true | 2019-01-15T14:41:16.686Z
67d8b855-bd13-4cb5-b937-c330eb88b767 | 0000000000000002 | 2019-01-15T14:46:32.857Z | 0 | TaskScheduled | 0ccf5cd3f5ee4c84aaf93db34479e430 | false | 2019-01-15T14:46:32.642Z
67d8b855-bd13-4cb5-b937-c330eb88b767 | 0000000000000003 | 2019-01-15T14:46:32.857Z | -1 | OrchestratorCompleted | 0ccf5cd3f5ee4c84aaf93db34479e430 | false | 2019-01-15T14:46:32.642Z
67d8b855-bd13-4cb5-b937-c330eb88b767 | 0000000000000004 | 2019-01-15T14:46:39.576Z | -1 | OrchestratorStarted | 0ccf5cd3f5ee4c84aaf93db34479e430 | false | 2019-01-15T14:46:39.551Z
[...]

Issue #378, opened 2018-07-03
[Enhancement] Enhance external event usecases to support classic scenarios


SimonLuckenuik:
As explained in the warning [here](https://docs.microsoft.com/en-us/azure/azure-functions/durable-functions-external-events#send-events): _"If there is no orchestration instance with the specified instance ID or if the instance is not waiting on the specified event name, the event message is discarded."_

Considering that limitation:
1) It becomes impossible to handle orchestrations with variable (undefined) number of events (because a new event could be received between two WaitForExternalEvent calls)
2) Unless ordering is possible with the trigger mechanism that raised the external events, all expected events need to be WaitForExternalEvent at the very beginning to make sure that none will be discarded.

Suggestions:
A) Considering that Events are **the** way to interact with an executing orchestration, I would suggest to enhance the event behavior to be more flexible. Currently, with these limitations, I cannot for example notify an orchestration each time a message is received on a queue, unless I know upfront the number of messages that will be received.
B) Maybe adding a new concept like an Orchestration Queue would be acceptable?

To display how easy it is to lose and event, consider the following code, only first event will be processed. I guess if the orchestration is not executed right away, I would probably lose the first event as well.

Versions:
- Microsoft.Azure.WebJobs.Extensions.DurableTask: 1.5.0
- Microsoft.NET.Sdk.Functions: 1.0.14

```
using System.Collections.Generic;
using System.Linq;
using System;
using System.Net;
using System.Net.Http;
using System.Threading.Tasks;
using Microsoft.Azure.WebJobs;
using Microsoft.Azure.WebJobs.Extensions.Http;
using Microsoft.Azure.WebJobs.Host;
using Microsoft.Extensions.Logging;

namespace FunctionAppStartAndEvent
{
    public static class OrchestratorFunction
    {
        [FunctionName(nameof(OrchestratorFunction))]
        public static async Task<DateTime> Run([OrchestrationTrigger] DurableOrchestrationContextBase context, ILogger logger)
        {
            var eventCount = 0;
            do
            {
                var newEvent = await context.WaitForExternalEvent<MyEvent>(nameof(MyEvent));
                if(!context.IsReplaying) logger.LogInformation($"** Event {eventCount} received. Message: {newEvent.Message}");
                if (newEvent.Message == "stop")
                {
                    break;
                }
                await context.CallActivityAsync<string>(nameof(LogActivityFunction), newEvent.Message);
            } while (true);
            return context.CurrentUtcDateTime;
[...]

PR #578, opened 2019-01-18
Save dropped events

glennamanns:
**Problem:** If no orchestrator is listening for a specific event at a given time, then that event will be discarded. This causes dropped messages, even when there is an orchestrator listening for that event at around the same time (due to the race conditions seen in #573). This PR should address the main issue brought up in #378.

**Solution:** Messages that would have been dropped before are now added to an (in-memory) queue. When `WaitForExternalEvent(name)` is called, it checks its queues for an event associated with `name.` If found, `RaiseEvent` is called on it immediately.

**Notes**: This is my first stab at solving this problem - I'm not sure if it fully solves issue 573, but it does make the 10 activities/events execute in the proper order in 378. 

I also need to think about how to cleanup the queue. If someone calls `RaiseEvent` with an event name that an orchestrator will never listen for, all of those messages may stick around forever in the `DurableOrchestrationContext`'s `bufferedExternalEvents` object. Presumably, we do want to drop some messages (given that we had logic in place to do so), so discerning between messages we will eventually want and messages we will never want is something I should think about.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
@@ -301,20 +301,27 @@ namespace Microsoft.Azure.WebJobs
                     }
                 }
 
-                this.config.TraceHelper.FunctionListening(
-                    this.config.Options.HubName,
-                    this.orchestrationName,
-                    this.InstanceId,
-                    reason: $"WaitForExternalEvent:{name}",
-                    isReplay: this.innerContext.IsReplaying);
-
                 // Check the queue to see if any events came in before the orchestrator was listening
                 if (this.bufferedExternalEvents.TryGetValue(name, out Queue queue))
                 {
                     object input = queue.Dequeue();
 
+                    if (queue.Count == 0)
+                    {
+                        this.bufferedExternalEvents.Remove(name);
+                    }
+
                     // We can call raise event right away, since we already have an event's input
                     this.RaiseEvent(name, input.ToString());
+                }
+                else
+                {
+                    this.config.TraceHelper.FunctionListening(
+                    this.config.Options.HubName,
+                    this.orchestrationName,
+                    this.InstanceId,
+                    reason: $"WaitForExternalEvent:{name}",
+                    isReplay: this.innerContext.IsReplaying);
                 }
 
                 return tcs.Task;
@@ -548,6 +555,13 @@ namespace Microsoft.Azure.WebJobs
                     }
 
                     bufferedEvents.Enqueue(input);
+
+                    this.config.TraceHelper.ExternalEventSaved(
+                        this.HubName,
+                        this.Name,
+                        this.InstanceId,
+                        name,
+                        this.IsReplaying);
                 }
             }
         }
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
@@ -279,26 +279,6 @@ namespace Microsoft.Azure.WebJobs
                 Stack taskCompletionSources;
                 TaskCompletionSource<T> tcs;
 
-                // Check the queue to see if any events came in while an orchestrator wasn't listening
-                Queue queue;
-                if (this.bufferedExternalEvents.TryGetValue(name, out queue))
-                {
-                    object input = queue.Dequeue();
-
-                    tcs = new TaskCompletionSource<T>();
-                    Type buffType = tcs.GetType();
-                    Type genericTypeArgument = buffType.GetGenericArguments()[0];
-
-                    if (queue.Count == 0)
-                    {
-                        this.bufferedExternalEvents.Remove(name);
-                    }
-
-                    object deserialized = MessagePayloadDataConverter.Default.Deserialize(input.ToString(), genericTypeArgument);
-                    MethodInfo trySetResult = buffType.GetMethod("TrySetResult");
-                    trySetResult.Invoke(tcs, new[] { deserialized });
-                }
-
                 // Set up the stack for listening to external events
                 if (!this.pendingExternalEvents.TryGetValue(name, out taskCompletionSources))
                 {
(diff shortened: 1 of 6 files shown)


cgillum:
Thanks Glenna! One thing I think we should do is write a test which makes sure your fix is working correctly. Basically, start an orchestration, define it do a for-loop where it waits for an event and then calls an activity function, and then raise 10 events to it at the same time. We should verify that all 10 of those events get processed successfully.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -304,6 +308,15 @@ public override Task<T> WaitForExternalEvent<T>(string name)
                     reason: $"WaitForExternalEvent:{name}",
                     isReplay: this.innerContext.IsReplaying);
 
+                // Check the queue to see if any events came in before the orchestrator was listening
+                if (this.bufferedExternalEvents.TryGetValue(name, out Queue queue))
+                {
+                    object input = queue.Dequeue();
+
+                    // We can call raise event right away, since we already have an event's input
+                    this.RaiseEvent(name, input.ToString());
```
cgillum: If we're able to raise the event right away, then we probably don't need to trace the `FunctionListening` event above.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -527,13 +540,14 @@ internal void RaiseEvent(string name, string input)
                 }
                 else
                 {
-                    // The orchestrator was not waiting for any event by this name, so the event will be dropped.
-                    this.config.TraceHelper.ExternalEventDropped(
```
cgillum: Yay! Since we're not dropping events anymore, can we repurpose this trace to say that an event has been buffered?

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -304,6 +308,15 @@ public override Task<T> WaitForExternalEvent<T>(string name)
                     reason: $"WaitForExternalEvent:{name}",
                     isReplay: this.innerContext.IsReplaying);
 
+                // Check the queue to see if any events came in before the orchestrator was listening
+                if (this.bufferedExternalEvents.TryGetValue(name, out Queue queue))
+                {
+                    object input = queue.Dequeue();
```
cgillum: If the queue is empty after dequeuing, I wonder if we should consider removing it from the `bufferedExternalEvents` dictionary. My worry is that someone could write an eternal orchestration that listens for dynamically named events, and their instance could run out of memory. It might not be very realistic, but I don't think it would hurt us to clean these up.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -301,21 +301,28 @@ public override Task<T> WaitForExternalEvent<T>(string name)
...
                     this.RaiseEvent(name, input.ToString());
                 }
+                else
+                {
+                    this.config.TraceHelper.FunctionListening(
+                    this.config.Options.HubName,
+                    this.orchestrationName,
+                    this.InstanceId,
+                    reason: $"WaitForExternalEvent:{name}",
+                    isReplay: this.innerContext.IsReplaying);
```
cgillum: nit: indent parameters.

src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
```diff
@@ -240,7 +240,7 @@ private EtwEventSource()
         }
 
         [Event(215, Level = EventLevel.Warning)]
```
cgillum: Let's change this to an informational trace and increase the `Version` number:

```cs
[Event(215, Level = EventLevel.Information, Version = 2)]
```

cgillum:
I just now realized that your PR is going against **master**. Can you point it to **dev** instead? We generally only touch **master** when we're doing releases.

glennamanns:
@cgillum Whoops, did not notice it was targeting master. Thanks for catching that. :)

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -365,7 +365,7 @@ public void ExtensionWarningEvent(string hubName, string functionName, string in
             if (this.ShouldLogEvent(isReplay: false))
             {
                 this.logger.LogInformation(
-                    "{instanceId}: Function '{functionName} ({functionType})' dropped a '{eventName}' event. State: {state}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
+                    "{instanceId}: Function '{functionName} ({functionType})' saved a '{eventName}' event to an in-memory queue. State: {state}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
                     instanceId, functionName, functionType, eventName, FunctionState.ExternalEventDropped, hubName,
```
nytian, line 369: @cgillum Chris, I am wondering why we have the FunctionState named as `ExeternalEventDropped`. 

I am asking because I had GH issue and also IcM tickets regrading this. Customers thought the external event was dropped/ not processed. But actually, it's a normal behavior and the event was saved to the in-memory buffer.  Just the naming is misleading.  Also, considering this is the only scenario that this FunctionState will be called. Do you think it would be better for us to update the FunctionState as `ExeternalEventSaved`?

nytian: Thanks for the reply, @cgillum !

What I want to update is [here](https://github.com/Azure/azure-functions-durable-extension/blob/5453308c9394b7f33633912802a36c8c0e577809/src/WebJobs.Extensions.DurableTask/FunctionState.cs#L16) to "ExternalEventSaved". We did rename the TaskName as "ExternalEventSaved", which is seen at Kusto logs. Customers can still see function state as "ExternalEventDropped" like this [post](https://github.com/Azure/azure-functions-durable-extension/issues/2482#issuecomment-1584178881). 

I have a ticket again today about this, so I am asking if we should update the FunctionState enum item name. Let me know if any is unclear to you. Thank you!

PR #614, opened 2019-02-21
Out-of-proc handling: don't complete Execute until orchestration completes.

kashimiz:
Bugfix: TaskOrchestrationShim.Execute calls for out-of-procedure orchestrators will not complete unless the orchestration is complete. This fixes the case of waiting on multiple waitForExternalEvent calls to complete. Currently the orchestrator is marked Completed after the first event is received.

The Durable Task Framework currently does something similar, by abandoning an orchestration's Task if the orchestration is not completed.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
@@ -213,6 +213,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             {
                 this.Context.SetOutput(execution.Output);
             }
+            else
+            {
+                // Don't return executions unless the orchestrator has completed.
+                await Task.Delay(-1);
+            }
         }
 
         private async Task ProcessAsyncActions(AsyncAction[][] actions)
--- a/src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask.xml
+++ b/src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask.xml
@@ -305,6 +305,12 @@
             Boolean value specifying if the replay events should be logged.
             </value>
         </member>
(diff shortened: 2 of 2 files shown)


src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -213,6 +213,10 @@ private async Task HandleOutOfProcExecution(JObject executionResult)
             {
                 this.Context.SetOutput(execution.Output);
             }
+            else
+            {
+                await Task.Delay(-1);
```
cgillum: Let's add a comment here explaining why we're doing this, lest we forget in the future.

src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask.xml
```diff
@@ -22,11 +22,6 @@
             Configuration for the Durable Functions extension.
             </summary>
         </member>
-        <member name="M:Microsoft.Azure.WebJobs.Extensions.DurableTask.DurableTaskExtension.#ctor">
```
cgillum: Let's remove this file from the PR. I think it randomly changes based on the order of the build and I don't want to clutter the history with random changes to this file.

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
kashimiz: Done.

src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask.xml
kashimiz: Removed.

cgillum:
Erm...next time please wait until the CI has actually completed before merging. :)

kashimiz:
Oops. Over-excitement.

Issue #602, opened 2019-02-05
Introduce Custom LifeCycleNotification interface


shibayan:
I want to add custom status to ETW logging.

In the current ETW logging, since I can not identify at which timing within the orchestrator it failed, I want to make it possible to identify by using custom status before executing the activity.

Thanks

cgillum:
Can you provide a few more details?

* Are you asking about tracing [this custom orchestration status](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-custom-orchestration-status) data or something else?
* If it's the custom orchestration status payload, do you need the full payload (up to 16 KB) to be written to the trace?
* Does it need to be ETW or can it just be Application Insights? Azure App Service consumes the ETW events internally, and I would prefer not put potentially sensitive customer data there.

cgillum:
Ah, I see. You wrote an ETW listener to help debug orchestrator problems, but need more information to be included in the ETW events. Is that correct?

If you just need more information, one thing we do have is a `traceInputsAndOutputs` setting in [host.json](https://docs.microsoft.com/azure/azure-functions/durable/durable-functions-bindings#host-json). If you set this to `true` then you can see the input and output data for each function included in the Application Insights traces.

Otherwise, the only thing we have for custom error processing is our [Event Grid notifications](https://docs.microsoft.com/azure/azure-functions/durable/durable-functions-event-publishing), which you mentioned already in your blog post.

shibayan:
>need more information to be included in the ETW events. Is that correct?
Yes!

>traceInputsAndOutputs
I did not know this option. I will try it.

>EventGrid notifications
I want to pluggable notification processing to EventGrid. (introduce Interface?)

Thanks

shibayan:
The extension of LifeCycleNotificationHelper seemed to be good, I wrote a prototype.

https://github.com/Azure/azure-functions-durable-extension/compare/dev...shibayan:lifecycle-pluggable

shibayan:
@cgillum I modified issue topic. Relative issue #613

Thanks!

cgillum:
I just now looked at your prototype. This seems really nice. If you convert it into a PR, I would be happy to review it and we can make it part of the next feature release.

shibayan:
I will create a pull request. Thank you!

shibayan:
It was released in v1.8, it closes. Thank you.

PR #621, opened 2019-02-23
Adding custom LifeCycle notification helper

shibayan:
Change LifeCycle notification helper to extensible. #602 

- Introduced `ILifeCycleNotificationHelper` and `NullLifeCycleNotificationHelper`
- Renamed `LifeCycleNotificationHelper` to `EventGridLifeCycleNotificationHelper`
- Added `CustomLifeCycleNotificationHelperType` option
- Added test cases

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
@@ -204,6 +204,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// </value>
         public bool LogReplayEvents { get; set; }
 
+        public string CustomLifeCycleNotificationHelperType { get; set; }
+
         // Used for mocking the lifecycle notification helper.
         internal HttpMessageHandler NotificationHandler { get; set; }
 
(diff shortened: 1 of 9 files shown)


src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
```diff
@@ -204,6 +204,8 @@ public class DurableTaskOptions
         /// </value>
         public bool LogReplayEvents { get; set; }
 
+        public string CustomLifeCycleNotificationHelperType { get; set; }
```
cgillum: Please add documentation comments. Maybe something like "Gets or sets the type name of a custom `<see cref="ILifeCycleNotificationHelper"> to use for handling lifecycle notification events."

shibayan: Thank you for review. Added document comments to public apis.

shibayan:
The method name (`Orchestrator**Async`) and `FunctionType` were redundant, I deleted it from the interface methods and put `FunctionType` enum back  to internal.
(Even in EventGrid helper, `FunctionType` was not used and only `Orchestrator` was passed.)

cgillum:
I really like these new changes.

I turned on the CI and one test failed:

* ConfigurationWihtoutEventGridKeySettingName

Can you take a look? If we can fix that, then I will merge this PR.

shibayan:
Thank you for review!

I fixed broken unit test and passed CI.

Issue #361, opened 2018-06-22
FunctionFailedException.InnerException is always of type System.Exception


cgillum:
### Issue
When an activity function or a sub-orchestration throws an exception, it propogates up to the parent orchestrator function as a `FunctionFailedException`. It contains an `InnerException` value which has the same message as the original exception, but the type information is not preserved. Rather, it is always a `System.Exception`.

### Expected
Ideally, the original exception is preserved.

### Root cause
The problem is with the serialization settings between when the exception is observed by the activity function executor (type information is written) and the parent orchestrator (type information is ignored). The fix may require changes to the Durable Task Framework.

armadak:
Any update on this issue?
In my opinion this is vital for proper error handling for activity functions, especially when using CallActivityWithRetryAsync

cgillum:
@armadak Still trying to figure out the right way to fix this. It turns out the serialization of exceptions in .NET is quite tricky for a variety of different reasons.

andrew-vandenbrink:
```
"{\"ExceptionType\":\"Microsoft.Azure.WebJobs.FunctionFailedException\",\"InnerExceptionType\":\"DurableTask.Core.Exceptions.TaskFailedExceptionDeserializationException\"}"
```
TaskFailedExceptionDeserializationException
Instead of trying to serialize/deserialize the whole exception object, 

Can you not just throw the exception where it contains 

```
public class ActivityException : Exception{

public string ActivityName{get;private set;}

public string ExceptionTypeName{get;private set;}
public string ExceptionTypeMessage{get;private set;}

public string InnerExceptionTypeName{get;private set;}
public string InnerExceptionTypeMessage{get;private set;}
}
```

I need to have at least ActivityName or FunctionName where the exception being thrown

shibayan:
@cgillum I tried simple solution. Adding TypeNameHandling to exception serializer settings.
Exception objects always inherit `Exception` class, they are serialized to base class even if they do not exist.
https://github.com/Azure/azure-functions-durable-extension/compare/dev...shibayan:exception-serializer-setting

```csharp
try
{
    await context.CallActivityWithRetryAsync<string>("Function1_Hello", new RetryOptions(TimeSpan.FromSeconds(5), 1)
    {
        Handle = ex =>
        {
            // It's work
            return ex.InnerException is InvalidOperationException;
        }
    }, null);
}
catch (FunctionFailedException ex) when (ex.InnerException is InvalidOperationException)
{
[...]

PR #622, opened 2019-02-26
Adding TypeNameHandling to Exception serializer settings

shibayan:
Keep actual `Exception` type in serialization. #361

#### Discussion needed

- Check if the exception was defined in C# script
  - Use type namespace (ex. `Submission#0.TestException`) or assembly location (`Assembly.Location` is empty from C# script)
  - Switch `TypeNameHandling.None` for exceptions defined in C# script


cgillum:
With this change, do you know what will happen if someone throws an exception in C# script?

shibayan:
`JsonSerializationException` is thrown in the current PR. Maybe in C# script it will be impossible to keep the exception type.

When using C# script, it seems to be a parent-child relationship of `FunctionInvocationException` => `TargetInvocationException` => (Actual thrown exception).

cgillum:
Maybe we could consider catching that `JsonSerializationException` and returning a more generic exception. Maybe that can be a separate PR.

It looks like the `OrchestratorThrowWithRetry` test failed. However, the failure is because you improved the error message, which is good! Can you update the test so that it passes?

Issue #618, opened 2019-02-22
Configurable queue polling delays


cgillum:
### Background
There have been several issues opened relating to queue polling in Durable Functions. They fall under two general categories:

* Wanting longer polling delays to mitigate cost
* Wanting shorter polling delays to mitigate end-to-end latency

Of the two, cost mitigation is the more common ask, but both are very much valid.

There are three sources of polling in the end-to-end system:

1. Individual control queue polling (there is one polling loop *per partition*)
2. Work item queue polling
3. Scale Controller polling in the Azure Functions Consumption and Premium plans

### Issue
The Durable Functions extension does not currently provide any mechanism to configure the queue polling delays, which means that customers have no control. This work item tracks allowing configuration for the above delays so that both cost-savings and end-to-end latency can be controlled by customers.

**Related:**
* https://github.com/Azure/azure-functions-durable-extension/issues/606
* https://github.com/Azure/azure-functions-durable-extension/issues/391

cgillum:
There are at least two places needed to make this work:

#### Durable Task Framework - Azure Storage Provider
Currently a hardcoded polling interval is used [here](https://github.com/Azure/durabletask/blob/7938ceb46c784adbc1c3a69b388506d3d81ad9f3/src/DurableTask.AzureStorage/Messaging/TaskHubQueue.cs#L52). Instead of being hardcoded, it should come from [AzureStorageOrchestrationServiceSettings](https://github.com/Azure/durabletask/blob/master/src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs). The new property can be named the same as the existing constant: `MaxQueuePollingDelay`. The constant can then be moved into `AzureStorageOrchestrationServiceSettings` and used as the default value.

#### Durable Task Extension for Azure Functions
A new property needs to be added to `DurableTaskExtension` which controls the polling behavior. By simply adding a new public property, it automatically becomes configurable via `host.json`. The value of this new property needs to be assigned to the corresponding property in the `AzureStorageOrchestrationServiceSettings` object [here](https://github.com/Azure/azure-functions-durable-extension/blob/f4b8f4575f792fb1049b14e0da910f0512d1cd85/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs#L451)

cgillum:
Resolved with [v1.8.0](https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Extensions.DurableTask/1.8.0) release.

PR #627, opened 2019-02-28
Allow to configure max polling interval via host.json

myarotskaya:
Makes the maximum queue polling interval configurable via host.json.

Second step to resolving [Azure/azure-functions-durable-extension#618](https://github.com/Azure/azure-functions-durable-extension/issues/618).

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
@@ -210,6 +210,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// <value>Assembly qualified class name that implements <see cref="ILifeCycleNotificationHelper">ILifeCycleNotificationHelper</see>.</value>
         public string CustomLifeCycleNotificationHelperType { get; set; }
 
+        /// <summary>
+        /// Gets or sets the maximum queue polling interval.
+        /// </summary>
+        /// <value>Maximum interval for polling control and work-item queues.</value>
+        public TimeSpan MaxQueuePollingInterval { get; set; } = TimeSpan.FromSeconds(30);
+
         // Used for mocking the lifecycle notification helper.
         internal HttpMessageHandler NotificationHandler { get; set; }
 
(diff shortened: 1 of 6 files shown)


src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
```diff
@@ -210,6 +210,12 @@ public class DurableTaskOptions
         /// <value>Assembly qualified class name that implements <see cref="ILifeCycleNotificationHelper">ILifeCycleNotificationHelper</see>.</value>
         public string CustomLifeCycleNotificationHelperType { get; set; }
 
+        /// <summary>
+        /// Gets or sets the maximum queue polling interval.
+        /// </summary>
+        /// <value>Maximum interval for polling control and work-item queues.</value>
+        public TimeSpan MaxQueuePollingInterval { get; set; } = TimeSpan.FromSeconds(30);
```
myarotskaya: @cgillum I think it should come from here https://github.com/Azure/durabletask/blob/azure-storage/src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs#L28
What do you think?

cgillum: I think it's fine to keep this the way it is.

cgillum: Let's also add this property value to the `GetDebugString()` method further below. This is important for diagnostics,

myarotskaya: Done!

myarotskaya:
@cgillum I noticed that `DurableTaskEndToEndTests.ActorOrchestration` test is failing when `extendedSessions` parameter is true. It looks like the reason is in the new version of `Microsoft.Azure.DurableTask.AzureStorage` package (Job host started to generate log messages with `IsReplay: True` while according to test it shouldn't). What would you suggest to do?

cgillum:
Yes I noticed this as well. It's a bug in the latest version of DurableTask.Core that I'm currently investigating. My hope is to resolve this today and re-run the CI so that your PR becomes green.

jeffhollan:
@myarotskaya this is awesome! Thanks a ton for contributing 🥳

myarotskaya:
@cgillum @jeffhollan thanks guys! I'm so glad that we figured it out so quickly!

Issue #459, opened 2018-10-04
Failed to deserialize exception from TaskActivity


adobedoe:
Microsoft.Azure.WebJobs.Extensions.DurableTask v1.6.2

### Issue

If you have an orchestrator call a sub orchestrator and that sub orchestrator calls an activity function which throws an exception, then the exception cannot be deserialised and you get "Failed to deserialize exception from TaskActivity"

Easy to reproduce like this:

`
public class MyOrchestrator
`

    {
        [FunctionName("MyTimer")]
        public static async Task RunTimerAsync([TimerTrigger("0 */5 * * * *", RunOnStartup = true)]TimerInfo myTimer, [OrchestrationClient]DurableOrchestrationClient orchestrationClient, ILogger log)
        {
            log.LogInformation($"Executing");

            string instanceId = await orchestrationClient.StartNewAsync("MyOrchestrator", "Test");

            
        }

        [FunctionName("MyOrchestrator")]
        public static async Task RunOrchestratorAsync([OrchestrationTrigger] DurableOrchestrationContext context, ILogger logger)
        {
            try
            {
                await context.CallSubOrchestratorAsync("MySubOrchestrator", "Booooo");
            }
            catch (Exception ex)
            {
                //The exception here is Failed to deserialize exception from TaskActivity:
            }
        }

        [FunctionName("MySubOrchestrator")]
        public static async Task RunSubOrchestratorAsync([OrchestrationTrigger] DurableOrchestrationContext context, ILogger logger)
        {
            await context.CallActivityAsync("MyActivityFunction", "Booooo");
        }

        [FunctionName("MyActivityFunction")]
        public static void RunActivityFunction([ActivityTrigger] DurableActivityContext context, ILogger logger)
        {
            throw new Exception("WTF");
        }
    }

thargenediad:
please fix this!

thomas-schreiter:
Any updates on this? Looks like it might be related to #313 .

cgillum:
We have a PR which addresses this currently in review: https://github.com/Azure/azure-functions-durable-extension/pull/622

cgillum:
I see this issue has received a lot of votes! I'll take a look and see if we can get a fix in for the next release.

cgillum:
This appears to have been recently fixed by PR https://github.com/Azure/azure-functions-durable-extension/pull/628. I'm adding an automated test case to confirm and ensure it doesn't regress again. The fix will go out with the next release.

[...]

Issue #67, opened 2017-09-22
Message loss due to race conditions with ContinueAsNew


cgillum:
## Summary
Message loss or instance termination may be observed if an orchestrator completes after calling `ContinueAsNew` and subsequently processes any other message before restarting.

## Details
When the orchestrator completes execution after `ContinueAsNew` is called, the status is internally marked as "completed" and a new message is enqueued to restart it. During this window between completion and restart, it's possible for other messages to arrive (for example, raised events or termination messages). Because the internal state of the orchestration is completed, those messages will be dropped. It's also possible for the DTFx runtime to terminate the instance claiming possible state corruption.

## Repro
1. Start with the counter sample
2. Create a new instance
3. Call RaiseEventAsync("operation", "incr") multiple times without waiting

**Expected:**
Calling ContinueAsNew multiple times in quick succession should never be an issue. Many workloads may require this, especially actor-style workloads.

**Workaround:**
Have the client wait a few seconds before sending events that may cause the orchestrator to do a ContinueAsNew. This give the instance time to get a new execution ID.

ericleigh007:
I guess the workaround to make things a bit more reliable is to use a while or for in the orchestrator, thereby staving off the ContinueAsNew until N number of events.
Otherwise, Durable functions are really great stuff.
When do you think this extreme fragility with one of the guiding principles of DF's will be addressed?  Do we need major function exec changes or changes within DF's, or something else..

Thanks again for the great work.

Hassmann:
Yes, please, give us a solution. I am using a singleton and it is not reliable. I would have to run a web app service alone for a very small task, if this is not handled. Or give an indication that you do not plan to solve this in the near future, please.

cgillum:
This one definitely needs to be solved, and it's high on the list since it impacts reliability for an important scenario. The change needs to be made in the Azure Storage extension of the Durable Task Framework: https://github.com/Azure/durabletask/blob/azure-functions/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs

christiansparre:
Yes, this one needs to be fixed. Was just playing around and ran into this immediately when creating a singleton function :(

cgillum:
I spent more time looking at this and experimenting with different fixes. Ultimately, I came to the conclusion that fixing this would be far more difficult than I originally anticipated. The short version is that the Durable Task Framework just wasn't designed for RPC/actor-style messaging workloads, and this issue is one example of why.

[...]

PR #631, opened 2019-03-04
Test updates to confirm some bug fixes

cgillum:
This PR adds tests to confirm the fixes for the following issues:

* [Message loss due to race conditions with ContinueAsNew (#67)](#67)
* [Failed to deserialize exception from TaskActivity (#459)](#459)

It also updates the test csproj files to point to more updated versions of some nuget packages.

### DIFF
--- a/test/Common/DurableTaskEndToEndTests.cs
+++ b/test/Common/DurableTaskEndToEndTests.cs
@@ -551,14 +551,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
         /// End-to-end test which validates the ContinueAsNew functionality by implementing a counter actor pattern.
         /// </summary>
         [Theory]
-        [Trait("Category", PlatformSpecificHelpers.FlakeyTestCategory)]
         [Trait("Category", PlatformSpecificHelpers.TestCategory + "_BVT")]
         [InlineData(true)]
         [InlineData(false)]
         public async Task ActorOrchestration(bool extendedSessions)
         {
-            const string testName = nameof(this.ActorOrchestration);
-
             string instanceId;
             using (JobHost host = TestHelpers.GetJobHost(
                 this.loggerProvider,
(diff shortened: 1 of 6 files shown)


test/Common/TestOrchestratorClient.cs
```diff
@@ -155,7 +155,7 @@ public async Task<DurableOrchestrationStatus> WaitForCustomStatusAsync(TimeSpan
             while (sw.Elapsed < timeout);
 
             output.WriteLine($"Timeout expired. {status.Name} ({status.InstanceId}) currently shows a status of '{status.CustomStatus}'.");
-            Assert.Equal(expectedStatus, status.CustomStatus);
+            Assert.Equal(expectedStatus.ToString(), status.CustomStatus?.ToString());
```
ConnorMcMahon: Is this just to give more clear output in the case of a failure?

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -622,6 +616,44 @@ public async Task ActorOrchestration(bool extendedSessions)
...
+                await client.RaiseEventAsync("operation", "incr", this.output);
+                await client.RaiseEventAsync("operation", "end", this.output);
+
+                var status = await client.WaitForCompletionAsync(TimeSpan.FromSeconds(10), this.output);
+
+                Assert.Equal(OrchestrationRuntimeStatus.Completed, status?.RuntimeStatus);
+                Assert.Equal(3, (int?)status?.Output);
+
+                // When using ContinueAsNew, the original input is discarded and replaced with the most recent state.
+                Assert.NotEqual(initialValue, status?.Input);
```
ConnorMcMahon: Are events guaranteed to be processed in the order they are raised? If so, you can strengthen this assertion to make it `Assert.Equal(2, status?.Input)`.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -1483,6 +1515,54 @@ public async Task UnhandledActivityException(bool extendedSessions)
...
+                    this.output);
+                var status = await client.WaitForCompletionAsync(timeout, this.output);
+
+                Assert.Equal(OrchestrationRuntimeStatus.Failed, status?.RuntimeStatus);
+
+                string output = (string)status?.Output;
+                this.output.WriteLine($"Orchestration output string: {output}");
+                Assert.StartsWith(
+                    string.Format(
+                        "Orchestrator function '{0}' failed: The orchestrator function '{1}' failed: \"The activity function '{2}' failed: \"{3}\"",
```
ConnorMcMahon: Out of curiosity, how far does this recursion go? Can your sub-orchestrations start their own orchestrations? If so, are orchestrations at recursion of depth 2 or greater properly handled?

test/Common/TestOrchestratorClient.cs
cgillum: Yeah, it wasn't actually printing the JValue values for some reason, so I switched it to `ToString()`.

test/Common/DurableTaskEndToEndTests.cs
cgillum: There isn't actually any recursion. In this case it's _orchestrator_ --> _subOrchestrator_ --> _activity_. It's possible for us to write a test which does recursion and/or goes deeper than 2, but I didn't think it would be necessary to cover this (I'm being optimistic that there aren't other more subtle issues that pop up as you go deeper).

test/Common/DurableTaskEndToEndTests.cs
cgillum: That's a very good point. I'll do that.

Issue #577, opened 2019-01-16
ArgumentException "The function 'XYZ' doesn't exist" when the functions host recycles


cgillum:
### Issue
Given sufficient load on a function app, it's possible that an orchestrator function will fail with the following exception during a host recycle:

```
{InstanceID}: Function 'XYZ (Orchestrator)' failed with an error. Reason: System.ArgumentException: The function 'ABC' doesn't exist, is disabled, or is not an activity function. Additional info: No activity functions are currently registered!
   at Microsoft.Azure.WebJobs.Extensions.DurableTask.DurableTaskExtension.ThrowIfFunctionDoesNotExist(String name, FunctionType functionType)
   at Microsoft.Azure.WebJobs.DurableOrchestrationContext.CallDurableTaskFunctionAsync[TResult](String functionName, FunctionType functionType, String instanceId, RetryOptions retryOptions, Object input)
```
This is caused by an incorrect validation check that occurs when the functions host is shutting down. The result is that the orchestration transitions into a Failed state.

This issue is expected to impact all versions of the Durable Functions extension.

### Workaround
No known workarounds at this time. Any failed orchestrations will need to be restarted from the beginning.

### Technical Details
The problem is that the [DurableTaskListener](https://github.com/Azure/azure-functions-durable-extension/blob/f6f1dce716b68d8baaa99ed64a9db1306577c58d/src/WebJobs.Extensions.DurableTask/Listener/DurableTaskListener.cs#L44-L54) "deregisters" a function when its listener is stopped by the functions runtime. It's possible that an orchestrator function may still be running after the listener has been stopped, in which case the `CallActivityAsync` and `CallSubOrchestratorAsync` APIs will [throw an exception because the target function is no longer listed](https://github.com/Azure/azure-functions-durable-extension/blob/f6f1dce716b68d8baaa99ed64a9db1306577c58d/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs#L385).

The fix should be to relax this validation so that we don't throw in the case where a function has been "delisted", but rather allow the `CallActivityAsync` or `CallSubOrchestratorAsync` call to succeed. This will allow the message to be written to the queue, and once the listeners are startup up again, the function-call message can be picked up.

olitomlinson:
I've just experienced this exact same exception under high-load.

**DF 1.7.1 **
**Functions V2**

[...]

PR #636, opened 2019-03-08
Remove validation for deregistered functions

cgillum:
Fixes https://github.com/Azure/azure-functions-durable-extension/issues/577

This PR addresses a race condition that happens during Functions host shutdown. Instead of removing deregistered functions from internal dictionary, we keep them in that dictionary and just set a `IsDeregistered` bit to `true`. This avoids the exception which can happen if an orchestrator function tries to schedule an activity function at the same time as a host shutdown.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -467,6 +467,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         internal void RegisterOrchestrator(FunctionName orchestratorFunction, RegisteredFunctionInfo orchestratorInfo)
         {
+            orchestratorInfo.IsDeregistered = false;
+
             if (this.knownOrchestrators.TryAdd(orchestratorFunction, orchestratorInfo))
             {
                 this.TraceHelper.ExtensionInformationalEvent(
@@ -485,7 +487,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         internal void DeregisterOrchestrator(FunctionName orchestratorFunction)
         {
             RegisteredFunctionInfo existing;
-            if (this.knownOrchestrators.TryGetValue(orchestratorFunction, out existing))
+            if (this.knownOrchestrators.TryGetValue(orchestratorFunction, out existing) && !existing.IsDeregistered)
             {
                 existing.IsDeregistered = true;
 
@@ -529,7 +531,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         internal void DeregisterActivity(FunctionName activityFunction)
         {
             RegisteredFunctionInfo info;
-            if (this.knownActivities.TryGetValue(activityFunction, out info))
+            if (this.knownActivities.TryGetValue(activityFunction, out info) && !info.IsDeregistered)
             {
                 info.IsDeregistered = true;
 
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -467,7 +467,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         internal void RegisterOrchestrator(FunctionName orchestratorFunction, RegisteredFunctionInfo orchestratorInfo)
         {
-            orchestratorInfo.IsDeregistered = false;
+            if (orchestratorInfo != null)
+            {
+                orchestratorInfo.IsDeregistered = false;
+            }
 
             if (this.knownOrchestrators.TryAdd(orchestratorFunction, orchestratorInfo))
             {
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Reflection;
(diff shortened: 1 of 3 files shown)


src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -464,85 +465,92 @@ protected internal virtual DurableOrchestrationClient GetClient(OrchestrationCli
...
-                functionName: activityFunction.Name,
-                message: $"Deregistering orchestrator function named {activityFunction}.",
-                writeToUserLogs: false);
+            RegisteredFunctionInfo info;
+            if (this.knownActivities.TryGetValue(activityFunction, out info))
+            {
+                info.IsDeregistered = true;
 
-            this.registeredActivities.TryRemove(activityFunction, out _);
+                this.TraceHelper.ExtensionInformationalEvent(
```
ConnorMcMahon: NIT: To keep parity with the old behavior, we should technically only log this if the known activity is transitioning from `IsDeregistered` from `false` to `true`.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -295,7 +296,7 @@ TaskActivity INameVersionObjectManager<TaskActivity>.GetObject(string name, stri
                 throw new InvalidOperationException(message);
             }
 
-            return new TaskActivityShim(this, executor, name);
+            return new TaskActivityShim(this, info.Executor, name);
```
ConnorMcMahon: When we return this value, we are losing information about whether the Task is registered or not. I'm assuming this is intentional, but I think I need more context to understand the ramifications of this.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -317,8 +318,8 @@ private async Task OrchestrationMiddleware(DispatchMiddlewareContext dispatchCon
 
             FunctionName orchestratorFunction = new FunctionName(context.Name);
 
-            OrchestratorInfo info;
-            if (!this.registeredOrchestrators.TryGetValue(orchestratorFunction, out info))
+            RegisteredFunctionInfo info;
+            if (!this.knownOrchestrators.TryGetValue(orchestratorFunction, out info))
```
ConnorMcMahon: Because we are no longer throwing an exception here for deregistered orchestrators, we will try executing them below. Again, I am assuming this is intentional, but I'm a little confused as to what "deregistered" means at this point.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -464,85 +465,92 @@ protected internal virtual DurableOrchestrationClient GetClient(OrchestrationCli
             };
         }
 
-        internal void RegisterOrchestrator(FunctionName orchestratorFunction, OrchestratorInfo orchestratorInfo)
+        internal void RegisterOrchestrator(FunctionName orchestratorFunction, RegisteredFunctionInfo orchestratorInfo)
         {
-            if (!this.registeredOrchestrators.TryUpdate(orchestratorFunction, orchestratorInfo, null))
```
ConnorMcMahon: I am very confused by the old logic here. How is `TryUpdate(orchestratorFunction, orchestratorInfo, null)` different than `TryAdd(orchestratorFunction, orchestratorInfo)`?

In regards to the new logic, I am a little confused why we use `TryAdd()` with a `[]` update if we couldn't add. `TryAddOrUpdate` seems to accomplish the same thing.  Is this so that we only log in the case where we are adding a new orchestrator as opposed to updating an existing one?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -464,85 +465,92 @@ protected internal virtual DurableOrchestrationClient GetClient(OrchestrationCli
             };
         }
 
-        internal void RegisterOrchestrator(FunctionName orchestratorFunction, OrchestratorInfo orchestratorInfo)
+        internal void RegisterOrchestrator(FunctionName orchestratorFunction, RegisteredFunctionInfo orchestratorInfo)
```
ConnorMcMahon: Should we consider verifying that `orchestratorInfo.IsDerigistered=false`?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
ConnorMcMahon: This also applies to the `DerigisterOrchestrator()` logic as well.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -551,9 +559,9 @@ internal void ThrowIfFunctionDoesNotExist(string name, FunctionType functionType
         internal string GetInvalidActivityFunctionMessage(string name)
```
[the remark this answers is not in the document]
cgillum: With this change, yes. Previously they were considered invalid, and now they are considered valid.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -566,9 +574,9 @@ internal string GetInvalidActivityFunctionMessage(string name)
         internal string GetInvalidOrchestratorFunctionMessage(string name)
```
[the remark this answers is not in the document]
cgillum: Yep, same thing here.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
cgillum: Fair enough. I'll add a check to see whether it's already been deregistered.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
cgillum: You are correct. Yes, the meaning of "deregistered" is a little confusing now. The only reason we need to keep track of this status is so that we do the right thing in `StopTaskHubWorkerIfIdleAsync` - basically shut down the task hub worker when all triggers are "deregistered". I'm happy to use a different property name if you have good ideas. :)

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
cgillum: I was confused about this too, but I think what happened. Previously this method was implemented the same way as `RegisterActivity`, and that use of `TryUpdate` actually made sense. Katy later made a change where instead of `ITriggeredFunctionExecutor` we used `OrchestratorInfo` and I think that caused this old logic to no longer work the way it was intended.

The use of `[]` over `TryAddOrUpdate` was somewhat arbitrary. I felt the behavior was properly understood and less verbose, which is why I went with it.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
cgillum: I guess I can just explicitly set it to `false` here.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
cgillum: Oh, I might have misuderstood your second question. The reason I do TryAdd followed by a direct `[]` update is because I want to distinguish between add and update, basically so I can write the trace only in the add case.

PR #639, opened 2019-03-11
Added TrackingStoreConnectionStringName and TrackingStoreNamePrefix settings

cgillum:
This PR adds two new properties to the `DurableTaskOptions` class. It allows multiple function apps to share a common History and Instances table, while having separate queues for polling.

The two main scenarios for this PR are:

* Zero downtime deployments (side-by-side versioning)
* Improved scalability - multiple function apps can all use separate storage accounts for queues and a common storage account for tables.

It is the followup change to https://github.com/Azure/durabletask/pull/252

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -462,6 +462,28 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 ExtendedSessionsEnabled = this.Options.ExtendedSessionsEnabled,
                 ExtendedSessionIdleTimeout = extendedSessionTimeout,
                 MaxQueuePollingInterval = this.Options.MaxQueuePollingInterval,
+                TrackingStoreStorageAccountDetails = this.GetStorageAccountDetailsOrNull(
+                    this.Options.TrackingStoreConnectionStringName),
+                TrackingStoreNamePrefix = this.Options.TrackingStoreNamePrefix,
+            };
+        }
+
+        private StorageAccountDetails GetStorageAccountDetailsOrNull(string connectionName)
+        {
+            if (string.IsNullOrEmpty(connectionName))
+            {
+                return null;
+            }
+
+            string resolvedStorageConnectionString = this.connectionStringResolver.Resolve(connectionName);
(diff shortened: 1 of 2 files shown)


src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -462,6 +462,28 @@ protected internal virtual DurableOrchestrationClient GetClient(OrchestrationCli
                 ExtendedSessionsEnabled = this.Options.ExtendedSessionsEnabled,
                 ExtendedSessionIdleTimeout = extendedSessionTimeout,
                 MaxQueuePollingInterval = this.Options.MaxQueuePollingInterval,
+                TrackingStoreStorageAccountDetails = this.GetStorageAccountDetailsOrNull(
+                    this.Options.TrackingStoreConnectionStringName),
+                TrackingStoreNamePrefix = this.Options.TrackingStoreNamePrefix,
```
ConnorMcMahon: If `this.Options.TrackingStoreNamePrefix` is null, then we are overwriting the default value for `AzureStorageOrchestrationServiceSettings.TrackingStoreNamePrefix`, which is `"DurableTask"`. There are several options to fix this, but ideally we would do it without hardcoding "DurableTask" as the default value here as well, considering it is already in `DurableTask.AzureStorage`.

src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
```diff
@@ -97,6 +97,32 @@ public class DurableTaskOptions
         /// <value>The name of a connection string that exists in the app's application settings.</value>
         public string AzureStorageConnectionStringName { get; set; }
```
TsuyoshiUshio: Do we need Validation for this class for kind error message related ConnectionString? 
What do you think?

cgillum: What kind of validation?

TsuyoshiUshio: DurableTaskOptions.cs has a validate method for the property.  For example, we can validate if it is configured(not null). Then people might realize that the ConnectionString has not set.  However, it is  not mandatory. So I ask you the question.

Issue #383, opened 2018-07-07
Task hub name should default to the function app name


cgillum:
Currently task hub names default to "DurableFunctionsHub". Instead, the default task hub name should derive from the function app name. This will have the following benefits:

1. It will make it much harder for developers to accidentally run multiple function apps in the same task hub (see https://github.com/Azure/azure-functions-durable-extension/issues/379).
2. Customer support will have an easier time collecting all relevant logs for a function app if both extension-level and DTFx-level traces use the same distinct task hub name (see https://github.com/Azure/azure-functions-durable-extension/issues/326).

This is a breaking change, so it would need to be introduced as part of a breaking change release.

ConnorMcMahon:
Unfortunately, any default we tried brainstorming up with doesn't play nicely with slots. 

The answer seems to be to remove the default task hub name, and enforce that developers provide a hub name. If we are able to come up with a default that works well with slots in the future, we can always add that in the future without it being a breaking change.

ConnorMcMahon:
This was addressed by PR #652.

PR #649, opened 2019-03-13
Remove default value of task hub name

ConnorMcMahon:
Addresses #383

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
@@ -14,12 +14,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
     public class DurableTaskOptions
     {
         /// <summary>
-        /// The default task hub name to use when not explicitly configured.
-        /// </summary>
-        internal const string DefaultHubName = "DurableFunctionsHub";
-
-        /// <summary>
-        /// Gets or sets default task hub name to be used by all <see cref="DurableOrchestrationClient"/>,
+        /// Gets or sets task hub name to be used by all <see cref="DurableOrchestrationClient"/>,
         /// <see cref="DurableOrchestrationContext"/>, and <see cref="DurableActivityContext"/> instances.
         /// </summary>
         /// <remarks>
@@ -27,7 +22,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// multiple Durable Functions applications from each other, even if they are using the same storage backend.
         /// </remarks>
         /// <value>The name of the default task hub.</value>
-        public string HubName { get; set; } = DefaultHubName;
+        public string HubName { get; set; }
 
         /// <summary>
         /// Gets or sets the number of messages to pull from the control queue at a time.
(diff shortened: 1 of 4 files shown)


test/Common/HttpApiHandlerTests.cs
```diff
@@ -804,9 +804,11 @@ public async Task StartNewInstance_Returns_HTTP_400_On_Missing_Function()
 
         private static DurableTaskExtension GetTestExtension()
         {
-            var options = new DurableTaskOptions();
-            options.NotificationUrl = new Uri(TestConstants.NotificationUrl);
-
+            var options = new DurableTaskOptions()
+            {
+                NotificationUrl = new Uri(TestConstants.NotificationUrl),
+                HubName = "DurableFunctionsHub",
```
ConnorMcMahon: These tests depended on the old default value, so set it for them.

src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
```diff
@@ -14,20 +14,29 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
...
+        /// no value for HubName was already provided.
+        /// </summary>
+        /// <remarks>
+        /// This method utilizes the WEBSITE_SITE_NAME environment variables that is set by the Azure App Service infrastructure.
+        /// If this environment variable is not present (i.e. running locally or on non-App Service infrastructure), we default
+        /// to localhost.
+        /// </remarks>
+        internal void SetDefaultHubName(INameResolver nameResolver)
+        {
+            this.HubName = nameResolver.Resolve("WEBSITE_SITE_NAME") ?? "localhost";
```
ConnorMcMahon: I considered using `WEBSITE_IIS_SITE_NAME` to allow differentiation between the site and its slots, but decided against it for a few reasons:

1. It prepends "~1" to the sitename, which could be confusing.
2. It doesn't exist on Linux.
3. The suffix for slot names isn't deterministic to the best of my knowledge. It may not even be consistent between site start-ups or slot swaps.  This could lead to some confusing behavior for the customers.

1 could be easily fixed, and in the case of 2 we could just default to `WEBSITE_SITE_NAME` if `WEBSITE_IIS_SITE_NAME` doesn't exist. The third point is my largest concern. It might be best just to insist that customers set their own TaskHub names if they want to use slots, and provide best practices for them.

cgillum: What about a combination of `WEBSITE_SITE_NAME` and `WEBSITE_SLOT_NAME`? And maybe if it's the production slot, we leave that suffix out and just make it `WEBSITE_SITE_NAME`?

We tried the guidance approach already, but unfortunately too many people don't read this far into the docs and end up getting into a bad situation, which is why I want to have reasonable defaults.

ConnorMcMahon: As per our offline discussion, decided to remove the default for now. We can always add one later without it being a breaking change if we find a good value to use.

ConnorMcMahon:
@cgillum There actually was already validation for the case where HubName was set to null in `DurableTaskOptions.Validate()`. A test that verifies that the runtime throws an exception right away without that value would be a good idea.

cgillum:
Hey @ConnorMcMahon, would you mind resolving the conflicts in this PR? I'd like to go ahead and get it merged.

ConnorMcMahon:
I actually believe this issue was fully addressed by PR #652. I am going to close this PR.

PR #650, opened 2019-03-13
 Improve feedback from RaiseEventAsync() and TerminateAsync() APIs 

ConnorMcMahon:


test/Common/DurableOrchestrationClientBaseTests.cs
```diff
@@ -23,5 +27,103 @@ public async Task StartNewAsync_is_calling_overload_method()
...
+        [Fact]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        public async Task RaiseEventAsync_InvalidInstanceId_ThrowsException()
+        {
+            var instanceId = Guid.NewGuid().ToString();
+            var orchestrationServiceClientMock = new Mock<IOrchestrationServiceClient>();
+            orchestrationServiceClientMock.Setup(x => x.GetOrchestrationStateAsync(It.IsAny<string>(), It.IsAny<bool>())).ReturnsAsync(GetInvalidInstanceState());
+            var durableOrchestrationClient = new DurableOrchestrationClient(orchestrationServiceClientMock.Object, GetDurableTaskExtension(), new OrchestrationClientAttribute { });
+
+            await Assert.ThrowsAnyAsync<ArgumentException>(async () => await durableOrchestrationClient.RaiseEventAsync("invalid_instance_id", "anyEvent", new { message = "any message" }));
```
ConnorMcMahon: I don't love adding a dependency on what type of exception we are throwing in the test, as it makes the test more fragile if we change the exception type in the future.

That being said, it seems preferable to allowing any exception, as if future code changes cause the test to break (like `RaiseEventAsync` encountering a `NullReferenceException`), the test would still pass if we allowed any exception.

cgillum: I think it's better to be specific. If we change the type of exception we throw, that could be considered a breaking change and we'd want our tests to catch this.

test/Common/HttpApiHandlerTests.cs
```diff
@@ -581,7 +581,7 @@ public async Task TerminateInstanceWebhook()
             var clientMock = new Mock<DurableOrchestrationClientBase>();
             clientMock
                 .Setup(x => x.TerminateAsync(It.IsAny<string>(), It.IsAny<string>()))
-                .Returns(Task.CompletedTask)
+                .Returns(Task.FromResult(true))
```
cgillum: Is this change still needed?

ConnorMcMahon: Nope! I thought I missed something. It's weird that this didn't cause a build issue.

Issue #643, opened 2019-03-12
Sub-orchestration's FunctionFailedException.InnerException is null


kijujjav:
**Issue**
I am throwing TaskCanceledException() from the action to the orchestrator and expecting FunctionFailedException.InnerException to have the exception, instead it has null.

**Expected**
FunctionFailedException.InnerException should have TaskCanceledException().

**Additional Details**

In action
throw new TaskCanceledException();

In Orchestrator
catch (Exception ex)
            {
                
                if (ex.InnerException is TaskCanceledException)
                {
                    log.LogError(ex.Message);
                }
             }
![InnerExceprionNull](https://user-images.githubusercontent.com/48494090/54231736-48f8f780-44c6-11e9-9ccb-ba525f8a7202.png)

cgillum:
Do you get the same behavior with any other exception type? I wonder if `TaskCanceledException` is being inadvertently handled somewhere in the runtime.

kijujjav:
InnerException is null no matter what type of exception I throw.

cgillum:
I'm not able to reproduce this behavior:
![image](https://user-images.githubusercontent.com/2704139/54324040-23004f80-45b9-11e9-9805-b896a6745acb.png)

Can you share more details about your app? For example, the version of the Functions runtime (1.0 or 2.0) and the version of the nuget packages you are using?

kijujjav:
We are calling context.CallSubOrchestratorAsync and the exception is thrown in the sub-orchestration, does the behavior differ for activity and sub-orchestration?

kijujjav:
We are using Functions 2.0 and azure-functions-durable-extension 1.7.1

cgillum:
I created the following test case orchestrator function to see whether this is reproduceable in the latest code:

```csharp
public static async Task SubOrchestrationThrow([OrchestrationTrigger] DurableOrchestrationContext ctx)
{
    string message = ctx.GetInput<string>();

    try
    {
        await ctx.CallSubOrchestratorAsync(nameof(TestOrchestrations.ThrowOrchestrator), message);
    }
    catch (FunctionFailedException e)
    {
        if (e.InnerException == null ||
            e.GetBaseException().GetType() != typeof(InvalidOperationException) ||
            !e.InnerException.Message.Contains(message))
        {
            throw new Exception("InnerException was not the expected value.");
        }

        // rethrow the original exception
        throw;
    }
}
```

[...]

PR #654, opened 2019-03-15
Various fixes and test tweaks

cgillum:
Added a new test to validate #643, plus made a few other small tweaks that I discovered while updating docs.

Note the `DurableTaskOptions.cs` has a bunch of newline changes, which will make the diff harder to look at. You'll want to turn off whitespace comparisons for that.


test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -1115,6 +1115,36 @@ public async Task HandledActivityException(bool extendedSessions)
...
+            using (JobHost host = TestHelpers.GetJobHost(
+                this.loggerProvider,
+                nameof(this.HandledActivityException),
+                enableExtendedSessions: true))
+            {
+                await host.StartAsync();
+
+                string message = $"Failure ID: {Guid.NewGuid()}";
+
+                // Empty string input should result in ArgumentNullException in the orchestration code.
```
ConnorMcMahon: Is this comment meant for this test?

test/Common/TestActivities.cs
```diff
@@ -53,7 +53,7 @@ public static long GetFileSize([ActivityTrigger] DurableActivityContext ctx)
         public static void ThrowActivity([ActivityTrigger] DurableActivityContext ctx)
         {
             string message = ctx.GetInput<string>();
-            throw new Exception(message);
+            throw new InvalidOperationException(message);
```
ConnorMcMahon: Was this change required? I thought `SubOrchestrationThrow` calls `ThrowOrchestrator`, not `ThrowActivity`.

cgillum: This is not required. But I added it because at one point we were having issues where exceptions were not being deserialized to the correct type. Throwing a non-generic Exception would have helped us catch that, so I figure it's worth adding it now.

test/Common/DurableTaskEndToEndTests.cs
cgillum: Nope. Copy/paste error. Good catch.

PR #663, opened 2019-03-21
initial implementation of durable actors for v2

sebastianburckhardt:
This is the current state of the actor implementation. It is not complete nor thoroughly tested, but should be sufficiently operable to run perf tests with. Also, since this makes some significant changes to the existing code (in terms of class organization and where functionality is placed in those classes), it may be good to merge this into v2 sooner rather than later to avoid difficult merge conflicts later on.

- introduces actor as a new type of function, with the API as discussed, and implemented using an orchestrator loop and the newly added auto-start feature of DTFx.
- splits TaskOrchestrationShim into three classes (TaskOrchestrationShim, TaskActorShim, TaskCommonShim) where the latter is a superclass of the two former
- splits DurableOrchestrationContext into three classes (DurableOrchestrationContext, DurableActorContext, DurableCommonContext) where the latter is a superclass of the two former
- changes the way that context implementations are exposed from abstract classes to interfaces. Concretely, it replaces the (DurableXXX + DurableXXXBase) pattern with (DurableXXX + IDurableXXX) pattern for XXX = {OrchestrationContext, ActivityContext, OrchestrationClient}. Also, all the DurableXXX classes are now internal, only the IDurableXXX interfaces remain public.
- use solution folders {ContextImplementations, ContextInterfaces} to organize all these contexts and interfaces.


src/WebJobs.Extensions.DurableTask/ActorScheduler/ActorRef.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// <summary>
+        /// Create an actor reference for an actor.
+        /// </summary>
+        /// <param name="actorClass">the name of the actor class.</param>
+        /// <param name="actorKey">the actor key.</param>
+        public ActorRef(string actorClass, string actorKey)
+        {
+            if (string.IsNullOrEmpty(actorClass))
+            {
+                throw new ArgumentException("invalid actor reference: actor class must not be a null or empty string", actorClass);
```
cgillum: Typically we just use `ArgumentNullException` for null or empty strings.

src/WebJobs.Extensions.DurableTask/ActorScheduler/ActorRef.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            this.ActorClass = actorClass;
+            this.ActorKey = actorKey;
+        }
+
+        /// <summary>
+        /// The name of the actor class.
+        /// </summary>
+        [JsonProperty("class")]
+        public string ActorClass { get; set; }
```
cgillum: Would it make sense to make these properties immutable? That way we can guarantee, for example, that `ActorClass` will never be null.  You could just use this syntax:

```csharp
public string ActorClass { get; }
```

src/WebJobs.Extensions.DurableTask/ActorScheduler/ActorRef.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// A reference to an actor, used to identify actor instances.
+    /// </summary>
+    public struct ActorRef : IEquatable<ActorRef>, IComparable
+    {
+        /// <summary>
+        /// Create an actor reference for an actor.
+        /// </summary>
+        /// <param name="actorClass">the name of the actor class.</param>
+        /// <param name="actorKey">the actor key.</param>
+        public ActorRef(string actorClass, string actorKey)
```
cgillum: Naming: One potential worry I have about the name `ActorRef` is that it gives the impression that it is a pointer to an actor, just like in C# how reference parameters are pointers to a variable defined elsewhere. I think we observed this same confusion when reviewing early iterations of this with the team (someone even thought it was a like a C# proxy type). In fact, I think akka uses [ActorRef](https://doc.akka.io/api/akka/2.0/akka/actor/ActorRef.html) as a proxy type as well.

Could we consider an alternate name, like `ActorId`? Service Fabric uses the term [ActorId](https://docs.microsoft.com/en-us/dotnet/api/microsoft.servicefabric.actors.actorid?view=azure-dotnet), as do other implementations, so there is some prior art here. I feel like the meaning will be generally more easily understood by developers.

src/WebJobs.Extensions.DurableTask/ActorScheduler/ActorRef.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        /// <summary>
+        /// The name of the actor class.
+        /// </summary>
+        [JsonProperty("class")]
+        public string ActorClass { get; set; }
+
+        /// <summary>
+        /// The actor key.
```
cgillum: I would add some additional description here. For example, we should call out that this is expected to be some _unique_ key for the actor.

src/WebJobs.Extensions.DurableTask/ActorScheduler/OperationMessage.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                this.Argument = jtoken.ToString(Formatting.None);
+            }
+            else
+            {
+                this.Argument = MessagePayloadDataConverter.Default.Serialize(obj);
+            }
+        }
+
+        public TArgument GetArgument<TArgument>()
```
cgillum: Elsewhere in the codebase, we just use `T` the generic parameter type argument name when there is just one and when it is unconstrained. I suggest we do the same here for consistency.

src/WebJobs.Extensions.DurableTask/ActorScheduler/ResponseMessage.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    internal class ResponseMessage
+    {
+        public string Result { get; set; }
```
cgillum: Is this property intended to be serialized? If so, let's decorate all members, similar to how you are doing with the `OperationMessage` class.

src/WebJobs.Extensions.DurableTask/ActorScheduler/OperationMessage.cs
```diff
@@ -0,0 +1,54 @@
...
+    internal class OperationMessage
+    {
+        [JsonProperty(PropertyName = "op")]
+        public string Operation { get; set; }
+
+        [JsonProperty(PropertyName = "signal")]
+        public bool IsSignal { get; set; }
+
+        [JsonProperty(PropertyName = "arg")]
+        public string Argument { get; set; }
```
cgillum: Have you considered other names, like `Data` or `Payload`? I wonder if those might be a bit more clear compared to `Argument`.

sebastianburckhardt: Yes, I agree. We just need to agree on a better term. Data is perhaps a bit generic also. How about  `content`? This would be in analogy to the content of http requests.

cgillum: I like `Content` as well.

src/WebJobs.Extensions.DurableTask/ActorScheduler/ResponseMessage.cs
```diff
@@ -0,0 +1,76 @@
+﻿using System;
```
cgillum: This file needs the copyright header.

src/WebJobs.Extensions.DurableTask/ActorScheduler/ResponseMessage.cs
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        public void SetExceptionResult(Exception exception, string operation, ActorRef actor)
+        {
+            this.ExceptionType = exception.GetType().AssemblyQualifiedName;
+
+            this.Result = MessagePayloadDataConverter.ErrorConverter.Serialize(exception);
+        }
+
+        public TResult GetResult<TResult>()
```
cgillum: Same suggestion about just using `<T>` here.

src/WebJobs.Extensions.DurableTask/ActorScheduler/SchedulerState.cs
```diff
@@ -0,0 +1,140 @@
+﻿using System;
```
cgillum: This file needs the copyright header.

src/WebJobs.Extensions.DurableTask/ActorScheduler/SchedulerState.cs
```diff
@@ -0,0 +1,140 @@
...
+    {
+        /// <summary>
+        /// Whether this actor exists or not.
+        /// </summary>
+        public bool ActorExists { get; set; }
+
+        /// <summary>
+        /// The serialized actor state. This can be stale while CurrentStateView != null.
+        /// </summary>
+        public string ActorState { get; set; }
```
cgillum: If these properties are intended to be serialized, let's explicitly mark them with `[JsonProperty]`. I would prefer it if we can set the `PropertyName` as well using a camelCase naming convention, which we typically use for all JSON payloads.

src/WebJobs.Extensions.DurableTask/ActorScheduler/SchedulerState.cs
```diff
@@ -0,0 +1,140 @@
...
+
+        /// <summary>
+        /// The queue of waiting operations, or null if none.
+        /// </summary>
+        public Queue<OperationMessage> Queue { get; set; }
```
cgillum: Does the `Queue` property ever get assigned from within another class? If not, I suggest we change `set` into `private set`.

src/WebJobs.Extensions.DurableTask/ActorScheduler/SchedulerState.cs
cgillum: I do see that you're using a custom JSON converter. I'll admit that I don't have much experience with those so I don't know whether `[JsonProperty]` is meant to be useful when custom converters are at play.

src/WebJobs.Extensions.DurableTask/ActorScheduler/SchedulerState.cs
```diff
@@ -0,0 +1,140 @@
...
+
+            return true;
+        }
+
+        public override string ToString()
+        {
+            return $"state.length={((this.ActorState != null) ? this.ActorState.Length : 0)} queue.count={((this.Queue != null) ? this.Queue.Count : 0)}";
+        }
+
+        public void FromJson(JsonReader reader, JsonSerializer serializer)
```
cgillum: A couple suggestions regarding custom serialization logic:
* I suggest we add a comment explaining why we use custom serialization instead of the reflection-based serialization that most people will expect.
* We should also add a comment giving an example of what the JSON payload we're parsing is expected to look like. That will make human validation of the code easier.

src/WebJobs.Extensions.DurableTask/ActorScheduler/TypedStateView.cs
```diff
@@ -0,0 +1,82 @@
+﻿using System;
```
cgillum: This file needs a copyright header.

src/WebJobs.Extensions.DurableTask/ActorScheduler/SchedulerState.cs
sebastianburckhardt: Good catch. [JsonProperty] has no effect here because I am using a custom JsonConverter now. I will remove it.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/DurableContextExtensions.cs
```diff
@@ -0,0 +1,402 @@
...
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// Defines convenient overloads for calling the context methods, for all the contexts.
```
cgillum: This is an interesting approach for extending interfaces. It's a little unfortunate that it has to be in a separate class since that makes it harder for users to find these methods when browsing the API documentation. I wonder if [C# 8's Default Interface Implementation](https://www.codejourney.net/2019/02/csharp-8-default-interface-methods/) can help with this. If so, I think we should consider this as an alternative approach (of course, this would require us to take a dependency on the newer tools which are not yet GA).

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDeterministicExecutionContext.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        /// <summary>
+        /// Determines whether the current context is locked, and if so, what locks are currently owned.
+        /// </summary>
+        /// <param name="ownedLocks">the collection of owned locks.</param>
+        /// <remarks>
+        /// Note that the collection of owned locks can be empty even if the context is locked. This happens
+        /// if an orchestration calls a suborchestration without lending any locks.
+        /// </remarks>
+        /// <returns>true if the context already holds some locks.</returns>
```
cgillum: `true` and `false` literals should be surrounded by `<c>` tags. See `IsReplaying` above for an example.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDeterministicExecutionContext.cs
```diff
@@ -0,0 +1,124 @@
...
+        /// </remarks>
+        /// <value>
+        /// <c>true</c> if the orchestration or operation is currently being replayed; otherwise <c>false</c>.
+        /// </value>
+        bool IsReplaying { get; }
+
+        /// <summary>
+        /// Determines whether the current context is locked, and if so, what locks are currently owned.
+        /// </summary>
+        /// <param name="ownedLocks">the collection of owned locks.</param>
```
cgillum: Parameter description sentences should start with a capital letter.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableActorContext.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        /// <summary>
+        /// Gets the name of the operation that was called.
+        /// </summary>
+        string OperationName { get; }
+
+        /// <summary>
+        /// A logger for logging information about this actor.
+        /// </summary>
+        ILogger Logger { get; }
```
cgillum: What is the scenario for this?

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableActorContext.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <typeparam name="TState">the JSON-serializable type of the actor state.</typeparam>
+        /// <returns>a typed view that allows reading and updating.</returns>
+        IStateView<TState> GetStateAs<TState>(Formatting formatting = Formatting.Indented, JsonSerializerSettings settings = null);
+
+        /// <summary>
+        /// Gets the content that was passed to this operation as a deserialized value.
+        /// </summary>
+        /// <typeparam name="T">The JSON-serializable type used for the operation content.</typeparam>
+        /// <returns>the operation content, or default(<typeparamref name="T"/>) if none.</returns>
+        T GetOperationContent<T>();
```
cgillum: I think we'll want to add a `<remarks>` paragraph explaining what exactly operation context is and how it is intended to be used. I don't even know what it is. :)

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableActorContext.cs
```diff
@@ -0,0 +1,112 @@
...
+    /// <summary>
+    /// A view of the current state of the actor.
+    /// </summary>
+    public interface IStateView : IDisposable
+    {
+        /// <summary>
+        /// Serializes the current state to JSON.
+        /// </summary>
+        void WriteBack();
+    }
```
cgillum: I suggest putting both of these interfaces into their own file, `IStateView.cs`. That's the recommendation of the .NET Framework design guidelines.

src/WebJobs.Extensions.DurableTask/Listener/TaskActorShim.cs
```diff
@@ -0,0 +1,323 @@
...
+        {
+            if (this.lastStartedOperation.IsCompleted)
+            {
+                return "Available";
+            }
+            else
+            {
+                if (this.context.State.Queue != null)
+                {
+                    return $"Busy ({this.context.State.Queue.Count} Queued)";
```
cgillum: nit: I would make "Queued" lowercase.

src/WebJobs.Extensions.DurableTask/Listener/TaskActorShim.cs
```diff
@@ -0,0 +1,323 @@
...
+                }
+                else
+                {
+                    return "Busy";
```
cgillum: What does it mean for an actor to be processing an operation but not have anything queued? Can you add some comments differenciating these cases?

src/WebJobs.Extensions.DurableTask/Listener/TaskActorShim.cs
```diff
@@ -0,0 +1,323 @@
...
+                }
+            }
+        }
+
+        private void SignalShutdown()
+        {
+            MethodInfo trySetResult = typeof(TaskCompletionSource<object>).GetMethod("TrySetResult");
+            trySetResult.Invoke(this.shutDownSignal, new[] { (object)null });
```
cgillum: Why do we need reflection here? It seems to me we can just call `TrySetResult` directly, like we are in the line below.

src/WebJobs.Extensions.DurableTask/Listener/TaskActorShim.cs
```diff
@@ -0,0 +1,323 @@
...
+        public override async Task<string> Execute(OrchestrationContext innerContext, string serializedInput)
+        {
+            if (this.FunctionInvocationCallback == null)
+            {
+                throw new InvalidOperationException($"The {nameof(this.FunctionInvocationCallback)} has not been assigned!");
+            }
+
+            if (this.GetFunctionInfo().IsOutOfProc)
+            {
+                throw new NotImplementedException("out-of-proc actor support is not implemented yet");
```
cgillum: FYI @kashimiz looks like this is where we need to plug-in out-of-proc support.

test/Common/TestActivities.cs
```diff
@@ -12,50 +14,73 @@ internal static class TestActivities
...
         }
 
-        public static void ThrowActivity([ActivityTrigger] DurableActivityContext ctx)
+        public static void ThrowActivity([ActivityTrigger] IDurableActivityContext ctx)
         {
             string message = ctx.GetInput<string>();
             throw new Exception(message);
         }
 
+        public static Task<string> ReadStringFromTextFile([ActivityTrigger] IDurableActivityContext ctx)
```
cgillum: It looks like there are a few activity functions and actor functions which aren't being used yet. I assume that's intentional, but wanted to call it out just in case.

cgillum:
Overall I feel like this PR is really nicely done. Thank you for the thoughtful design and careful refactoring of existing code. I feel like the new actor features will fit in nicely.

There are a couple more naming things I think we could potentially improve, but I think they are minor. The main thing that would be helpful for me is a high-level description of the dispatcher logic. This will be important for both reasoning about correctness and also making predictions on the performance characteristics.

Other than that, I think one big missing feature is the ability to interact with actors from `IDurableOrchestrationClient` or some equivalent client-side binding (maybe `IDurableActorClient`?). I feel that customers will expect to be able to interact with actors from non-orchestrator functions, like HTTP triggers. I realize that request-response will be very challenging, but I think we should at least strive to enable one-way messaging to (or signaling of) their actors.

src/WebJobs.Extensions.DurableTask/ActorScheduler/ActorId.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// <param name="actorKey">the actor key.</param>
+        public ActorId(string actorClass, string actorKey)
+        {
+            if (string.IsNullOrEmpty(actorClass))
+            {
+                throw new ArgumentNullException("invalid actor reference: actor class must not be a null or empty string", actorClass);
+            }
+
+            this.ActorClass = actorClass;
+            this.ActorKey = actorKey;
```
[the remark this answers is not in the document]
sebastianburckhardt: Yes, I think so. I will be validating that the key is not null, but I think we should allow the empty string as a key. For example, an empty key is appropriate for singleton actors.

src/WebJobs.Extensions.DurableTask/ActorScheduler/SchedulerState.cs
```diff
@@ -0,0 +1,153 @@
...
+            return true;
+        }
+
+        public override string ToString()
+        {
+            return $"state.length={((this.ActorState != null) ? this.ActorState.Length : 0)} queue.count={((this.Queue != null) ? this.Queue.Count : 0)}";
+        }
+
+        // To save space (e.g. avoid embedded CLR typenames) and for better human readability
+        // of the history we customize the JSON representation of the actor scheduler.
```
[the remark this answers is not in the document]
sebastianburckhardt: I am pretty sure I saw the typename being included somewhere, but after testing this again, it does not seem to be happening. So I agree it is not maintaining the custom serialization (it is indeed error prone and would require adding unit tests), we can get almost the same result using simple json serialization attributes.

src/WebJobs.Extensions.DurableTask/ActorScheduler/TypedStateView.cs
```diff
@@ -0,0 +1,85 @@
...
+
+            var actorState = context.State.ActorState;
+
+            if (actorState == null)
+            {
+                this.state = default(TState);
+            }
+            else
+            {
+                this.state = JsonConvert.DeserializeObject<TState>(actorState);
```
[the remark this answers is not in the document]
sebastianburckhardt: Of course, yes. Good catch.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActorContext.cs
```diff
@@ -0,0 +1,121 @@
...
+            this.self = actor;
+        }
+
+        internal SchedulerState State { get; set; }
+
+        internal OperationMessage CurrentOperation { get; set; }
+
+        internal ResponseMessage CurrentOperationResponse { get; set; }
+
+        internal bool Fresh { get; set; }
```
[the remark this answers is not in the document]
sebastianburckhardt: no reason. I will change it to use the same name.

src/WebJobs.Extensions.DurableTask/Bindings/ActorTriggerAttributeBindingProvider.cs
```diff
@@ -0,0 +1,164 @@
...
+
+            private static string ActorContextToString(DurableActorContext arg)
+            {
+                var history = JArray.FromObject(arg.History);
+
+                // TODO figure out what exactly is needed here
+                var contextObject = new JObject(
+                    new JProperty("history", history),
+                    new JProperty("actor", ((IDurableActorContext)arg).Self),
+                    new JProperty("isReplaying", arg.IsReplaying));
```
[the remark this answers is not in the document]
sebastianburckhardt: Yes, support for out-of-proc is not yet implemented in this PR.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActorContext.cs
```diff
@@ -0,0 +1,121 @@
...
+        string IDurableActorContext.OperationName
+        {
+            get
+            {
+                this.ThrowIfInvalidAccess();
+                return this.CurrentOperation.Operation;
+            }
+        }
+
+        ILogger IDurableActorContext.Logger => throw new NotImplementedException();
```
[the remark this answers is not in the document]
sebastianburckhardt: This was meant to make it convenient to trace things from actors. However, I did not actually implement it yet. I think I will just remove it for now, it is not essential for this PR. We can always add it back later.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActorContext.cs
```diff
@@ -0,0 +1,121 @@
...
+            this.ThrowIfInvalidAccess();
+            return this.CurrentOperation.GetContent(argumentType);
+        }
+
+        IStateView<TState> IDurableActorContext.GetStateAs<TState>(Formatting formatting, JsonSerializerSettings settings)
+        {
+            this.ThrowIfInvalidAccess();
+            if (this.State.CurrentStateView != null)
+            {
+                if (this.State.CurrentStateView is IStateView<TState> view)
```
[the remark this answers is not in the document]
sebastianburckhardt: Sure. The point is that applications can choose to use multiple different .NET types to create different views of the same JSON object. This can happen when applications are migrating between versions, for example. Also, applications can create views on part of the state, rather than the whole state, by customizing the JSON serialization settings so they work on a subset of the fields of the object.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActorContext.cs
```diff
@@ -0,0 +1,121 @@
...
+            return newView;
+        }
+
+        void IDurableActorContext.Return<TResult>(TResult result)
+        {
+            this.ThrowIfInvalidAccess();
+            this.CurrentOperationResponse.SetResult(result);
+        }
+
+        void IDurableActorContext.Return(object result, Type resultType)
```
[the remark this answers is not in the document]
sebastianburckhardt: In my own framework I am always passing the type because it is required for some serialization settings to work correctly (e.g. TypenameHandling.Automatic). However, since this is not how it is done anywhere else in your APIs, nor is that setting really supported, I agree that I should simplify it as you suggest.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
```diff
@@ -0,0 +1,450 @@
...
+        {
+            return this.NewGuid();
+        }
+
+        /// <inheritdoc/>
+        void IDeterministicExecutionContext.SignalActor(ActorId actor, string operationName, object operationContent)
+        {
+            this.ThrowIfInvalidAccess();
+            var alreadyCompletedTask = this.CallDurableTaskFunctionAsync<object>(actor.ActorClass, FunctionType.Actor, true, TaskActorShim.GetSchedulerIdFromActorId(actor), operationName, null, operationContent);
+            var ignoredValue = alreadyCompletedTask.Result; // just so we see exceptions during testing
```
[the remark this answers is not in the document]
sebastianburckhardt: Sure.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
```diff
@@ -0,0 +1,450 @@
...
+        Guid IDeterministicExecutionContext.NewGuid()
+        {
+            return this.NewGuid();
+        }
+
+        /// <inheritdoc/>
+        void IDeterministicExecutionContext.SignalActor(ActorId actor, string operationName, object operationContent)
+        {
+            this.ThrowIfInvalidAccess();
+            var alreadyCompletedTask = this.CallDurableTaskFunctionAsync<object>(actor.ActorClass, FunctionType.Actor, true, TaskActorShim.GetSchedulerIdFromActorId(actor), operationName, null, operationContent);
```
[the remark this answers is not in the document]
sebastianburckhardt: Sure. Is `System.Diagnostics.Debug.Assert` o.k. to appear in this project? I noticed them in various places so I assume they are. However, they can be problematic in service environments so I wanted to confirm.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
```diff
@@ -0,0 +1,450 @@
...
+                    var request = new OperationMessage()
+                    {
+                        ParentInstanceId = this.InstanceId,
+                        Id = guid,
+                        IsSignal = oneWay,
+                        Operation = operation,
+                    };
+                    request.SetContent(input);
+                    var jrequest = JToken.FromObject(request, MessagePayloadDataConverter.DefaultSerializer);
+                    this.InnerContext.SendEvent(target, "op", jrequest);
```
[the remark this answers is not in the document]
sebastianburckhardt: This was meant to solve the problem with type names appearing in the JSON. However, I can no longer repro that problem, so this solution is no longer necessary.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
```diff
@@ -0,0 +1,450 @@
...
+            catch (Exception e)
+            {
+                exception = e;
+                throw;
+            }
+            finally
+            {
+                if (exception != null && this.InnerContext.IsReplaying)
+                {
+                    // If this were not a replay, then the activity function trigger would have already
```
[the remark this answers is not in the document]
sebastianburckhardt: I am not sure I fully understand the comment, but I think it should apply to all three (orchestrator, activity, actor). I don't have a particularly good understanding of how the tracing works so I was mostly just trying to reuse existing code with minimal changes.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/DurableContextExtensions.cs
sebastianburckhardt: Which API browsing experience are you referring to?
-  I tested the normal Intellisense, it works great (i.e. I get the full API list, including the extension methods, when typing a dot after a context-typed expression in VS).  
- Inspection in the object browser in VS is definitely not nice, not just because of the extension class, but also because of the many layers of interfaces.
- Not sure if there are any automatically generated API documentations.

C# 8 default interfaces would definitely be great for this.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableActorContext.cs
sebastianburckhardt: As mentioned above, removing this for now.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IInterleavingContext.cs
```diff
@@ -0,0 +1,149 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// Provides functionality available to orchestration code and reentrant actor code.
+    /// </summary>
+    public interface IInterleavingContext : IDeterministicExecutionContext
```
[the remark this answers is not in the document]
sebastianburckhardt: Yes, the name `IInterleavingContext` is not very good, I would like to change it for something less confusing. I couldn't really come up with a good one yet though.

I am not sure about `IAsynchronousExecutionContext` - it seems a bit approximate because even the actor operations can still make asynchronous calls to activities.  Also,  "asynchronous" already has a lot of meanings (e.g. Async suffix for methods returning tasks) which can be confusing.

The point of this interface is to capture functionality that is common between normal orchestrations and "actor-orchestrations", which is really what the "reentrant/always-interleave/non-locked" operations are... they are orchestrations that run inside an actor, and have a similar API to normal orchestrations, but plus the actor stuff, and minus some stuff that only makes sense for classic standalone orchestrations (get input, continue as new instance id, parent instance id, set custom status). 

Perhaps we introduce the term "actor-orchestration" (instead of calling them interleaving operations) and then use `IOrchestrationContextCommon`. Not sure.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -203,8 +209,14 @@ void INameVersionObjectManager<TaskOrchestration>.Add(ObjectCreator<TaskOrchestr
         /// <returns>An orchestration shim that delegates execution to an orchestrator function.</returns>
         TaskOrchestration INameVersionObjectManager<TaskOrchestration>.GetObject(string name, string version)
         {
-            var context = new DurableOrchestrationContext(this, name);
-            return new TaskOrchestrationShim(this, context);
+            if (name.StartsWith("@"))
```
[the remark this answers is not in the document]
sebastianburckhardt: That's right, yes. I will validate it in the binding. 

Also, I will validate that instance ids passed to StartNewOrchestration do not start with @.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -517,6 +531,41 @@ internal void DeregisterActivity(FunctionName activityFunction)
             }
         }
 
+        internal void RegisterActor(FunctionName actorFunction, RegisteredFunctionInfo actorInfo)
+        {
+            if (!this.knownActors.TryUpdate(actorFunction, actorInfo, null))
```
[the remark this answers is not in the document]
sebastianburckhardt: Ok, I think I see what you mean. I see small changes to RegisterOrchestrator in the latest version of v2. I have copied them over.

src/WebJobs.Extensions.DurableTask/Listener/TaskActorShim.cs
```diff
@@ -0,0 +1,323 @@
...
+    internal class TaskActorShim : TaskCommonShim
+    {
+        private readonly DurableActorContext context;
+
+        private TaskCompletionSource<object> shutDownSignal
+            = new TaskCompletionSource<object>();
+
+        private Task lastStartedOperation = Task.CompletedTask;
+
+        private Dictionary<Guid, Task> currentReentrantRequests = new Dictionary<Guid, Task>();
```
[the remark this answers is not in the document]
sebastianburckhardt: right, this is left over from a preliminary attempt to support reentrant operations which are however not implemented yet. I will remove this.

src/WebJobs.Extensions.DurableTask/Listener/TaskActorShim.cs
```diff
@@ -0,0 +1,323 @@
...
+
+        public override DurableCommonContext Context => context;
+
+        public string SchedulerId { get; private set; }
+
+        public ActorId ActorId { get; private set; }
+
+        public static string GetSchedulerIdFromActorId(ActorId actor)
+        {
+            return $"@{@actor.ActorClass}@{actor.ActorKey}";
```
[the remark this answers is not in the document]
sebastianburckhardt: hmm, not sure how that got there. It is not needed.

src/WebJobs.Extensions.DurableTask/Listener/TaskActorShim.cs
sebastianburckhardt: Sure, I will add some comments.
"busy" means an operation is currently in progress, e.g. waiting for an activity to complete. You can thus have a busy actor but no other operations waiting in the queue.

src/WebJobs.Extensions.DurableTask/Listener/TaskActorShim.cs
sebastianburckhardt: This is a strange trick I copied from your code: calling using reflection forces the completion to be executed inline, i.e. the same thread will continue at the await. Not sure why it works, but I was assuming you knew. I actually meant to ask you about this.

(I tried without that but ran into trouble because the DTFx framework expects all the orchestrator actions to be produced before RaiseEvents returns).

Of course there is still a bug here, I should not be doing this twice, the last line is superfluous.

src/WebJobs.Extensions.DurableTask/Listener/TaskActorShim.cs
```diff
@@ -0,0 +1,323 @@
...
+                this.Context.InstanceId,
+                eventName,
+                this.Config.GetIntputOutputTrace(serializedEventData),
+                this.Context.IsReplaying);
+
+            if (eventName == "op")
+            {
+                var operationMessage = JsonConvert.DeserializeObject<OperationMessage>(serializedEventData);
+
+                if (this.lastStartedOperation.IsCompleted && this.remainingReceiveLimit > 0)
```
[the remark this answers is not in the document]
sebastianburckhardt: yes, there is an interesting design question here. I will add comments. 

The point is that we need to prevent the history from growing too large. I currently do that by counting how many requests were received, and calling continueAsNew after handling 20 requests. But we probably want to take another look at this heuristic once we get some perf experiments going.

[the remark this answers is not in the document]
sebastianburckhardt: I think I will rename this to `numberOperationsProcessed` which should be more readable.

src/WebJobs.Extensions.DurableTask/Listener/TaskActorShim.cs
```diff
@@ -0,0 +1,323 @@
...
+            // destruct actor state if application code requested it
+            if (destructOnExit)
+            {
+                this.context.State.ActorExists = false;
+                this.context.State.CurrentStateView = null;
+                this.context.State.ActorState = null;
+            }
+
+            // start next request waiting in the queue if there is one
+            if (this.context.State.TryDequeue(out var operationMessage))
```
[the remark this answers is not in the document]
sebastianburckhardt: The actor is always immediately destructed, whether there are operations pending or not (see the code after `if (destructOnExit` which erases the actor state).

The only thing that is delayed is the call to ContinueAsNew. Note that ContinueAsNew is not observable from the actor semantics viewpoint... it is simply a way to manage history size in the DTFx framework. The idea here is that if you are deleting an actor, you probably also want to delete the history, but you don't necessarily have to do that right away.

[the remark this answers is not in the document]
sebastianburckhardt: I will rewrite/comment this code a bit to make the intention clearer.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActivityContext.cs
```diff
@@ -1,145 +1,145 @@
...
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// The default parameter type for activity functions.
+    /// </summary>
+    internal class DurableActivityContext : IDurableActivityContext
```
kashimiz: How does moving from `DurableActivityContextBase` to `IDurableActivityContext` affect unit testing, ex. the example unit tests in our samples?

src/WebJobs.Extensions.DurableTask/Listener/TaskActorShim.cs
```diff
@@ -0,0 +1,323 @@
...
+            {
+                this.lastStartedOperation = this.ProcessRequest(operationMessage);
```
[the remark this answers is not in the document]
sebastianburckhardt: There is an upper limit on the number of operations that are processed between continueAsNew calls (default is 20), so this should not matter.

I am moving that test to this location so it is more clear when looking at the code.

src/WebJobs.Extensions.DurableTask/Listener/TaskActorShim.cs
```diff
@@ -0,0 +1,323 @@
...
+                    return $"Busy ({this.context.State.Queue.Count} Queued)";
+                }
+                else
+                {
+                    return "Busy";
+                }
+            }
+        }
+
+        private void SignalShutdown()
```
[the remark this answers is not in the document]
sebastianburckhardt: Right. I will rename this to `SignalContinueAsNew`, which is really about controlling the history size. The actor scheduler orchestration never actually terminates in the current design, it only ever calls ContinueAsNew.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActorContext.cs
```diff
@@ -0,0 +1,121 @@
...
+            return this.CurrentOperation.GetArgument<TArgument>();
+        }
+
+        object IDurableActorContext.GetArgument(Type argumentType)
+        {
+            this.ThrowIfInvalidAccess();
+            return this.CurrentOperation.GetArgument(argumentType);
+        }
+
+        IStateView<TState> IDurableActorContext.GetStateAs<TState>(Formatting formatting, JsonSerializerSettings settings)
```
kashimiz: I think `TState` can be simplified to `T` per the convention Chris mentioned in his comments.

src/WebJobs.Extensions.DurableTask/TriggerAttributes/ActorTriggerAttribute.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <remarks>
+        /// If not specified, the function name is used as the name of the actor class.
+        /// </remarks>
+        /// <value>
+        /// The name of the actor class or <c>null</c> to use the function name.
+        /// </value>
+#pragma warning disable CS0618 // Type or member is obsolete
+        [AutoResolve]
+#pragma warning restore CS0618 // Type or member is obsolete
+        public string ActorClassName { get; set; }
```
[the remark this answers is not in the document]
sebastianburckhardt: Yes, I have gone back and forth between the two, actually. I think I ultimately chose class because it lets me talk about "instances" of an actor when documenting the behavior.

src/WebJobs.Extensions.DurableTask/Bindings/ActorTriggerAttributeBindingProvider.cs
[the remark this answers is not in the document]
kashimiz: See [my comment below.](https://github.com/Azure/azure-functions-durable-extension/pull/663#discussion_r268363286)

src/WebJobs.Extensions.DurableTask/TriggerAttributes/ActorTriggerAttribute.cs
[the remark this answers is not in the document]
sebastianburckhardt: Yes. This will need to be extended when we implement out-of-proc. Also, both actors and orchestrations should be able to call StartNewOrchestration.

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -257,11 +254,15 @@ private async Task ProcessAsyncActions(AsyncAction[][] actions)
                         case AsyncActionType.CallSubOrchestratorWithRetry:
                             tasks.Add(this.context.CallSubOrchestratorWithRetryAsync(action.FunctionName, action.RetryOptions, action.InstanceId, action.Input));
                             break;
+                        case AsyncActionType.CallActor:
+                            var actorId = TaskActorShim.GetActorIdFromSchedulerId(action.InstanceId);
+                            tasks.Add(((IInterleavingContext)this.context).CallActorAsync(actorId, action.ExternalEventName, action.Input));
+                            break;
```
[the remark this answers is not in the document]
sebastianburckhardt: Yes. This will need to be extended when we implement out-of-proc. Also, both actors and orchestrations should be able to call StartNewOrchestration.

test/Common/TestActivities.cs
sebastianburckhardt: Yes, this is actually the one which I was working on last week. I have a test / example implemented in a different branch but I am not sure I should add it to this PR.

src/WebJobs.Extensions.DurableTask/Bindings/ActorTriggerAttributeBindingProvider.cs
```diff
@@ -0,0 +1,164 @@
...
+                    context.Executor,
+                    FunctionType.Actor);
+                return Task.FromResult<IListener>(listener);
+            }
+
+            private static string ActorContextToString(DurableActorContext arg)
+            {
+                var history = JArray.FromObject(arg.History);
+
+                // TODO figure out what exactly is needed here
```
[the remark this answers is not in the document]
sebastianburckhardt: Yes. The idea is that they are all derived from the history, i.e. actors use the exact same deterministic replay mechanism as is used for orchestrations.

src/WebJobs.Extensions.DurableTask/ActorScheduler/ActorRef.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// <param name="actorKey">the actor key.</param>
+        public ActorRef(string actorClass, string actorKey)
+        {
+            if (string.IsNullOrEmpty(actorClass))
+            {
+                throw new ArgumentException("invalid actor reference: actor class must not be a null or empty string", actorClass);
+            }
+
+            this.ActorClass = actorClass;
+            this.ActorKey = actorKey;
```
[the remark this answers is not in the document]
sebastianburckhardt: Yes, I am adding a null check for actorKey.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActivityContext.cs
sebastianburckhardt: It breaks everything, I am afraid.

I have already updated all the unit tests in the solution, though of course there may be a lot of other places that I am not aware of.

src/WebJobs.Extensions.DurableTask/Listener/TaskActorShim.cs
```diff
@@ -0,0 +1,323 @@
...
+                    this.context.Name,
+                    this.context.InstanceId,
+                    this.context.State.ToString(),
+                    continueAsNew,
+                    FunctionType.Orchestrator,
+                    this.context.IsReplaying);
+
+                if (!this.context.IsReplaying)
+                {
+                    this.context.AddDeferredTask(() => this.Config.LifeCycleNotificationHelper.OrchestratorCompletedAsync(
```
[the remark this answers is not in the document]
sebastianburckhardt: ok.

BTW, in current implementation, continueAsNew is always true.

src/WebJobs.Extensions.DurableTask/ActorScheduler/SchedulerState.cs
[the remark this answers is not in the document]
sebastianburckhardt: I have removed the custom serialization code.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDeterministicExecutionContext.cs
sebastianburckhardt: I see now that I missed a lot of those all over the place. I will check more carefully.

src/WebJobs.Extensions.DurableTask/ActorScheduler/SchedulerState.cs
sebastianburckhardt: The custom JSON serialization has now been removed, and we are in fact using [JsonProperty] attributes.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
```diff
@@ -0,0 +1,450 @@
...
+                throw new ArgumentNullException(nameof(retryOptions));
+            }
+
+            return this.CallDurableTaskFunctionAsync<TResult>(functionName, FunctionType.Activity, false, null, null, retryOptions, input);
+        }
+
+        /// <inheritdoc/>
+        bool IDeterministicExecutionContext.IsLocked(out IReadOnlyList<string> ownedLocks)
+        {
+            throw new NotImplementedException();
```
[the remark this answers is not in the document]
sebastianburckhardt: This is really just a placeholder for functionality yet to be implemented. I have added a comment that clarifies that.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
```diff
@@ -0,0 +1,450 @@
...
+                            input);
+                    }
+
+                    break;
+
+                case FunctionType.Actor:
+                    System.Diagnostics.Debug.Assert(!string.IsNullOrEmpty(operation), "The operation parameter is required.");
+                    System.Diagnostics.Debug.Assert(retryOptions == null, "retries are not supported for actor calls");
+
+                    var guid = this.NewGuid(); // deterministically replayable unique id for this request
```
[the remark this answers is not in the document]
sebastianburckhardt: Yes, indeed.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableActorContext.cs
sebastianburckhardt: I have added remarks to this method (as well as two others) that clarify the meaning of operation (name, content).

sebastianburckhardt:
Just to track it somewhere: here is a list of open problems or unimplemented pieces that we may need to or want to address (many of these probably in a separate PR). 

- Implement out-of-proc 
- Implement locking API
- Write more tests and samples
- Implement StartNewOrchestration support
- Figure out how to get actor responses back to clients
- Figure out how to avoid message duplication in the storage provider
- Add easy logging with ILogger
- Explore heuristics and design for ContinueAsNew decision
- Add support for interleaving actor operations, finalize naming and API around this

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IInterleavingContext.cs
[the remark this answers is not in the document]
cgillum: Yeah, I'm not sure either, so I'm inclined to just leave it as-is for now.

cgillum:
Hi @sebastianburckhardt is this PR ready to be merged? If so, I can take care of it now.

sebastianburckhardt:
Yes, I think it is ready.

Issue #652, opened 2019-03-14
Allow storage providers other than DurableTask.AzureStorage


cgillum:
The Durable Task Framework allows using an extensible set of storage providers. The currently supported providers include:

* [DurableTask.AzureStorage](https://www.nuget.org/packages/Microsoft.Azure.DurableTask.AzureStorage/)
* [DurableTask.ServiceBus](https://www.nuget.org/packages/Microsoft.Azure.DurableTask.ServiceBus/)
* [DurableTask.Emulator](https://www.nuget.org/packages/Microsoft.Azure.DurableTask.Emulator)

An official Service Fabric provider is also currently under development.

However, the Durable Task extension for Azure Functions is hardcoded to use **DurableTask.AzureStorage**. This is problematic particularly for those who want to run Durable Functions in a non-Azure environment (which some refer to as "on-premises" or "the Edge").

This work item tracks allowing users to choose their own storage provider, whether it's one of the official supported providers listed above (P1), or an arbitrary provider that a user may have developed themselves (P2).

Related issue (host.json changes): https://github.com/Azure/azure-functions-durable-extension/issues/641

olitomlinson:
A cosmosDb provider would be really great. 

Reason being Azure storage replication is async, so it is possible that writes could go missing in the event of a failover to another region, whereas CosmosDb has guaranteed replication of writes.

SimonLuckenuik:
SQL Server provider would be interesting, so that we can easily leverage a provider with built-in backup\restore.

Issue #641, opened 2019-03-12
Host.json configuration should be grouped


cgillum:
The number of durable-specific settings in host.json are fairly large. We should group them to make it easier to continue adding more.

There's also a desire to be able to choose different storage providers, in which case provider-specific settings will need to be distinguished from each other.  A (brainstorm) proposal:

```json
{
  "durableTask": {
    "hubName": "MyTaskHub",
    "provider": {
        "azureStorage": {
            "controlQueueBatchSize": 32,
            "partitionCount": 4,
            "controlQueueVisibilityTimeout": "00:05:00",
            "workItemQueueVisibilityTimeout": "00:05:00",
            "maxConcurrentActivityFunctions": 10,
            "maxConcurrentOrchestratorFunctions": 10,
            "connectionStringName": "AzureWebJobsStorage",
        }
    },
    "tracing": {
        "traceInputsAndOutputs": false,
        "traceReplayEvents": false,
    },
    "notifications": {
        "eventGrid": {
            "topicEndpoint": "https://topic_name.westus2-1.eventgrid.azure.net/api/events",
            "keySettingName":  "EventGridKey",
            "publishRetryCount": 3,
            "publishRetryInterval": "00:00:30",
            "publishEventTypes": ["Started", "Pending", "Failed", "Terminated"]
        }
    }
  }
}
```

CarlosSardo:
Hi @cgillum 

I would like to help out by picking up this issue, if possible.

Thanks!

cgillum:
@CarlosSardo that would be super helpful! Connor has already done the storage provider part but I would love help with the others. If you'd like to submit a PR, this work should currently target the v2 branch.

CarlosSardo:
@cgillum @ConnorMcMahon 

Here's a [draft PR](https://github.com/CarlosSardo/azure-docs/pull/1) to the [preview docs](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-preview#hostjson-schema), regarding this breaking change. It can be submitted, once you decide to ship it.

cgillum:
This change has been merged and will go out with our `2.0.0-beta1` release.

CarlosSardo:
The related PR with updates to the **official docs** can be found [here](https://github.com/MicrosoftDocs/azure-docs/pull/34753).

PR #664, opened 2019-03-22
Support for storage providers other than Azure Storage

ConnorMcMahon:
Proof of concept for #652 with DurableTask.Emulator provider. Will eventually add work for #641 as well. 

Only added tests for the emulator provider to a few tests for now, as I didn't want to effectively double the amount of time it takes for the CI to run.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Concurrent;
 using System.Linq;
-using System.Net;
 using System.Net.Http;
 using System.Reflection;
 using System.Text;
@@ -18,7 +17,6 @@ using Microsoft.Azure.WebJobs.Host;
 using Microsoft.Azure.WebJobs.Host.Config;
 using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Azure.WebJobs.Logging;
-using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Newtonsoft.Json.Linq;
@@ -53,11 +51,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private readonly AsyncLock taskHubLock = new AsyncLock();
 
-        private INameResolver nameResolver;
-        private AzureStorageOrchestrationService orchestrationService;
-        private AzureStorageOrchestrationServiceSettings orchestrationServiceSettings;
+        private readonly IOrchestrationServiceFactory orchestrationServiceFactory;
+        private readonly INameResolver nameResolver;
+        private IOrchestrationService orchestrationService;
         private TaskHubWorker taskHubWorker;
-        private IConnectionStringResolver connectionStringResolver;
         private bool isTaskHubWorkerStarted;
 
         /// <summary>
@@ -66,16 +63,15 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// <param name="options">The configuration options for this extension.</param>
         /// <param name="loggerFactory">The logger factory used for extension-specific logging and orchestration tracking.</param>
         /// <param name="nameResolver">The name resolver to use for looking up application settings.</param>
-        /// <param name="connectionStringResolver">The resolver to use for looking up connection strings.</param>
+        /// <param name="orchestrationServiceFactory">The factory used to create orchestration service based on the configured storage provider.</param>
         public DurableTaskExtension(
             IOptions<DurableTaskOptions> options,
             ILoggerFactory loggerFactory,
             INameResolver nameResolver,
-            IConnectionStringResolver connectionStringResolver)
+            IOrchestrationServiceFactory orchestrationServiceFactory)
         {
             this.Options = options.Value;
             this.nameResolver = nameResolver ?? throw new ArgumentNullException(nameof(nameResolver));
-            this.connectionStringResolver = connectionStringResolver ?? throw new ArgumentNullException(nameof(connectionStringResolver));
 
             if (loggerFactory == null)
             {
@@ -87,6 +83,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.TraceHelper = new EndToEndTraceHelper(logger, this.Options.LogReplayEvents);
             this.HttpApiHandler = new HttpApiHandler(this, logger);
             this.LifeCycleNotificationHelper = this.CreateLifeCycleNotificationHelper();
+            this.orchestrationServiceFactory = orchestrationServiceFactory;
         }
 
         internal DurableTaskOptions Options { get; }
@@ -144,8 +141,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             context.AddBindingRule<ActorTriggerAttribute>()
                 .BindToTrigger(new ActorTriggerAttributeBindingProvider(this, context, this.TraceHelper));
 
-            this.orchestrationServiceSettings = this.GetOrchestrationServiceSettings();
-            this.orchestrationService = new AzureStorageOrchestrationService(this.orchestrationServiceSettings);
+            this.orchestrationService = this.orchestrationServiceFactory.GetOrchestrationService();
+
             this.taskHubWorker = new TaskHubWorker(this.orchestrationService, this, this);
             this.taskHubWorker.AddOrchestrationDispatcherMiddleware(this.OrchestrationMiddleware);
         }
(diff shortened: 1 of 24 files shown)


src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -344,97 +339,13 @@ protected internal virtual DurableOrchestrationClient GetClient(OrchestrationCli
...
-                        // as the host when possible to ensure we any send operations can be picked up
-                        // immediately instead of waiting for the next queue polling interval.
-                        innerClient = this.orchestrationService;
-                    }
-                    else
-                    {
-                        innerClient = new AzureStorageOrchestrationService(settings);
-                    }
-
+                    var innerClient = this.orchestrationServiceFactory.GetOrchestrationClient(attribute);
```
cgillum: I prefer we use actual class names in variable decorations when the right-hand expression doesn't already contain it. It makes the code a bit more understandable.

src/WebJobs.Extensions.DurableTask/IOrchestrationServiceFactory.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
```
cgillum: This new file needs a copyright header

src/WebJobs.Extensions.DurableTask/Options/AzureStorageOptions.cs
```diff
@@ -0,0 +1,122 @@
+﻿using System;
```
cgillum: This new file and other new files all need copywrite headers.

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
```diff
@@ -27,7 +26,8 @@ public static IWebJobsBuilder AddDurableTask(this IWebJobsBuilder builder)
 
             builder.AddExtension<DurableTaskExtension>()
                 .BindOptions<DurableTaskOptions>()
-                .Services.AddSingleton<IConnectionStringResolver, WebJobsConnectionStringProvider>();
+                .Services.AddSingleton<IConnectionStringResolver, WebJobsConnectionStringProvider>()
+                         .AddSingleton<IOrchestrationServiceFactory, OrchestrationServiceFactory>();
```
cgillum: With this model, do you think it would be possible for customers to provide their own `IOrchestrationServiceFactory` in the future via our DI support?

src/WebJobs.Extensions.DurableTask/IOrchestrationServiceFactory.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Collections.Generic;
+using System.Text;
+using DurableTask.Core;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask.Options;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    public interface IOrchestrationServiceFactory
```
cgillum: All public types and members need to be documented with XML doc comments. I believe StyleCop and/or the compiler should be flagging this. If it's too much work to do it now, let's open an issue to track adding these (or a generic issue to track fixing all warnings).

src/WebJobs.Extensions.DurableTask/Options/AzureStorageOptions.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// </summary>
+        /// <value>Maximum interval for polling control and work-item queues.</value>
+        public TimeSpan MaxQueuePollingInterval { get; set; } = TimeSpan.FromSeconds(30);
+
+
+        internal override void ValidateHubName(string hubName)
+        {
+            try
+            {
+                NameValidator.ValidateBlobName(hubName);
```
cgillum: In your other PR, we're making it so that the task hub name has no default. In that case, should we be checking for a null-or-empty `hubName` here?

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -27,46 +23,10 @@ public class DurableTaskOptions
...
-        public int PartitionCount { get; set; } = 4;
-
-        /// <summary>
-        /// Gets or sets the visibility timeout of dequeued control queue messages.
-        /// </summary>
-        /// <value>
-        /// A <c>TimeSpan</c> configured by the host. The default is 5 minutes.
-        /// </value>
-        public TimeSpan ControlQueueVisibilityTimeout { get; set; } = TimeSpan.FromMinutes(5);
+        /// TODO: move this back into DurableTaskOptions.cs
```
cgillum: Is this `TODO` valid?

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -244,27 +169,13 @@ public class DurableTaskOptions
...
-        /// <value>Maximum interval for polling control and work-item queues.</value>
-        public TimeSpan MaxQueuePollingInterval { get; set; } = TimeSpan.FromSeconds(30);
-
         // Used for mocking the lifecycle notification helper.
         internal HttpMessageHandler NotificationHandler { get; set; }
 
         internal string GetDebugString()
         {
             var sb = new StringBuilder(4096);
             sb.AppendLine("Initializing extension with the following settings:");
```
cgillum: What is the plan to get the other Options types to participate in populating this debug string?

src/WebJobs.Extensions.DurableTask/Options/EmulatorStorageOptions.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Options
+{
+    public class EmulatorStorageOptions : CommonStorageProviderOptions
+    {
+    }
```
cgillum: Add a coment inside the class definition explaining that there are no settings.

src/WebJobs.Extensions.DurableTask/Options/StorageProviderOptions.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        internal CommonStorageProviderOptions GetConfiguredProvider()
+        {
+            if (this.configuredProvider == null)
+            {
+                var storageProviderOptions = new CommonStorageProviderOptions[] { this.AzureStorage, this.Emulator };
+                var activeProviders = storageProviderOptions.Where(provider => provider != null);
+                if (activeProviders.Count() != 1)
+                {
+                    throw new InvalidOperationException("Must have one and only one storage provider configured.");
```
cgillum: I would rephrase this slightly: `There must be exactly one storage provider configured.`

src/WebJobs.Extensions.DurableTask/Options/EmulatorStorageOptions.cs
ConnorMcMahon: Instead I added a remark in the xml that there is currently no configuration available for the Emulator storage provider. If you still think a comment inside the class definition is necessary, I will add one.

src/WebJobs.Extensions.DurableTask/Options/AzureStorageOptions.cs
ConnorMcMahon: The implementations of `CommonStorageProviderOptions.ValidateHubName()` are just going to be the storage provider specific limitations on task name. I restored the null check in the `DurableTaskOptions.Validate()` method, as that is a universal requirement.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
ConnorMcMahon: I think I have addressed this now.

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
ConnorMcMahon: I'm not quite sure what our customer facing DI support will look like, and especially how they will interact with extensions. I don't see why this approach wouldn't allow customers to provide their own `IOrchestrationServiceFactory` though.

test/FunctionsV2/TestDataGenerator.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
+{
+    public class TestDataGenerator
```
cgillum: This is pretty cool. I did not know about this.

cgillum:
Looks like this PR needs to be updated to fix some merge conflicts, most likely caused by Sebastian's changes.

PR #675, opened 2019-03-26
Add Functions v1 test covereage for basic Event Grid scenario

cgillum:
This also fixes some StyleCop regressions from previous PRs.

The change appears to be bigger than it really is because it also fixes newline inconsistencies that were introduced in other PRs. Turning off whitespace comparisons should help make it readable.

### DIFF
--- a/test/Common/DurableTaskLifeCycleNotificationTest.cs
+++ b/test/Common/DurableTaskLifeCycleNotificationTest.cs
@@ -32,7 +32,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
         }
 
         [Theory]
-        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory + "_BVT")]
         [InlineData(true)]
         [InlineData(false)]
         public async Task OrchestrationStartAndCompleted(bool extendedSessionsEnabled)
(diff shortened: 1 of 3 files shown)


test/Common/DurableTaskLifeCycleNotificationTest.cs
```diff
@@ -1,1340 +1,1340 @@
...
-        [InlineData(true)]
-        [InlineData(false)]
-        public async Task OrchestrationStartedOptOutOfEvent(bool extendedSessionsEnabled)
-        {
-            var testName = nameof(this.OrchestrationStartAndCompleted);
-            var functionName = nameof(TestOrchestrations.SayHelloInline);
-            var eventGridKeyValue = "testEventGridKey";
+
+        [Theory]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory + "_BVT")]
```
ConnorMcMahon: Out of curiosity, what is the difference between `PlatformSpecificHelpers.TestCategory` and `PlatformSpecificHelpers.TestCategory + "_BVT"`.

cgillum: It has to do with how we set up our CI to run automated tests. Here is a screenshot:

image

Basically, anything that has `_BVT` at the end of it will be run in both the V1 and V2 test projects. Anything without this will only run in the V2 test projects. We had to do this because the V1 tests were running extremely slowly over time, and becoming unreliable. So basically, we only run the most important tests against Functions V1.

Issue #676, opened 2019-03-27
Events added to in-memory queue are not being raised after ContinueAsNew called.


revoxover:
Perhaps I misunderstood the functionality added in version 1.8.0 to address the "Race conditions causing external events to be dropped" issues but I expected the following scenario to work.

I have a Orchestration running with the following code:
```
[FunctionName("RunOrchestrator")]
public static async Task RunOrchestrator(
	[OrchestrationTrigger] DurableOrchestrationContext context,
	ILogger log)
{

	//wait for event
	log.LogWarning($"Waiting for event.");
	var raisedEvent = await context.WaitForExternalEvent<string>("Event");

	//handle event
	log.LogWarning($"Event caught: {raisedEvent}");
	await context.CallActivityAsync("HandleEvent", raisedEvent);

	//call continue
	log.LogWarning($"Calling ContinueAsNew().");
	context.ContinueAsNew(null);
}
```

In the "HandleEvent" activity I have an **await Task.Delay(10000);** to simulate some work being done. When I raise events in quick succession only the first event is handled and the others are added to an in memory queue as expected with the following output message:

_"Function 'RunOrchestrator (Orchestrator)' saved a 'Event' event to an in-memory queue. State: ExternalEventDropped."_

However once the first handled event completes and **ContinueAsNew()** is called the events that are in the queue are not being raised once the **context.WaitForExternalEvent** line is hit again.

Is this scenario supposed to work? Thanks in advance.

cgillum:
Yes, this scenario is supposed to work. We have a test intended to validate it [here](https://github.com/Azure/azure-functions-durable-extension/blob/cf6b0b37ad646fe5c65d0965653b0cb136763174/test/Common/DurableTaskEndToEndTests.cs#L627).

I'll try modifying our test to better match yours and see if I can reproduce the issue.

cgillum:
I confirmed that adding the delay like you mentioned seems to cause an issue. I'll investigate further and see if we can make a fix.

revoxover:
Thanks you, much appreciated.

cgillum:
This fix has been released in the [v1.8.1 release](https://github.com/Azure/azure-functions-durable-extension/releases/tag/v1.8.1).

Issue #646, opened 2019-03-13
End-to-end tests have become very flakey


cgillum:
The automated end-to-end tests have become very flakey. A wide variety of tests seem to be randomly failing due to orchestrations either not starting or failing to complete. This needs to be investigated to see if there could be a regression. Anecdotally, the failures mostly seem to involve timer messages, but this is not always the case.

Recent test failures:
* WaitForExternalEventWithTimeout ([1](https://ci.appveyor.com/project/appsvc/azure-functions-durable-extension/builds/23431033/job/anroksi3e00uh78l/tests), [2](https://ci.appveyor.com/project/appsvc/azure-functions-durable-extension/builds/23432577/job/kns1e0m6hl3cwmo1/tests))
* VerifyUniqueGuids ([1](https://ci.appveyor.com/project/appsvc/azure-functions-durable-extension/build/job/6p5cq1fh3hoioa3e/tests))
* RaiseEventToSubOrchestration ([1](https://ci.appveyor.com/project/appsvc/azure-functions-durable-extension/build/job/nempcb4pu3k8qe14/tests))

olitomlinson:
Agreed, I was getting this a hell of a lot when running the tests locally in VS2017  against remote Azure Storage.

cgillum:
@olitomlinson that's actually good to know that you observed this locally. I'll try to reproduce it locally as well as see if I can better debug the issue.

blackalbum:
I'm having this issue too. When I start End to End tests, it fails due to the fact that it cannot find the local azure storage connection string. Any solution?

cgillum:
@blackalbum Is that a flakiness issue? Or do the tests just not run at all? Did you follow the local testing instructions in the [contribution guide](https://github.com/Azure/azure-functions-durable-extension/blob/dev/CONTRIBUTING.md#running-the-tests-visual-studio)?

The issue I'm observing is that sometimes orchestrations don't seem to start or don't seem to complete in the specified timeout. Looking through the Azure Storage logs and history tables, it seems like some of the internal messages are being heavily delayed, but I have yet to understand why.

cgillum:
I did not mean to close this - the tests are still flakey.

TsuyoshiUshio:
Once Azure DevOps guy fix the Anonymous PR issue, how about moving Azure DevOps and use replay feature?

cgillum:
[...]

PR #690, opened 2019-04-05
Ability to carry over buffered external events on ContinueAsNew

cgillum:
Fixes https://github.com/Azure/azure-functions-durable-extension/issues/676 (Events added to in-memory queue are not being raised after ContinueAsNew called)

Included in this PR are also some test fixes which may help resolve https://github.com/Azure/azure-functions-durable-extension/issues/646 (integration test flakiness)

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContextBase.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContextBase.cs
@@ -455,6 +455,22 @@ namespace Microsoft.Azure.WebJobs
         public abstract void ContinueAsNew(object input);
 
         /// <summary>
+        /// Restarts the orchestration by clearing its history.
+        /// </summary>
+        /// <remarks>
+        /// <para>Large orchestration histories can consume a lot of memory and cause delays in
+        /// instance load times. This method can be used to periodically truncate the stored
+        /// history of an orchestration instance.</para>
+        /// </remarks>
+        /// <param name="input">The JSON-serializeable data to re-initialize the instance with.</param>
+        /// <param name="preserveUnprocessEvents">
+        /// If set to <c>true</c>, re-adds any unprocessed external events into the new execution
+        /// history when the orchestration instance restarts. If <c>false</c>, any unprocessed
+        /// external events will be discarded when the orchestration instance restarts.
+        /// </param>
+        public virtual void ContinueAsNew(object input, bool preserveUnprocessEvents) => throw new NotImplementedException();
+
+        /// <summary>
         /// Sets the JSON-serializeable status of the current orchestrator function.
         /// </summary>
         /// <remarks>
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContextBase.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContextBase.cs
@@ -463,12 +463,12 @@ namespace Microsoft.Azure.WebJobs
         /// history of an orchestration instance.</para>
         /// </remarks>
         /// <param name="input">The JSON-serializeable data to re-initialize the instance with.</param>
-        /// <param name="preserveUnprocessEvents">
+        /// <param name="preserveUnprocessedEvents">
         /// If set to <c>true</c>, re-adds any unprocessed external events into the new execution
         /// history when the orchestration instance restarts. If <c>false</c>, any unprocessed
         /// external events will be discarded when the orchestration instance restarts.
         /// </param>
-        public virtual void ContinueAsNew(object input, bool preserveUnprocessEvents) => throw new NotImplementedException();
+        public virtual void ContinueAsNew(object input, bool preserveUnprocessedEvents) => throw new NotImplementedException();
 
         /// <summary>
         /// Sets the JSON-serializeable status of the current orchestrator function.
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -51,9 +51,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         private readonly ConcurrentDictionary<FunctionName, RegisteredFunctionInfo> knownActivities =
             new ConcurrentDictionary<FunctionName, RegisteredFunctionInfo>();
 
-        private readonly ConcurrentDictionary<string, DurableOrchestrationContext> continueAsNewContexts =
-            new ConcurrentDictionary<string, DurableOrchestrationContext>(StringComparer.OrdinalIgnoreCase);
-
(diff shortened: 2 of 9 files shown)


src/WebJobs.Extensions.DurableTask/DurableOrchestrationContextBase.cs
```diff
@@ -454,6 +454,22 @@ public virtual Task CreateTimer(DateTime fireAt, CancellationToken cancelToken)
...
+        /// instance load times. This method can be used to periodically truncate the stored
+        /// history of an orchestration instance.</para>
+        /// </remarks>
+        /// <param name="input">The JSON-serializeable data to re-initialize the instance with.</param>
+        /// <param name="preserveUnprocessEvents">
+        /// If set to <c>true</c>, re-adds any unprocessed external events into the new execution
+        /// history when the orchestration instance restarts. If <c>false</c>, any unprocessed
+        /// external events will be discarded when the orchestration instance restarts.
+        /// </param>
+        public virtual void ContinueAsNew(object input, bool preserveUnprocessEvents) => throw new NotImplementedException();
```
ConnorMcMahon: Is there a reason we went with a virtual method that throws `NotImplementedException` as opposed to just making the method abstract?

ConnorMcMahon: Is there a reason we used a virtual method here with `NotImplementedException` as opposed to making this an abstract method?

src/WebJobs.Extensions.DurableTask/OrchestrationRuntimeStatus.cs
```diff
@@ -11,6 +11,11 @@ namespace Microsoft.Azure.WebJobs
     /// </summary>
     public enum OrchestrationRuntimeStatus
     {
+        /// <summary>
+        /// The status of the orchestration could not be determined.
+        /// </summary>
+        Unknown = -1,
```
ConnorMcMahon: I take it this is unrelated to this fix, but more related to [this PR ](https://github.com/Azure/durabletask/pull/277#pullrequestreview-223391133)?

test/Common/LogEventTraceListener.cs
```diff
@@ -129,7 +129,7 @@ private static bool ShouldExcludeEvent(TraceEvent traceEvent, IDictionary<string
         public void Stop()
         {
             this.currentSession?.Source.StopProcessing();
-            this.backgroundTraceThread.Join(TimeSpan.FromSeconds(10));
+            this.backgroundTraceThread?.Join(TimeSpan.FromSeconds(10));
```
ConnorMcMahon: Was this causing some exceptions in some cases?

test/Common/TestActivities.cs
```diff
@@ -100,6 +100,11 @@ public static double BindToDouble([ActivityTrigger] double value)
             return httpManagementPayload;
         }
 
+        public static Task Delay([ActivityTrigger] TimeSpan duration)
```
ConnorMcMahon: Is this being used somewhere?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -50,6 +51,9 @@ public class DurableTaskExtension :
         private readonly ConcurrentDictionary<FunctionName, RegisteredFunctionInfo> knownActivities =
             new ConcurrentDictionary<FunctionName, RegisteredFunctionInfo>();
 
+        private readonly ConcurrentDictionary<string, DurableOrchestrationContext> continueAsNewContexts =
```
ConnorMcMahon: This seems to be unused.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContextBase.cs
ConnorMcMahon: NIT: would prefer naming the bool `preserveUnprocessedEvents`.

ConnorMcMahon: What is our plan to document this new method? It seems using this new API is necessary to actually avoid the issue this PR sets out to address, but how will a customer know if they should use this new API or not?

cgillum: Ideally it would be abstract, but that's a breaking change for customers. I made it virtual so that it is non-breaking.

src/WebJobs.Extensions.DurableTask/OrchestrationRuntimeStatus.cs
cgillum: Yes, this is related to the other PR.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContextBase.cs
ConnorMcMahon: Ah, that is because the class is abstract for unit test purposes right?

ConnorMcMahon: We can maybe change it to be abstract when we merge this commit over to V2.

test/Common/LogEventTraceListener.cs
cgillum: The ETW tracing was breaking for me so I had to disable it by commenting out some code. Commenting out the code caused this null-ref. I felt I should leave this null-check in, however, because in theory the `Stop()` method could get called (e.g. via `Dispose()`) before this variable is initialized, so it's actually a correct thing to check.

test/Common/TestActivities.cs
cgillum: Not anymore. I created it to test something. I can remove it.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
cgillum: Thanks, I _knew_ I forgot to delete something but couldn't find it. :) This is a leftover from an alternative fixed I tried (which didn't work and had to be removed).

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContextBase.cs
cgillum: The current name is a typo. I'll fix it based on your suggestion.

cgillum: We'll need to rev the autogenerated API documentation after checking into `master`. I didn't want to make this the default behavior because it could lead to undesirable effects in some circumstances (e.g. a never-ending list of events that keep piling on). The main use case for this will be actors, which means our own code will need to call it. But we'll also suggest it to people who want to do actor-like things and preserve events.

PR #711, opened 2019-04-17
Signal or call actors using interfaces and reflection (step 1)

sebastianburckhardt:
This PR adds a new feature that allows actors to be called or signaled using C# interfaces and reflection. 

On the actor side, incoming operations can be dispatched to the object type representing the actor using reflection, e.g.

```csharp
   ctx.Dispatch<Room>()
```

On the caller side, an operation is chosen by providing a delegate that performs the invocation on an interface, e.g.
```csharp
   ctx.SignalActor<IRoom>(actorId, r => r.MethodToBeCalled(typedArg1, typedArg2))
```
or
```csharp
   var result = await ctx.CallActorAsync<IRoom,int>(actorId, r => r.MethodReturningInteger())
```

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/EntityScheduler/Entity.cs
+++ b/src/WebJobs.Extensions.DurableTask/EntityScheduler/Entity.cs
@@ -1,0 +1,40 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// Statically accessible context for entity operations.
+    /// </summary>
+    public static class Entity
+    {
+        private static readonly AsyncLocal<IDurableEntityContext> EntityContext
+            = new AsyncLocal<IDurableEntityContext>();
+
+        /// <summary>
+        /// The context of the currently executing entity.
+        /// </summary>
+        public static IDurableEntityContext Context => EntityContext.Value;
+
+        /// <summary>
+        /// The key of the currently executing entity.
+        /// </summary>
+        public static string Key => EntityContext.Value.Key;
+
+        /// <summary>
+        /// The entity reference for the currently executing entity.
+        /// </summary>
+        public static EntityId Self => EntityContext.Value.Self;
+
+        internal static void SetContext(IDurableEntityContext context)
+        {
+            EntityContext.Value = context;
+        }
+    }
+}
--- a/src/WebJobs.Extensions.DurableTask/EntityScheduler/Entity.cs
+++ b/src/WebJobs.Extensions.DurableTask/EntityScheduler/Entity.cs
@@ -20,21 +20,25 @@ namespace Microsoft.Azure.WebJobs
         /// <summary>
         /// The context of the currently executing entity.
         /// </summary>
-        public static IDurableEntityContext Context => EntityContext.Value;
-
-        /// <summary>
-        /// The key of the currently executing entity.
-        /// </summary>
-        public static string Key => EntityContext.Value.Key;
-
-        /// <summary>
-        /// The entity reference for the currently executing entity.
-        /// </summary>
-        public static EntityId Self => EntityContext.Value.Self;
+        public static IDurableEntityContext Current => EntityContext.Value;
 
         internal static void SetContext(IDurableEntityContext context)
         {
             EntityContext.Value = context;
+        }
+
+        /// <summary>
+        /// Sets the current context to a mocked context for unit testing.
+        /// </summary>
+        /// <param name="mockContext">The mocked context.</param>
+        public static void SetMockContext(IDurableEntityContext mockContext)
+        {
+            if (mockContext is DurableEntityContext)
+            {
+                throw new InvalidOperationException("Only mocked entity contexts are supported, not real ones.");
+            }
+
+            EntityContext.Value = mockContext;
         }
     }
 }
--- a/src/WebJobs.Extensions.DurableTask/EntityScheduler/TypedInvocationExtensions.cs
+++ b/src/WebJobs.Extensions.DurableTask/EntityScheduler/TypedInvocationExtensions.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Azure.WebJobs
         /// <returns>A task that completes when the dispatched operation has finished.</returns>
         /// <exception cref="AmbiguousMatchException">If there is more than one method with the given operation name.</exception>
         /// <exception cref="MissingMethodException">If there is no method with the given operation name.</exception>
-        /// <exception cref="InvalidCastException">If the method has more than one argument.</exception>
+        /// <exception cref="InvalidOperationException">If the method has more than one argument.</exception>
         /// <remarks>
         /// If the entity's state is null, an object of type <typeparamref name="T"/> is created first. Then, reflection
         /// is used to try to find a matching method. This match is based on the method name
@@ -46,7 +46,7 @@ namespace Microsoft.Azure.WebJobs
             var parameters = method.GetParameters();
             if (parameters.Length > 1)
             {
-                throw new InvalidCastException("only a single argument can be used for operation input");
+                throw new InvalidOperationException("Only a single argument can be used for operation input.");
             }
 
             object[] args;
--- a/src/WebJobs.Extensions.DurableTask/EntityScheduler/TypedInvocationExtensions.cs
+++ b/src/WebJobs.Extensions.DurableTask/EntityScheduler/TypedInvocationExtensions.cs
@@ -1,0 +1,87 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Runtime.Serialization.Formatters.Binary;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// Extends the durable entity context to support reflection-based invocation of entity operations.
+    /// </summary>
+    public static class TypedInvocationExtensions
+    {
+        /// <summary>
+        /// Dynamically dispatches the incoming entity operation using reflection.
+        /// </summary>
+        /// <typeparam name="T">The class to use for entity instances.</typeparam>
+        /// <returns>A task that completes when the dispatched operation has finished.</returns>
+        /// <exception cref="AmbiguousMatchException">If there is more than one method with the given operation name.</exception>
+        /// <exception cref="MissingMethodException">If there is no method with the given operation name.</exception>
+        /// <exception cref="InvalidCastException">If the method has more than one argument.</exception>
+        /// <remarks>
+        /// If the entity's state is null, an object of type <typeparamref name="T"/> is created first. Then, reflection
+        /// is used to try to find a matching method. This match is based on the method name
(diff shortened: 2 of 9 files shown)


sebastianburckhardt:
I agree. This is a bit of a quick hack that does not address all requirements. We should take the time and come up with something that does not only satisfy the C#-to-C# actor scenario, but also language interop.

cgillum:
For the purposes of completing this PR, I suggest we do the following:

1. Put more restrictions on the parameter types for entity operations. My recommendation is that we only allow a single parameter, and it must be a JSON-serializeable parameter. My expectation is that this change will make supporting non-C# clients very simple and will result in nice parity with the programming model for activities and orchestrations - basically we can deserialize the JSON operation input directly into the single argument.

2. Let's move the client-side piece into a separate PR. If we go with suggestion (1) above, then our existing client interfaces will be compatible with typed entities so there isn't as much of an urgent need.

Does this sound like a good next step to you?

sebastianburckhardt:
I agree with this plan, this is what I had in mind as well.

One thing we could consider is to allow multiple JSON-serializable parameters and then accept a matching JSON object, e.g. if the interface looks like this

```  csharp
void SomeCall(int someNumber, SomeJsonSerializableType someObject, DateTime timestamp)
```

then we could accept an incoming JSON payload of the form
```json
{
   "someNumber" : "33",
   "someObject" : { ... },
   "timestamp" : "2019-05-16T16:12:15.678"
}
```
Of course we can always add this later, it would apply only if there is more than one argument.

cgillum:
I think my main concern with the multi-parameter approach is that it could create confusion for users when they have a single parameter. For example, suppose I have the following signature:

```csharp
void SomeCall(SomeJsonSerializableType someObject);
```

If a payload arrives that looks like `{ "someObject": { ... }, "someOtherObject": { ... } }`, do you try to deserialize the whole JSON payload into the `someObject` C# parameter or do you take the value of the embedded `"someObject"` JSON field and deserialize that into the C# parameter? We could have a convention for this, but it might not be obvious to users.

The other issue is that you also create an inconsistency between activity function parameters and entity operation parameters (activity functions can only support a single parameter), and that could also result in confusion.

My vote is to start with the simple single parameter approach and think more about multi-parameter later. :)

sebastianburckhardt:
> do you try to deserialize the whole JSON payload into the someObject C# parameter or do you take the value of the embedded "someObject" JSON field and deserialize that into the C# parameter?

My plan was to not look at the incoming object but only at the signature of the operation to make that decision. If there is one argument, we try to deserialize that type. If there are multiple argument, we write a custom deserializer that deserializes each field according to the expected type.  

But I agree, better to get the simple case done first.

sebastianburckhardt:
Following the plan outlined above and in #763, I have stripped down this PR to "step 1", i.e. the minimum needed to support reflection-based operation dispatch.

src/WebJobs.Extensions.DurableTask/EntityScheduler/TypedInvocationExtensions.cs
```diff
@@ -0,0 +1,87 @@
...
+                System.Reflection.BindingFlags.IgnoreCase
+                | System.Reflection.BindingFlags.Public
+                | System.Reflection.BindingFlags.NonPublic
+                | System.Reflection.BindingFlags.Instance);
+
+            // check that the number of arguments is zero or one
+            var parameters = method.GetParameters();
+            if (parameters.Length > 1)
+            {
+                throw new InvalidCastException("only a single argument can be used for operation input");
```
cgillum: nit: Capitalize "Only" and add a period at the end of the sentence.

src/WebJobs.Extensions.DurableTask/EntityScheduler/TypedInvocationExtensions.cs
```diff
@@ -0,0 +1,87 @@
...
+    public static class TypedInvocationExtensions
+    {
+        /// <summary>
+        /// Dynamically dispatches the incoming entity operation using reflection.
+        /// </summary>
+        /// <typeparam name="T">The class to use for entity instances.</typeparam>
+        /// <returns>A task that completes when the dispatched operation has finished.</returns>
+        /// <exception cref="AmbiguousMatchException">If there is more than one method with the given operation name.</exception>
+        /// <exception cref="MissingMethodException">If there is no method with the given operation name.</exception>
+        /// <exception cref="InvalidCastException">If the method has more than one argument.</exception>
```
cgillum: `InvalidCastException` has a pretty well-known meaning, and I feel like this use of it is confusing. How about `InvalidOperationException`? That one coincidentally has a double meaning in the sense that this is an invalid entity operation definition.

src/WebJobs.Extensions.DurableTask/EntityScheduler/Entity.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// </summary>
+    public static class Entity
+    {
+        private static readonly AsyncLocal<IDurableEntityContext> EntityContext
+            = new AsyncLocal<IDurableEntityContext>();
+
+        /// <summary>
+        /// The context of the currently executing entity.
+        /// </summary>
+        public static IDurableEntityContext Context => EntityContext.Value;
```
cgillum: Other APIs in .NET name static properties like this `Current`. For example, `Thread.CurrentThread`, `ClaimsPrincipal.Current`, and WCF's `OperationContext.Current`, etc. I suggest we follow this convention, and name this class `EntityContext` with a property named `Current` - e.g. `EntityContext.Current`.

src/WebJobs.Extensions.DurableTask/EntityScheduler/Entity.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        /// <summary>
+        /// The key of the currently executing entity.
+        /// </summary>
+        public static string Key => EntityContext.Value.Key;
```
cgillum: Since `Key` and `Self` are already defined in `IDurableEntityContext`, are we really saving much by adding these extra properties? I almost wonder if code is actually more readable if written as `EntityContext.Current.Key` and `EntityContext.Current.Id` instead of `EntityContext.Key` and `EntityContext.Self`.

src/WebJobs.Extensions.DurableTask/EntityScheduler/Entity.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        /// <summary>
+        /// The entity reference for the currently executing entity.
+        /// </summary>
+        public static EntityId Self => EntityContext.Value.Self;
```
cgillum: I know I tend to be picky about names, but I do wonder if `Self` is a little too ambiguous. In languages like Python, `self` is essentially the same thing as the `this` pointer in C-style languages. For that reason, it confuses me when something named `Self` refers to an `EntityId` rather than an `IDurableEntityContext` or even the current entity class instance (e.g. `ChatRoom`). I feel it might be more obvious what the type of the property is if it were instead simply named `Id`.

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -310,6 +310,8 @@ private async Task ProcessOperationRequestAsync(RequestMessage request)
             this.context.DestructOnExit = false;
             this.context.IsCompleted = false;
 
+            Entity.SetContext(this.context);
```
cgillum: Let's add a comment here explaining what we're doing.

src/WebJobs.Extensions.DurableTask/EntityScheduler/Entity.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// Statically accessible context for entity operations.
+    /// </summary>
+    public static class Entity
+    {
+        private static readonly AsyncLocal<IDurableEntityContext> EntityContext
+            = new AsyncLocal<IDurableEntityContext>();
```
cgillum: I appreciate that `AsyncLocal<T>` enables access to the current entity context from within the typed actor implementation. However, like with all static variables, what is the impact to unit testing? Will users be able to mock this?

src/WebJobs.Extensions.DurableTask/EntityScheduler/TypedInvocationExtensions.cs
sebastianburckhardt: yeah, I could not find one that I was completely happy with but `InvalidOperationException` is fine.

src/WebJobs.Extensions.DurableTask/EntityScheduler/Entity.cs
sebastianburckhardt: I am fine with `Current`. This would then be called `Entity.Current` which I think is o.k.

BTW, I prefer to keep the static class name as `Entity` as opposed to changing it to `EntityContext` because I think length matters here , see next comment.

src/WebJobs.Extensions.DurableTask/EntityScheduler/Entity.cs
sebastianburckhardt: I have removed this field from this class.

Are you making the case for changing the property`Self` in `IDurableEntityContext` as well? It would be a breaking change.

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
sebastianburckhardt: Sure.

src/WebJobs.Extensions.DurableTask/EntityScheduler/Entity.cs
sebastianburckhardt: I am not really familiar with the problem or how one would solve that in general. 

I suppose we could use dependency injection as an alternative mechanism for people who want to unit test their code. This is what Orleans does. Not sure how we pull that off though. As far as I know support for dependency injection in AF was just announced, but it may take some work to figure out how to hook it up with the DF piece.

src/WebJobs.Extensions.DurableTask/EntityScheduler/Entity.cs
cgillum: I would prefer we update `IDurableEntityContext` as well. I realize it is a breaking change, though I assume it's not commonly used yet. My hope is that we can have most of the breaking changes done by the next release (more breaking changes in other areas are already coming).

src/WebJobs.Extensions.DurableTask/EntityScheduler/Entity.cs
cgillum: One simple option would be to make `Entity.Current` settable (that or `Entity.SetContext(...))`. Someone writing a unit test could then set the static context themselves to avoid null-references. But of course this has the negative side effect of encouraging developers to set this property in production code as well - so we would have to discourage this in documentation.

Another option would be to allow the entity constructor to take an `IDurableEntityContext` parameter. If present, we could call that constructor and the user could save a reference to it for use in their operation handling logic. That solves the problem as well, though requires the user to write more boilerplate code if they want to enable unit testing.

sebastianburckhardt: I suppose I like the first suggestion better. The second one has some pitfalls (e.g. users store IDurableEntityContext in a field then get SerializationException because they did not mark it as `[JsonIgnore]`) and is a bit counterintuitive since the context is not really part of the actor state.

Perhaps provide `Entity.SetMockContext(…)` to indicate that it is for testing purposes? We could even throw an exception if users use it in a real situation, although I am not sure we need to worry too much about it.

src/WebJobs.Extensions.DurableTask/EntityScheduler/Entity.cs
sebastianburckhardt: Sure, I agree. I will put the rename in a separate PR.

src/WebJobs.Extensions.DurableTask/EntityScheduler/Entity.cs
cgillum: I've used this approach in the past, and I like your implementation of it since it's able to distinguish between a real context object and a mock one.

Issue #712, opened 2019-04-18
State management for actors should be simplified


cgillum:
I'm writing a simple actor test case. I noticed that `IDurableActorContext.GetState<T>` returns an `IStateView<T>`. My first reaction was "Why do we need `IStateView<T>` instead of just `T`"?  It took me a minute, but eventually I reasoned that it must be because we require the user to call the `WriteBack` method in the case where the state object is mutable.

To be honest, the experience felt a bit cumbersome, and I had to stop coding to spend time trying to understand what the expectations were for `IStateView<T>` - e.g. do I have to check if it can be null? What kind of concrete types are there? What are these other methods for? How do I mock this for unit testing and how much work will it be? Etc.

I was wondering if we could simplify the experience a bit. Here is my suggestion:

* Change `public IStateView<T> GetState<T>()` to simply `public T GetState<T>()`. This will simplify the process of accessing state.
* Create a new method `void SetState(object state)` for writing back the state. I think this is more discoverable and clear than `IStateView<T>.WriteBack()`.
* Remove `IStateView<T>`.

I believe these changes will result in a lot less cognative overhead for the developer (fewer methods and types that need to be reasoned about), simplified mocking, and is generally more consistent with existing durable APIs - which I think is important. It's also more likely to be consistent with the experience we create for other languages, which are less typed than .NET.

I think we should consider this for the initial release. I can put together a before/after example if that would help illustrate the increase in simplicity.

sebastianburckhardt:
I agree that we should simplify this since it is causing unnecessary mental overhead. Some minor advantages to the old scheme were:

- for simple types such as the  integer in the counter example, you can write ```state.Value++``` instead of context.SetState(context.GetState<int> + 1). Certainly not a big deal.

- users can exert full  control over when and how things get serialized and deserialized, and can keep this independent of reading and updating the state.  For example,
```csharp
using (var s = ctx.GetState<T>( /* serialization settings */))
{
   /// read and update s as many times as you want
} 
```
This may be useful in some advanced scenarios, for example when using multiple alternative C# types for the same actor, but I agree it is more complicated than necessary for the simple stuff. 

BTW, the user is not really required to call the writeback, we are implicitly calling at the end of the operation. Another reason not to expose this at all.

cgillum:
Regarding this:

> Users can exert full control over when and how things get serialized and deserialized, and can keep this independent of reading and updating the state.

[...]

PR #714, opened 2019-04-18
Simplify actor state management

sebastianburckhardt:
Addresses issue #712.

State management is simplified by eliminating the `TypedStateView` class altogether. Instead, simple `GetState<T>()` and `SetState(obj)` are used.

I also removed the serialization customization settings, for now, to keep things more uniform. We can bring this back later when needed.


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActorContext.cs
```diff
@@ -82,25 +84,42 @@ object IDurableActorContext.GetOperationContent(Type argumentType)
...
+        T IDurableActorContext.GetState<T>()
         {
             this.ThrowIfInvalidAccess();
-            if (this.CurrentStateView != null)
+
+            if (!this.StateWasAccessed)
+            {
+                var result = (this.State.ActorState == null)
+                    ? default(T)
+                    : JsonConvert.DeserializeObject<T>(this.State.ActorState);
```
cgillum: I believe we want to use `MessagePayloadDataConverter.Default` here.

sebastianburckhardt: yes.

Issue #719, opened 2019-04-19
ReadActorState returns incorrect content for large states


sebastianburckhardt:
When querying the state of a large actor from the client, a blob reference is returned instead of the actual state. I believe we should automatically load the blob contents and return them. 

The root cause for this is Azure/durabletask#274.

sebastianburckhardt:
I have implemented a fix in respective PRs: #719 and Azure/durabletask#282 (for the DTFx piece).

With this change, the previously failing test `DurableActor_LargeActor` now passes.

PR #720, opened 2019-04-19
Fix bug in ReadActorState (#719)

sebastianburckhardt:
This bug fix uses the newly proposed support in the AzureStorageOrchestrationService for downloading and decompressing fields in the orchestration status that were too large (see Azure/durabletask#282).


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationClient.cs
```diff
@@ -360,9 +360,20 @@ private async Task<ActorStateResponse<T>> ReadActorState<T>(TaskHubClient client
...
             {
-                var serializedState = state.Input;
+                string serializedState;
 
-                var schedulerState = JsonConvert.DeserializeObject<SchedulerState>(state.Input, MessagePayloadDataConverter.MessageSettings);
+                if (this.client.ServiceClient is AzureStorageOrchestrationService service
+                    && state.Input.StartsWith("http"))
+                {
+                    // the input was compressed... read it back from blob
+                    serializedState = await service.DownloadLargeStatusField(state.Input);
```
cgillum: I had some feedback in your other PR that this API should be suffixed with "Async".

sebastianburckhardt:
I have renamed the method as suggested. We may want to wait with merging until we get DTFx changed, otherwise this breaks the build.

PR #723, opened 2019-04-22
Restore v1 Azure Storage default behavior to v2

ConnorMcMahon:
This PR addresses stops throwing exceptions when the host.json either has `storageProvider` as `null` or `{}`.

The new behavior is to default to `Storage`, which is the same behavior as v1 if no connection string was provided.

### DIFF
--- a/test/Common/TestHelpers.cs
+++ b/test/Common/TestHelpers.cs
@@ -21,6 +21,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
         public const string AzureStorageProviderType = "azure_storage";
         public const string EmulatorProviderType = "emulator";
         public const string LogCategory = "Host.Triggers.DurableTask";
+        public const string EmptyStorageProviderType = "empty";
 
         public static JobHost GetJobHost(
             ILoggerProvider loggerProvider,
@@ -41,7 +42,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
         {
             var durableTaskOptions = new DurableTaskOptions
             {
-                StorageProvider = new StorageProviderOptions(),
+
                 HubName = GetTaskHubNameFromTestName(testName, enableExtendedSessions),
                 TraceInputsAndOutputs = true,
                 EventGridKeySettingName = eventGridKeySettingName,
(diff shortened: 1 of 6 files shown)


test/Common/TestHelpers.cs
```diff
@@ -655,7 +658,7 @@ public IReadOnlyList<Type> GetTypes()
             }
         }
 
-        private class TestNameResolver : INameResolver
+        public class TestNameResolver : INameResolver
```
ConnorMcMahon: Don't need this as public anymore.

test/FunctionsV2/DurableOptionsConfigurationTests.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        public DurableOptionsConfigurationTests(ITestOutputHelper output)
+        {
+            this.output = output;
+            this.loggerProvider = new TestLoggerProvider(output);
+        }
+
+        [Fact]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        public async Task EmptyStorageProviderUsesAzureStorageDefaults()
```
ConnorMcMahon: Could probably reduce the code reuse going on here by adding some helper methods.

test/FunctionsV2/WebJobs.Extensions.DurableTask.Tests.V2.csproj
```diff
@@ -11,6 +11,7 @@
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="4.19.4" />
     <PackageReference Include="Microsoft.Azure.DurableTask.AzureStorage" Version="1.6.0" />
+    <PackageReference Include="Microsoft.Azure.Storage.Blob" Version="10.0.1" />
```
cgillum: This reference makes me a little nervous since it's another flavor of the WindowsAzure.Storage nuget that we're already importing via `Microsoft.Azure.DurableTask.AzureStorage`. Not a big deal, but I wonder if we should avoid adding this dependency to avoid the possibility of conflicts.

ConnorMcMahon: Good point. Just tried the tests without adding that dependency and they still pass, so I will remove this addition.

PR #726, opened 2019-04-24
Actor scheduler overhaul

sebastianburckhardt:
Several fixes and improvements to the new actor scheduler.
- add locks to fix race condition between finishing/queueing operations
- keep list of outgoing messages instead of list of orchestrator actions
- change continue-as-new logic so it always processes exactly the batch of messages being delivered
- remove operation limit
- fix the counting of messages, and the waiting for pending operations

Also, some other changes:
- remove `IDeterministicExecutionContext` from `IDurableActorContext` (since it is not deterministic anymore)
- add more precise tracing for actor messages

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ActorScheduler/ReleaseMessage.cs
+++ b/src/WebJobs.Extensions.DurableTask/ActorScheduler/ReleaseMessage.cs
@@ -15,5 +15,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         [JsonProperty(PropertyName = "id")]
         public string LockRequestId { get; set; }
+
+        public override string ToString()
+        {
+            return $"[Release lock {this.LockRequestId} by {this.ParentInstanceId}]";
+        }
     }
 }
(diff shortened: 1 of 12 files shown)


src/WebJobs.Extensions.DurableTask/ActorScheduler/ReleaseMessage.cs
```diff
@@ -15,5 +15,10 @@ internal class ReleaseMessage
 
         [JsonProperty(PropertyName = "id")]
         public string LockRequestId { get; set; }
+
+        public override string ToString()
+        {
+            return $"[Release lock {this.LockRequestId} by {this.ParentInstanceId}]";
```
cgillum: These will be very helpful!

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActorContext.cs
```diff
@@ -111,5 +123,55 @@ internal override void ThrowIfInvalidAccess()
                 throw new InvalidOperationException("No operation is being processed.");
             }
         }
+
+        internal override void SendActorMessage(OrchestrationInstance target, string eventName, object eventContent)
+        {
+            lock (this)
```
cgillum: The guidance I've seen recommends locking a `private readonly object` variable rather than `this` to avoid the possibility of some external code also locking the same object (and creating a hard-to-debug deadlock). You may want to do the same here as well.

sebastianburckhardt: Right. I think it can simplify this a bit.

Issue #733, opened 2019-04-27
REST APIs for Durable Entities


cgillum:
The patterns of interaction are different between entities and orchestrations. I think it makes sense to give entities their own REST APIs separate from orchestrations. Some ideas:

### Signaling an entity
```http
POST /runtime/webhooks/durabletask/entities/{name}/{key}?op={operation}

(JSON input)
```
Signals the `{name}` entity with key `{key}` with `5` as the operation input. For example:

```http
POST /runtime/webhooks/durabletask/entities/counter/device1?op=add

5
```

### Querying the state of an entity
```http
GET /runtime/webhooks/durabletask/entities/{name}/{key}
```
Returns the state of the entity named `{entity}` with key `{key}` as a JSON payload.

The nice thing about having HTTP APIs like this is that simple interaction scenarios require less coding - i.e. I don't have to write an HTTP trigger function to call the client APIs. It would also enable a simpler integration with the JavaScript client.

sebastianburckhardt:
Should we also support a `DELETE` verb? Not sure if it has much application but it would fit the pattern.

cgillum:
What would calling `DELETE` on an entity do? Remove the state? I think it makes sense to include this.

sebastianburckhardt:
There is some ambiguity about `DELETE`. It could delete just the entity, or delete all the metadata (e.g. queues, locks, scheduler, ...) as well. In this case I think it should only delete the entity state, and maybe we need another API call (more like purge) that destructively (and dangerously) removes all data including history.

Given this complication maybe it is better to keep `DELETE` out of the story for now, so we don't run into misunderstandings until we have the rest of the management functions figured out.

PR #734, opened 2019-04-29
Simple REST API for durable entities

sebastianburckhardt:
A basic implementation of the feature request #733.

Entities have a path of the form 

- GET operations return the state of the entity, as JSON content, or 404 (NotFound) if the entity does not exist.

- POST operations send a signal to the entity and return 202 (Accepted). 
   - The operation name is given by the query argument `?op=opname`. It is optional and defaults to the empty string.
  - The operation content is retrieved from the content. If the content type is 'application/json', we parse it to make sure it is valid JSON before sending the signal. Otherwise, we consider it to be plain string.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
@@ -136,6 +136,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         void IDeterministicExecutionContext.SignalEntity(EntityId entity, string operationName, object operationInput)
         {
             this.ThrowIfInvalidAccess();
+            if (operationName == null)
+            {
+                throw new ArgumentNullException(nameof(operationName));
+            }
+
             var alreadyCompletedTask = this.CallDurableTaskFunctionAsync<object>(entity.EntityName, FunctionType.Entity, true, EntityId.GetSchedulerIdFromEntityId(entity), operationName, null, operationInput);
             System.Diagnostics.Debug.Assert(alreadyCompletedTask.IsCompleted, "signaling entities is synchronous");
             alreadyCompletedTask.Wait(); // just so we see exceptions during testing
@@ -250,7 +255,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     break;
 
                 case FunctionType.Entity:
-                    System.Diagnostics.Debug.Assert(!string.IsNullOrEmpty(operation), "The operation parameter is required.");
+                    System.Diagnostics.Debug.Assert(operation != null, "The operation parameter is required.");
                     System.Diagnostics.Debug.Assert(oneWay || !isEntity, "Entities cannot call entities");
                     System.Diagnostics.Debug.Assert(retryOptions == null, "Retries are not supported for entity calls.");
                     System.Diagnostics.Debug.Assert(instanceId != null, "Entity calls need to specify the target entity.");
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
@@ -124,6 +124,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         void IDurableEntityContext.SignalEntity(EntityId entity, string operationName, object operationInput)
         {
             this.ThrowIfInvalidAccess();
+            if (operationName == null)
+            {
+                throw new ArgumentNullException(nameof(operationName));
+            }
+
             var alreadyCompletedTask = this.CallDurableTaskFunctionAsync<object>(entity.EntityName, FunctionType.Entity, true, EntityId.GetSchedulerIdFromEntityId(entity), operationName, null, operationInput);
             System.Diagnostics.Debug.Assert(alreadyCompletedTask.IsCompleted, "signalling entities is synchronous");
             alreadyCompletedTask.Wait(); // just so we see exceptions during testing
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationClient.cs
@@ -180,7 +180,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private async Task SignalEntityAsync(TaskHubClient client, string hubName, EntityId entityId, string operationName, object operationInput)
         {
-            if (string.IsNullOrEmpty(operationName))
+            if (operationName == null)
             {
                 throw new ArgumentNullException(nameof(operationName));
             }
--- a/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
+++ b/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
@@ -188,28 +188,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     }
                 }
 
-                if (i >= 0 && request.Method == HttpMethod.Post)
-                {
-                    string functionName;
-                    string instanceId = string.Empty;
-
-                    i += OrchestratorsControllerSegment.Length;
-                    nextSlash = path.IndexOf('/', i);
-
-                    if (nextSlash < 0)
-                    {
-                        functionName = path.Substring(i);
-                    }
-                    else
-                    {
-                        functionName = path.Substring(i, nextSlash - i);
-                        i = nextSlash + 1;
-                        instanceId = path.Substring(i);
-                    }
-
-                    return await this.HandleStartOrchestratorRequestAsync(request, functionName, instanceId);
-                }
-
                 i = path.IndexOf(InstancesControllerSegment, StringComparison.OrdinalIgnoreCase);
                 if (i < 0)
                 {
--- a/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
+++ b/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
@@ -23,6 +23,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
     {
         private const string InstancesControllerSegment = "/instances/";
         private const string OrchestratorsControllerSegment = "/orchestrators/";
+        private const string EntitiesControllerSegment = "/entities/";
         private const string TaskHubParameter = "taskHub";
         private const string ConnectionParameter = "connection";
         private const string RaiseEventOperation = "raiseEvent";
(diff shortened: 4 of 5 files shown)


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
```diff
@@ -250,7 +255,7 @@ internal virtual void ThrowIfInvalidAccess()
                     break;
 
                 case FunctionType.Entity:
-                    System.Diagnostics.Debug.Assert(!string.IsNullOrEmpty(operation), "The operation parameter is required.");
+                    System.Diagnostics.Debug.Assert(operation != null, "The operation parameter is required.");
```
ConnorMcMahon: Did something change to allow this to be `""` now?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationClient.cs
```diff
@@ -180,7 +180,7 @@ Task IDurableOrchestrationClient.SignalEntityAsync(EntityId entityId, string ope
 
         private async Task SignalEntityAsync(TaskHubClient client, string hubName, EntityId entityId, string operationName, object operationInput)
         {
-            if (string.IsNullOrEmpty(operationName))
+            if (operationName == null)
```
ConnorMcMahon: Same question here.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
```diff
@@ -136,6 +136,11 @@ Guid IDeterministicExecutionContext.NewGuid()
         void IDeterministicExecutionContext.SignalEntity(EntityId entity, string operationName, object operationInput)
         {
             this.ThrowIfInvalidAccess();
+            if (operationName == null)
```
kashimiz: Wouldn't this allow for empty or whitespace strings to be used as operations?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
kashimiz: Same comment as above re: whitespace/empty string ops.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
```diff
@@ -124,6 +124,11 @@ internal void Writeback()
         void IDurableEntityContext.SignalEntity(EntityId entity, string operationName, object operationInput)
         {
             this.ThrowIfInvalidAccess();
+            if (operationName == null)
```
kashimiz: Same question as in DurableCommonContext: do we want to allow whitespace/empty string ops?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationClient.cs
kashimiz: Same whitespace/empty string question.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -153,6 +154,62 @@ public async Task<HttpResponseMessage> HandleRequestAsync(HttpRequestMessage req
...
+                    {
+                        return await this.HandleGetEntityRequestAsync(request, entityId);
+                    }
+                    else
+                    {
+                        return await this.HandlePostEntityOperationRequestAsync(request, entityId);
+                    }
+                }
+
+                if (i >= 0 && request.Method == HttpMethod.Post)
```
ConnorMcMahon: Will this condition ever be `true`? If the condition at line 158 is true, then this method has already returned. If that condition is false, then `i` is still -1 or request.Method is not `Post`.

test/Common/HttpApiHandlerTests.cs
```diff
@@ -836,6 +836,140 @@ public async Task StartNewInstance_Returns_HTTP_400_On_Missing_Function()
...
+            else
+            {
+                Assert.Equal(HttpStatusCode.NotFound, actualResponse.StatusCode);
+            }
+        }
+
+        [Theory]
+        [InlineData(false, false, false)]
+        [InlineData(false, false, true, true)]
+        [InlineData(false, false, true, false)]
```
kashimiz: Why do some tests rely on the default value of `hasJsonContent` and others explicitly set it to `false`?

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -153,6 +154,62 @@ public async Task<HttpResponseMessage> HandleRequestAsync(HttpRequestMessage req
                     return await this.HandleStartOrchestratorRequestAsync(request, functionName, instanceId);
                 }
 
+                i = path.IndexOf(EntitiesControllerSegment, StringComparison.OrdinalIgnoreCase);
```
ConnorMcMahon: In general, this method is starting to get very difficult to read. @cgillum, at some point we should consider refactoring this method.

When I had to do a similar set of APIs in a functions extension, I wrote it as a series of if statements that were effectively a series of if/else-if statements where each condition checked if the requestUrl matched the expected pattern and that the request method matched expectations. This is almost certainly not as effecient, but it was fairly readable in my opinion.

As we keep adding more and more methods, I think we need some easily expandable and readable approach to handle this problem.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -594,6 +651,63 @@ private static int GetIntQueryParameterValue(NameValueCollection queryStringName
...
+            HttpRequestMessage request,
+            EntityId entityId)
+        {
+            IDurableOrchestrationClient client = this.GetClient(request);
+
+            string operationName = request.GetQueryNameValuePairs()["op"] ?? string.Empty;
+
+            if (request.Content == null)
+            {
+                await client.SignalEntityAsync(entityId, operationName);
```
ConnorMcMahon: What does this do for a non-existent entity?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
sebastianburckhardt: Yes, the idea was to now allow empty strings as operation names.

It makes sense for REST API, and I did not see a strong reason to disallow it in general.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -594,6 +651,63 @@ private static int GetIntQueryParameterValue(NameValueCollection queryStringName
...
+                    {
+                        return request.CreateErrorResponse(HttpStatusCode.BadRequest, "Could not parse JSON content: " + e.Message);
+                    }
+                }
+                else
+                {
+                    operationInput = requestContent;
+                }
+
+                await client.SignalEntityAsync(entityId, operationName, operationInput);
```
ConnorMcMahon: Same question here.

test/Common/HttpApiHandlerTests.cs
```diff
@@ -836,6 +836,140 @@ public async Task StartNewInstance_Returns_HTTP_400_On_Missing_Function()
...
+        [InlineData(false, true, true, true)]
+        [InlineData(false, true, true, false)]
+        [InlineData(true, false, false)]
+        [InlineData(true, false, true, true)]
+        [InlineData(true, false, true, false)]
+        [InlineData(true, true, false)]
+        [InlineData(true, true, true, true)]
+        [InlineData(true, true, true, false)]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        public async Task SignalEntity_Is_Success(bool hasKey, bool hasOp, bool hasContent, bool hasJsonContent = false)
```
ConnorMcMahon: Can we add a separate test (or separate condition on this test) for when an entity doesn't exist?

test/Common/HttpApiHandlerTests.cs
sebastianburckhardt: It is meant to convey that the the fourth argument (`hasJsonContent`) is irrelevant unless the third argument (`hasContent`) is true.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
sebastianburckhardt: Same answer.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
sebastianburckhardt: Again, same answer.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationClient.cs
sebastianburckhardt: Same answer again.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
ConnorMcMahon: Looks like you answered this above.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
sebastianburckhardt: entities get created automatically when they receive a signal or call.

test/Common/HttpApiHandlerTests.cs
sebastianburckhardt: The behavior of a signal is identical (whether the entity exists or not) so I don't think we need a separate test.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
sebastianburckhardt: Just that the REST API made me rethink the original design. I don't see a strong reason why we need to insist on an operation name being specified.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
sebastianburckhardt: I agree. At a minimum, create subhandlers for each prefix to avoid a super-long function body.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
sebastianburckhardt: Good catch - I must have accidentally duplicated the first paragraph of the function while editing.

This just goes to confirm that we should refactor this function.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
cgillum: I agree we should refactor this at some point. I think I went down this path originally because I figured it would be simple and highly optimized, but it ended up being not very maintainable. I opened an issue for tracking: https://github.com/Azure/azure-functions-durable-extension/issues/735

PR #741, opened 2019-04-30
Add support for Redis provider

ConnorMcMahon:


src/WebJobs.Extensions.DurableTask/OrchestrationServiceFactory.cs
```diff
@@ -179,12 +182,46 @@ public EmulaterOrchestrationServiceFactory(DurableTaskOptions options)
...
+                this.defaultTaskHubService = new RedisOrchestrationService(new RedisOrchestrationServiceSettings()
+                {
+                    TaskHubName = options.HubName,
+                    RedisConnectionString = this.redisConnectionString,
+                });
+            }
+
+            public IOrchestrationServiceClient GetOrchestrationClient(OrchestrationClientAttribute attribute)
+            {
+                if (string.IsNullOrEmpty(attribute.TaskHub))
```
cgillum: In addition to this, if a `TaskHub` value is specified and its the same name as `defaultTaskHubService.HubName`, we should also return `this.defaultTaskHubService`.

cgillum:
Looks like we need to add **DurableTask.Redis** to the myget feed.

src/WebJobs.Extensions.DurableTask/OrchestrationServiceFactory.cs
ConnorMcMahon: Good catch. That may have been the reason behind one of the scenarios not working that I thought should have.

PR #743, opened 2019-05-02
Add sample project using Redis storage provider

ConnorMcMahon:


samples/v2/redis/HttpStart.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace VSSample
+{
+    public static class HttpStart
+    {
+        [FunctionName("HttpStart")]
+        public static async Task<HttpResponseMessage> Run(
+            [HttpTrigger(AuthorizationLevel.Function, methods: "post", Route = "orchestrators/{functionName}")] HttpRequestMessage req,
+            [OrchestrationClient] IDurableOrchestrationClient starter,
+            string functionName,
```
cgillum: Since this sample only contains one orchestration, I suggest we hardcode a value for `functionName` to make it easier to run.

samples/v2/redis/HelloSequence.cs
```diff
@@ -0,0 +1,39 @@
...
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs;
+
+namespace VSSample
+{
+    public static class HelloSequence
+    {
+        [FunctionName("E1_HelloSequence")]
```
cgillum: I suggest we remove the "E1_" prefix from all of these. They were really only inteded for csx samples where I wanted file names to be ordered in the Azure portal. Since this is a fresh sample, that legacy probably shoudn't be carried forward.

samples/v2/redis/HelloSequence.cs
```diff
@@ -0,0 +1,39 @@
...
+            outputs.Add(await context.CallActivityAsync<string>("E1_SayHello", "Tokyo"));
+            outputs.Add(await context.CallActivityAsync<string>("E1_SayHello", "Seattle"));
+            outputs.Add(await context.CallActivityAsync<string>("E1_SayHello_DirectInput", "London"));
+
+            // returns ["Hello Tokyo!", "Hello Seattle!", "Hello London!"]
+            return outputs;
+        }
+
+        [FunctionName("E1_SayHello")]
+        public static string SayHello([ActivityTrigger] IDurableActivityContext context)
```
cgillum: Let's remove this overload and just use the "direct input" version. However, let's not call it "_DirectInput". I don't think these distinctions are important for the purposes of this simple sample.

Issue #318, opened 2018-05-23
Support for HttpRequest/IActionResult in HTTP-specific APIs


YodasMyDad:
The latest Azure functions code has moved away from using `Task<HttpResponseMessage>` and `HttpRequestMessage req` and has moved to this format

```
        public static async Task<IActionResult> Run(
                        [HttpTrigger(AuthorizationLevel.Function, "get", "post", Route = null)]
                        HttpRequest req,
                        TraceWriter log)
```

Returning an IActionResult and HttpRequest instead of HttpRequestMessage

gled4er:
Hello @cgillum,

I am interested in helping with this issue.

Do you have any recommendation or design guidelines apart from the description above?

Thank you!

cgillum:
Now that Functions 2.0 is GA, I think it's the perfect time to revisit this one.  I was originally thinking we needed to use `#ifdef` for this, but actually we might be able to reference the `Microsoft.AspNetCore.Http.Abstractions` for both the .NET 4.6.1 and the .NET Standard 2.0 implementations.  So basically, we will have two sets of HTTP APIs, one set that works with .NET Framework HTTP APIs and one that works with .NET Core HTTP APIs.

cgillum:
Bumping this one back up. We're planning a Durable Functions 2.0 release, where we drop support for Functions 1.0 and fully embrase Functions 2.0+. This means we can actually replace our legacy HTTP APIs with the new ones (`HttpRequest`, `IActionResult`, etc.)

davet44:
convert HttpResponseMessage to 
IActionResult

PR #748, opened 2019-05-08
Add .NET Core friendly HTTP types to IDurableOrchestrationClient

ConnorMcMahon:
#318

### DIFF
--- a/test/Common/DurableOrchestrationClientBaseTests.cs
+++ b/test/Common/DurableOrchestrationClientBaseTests.cs
@@ -3,14 +3,20 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
 using System.Threading.Tasks;
 using DurableTask.Core;
 using FluentAssertions;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
 using Microsoft.Azure.WebJobs.Extensions.DurableTask.Options;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
+using Microsoft.Extensions.Primitives;
 using Moq;
 using Xunit;
+using static Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests.HttpApiHandlerTests;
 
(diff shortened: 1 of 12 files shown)


test/Common/DurableOrchestrationClientBaseTests.cs
```diff
@@ -90,12 +99,72 @@ public async Task TerminateAsync_NonRunningOrchestrator_ThrowsException()
...
+            httpHandler.InnerClient = durableOrchestrationClient;
+
+            string sampleUrl = "https://samplesite.azurewebsites.net";
+            string sampleId = Guid.NewGuid().ToString();
+
+            var netFrameworkRequest = new HttpRequestMessage(HttpMethod.Get, sampleUrl);
+            HttpRequest netCoreRequest = await ConvertHttpRequestMessageAsync(netFrameworkRequest);
+
+            HttpResponseMessage netFrameworkResponse = durableOrchestrationClient.CreateCheckStatusResponse(netFrameworkRequest, sampleId);
+            HttpResponseMessage netCoreResponse = (HttpResponseMessage)((ObjectResult)durableOrchestrationClient.CreateCheckStatusResponse(netCoreRequest, sampleId)).Value;
```
cgillum: >HttpResponseMessage netCoreResponse [](start = 12, length = 35)

Based on the variable name, I would have expected this to be of type `HttpResponse`. Why do we cast it to `HttpResponseMessage`?

ConnorMcMahon: The easiest way to convert `HttpResponseMessage` into an `IActionResult` (which Controllers in .NET Core implicitly turn into `HttpResponse` I believe) is the following code:

```
var result = new ObjectResult(response);
result.Formatters.Add(new HttpResponseMessageOutputFormatter());
```

With that custom formatter, the Controller logic should automatically apply the `HttpRequestMessage` data to the `HttpResponse`. Since I don't have access to a controller to do that for me, the easiest way to validate is just to make sure that the `HttpResponseMessage` fed into the `ObjectResult` is identical to the `HttpResponseMessage` returned by our old method, and trust that the ` HttpResponseMessageOutputFormatter` does its job correctly.

PR #751, opened 2019-05-09
Add ridesharing sample

sebastianburckhardt:
I added a new ridesharing sample for durable entities that demonstrates two concepts which are a bit more advanced (and can be quite challenging to implement without the new features we added).

- An "Indexing" pattern which allows more efficient querying of entities whose state satisfies some property (in this case, search all users in a particular region)
- Use of a critical section to prevent inconsistencies resulting from races (e.g. two riders matching with the same driver)

I also reorganized the directory structure a bit, adding a entities-csharp subfolder for samples.


cgillum:
Hey @sebastianburckhardt, just checkin in to see if/when you'll be ready to merge this. Folks have been asking for samples and the more the merrier. :)

Issue #753, opened 2019-05-10
Cannot create FunctionFailedException in unit test


DarranShepherd:
I am creating unit tests for my orchestrator function, following the guidance on https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-unit-testing .

I wish to test the logic that handles one of my activity tasks throwing an exception. The orchestrator  handles this with
```C#
catch (FunctionFailedException ex) when (ex.InnerException is MyCustomException)
```

However, I am unable to mock the call to `CallActivityAsync` and have it throw as all constructors of `FunctionFailedException` are internal. As `Exception.InnerException` is not virtual, I also cannot use `Mock.Of<FunctionFailedException>(ex => ex.InnerException == new MyCustomException())`

I have been able to resolve this temporarily by using reflection, but I doubt this is the expected approach. Is this not something that should be tested for some reason, or is there an official way to do so?

```C#
private FunctionFailedException CreateException(string message, Exception inner)
{
    var ctor = typeof(FunctionFailedException).GetConstructor(BindingFlags.NonPublic | BindingFlags.Instance, null, new[] { typeof(string), typeof(Exception) }, null);
    return (FunctionFailedException)ctor.Invoke(new object[] { message, inner });
}
```

cgillum:
I think it makes sense for us to expose those constructors to make things more unit testable. @ConnorMcMahon what do you think?

cgillum:
This fix was released as part of [v1.8.2](https://github.com/Azure/azure-functions-durable-extension/releases/tag/v1.8.2).

PR #756, opened 2019-05-10
Expose FunctionFailedException constructors for easier unit testing

ConnorMcMahon:
Addresses issue #753

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/FunctionFailedException.cs
+++ b/src/WebJobs.Extensions.DurableTask/FunctionFailedException.cs
@@ -21,12 +21,21 @@ namespace Microsoft.Azure.WebJobs
         {
         }
 
-        internal FunctionFailedException(string message)
+        /// <summary>
+        /// Initializes a new instance of a <see cref="FunctionFailedException"/>.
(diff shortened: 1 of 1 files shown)


src/WebJobs.Extensions.DurableTask/FunctionFailedException.cs
```diff
@@ -21,12 +21,21 @@ internal FunctionFailedException()
         {
         }
 
-        internal FunctionFailedException(string message)
+        /// <summary>
+        /// Initializes a new instance of a <see cref="FunctionFailedException"/>.
+        /// </summary>
+        /// <param name="message">A message describing where to look for more details.</param>
+        public FunctionFailedException(string message)
```
ConnorMcMahon: Unsure if this method really should be public. It isn't currently used in our code (but the `[Serializable]` attribute might require the method exists).  Once we expose it as public, we can't remove it without it being a breaking change.

cgillum: It's pretty common to have exception types with public constructors like this. I think it's fine to make this public.

PR #765, opened 2019-05-16
Prevent sub-orchestrations from using the instance ID of the parent

### LINKED ISSUE #747: Throw exception when starting sub-orchestrator with instance ID of the parent
There have been a couple issues reported where users are starting sub-orchestrations using the instance ID of the parent. This causes the parent's history to be overwritten by the sub-orchestration.

We should update the sub-orchestration APIs to check for this condition and throw an exception if the user attempts to start a sub-orchestration with its parent's ID.

cgillum:
Resolves #747

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
@@ -438,6 +438,15 @@ namespace Microsoft.Azure.WebJobs
 
                     break;
                 case FunctionType.Orchestrator:
+                    // Instance IDs should not be reused when creating sub-orchestrations. This is a best-effort
+                    // check. We cannot easily check the full hierarchy, so we just look at the current orchestration
+                    // and the immediate parent.
+                    if (string.Equals(instanceId, this.InstanceId, StringComparison.OrdinalIgnoreCase) ||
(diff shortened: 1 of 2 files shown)


src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -438,6 +438,15 @@ public override void ContinueAsNew(object input, bool preserveUnprocessedEvents)
 
                     break;
                 case FunctionType.Orchestrator:
+                    // Instance IDs should not be reused when creating sub-orchestrations. This is a best-effort
+                    // check. We cannot easily check the full hierarchy, so we just look at the current orchestration
+                    // and the immediate parent.
+                    if (string.Equals(instanceId, this.InstanceId, StringComparison.OrdinalIgnoreCase) ||
+                        (this.ParentInstanceId != null && string.Equals(instanceId, this.ParentInstanceId, StringComparison.OrdinalIgnoreCase)))
```
ConnorMcMahon: Adding a check that the `ParentInstanceId` is not `null` prevents the exception from firing in the scenario where an orchestration without a parent (i.e. a top level orchestration) calls a suborchestration without a specified instance id.

PR #766, opened 2019-05-17
Add legacy compatibility for abstract base classes

sebastianburckhardt:
This is a little change meant to ease the transition to v2.

Rather than breaking v1 orchestration triggers based on `DurableOrchestrationContextBase` and `DurableActivityContextBase`, this change keeps them defined and functional, but marked as obsolete. So users get warnings that tell them what to do, that they can fix at their leisure.

I think this is better than reverting back to the old scheme because I think public interfaces are usually preferable to public base classes as they keep the options open for how things are implemented internally.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActivityContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActivityContext.cs
@@ -10,7 +10,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
     /// <summary>
     /// The default parameter type for activity functions.
     /// </summary>
-    internal class DurableActivityContext : IDurableActivityContext
+    internal class DurableActivityContext : IDurableActivityContext,
+#pragma warning disable 618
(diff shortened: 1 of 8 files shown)


sebastianburckhardt:
I understand the concern with "breaking changes" on extending interfaces in general, but I don't quite see how it applies in this specific case. 

The problem described in https://haacked.com/archive/2008/02/21/versioning-issues-with-abstract-base-classes-and-interfaces.aspx/ is for a different, more complex situation. The problem there is that some users may implement the interfaces in question, and then their implementations break whenever the interface gets extended. But this is not the case for `IDurableOrchestrationContext`, `IDurableOrchestrationClient`, `IDurableActivityContext`, or `IDurableEntityContext`. Nobody outside this dll is expected to ever implement these interfaces. They are just public contracts of implementations provided by this dll, meant purely for consumption (not implementation) by others. So, when we extend those interfaces, we can also extend all their implementations at the same time. 

Users may mock these interfaces in unit tests, but those mocks are also not going to break even if we add new members to the interfaces, as long as we don't call the new members from within the implementation.

PR #773, opened 2019-05-25
Refactor HTTP routing logic

### LINKED ISSUE #735: Refactor HTTP API URL matching code
Per [this discussion](https://github.com/Azure/azure-functions-durable-extension/pull/734#discussion_r279505911), we should refactor this code to make it easier to read and maintain.

ConnorMcMahon:
Resolves #735

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
+++ b/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
@@ -221,17 +221,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
                 if (InstancesRoute.TryMatch(path, routeValues))
                 {
-                    string instanceId = null;
-                    string operation = null;
-                    if (routeValues.ContainsKey(InstanceIdRouteParameter))
-                    {
-                        instanceId = (string)routeValues[InstanceIdRouteParameter];
-                    }
-
-                    if (routeValues.ContainsKey(OperationRouteParameter))
-                    {
-                        operation = (string)routeValues[OperationRouteParameter];
-                    }
+                    routeValues.TryGetValue(InstanceIdRouteParameter, out object instanceIdValue);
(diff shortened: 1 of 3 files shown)


cgillum:
Will this work for Functions 1.0? I noticed you're using types defined in ASP.NET Core namespaces. I'm getting signals that we should reconsider pulling support for Functions 1.0, so if there is a way we could refactor this that is compatible with .NET Framework, then that would be preferable.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -130,107 +176,96 @@ public async Task<HttpResponseMessage> HandleRequestAsync(HttpRequestMessage req
...
-                i = path.IndexOf(InstancesControllerSegment, StringComparison.OrdinalIgnoreCase);
-                if (i < 0)
+                if (InstancesRoute.TryMatch(path, routeValues))
                 {
-                    // Retrieve All Status or conditional query in case of the request URL ends e.g. /instances/
-                    if (request.Method == HttpMethod.Get
-                        && path.EndsWith(InstancesControllerSegment.TrimEnd('/'), StringComparison.OrdinalIgnoreCase))
+                    string instanceId = null;
+                    string operation = null;
+                    if (routeValues.ContainsKey(InstanceIdRouteParameter))
```
cgillum: Are you able to use `TryGetValue` instead? If available, that's always preferred compared to `Compare` + dictionary lookup.

ConnorMcMahon: The reason I think I avoided it the first time around was because the `TryGetValue()` maps to an `object`, not a `string`, so it made the code more awkward since I had to explicitly cast the value anyways.

I think you are right that I can make it cleaner though.

cgillum:
Looks good. Go ahead and merge when ready.

PR #793, opened 2019-06-04
Batch-processing of entity operations for out-of-proc support

sebastianburckhardt:
For out-of-proc, we need to batch multiple operations into a single call (since we cannot change any of the parameters after the binding has taken place, and can therefore not call the function multiple times). 

To support this 

 - I rewrote the scheduler a bit. It now collects all the operations to execute, before executing any of them. This is actually cleaner than what we had before, it gets rid of the concurrency control (no more locking needed).
- I modified the JSON fields on the context object passed to the out-of-proc function:
   - 'batch' is an array of objects describing the operations to execute
   - 'exists' is a boolean indicating whether the entity exists 
   - 'state' is the state of the entity as a string (if `exists`)

- I modified the JSON fields on the result object returned from the out-of-proc function:
   - 'entityExists' indicates whether the entity exists. This is true unless the last operation deleted it.
   - 'entityState' is the state of the entity as a string
   - 'signals' is an array of outgoing signals
   - 'responses' is an array of outgoing responses (one per operation that is not a signal)

I also changed some of the tracing calls: since the entity scheduler (unlike general orchestrations) does not return until the entity function has completed, we do not need the `TraceAwait`. And I lowered the log level for delivery tracing to 'Debug' so it can be suppressed via configuration.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
@@ -371,7 +371,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
                     if (functionType == FunctionType.Entity)
                     {
-                        this.Config.TraceHelper.OperationCompleted(
+                        this.Config.TraceHelper.OperationFailed(
                             this.Config.Options.HubName,
                             functionName,
                             this.InstanceId,
@@ -379,7 +379,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                             operationName,
                             input: "(replayed)",
                             output: "(replayed)",
-                            failed: true,
                             duration: 0,
                             isReplay: true);
                     }
@@ -411,20 +410,19 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                         operationName,
                         input: "(replayed)",
                         output: "(replayed)",
-                        failed: false,
                         duration: 0,
                         isReplay: true);
                 }
                 else
                 {
                     this.Config.TraceHelper.FunctionCompleted(
-                    this.Config.Options.HubName,
-                    functionName,
-                    this.InstanceId,
-                    output: "(replayed)",
-                    continuedAsNew: false,
-                    functionType: functionType,
-                    isReplay: true);
+                        this.Config.Options.HubName,
+                        functionName,
+                        this.InstanceId,
+                        output: "(replayed)",
+                        continuedAsNew: false,
+                        functionType: functionType,
+                        isReplay: true);
                 }
             }
 
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
@@ -75,11 +75,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         internal string RawInput { get; set; }
 
-        internal bool IsCompleted { get; set; }
-
         internal OrchestrationContext InnerContext { get; set; }
-
-        internal ExceptionDispatchInfo OrchestrationException { get; set; }
 
         internal string HubName => this.Config.Options.HubName;
 
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
@@ -368,15 +368,31 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 {
                     // If this were not a replay, then the orchestrator/activity/entity function trigger would have already
                     // emitted a FunctionFailed trace with the full exception details.
-                    this.Config.TraceHelper.FunctionFailed(
-                        this.Config.Options.HubName,
-                        functionName,
-                        this.InstanceId,
-                        operationId,
-                        operationName,
-                        reason: $"(replayed {exception.GetType().Name})",
-                        functionType: functionType,
-                        isReplay: true);
+
+                    if (functionType == FunctionType.Entity)
+                    {
+                        this.Config.TraceHelper.OperationCompleted(
+                            this.Config.Options.HubName,
+                            functionName,
+                            this.InstanceId,
+                            operationId,
+                            operationName,
+                            input: "(replayed)",
+                            output: "(replayed)",
+                            failed: true,
+                            duration: 0,
+                            isReplay: true);
+                    }
+                    else
+                    {
+                        this.Config.TraceHelper.FunctionFailed(
+                            this.Config.Options.HubName,
+                            functionName,
+                            this.InstanceId,
+                            reason: $"(replayed {exception.GetType().Name})",
+                            functionType: functionType,
+                            isReplay: true);
+                    }
                 }
             }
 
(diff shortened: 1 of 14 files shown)


src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -48,6 +45,20 @@ public TaskEntityShim(DurableTaskExtension config, string schedulerId)
...
+        internal List<RequestMessage> OperationBatch => this.operationBatch;
+
+        public void AddOperationToBatch(RequestMessage operationMessage)
+        {
+            this.operationBatch.Add(operationMessage);
+        }
+
+        public void AddLockRequestToBatch(RequestMessage lockRequest)
+        {
+            this.lockRequest = lockRequest;
```
cgillum: >this.lockRequest = lockRequest; [](start = 12, length = 31)

I see that we check for a null lock request before calling this method, but for defense-in-depth should we throw an exception (or assert) here if `this.lockRequest` is non-null?

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -254,19 +159,30 @@ public override async Task<string> Execute(OrchestrationContext innerContext, st
...
 
-        private async Task ProcessRequestAsync(RequestMessage request)
+        public async Task ExecuteBatch()
         {
-            if (request.IsLockMessage)
+            if (this.GetFunctionInfo().IsOutOfProc)
             {
-                this.ProcessLockRequest(request);
+                // process all operations in the batch using a single function call
+                await this.ExecuteOutOfProcBatch();
```
cgillum: >await this.ExecuteOutOfProcBatch(); [](start = 16, length = 35)

If I understand correctly, in .NET we will always have one function execution per operation, but in non-.NET (e.g. JavaScript) there will be one function execution for the entire batch. Is that correct? If so, it worries me a little from a consistency perspective, especially since that will impact billing, metering/timeouts, tracing, etc. Is there a technical reason we decided to do batching for out-of-proc or would it make sense for us to try and preserve the 1:1 operation to function execution?

sebastianburckhardt: It was purely a technical reason (tough one can argue that batching is the better design from a performance and billing viewpoint - for example, the latest EventHubs binding also allows batching).

From what I understood we cannot make multiple out-of-proc calls since we cannot pass different parameters to each call as we do in C#... all parameters must be known upfront at the time we call the `Executor.TryExecuteAsync` because that's when they get serialized into the JSON object.

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
sebastianburckhardt: Yes, that seems like a good idea.

cgillum:
Per our discussion, I think it's okay that the batching behavior of event processing differs between .NET and out-of-proc. However, we should try to keep the tracing consistent between the two worlds so that our tooling and diagnostics doesn't need to be aware of the language differences. The suggestion we landed on was to emit a start trace saying that we're starting a batch, and then emit a single message for the completion of each individual operation (rather than having distinct start and end traces for each operation).

sebastianburckhardt:
I updated the tracing code. Now the function tracing is completely decoupled from the operation tracing. This allowed me to revert some of the changes I made earlier in v2, simplifying the code.

Each executed operation now produces one trace entry, either OperationCompleted or OperationFailed, and this entry includes input, output/exception, and duration in milliseconds. I had to modify the JavaScript interface a bit so it can pass back the duration.

src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
```diff
@@ -341,6 +333,44 @@ private EtwEventSource()
...
+            string OperationId,
+            string OperationName,
+            string Input,
+            string Output,
+            double Duration,
+            string FunctionType,
+            string ExtensionVersion,
+            bool IsReplay)
+                {
+            this.WriteEvent(206, TaskHub, AppName, SlotName, FunctionName, InstanceId, OperationId, OperationName, Input ?? "(null)", Output ?? "(null)", Duration, FunctionType, ExtensionVersion, IsReplay);
```
cgillum: Change `206` to `221` here.

And for the sake of tidiness, we should probably fix the indentation of the `{` right above this line. :)

src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
```diff
@@ -341,6 +333,44 @@ private EtwEventSource()
...
+            string OperationId,
+            string OperationName,
+            string Input,
+            string Exception,
+            double Duration,
+            string FunctionType,
+            string ExtensionVersion,
+            bool IsReplay)
+        {
+            this.WriteEvent(206, TaskHub, AppName, SlotName, FunctionName, InstanceId, OperationId, OperationName, Input ?? "(null)", Exception, Duration, FunctionType, ExtensionVersion, IsReplay);
```
cgillum: Change `206` to `222` here.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
```diff
@@ -372,32 +368,64 @@ internal virtual void ThrowIfInvalidAccess()
...
+                    this.Config.TraceHelper.FunctionCompleted(
                     this.Config.Options.HubName,
                     functionName,
                     this.InstanceId,
-                    operationId,
-                    operationName,
                     output: "(replayed)",
                     continuedAsNew: false,
                     functionType: functionType,
                     isReplay: true);
```
cgillum: nit: need one more level of indentation for the parameter list.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
```diff
@@ -372,32 +368,64 @@ internal virtual void ThrowIfInvalidAccess()
...
-                        this.InstanceId,
-                        operationId,
-                        operationName,
-                        reason: $"(replayed {exception.GetType().Name})",
-                        functionType: functionType,
-                        isReplay: true);
+
+                    if (functionType == FunctionType.Entity)
+                    {
+                        this.Config.TraceHelper.OperationCompleted(
```
cgillum: Should this be `OperationFailed`?

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -388,60 +307,123 @@ private async Task ProcessOperationRequestAsync(RequestMessage request)
...
-                    this.context.CurrentOperation?.Operation ?? string.Empty,
-                    this.context.IsReplaying);
+                [JsonProperty("result")]
+                public string Result { get; set; }
+
+                [JsonProperty("isError")]
+                public bool IsError { get; set; }
+
+                [JsonProperty("duration")]
+                public double Duration { get; set; }
```
cgillum: What is the unit for `Duration`? Can we add a comment, or use the name of the field to make this clear?

sebastianburckhardt: Yes to both. I will add comments to all the fields.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
sebastianburckhardt: It's `OperationCompleted` with `failed:true`. I used the same call since all the parameters align.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
sebastianburckhardt: yes.

src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
sebastianburckhardt: yes.

src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
sebastianburckhardt: yes.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableCommonContext.cs
sebastianburckhardt: I think I will revert that to two separate calls so it is more in line with the rest of the tracing code.

PR #797, opened 2019-06-06
Updated JavaScript samples' dependencies:

kashimiz:
* Microsoft.Azure.WebJobs.Extensions.DurableTask: 1.8.0 -> 1.8.2
* durable-functions: 1.1.2 -> 1.2.2

This update puts the samples on the latest versions of the extension and JS library, taking advantage of several bugfixes and resolving a security vulnerability in the axios library.

### DIFF
--- a/samples/javascript/package-lock.json
+++ b/samples/javascript/package-lock.json
@@ -4,10 +4,28 @@
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
+        "@azure/functions": {
+            "version": "1.0.3",
+            "resolved": "https://registry.npmjs.org/@azure/functions/-/functions-1.0.3.tgz",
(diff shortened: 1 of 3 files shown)


samples/javascript/package-lock.json
```diff
@@ -552,25 +575,20 @@
             }
         },
         "follow-redirects": {
-            "version": "1.7.0",
-            "resolved": "https://registry.npmjs.org/follow-redirects/-/follow-redirects-1.7.0.tgz",
-            "integrity": "sha512-m/pZQy4Gj287eNy94nivy5wchN3Kp+Q5WgUPNy5lJSZ3sgkVKSYV/ZChMAQVIgx1SqfZ2zBZtPA2YlXIWxxJOQ==",
+            "version": "1.5.10",
```
cgillum: Are we intentionally moving backwards on this dependency?

kashimiz: Yes; that's the result of upgrading our axios dependency from 0.18.0 to 0.19.0. They dropped to follow-redirects 1.5.0 between those releases.

PR #640, opened 2019-03-11
Initial work on Durable V2

ConnorMcMahon:
This PR aims to do the following:

- Remove Functions V1 compatability
- Make extension development on VS Code more friendly
- Ensure tests work properly on Unix devices

PR #812, opened 2019-06-18
Restore functions v1 support to Durable 2.0

ConnorMcMahon:
Undo a lot of the work for #640 that removed Functions V1 support.

Note that there are key differences, even among dependencies' versions, due to some interactions between the Functions runtime assembly binding redirects as well as new code paths we have added.

@sebastianburckhardt, one of the actor tests is currently flakey, but only in Functions V1. Would love some help investigating why.


src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -150,6 +213,22 @@ void IExtensionConfigProvider.Initialize(ExtensionConfigContext context)
...
+            context.ApplyConfig(this.Options, "DurableTask");
+
+            ILogger logger = context.Config.LoggerFactory.CreateLogger(LoggerCategoryName);
+
+            this.TraceHelper = new EndToEndTraceHelper(logger, this.Options.Tracing.TraceReplayEvents);
+            this.HttpApiHandler = new HttpApiHandler(this, logger);
+            this.connectionStringResolver = new WebJobsConnectionStringProvider();
+            this.orchestrationServiceFactory = new OrchestrationServiceFactory(new OptionsWrapper<DurableTaskOptions>(this.Options), this.connectionStringResolver);
+            this.LifeCycleNotificationHelper = this.CreateLifeCycleNotificationHelper();
+            this.nameResolver = context.Config.NameResolver;
```
cgillum: Is this change based on the latest, or is it based on the time we forked v2? I recall there was an `ArgumentNullException` caused by the fact that `nameResolver` was initialized after `CreateLifeCycleNotificationHelper()`. The fix was to reorder things so that `nameResolver` was initialized before calling `CreateLifeCycleNotificationHelper()`.

ConnorMcMahon: Hmm, good call. I'll take a look.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -55,12 +60,31 @@ public class DurableTaskExtension :
...
-        private readonly INameResolver nameResolver;
+        private readonly bool isOptionsConfigured;
+
+        private IOrchestrationServiceFactory orchestrationServiceFactory;
+        private INameResolver nameResolver;
         private IOrchestrationService orchestrationService;
         private TaskHubWorker taskHubWorker;
         private bool isTaskHubWorkerStarted;
 
+#if !NETSTANDARD
```
cgillum: Is this a typo? I was expecting we'd only be looking for `NETSTANDARD2_0`.

src/WebJobs.Extensions.DurableTask/DurableTaskWebJobsStartup.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+#if NETSTANDARD2_0
```
cgillum: nit: I wonder if we should move this `#if` up one line. It seems weird to compile just a `using` statement in the case of .NET Framework.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -25,9 +24,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
     internal class HttpApiHandler
     {
         // Route segments
-        private const string InstancesControllerSegment = "/instances/";
-        private const string OrchestratorsControllerSegment = "/orchestrators/";
-        private const string EntitiesControllerSegment = "/entities/";
+        private const string InstancesControllerSegment = "instances/";
+        private const string OrchestratorsControllerSegment = "orchestrators/";
+        private const string EntitiesControllerSegment = "entities/";
```
cgillum: Any particular reason for these changes?

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -85,17 +84,28 @@ public HttpApiHandler(DurableTaskExtension config, ILogger logger)
...
+        private static RouteValueDictionary RouteValueDictionaryFromArray(KeyValuePair<string, object>[] values)
+        {
+            var routeValueDictionary = new RouteValueDictionary();
+            foreach (var pair in values)
+            {
+                routeValueDictionary[pair.Key] = pair.Value;
+            }
+
+            return routeValueDictionary;
+        }
```
cgillum: I assume this code was copied from somewhere. If so, we should add a comment indicating this. Or perhaps just mention why we can't call `RouteValueDictionary.FromArray()`.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0; net471</TargetFrameworks>
```
cgillum: This gives us [full support for .NET Standard 2.0](https://devblogs.microsoft.com/dotnet/announcing-the-net-framework-4-7-1/) correct? And the Azure Functions host supports this framework version?

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
```diff
@@ -52,6 +54,7 @@ public interface IDurableOrchestrationClient
         /// <param name="instanceId">The ID of the orchestration instance to check.</param>
         /// <returns>An HTTP 202 response with a Location header and a payload containing instance control URLs.</returns>
         IActionResult CreateCheckStatusResponse(HttpRequest request, string instanceId);
+#endif
```
cgillum: If we're upgrading our full framework target to .NET Framework 4.7.1, and it supports .NET Standard 2.0, shouldn't we also be able to use these types/methods in Functions v1?

sebastianburckhardt:
The `DurableEntity_BasicObjects` test is definitely flakey, not just in v1. I will need to change this test, it is simply not written quite right (I was lazy)... it depends on message timing.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
ConnorMcMahon: `RouteValueDictionary.FromArray()` doesn't exist in the `Microsoft.AspNetCore.Routing` we use for Functions V1.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
ConnorMcMahon: `TemplateMatcher` throws an exception when routes start with a `/` character in the version of `Microsoft.AspNetCore.Routing` we use in Functions V1.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
ConnorMcMahon: The Functions host does support this framework. Not sure about the "full support" or not.

PR #820, opened 2019-06-26
Add WaitForExternalEvent with CancellationToken overloads

### LINKED ISSUE #816: WaitForExternalEvent with timeout generates an uncancellable timer
I've been using `WaitForExternalEvent` with a timeout in an orchestrator function of mine, and am just getting around to debugging why the orchestration hangs until the event is raised (or the timeout is reached, *even* if I took a route through the orchestration that didn't necessitate it finishing.  The documentation for the non-timeout version of `WaitForExternalEvent` suggests (correctly) that you don't need to cancel those (also #553)...but the timeout overload uses `CreateTimer` inside and while it does the right thing to ensure the token is cancelled if the event is received before the timeout itself is reached...I must now wait for the event to be received, or the timeout to fire.

I think I have a few options.  I could roll the timeout code myself so that I can have access to the `CancellationTokenSource`, but that feels like a cop-out.  I could also detect the scenario and have my orchestration send itself the event, but that also feels like a cop-out.  The option I'm leaning towards is to contribute back a fix for this though, since then everyone can be happy and not have to try to dance around the library. :smile:

What is the right solution here?  Add another set of overloads that allows passing in a `CancellationTokenSource` instance for every overload that takes a `TimeSpan`?  I suspect there's some subtleties to think through, such as who's responsible for disposing it.  If the token source is passed in, then it's probably not the library's job to dispose it anymore maybe?

Assuming this is an API that the extension is willing to take on, I'm happy to think through it and put up a PR for the issue, I just want to make sure you agree it's the right direction.  Thanks!

AtOMiCNebula:
Previously, `WaitForExternalEvent` when invoked with a specified timeout value would create a Timer instance that was uncancelable by the surrounding orchestration.  This could have the practical effect of making orchestrations hang until either the event was received (causing the timer to be cancelled) or the timeout was reached (because the timer expired).

This change adds overloads for `TimeSpan`-accepting overloads of `WaitForExternalEvent` that allow providing a `CancellationToken`, which the calling orchestration can cancel if it no longer intends to wait for the event to be raised.

Fixes #816

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContextBase.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContextBase.cs
@@ -411,6 +411,22 @@ namespace Microsoft.Azure.WebJobs
         public virtual Task WaitForExternalEvent(string name, TimeSpan timeout) => this.WaitForExternalEvent<object>(name, timeout);
 
         /// <summary>
+        /// Waits asynchronously for an event to be raised with name <paramref name="name"/>.
+        /// </summary>
+        /// <remarks>
+        /// External clients can raise events to a waiting orchestration instance using
+        /// <see cref="DurableOrchestrationClient.RaiseEventAsync(string, string, object)"/> with the object parameter set to <c>null</c>.
+        /// </remarks>
+        /// <param name="name">The name of the event to wait for.</param>
+        /// <param name="timeout">The duration after which to throw a TimeoutException.</param>
+        /// <param name="cancelToken">The <c>CancellationToken</c> to use for cancelling the timer.</param>
+        /// <returns>A durable task that completes when the external event is received.</returns>
+        /// <exception cref="TimeoutException">
+        /// The external event was not received before the timeout expired.
+        /// </exception>
+        public virtual Task WaitForExternalEvent(string name, TimeSpan timeout, CancellationToken cancelToken) => throw new NotImplementedException();
+
+        /// <summary>
         /// Waits asynchronously for an event to be raised with name <paramref name="name"/> and returns the event data.
         /// </summary>
         /// <remarks>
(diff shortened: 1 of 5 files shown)


AtOMiCNebula:
I'd still like to add a test here, but wanted to get this up for review to help get a conversation started. :)

AtOMiCNebula:
Test added and verified locally.  Also, I dropped my "Normalize" commit, since I was seeing unexpected behavior with the XML locally; the build was non-deterministically generating different outputs.  I'm leaving that for project owners to clean up, if they choose to. :)

AtOMiCNebula:
I agree it feels a little clunky to me too, but I'm not sure of a better way to go about it either.  The only other (half-baked 😅) idea I had was replacing the `TimeSpan` with a timer from `CreateTimer`, but...replacing the `TimeSpan` with `Task`/`Task<T>` feels even clunkier.

Happy to help with v2 API cleanup for this if you'd like it once this is settled!

AtOMiCNebula:
Ping @cgillum @ConnorMcMahon!  Sorry to ping during OneWeek, but hopefully I've made this an easy-enough review for you. :smile:

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContextBase.cs
```diff
@@ -410,6 +410,22 @@ public virtual Task CreateTimer(DateTime fireAt, CancellationToken cancelToken)
         /// </exception>
         public virtual Task WaitForExternalEvent(string name, TimeSpan timeout) => this.WaitForExternalEvent<object>(name, timeout);
 
+        /// <summary>
+        /// Waits asynchronously for an event to be raised with name <paramref name="name"/>.
+        /// </summary>
+        /// <remarks>
+        /// External clients can raise events to a waiting orchestration instance using
+        /// <see cref="DurableOrchestrationClient.RaiseEventAsync(string, string, object)"/> with the object parameter set to <c>null</c>.
```
ConnorMcMahon: Not sure if this feedback is necessary, as I think it is fairly clear from the context of the signature here, but it might be a good idea to call out explicitly how the timeout and cancellation token interact. If you do decide to add that description to this comment, be sure to change it in the parameterized version of the method below.

AtOMiCNebula: Ah yes, I should have updated the comments here, instead of just adding `<param>`s.  Will update. :)

AtOMiCNebula:
@ConnorMcMahon Are those doc updates any better? :)

AtOMiCNebula:
Ping @ConnorMcMahon / @cgillum!

ConnorMcMahon:
Sorry for the delay @AtOMiCNebula! Thanks very much for the documentation changes, they really add some clarity. You have my approval.

AtOMiCNebula:
Great, thanks for your help reviewing this!  I don't have repo write access, so I need one of you to merge the PR for me. :tada:

gdodd1977:
Do I have to move to 1.0.29 to get this change? I too am having stuck orchestrations due to having multiple tasks in a Task.WhenAny(). I need to complete those tasks once one of them wins.

cgillum:
@gdodd1977 it's not clear from your description whether this PR would help your situation. Would you mind opening a new GitHub issue with a description that includes a code sample?

PR #832, opened 2019-07-02
Updates for v2-beta1 release

cgillum:
- Updated package versions
- Added `Counter` sample for precompiled v2
- Minor fixes for entity HTTP handling


cgillum:
FYI @jeffhollan, we have a build available here if you'd like to play with it before we post it to nuget.org: https://www.myget.org/feed/azure-appservice/package/nuget/Microsoft.Azure.WebJobs.Extensions.DurableTask/2.0.0-beta1

PR #836, opened 2019-07-03
Azure pipelines v2

amdeel:
Adding YAML file for Azure Pipelines

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,23 +1,16 @@
-# ASP.NET Core (.NET Framework)
-# Build and test ASP.NET Core projects targeting the full .NET Framework.
-# Add steps that publish symbols, save build artifacts, and more:
-# https://docs.microsoft.com/azure/devops/pipelines/languages/dotnet-core
-      
-trigger: none
+trigger:
+- v2
 
 jobs:
-  - job: BuildAndTest
 
+  - job: FunctionsV1Tests
     pool: 
       vmImage: 'windows-latest'
-
-    strategy:
-      parallel: 10
 
     variables:
       solution: 'WebJobs.Extensions.DurableTask.sln'
       buildPlatform: 'Any CPU'
-      buildConfiguration: 'Release'
+      buildConfiguration: 'Debug'
 
     steps:
     - task: NuGetToolInstaller@0
@@ -32,26 +25,69 @@ jobs:
     - task: VSBuild@1
       inputs:
         solution: '$(solution)'
-        msbuildArgs: '/p:DeployOnBuild=true /p:WebPublishMethod=Package /p:PackageAsSingleFile=true /p:SkipInvalidConfigurations=true'
         platform: '$(buildPlatform)'
         configuration: '$(buildConfiguration)'
 
     - task: VSTest@2
       inputs:
         testSelector: 'testAssemblies'
-        testAssemblyVer2: 'test/**'
-        testFiltercriteria: 'TestCategory!=DisabledInCI'
-        distributionBatchType: basedOnTestCases
+        testAssemblyVer2: |
+          **/*tests.v1.dll
+        testFiltercriteria: 'Category=FunctionsV1_BVT'
         diagnosticsEnabled: True
         rerunFailedTests: true
         rerunFailedThreshold: 10
         rerunMaxAttempts: 2
       env:
-        AzureWebJobsStorage: $(durablecitest)
+        AzureWebJobsStorage: $(AzureWebJobsStorage)
 
+  - job: FunctionsV2Tests
+    pool: 
+      vmImage: 'windows-latest'
 
+    strategy:
+      parallel: 9
+
+    variables:
+      solution: 'WebJobs.Extensions.DurableTask.sln'
+      buildPlatform: 'Any CPU'
+      buildConfiguration: 'Debug'
+
+    steps:
+    - task: NuGetToolInstaller@0
+
+    - task: DotNetCoreCLI@2
+      inputs:
+        command: 'restore'
+        projects: '**/**/*.csproj'
+        feedsToUse: 'config'
+        nugetConfigPath: '.nuget/nuget.config'
+
+    - task: VSBuild@1
+      inputs:
+        solution: '$(solution)'
+        platform: '$(buildPlatform)'
+        configuration: '$(buildConfiguration)'
+
+    - task: VSTest@2
+      inputs:
+        testSelector: 'testAssemblies'
+        testAssemblyVer2: |
+          **/*tests.v2.dll
+        testFiltercriteria:
+        distributionBatchType: basedOnExecutionTime
+        diagnosticsEnabled: True
+        rerunFailedTests: true
+        rerunFailedThreshold: 10
+        rerunMaxAttempts: 2
+      env:
+        AzureWebJobsStorage: $(AzureWebJobsStorage)
+
+    
   - job: PublishPipelineArtifact
-    dependsOn: BuildAndTest
+    dependsOn:
+      - FunctionsV1Tests
+      - FunctionsV2Tests
 
     pool: 
       vmImage: 'windows-latest'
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -0,0 +1,121 @@
+trigger:
+- v2
+
+jobs:
+
+  - job: V1Test
```
cgillum: I suggest we name this `Functions 1.0 Tests` or something similar. `V1` by itself can be a bit ambiguous.

azure-pipelines.yml
```diff
@@ -0,0 +1,121 @@
...
+      inputs:
+        command: 'restore'
+        projects: '**/**/*.csproj'
+        feedsToUse: 'config'
+        nugetConfigPath: '.nuget/nuget.config'
+
+    - task: VSBuild@1
+      inputs:
+        solution: '$(solution)'
+        msbuildArgs: '/p:DeployOnBuild=true /p:WebPublishMethod=Package /p:PackageAsSingleFile=true /p:SkipInvalidConfigurations=true'
```
cgillum: I wonder if we really need any of these `msbuildArgs`. They look like they might be just for web projects.

azure-pipelines.yml
```diff
@@ -0,0 +1,121 @@
...
+        platform: '$(buildPlatform)'
+        configuration: '$(buildConfiguration)'
+
+    - task: VSTest@2
+      inputs:
+        testSelector: 'testAssemblies'
+        testAssemblyVer2: |
+          **/*tests.v1.dll
+        testFiltercriteria: 'Category=FunctionsV1_BVT'
+        distributionBatchType: basedOnTestCases
```
cgillum: Since these are not run in parallel, maybe we should remove `distributionBatchType` from here?

azure-pipelines.yml
```diff
@@ -0,0 +1,121 @@
...
+        diagnosticsEnabled: True
+        rerunFailedTests: true
+        rerunFailedThreshold: 10
+        rerunMaxAttempts: 2
+      env:
+        AzureWebJobsStorage: $(durablecitest)
```
cgillum: Is `durablecitest` the name of the variable for the connection string? I wonder if we should name it `AzureWebJobsStorage` to match the environment variable that we're populating.

azure-pipelines.yml
```diff
@@ -0,0 +1,121 @@
...
+
+  - job: V2Test
```
cgillum: Related to my previous comment, I wonder if this should be `Functions 2.0 Tests`.

azure-pipelines.yml
```diff
@@ -0,0 +1,121 @@
...
+      inputs:
+        command: 'restore'
+        projects: '**/**/*.csproj'
+        feedsToUse: 'config'
+        nugetConfigPath: '.nuget/nuget.config'
+
+    - task: VSBuild@1
+      inputs:
+        solution: '$(solution)'
+        msbuildArgs: '/p:DeployOnBuild=true /p:WebPublishMethod=Package /p:PackageAsSingleFile=true /p:SkipInvalidConfigurations=true'
```
cgillum: Same comment here about `msbuildArgs`.

azure-pipelines.yml
```diff
@@ -0,0 +1,121 @@
...
+        testAssemblyVer2: |
+          **/*tests.v2.dll
+        testFiltercriteria:
+        distributionBatchType: basedOnExecutionTime
+        diagnosticsEnabled: True
+        rerunFailedTests: true
+        rerunFailedThreshold: 10
+        rerunMaxAttempts: 2
+      env:
+        AzureWebJobsStorage: $(durablecitest)
```
cgillum: Same comment about the variable name.

azure-pipelines.yml
```diff
@@ -0,0 +1,121 @@
...
+      inputs:
+        command: 'restore'
+        projects: '**/**/*.csproj'
+        feedsToUse: 'config'
+        nugetConfigPath: '.nuget/nuget.config'
+
+    - task: VSBuild@1
+      inputs:
+        solution: '$(solution)'
+        msbuildArgs: '/p:DeployOnBuild=true /p:WebPublishMethod=Package /p:PackageAsSingleFile=true /p:SkipInvalidConfigurations=true /p:DesktopBuildPackageLocation="$(build.artifactStagingDirectory)\WebApp.zip" /p:DeployIisAppPath="Default Web Site"'
```
cgillum: Same comment about the `msbuildArgs`.

azure-pipelines.yml
```diff
@@ -0,0 +1,121 @@
...
+jobs:
+
+  - job: V1Test
+    pool: 
+      vmImage: 'windows-latest'
+
+    variables:
+      solution: 'WebJobs.Extensions.DurableTask.sln'
+      buildPlatform: 'Any CPU'
+      buildConfiguration: 'Release'
```
cgillum: Since we're doing a separate build for running tests and doing the release, I wonder if we should actually use `Debug` as the `buildConfiguration` value for the V1 and V2 tests. That might provide us with better diagnostic information (exception traces, etc.) if something fails. We should continue to use the `Release` build configuration for the publish job.

azure-pipelines.yml
```diff
@@ -0,0 +1,117 @@
...
+    pool: 
+      vmImage: 'windows-latest'
+
+    strategy:
+      parallel: 9
+
+    variables:
+      solution: 'WebJobs.Extensions.DurableTask.sln'
+      buildPlatform: 'Any CPU'
+      buildConfiguration: 'Debug'
```
ConnorMcMahon: Similar comment about debug here.

azure-pipelines.yml
```diff
@@ -0,0 +1,117 @@
...
+        feedsToUse: 'config'
+        nugetConfigPath: '.nuget/nuget.config'
+
+    - task: VSBuild@1
+      inputs:
+        solution: '$(solution)'
+        platform: '$(buildPlatform)'
+        configuration: '$(buildConfiguration)'
+
+    - publish: 'src/WebJobs.Extensions.DurableTask/bin'
```
ConnorMcMahon: Does this publish to MyGet, or to NuGet directly?

azure-pipelines.yml
```diff
@@ -0,0 +1,117 @@
...
+        solution: '$(solution)'
+        platform: '$(buildPlatform)'
+        configuration: '$(buildConfiguration)'
+
+    - task: VSTest@2
+      inputs:
+        testSelector: 'testAssemblies'
+        testAssemblyVer2: |
+          **/*tests.v2.dll
+        testFiltercriteria:
```
ConnorMcMahon: Is there a reason we have filter criteria in Functions V1, but not in Functions V2?

azure-pipelines.yml
```diff
@@ -0,0 +1,117 @@
...
+        distributionBatchType: basedOnExecutionTime
+        diagnosticsEnabled: True
+        rerunFailedTests: true
```
ConnorMcMahon: Do we get any notifications of failed tests when this is set to `true`? If we have a consistently flaky test, that can indicate a race condition, and rerunning multiple times could hide that from us. That being said, it does make flakey test failures far less annoying.

azure-pipelines.yml
amdeel: I'm making sure we only get the FunctionsV1_BVT tests in V1 (this mirrors how AppVeyor is currently configured), but want all of the V2 tests (including flakey tests). I could delete the empty testFilterCriteria line under FunctionV2Tests, though, since there currently aren't any filters for those tests.

azure-pipelines.yml
amdeel: No, this publishes to pipeline artifacts, although I could publish to NuGet in the future using Azure Pipelines.

azure-pipelines.yml
```diff
@@ -0,0 +1,117 @@
...
+        rerunMaxAttempts: 2
+      env:
+        AzureWebJobsStorage: $(AzureWebJobsStorage)
+
+  - job: FunctionsV2Tests
+    pool: 
+      vmImage: 'windows-latest'
+
+    strategy:
+      parallel: 9
```
TsuyoshiUshio: It might be the root cause why this pipeline takes around 30 min.  In this case, FunctionsV2Test run with 9 agents parallel. However, FunctionV1Test run on single agent. You don't specify this for FunctionV1Test.  I don't know how much your organization enables Azure Pipeline run parallel.  FunctionV1Test and FunctionV2Test might be the same number of the test cases. If you have only 10 agents, then you can run FunctionV1Test with 5 agents, FunctionV2Test with 5 agents as well.

amdeel: I actually intentionally left FunctionsV1 Tests on a single agent. We're only running a subset of FunctionsV1 Tests and while testing, FunctionsV1 Tests finished before any of the 9 agents associated with FunctionsV2 Tests, so it isn't actually holding back performance.

PR #840, opened 2019-07-08
Added dll/nupkg Signing to Azure Pipelines

amdeel:


azure-pipelines.yml
```diff
@@ -112,5 +111,77 @@ jobs:
         platform: '$(buildPlatform)'
         configuration: '$(buildConfiguration)'
 
-    - publish: 'src/WebJobs.Extensions.DurableTask/bin'
-      artifact: bin
\ No newline at end of file
+    - task: EsrpCodeSigning@1
```
cgillum: Now that we have an official release pipeline, let's remove the signing from this one and do signing only for official releases. That way we're not spending time and resources doing signing for a package that doesn't really require it.

azure-pipelines.yml
```diff
@@ -100,6 +98,7 @@ jobs:
     - task: NuGetToolInstaller@0
```
cgillum: A few lines above this, I see we're doing a `Release` build configuration. Let's actually switch this back to `Debug` since this is not an official release package. That will make testing easier.

Issue #469, opened 2018-10-11
Durable HTTP


cgillum:
## Summary
In addition to being able to invoke activity functions, sub-orchestrations, and create durable timers, durable orchestrator functions would have an API for invoking HTTP endpoints in a durable, safe-for-replay way.
```csharp
public static async Task Run(DurableOrchestrationContext context)
{
    var res = await context.CallHttpAsync(...);
}
```

## Motivation
There are two top-level scenarios that would benefit from something like this:
1. **Cross function app communication**: There are many sub-scenarios for cross function-app communication including multi-language support (e.g. C# function app orchestrating Python HTTP functions in another app), security isolation (e.g. function apps with different MSI roles/identities), or general microservice-style development.

2. **Calling external services**: Most services on the internet are exposed over HTTP and can be invoked directly from an orchestrator function using a built-in API. This is a productivity improvement over writing an activity function which wraps HTTP calls. There can even be native integration with features such as Easy Auth and MSI for making outbound HTTP requests to AAD-protected services. Built-in integration with Application Insights (App Map, etc.) is also a potential benefit.

These things can technically be done today with activity functions but they are hard for developers because:

- Wrapping HTTP calls in activity functions is very boilerplate.
- Connection management best practices are not always known/followed.
- HTTP retry policies are hard/complicated.
- Client-side handling of long-running HTTP async APIs is *really* hard/complicated.
- OAuth/MSI integration is complicated.
- End-to-end tracing can be cumbersome.

Ideally, a built-in HTTP solution could make all of these pain points really simple.

### Assumptions
This design makes the following assumptions:

- **Streaming**: Orchestrator functions are not allowed do to I/O, so there isn't a situation where it will have access to an I/O stream. Because of this, stream-based HTTP APIs will not be useful.
- **API Interop**: Orchestrator functions won't have direct access to the underlying HTTP client, and orchestrator functions are not aware of any HTTP request which may have triggered them, so platform-specific HTTP API interoperability (e.g. `HttpRequestMessage` for .NET) will not be useful.

### API
New `CallHttpAsync` methods will be added to the `DurableOrchestrationContextBase` class.
```csharp
public abstract class DurableOrchestrationContextBase
{
    public abstract Task<DurableHttpResponse> CallHttpAsync(HttpMethod method, Uri uri);
    public abstract Task<DurableHttpResponse> CallHttpAsync(DurableHttpRequest req);
}
```
[...]

PR #866, opened 2019-07-18
Adding first class support for HTTP calls in Orchestration functions.

bachuv:
Includes CallHttpAsync API to DurableOrchestrationContext. Also, adds
support for Bearer tokens to the CallHttpAsync API. This API supports an
asynchronous 202 HTTP pattern.

Resolves https://github.com/Azure/azure-functions-durable-extension/issues/469

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -5,6 +5,8 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
+using System.Net.Http;
 using System.Reflection;
 using System.Runtime.ExceptionServices;
 using System.Text;
@@ -13,6 +15,8 @@ using System.Threading.Tasks;
 using DurableTask.Core;
 using DurableTask.Core.Exceptions;
 using DurableTask.Core.History;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask.Options;
+using Microsoft.Extensions.Primitives;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
@@ -30,6 +34,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private string serializedOutput;
         private string serializedCustomStatus;
+        private static JsonSerializer serializer = CreateDurableHttpResponseSerializer();
 
         private int newGuidCounter = 0;
 
@@ -156,6 +161,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             return this.serializedCustomStatus;
         }
 
+        private static JsonSerializer CreateDurableHttpResponseSerializer()
+        {
+            JsonSerializer serializer = new JsonSerializer();
+            serializer.Converters.Add(new DurableHttpResponseJsonConverter());
+            return serializer;
+        }
+
         /// <inheritdoc />
         Task<TResult> IInterleavingContext.CallSubOrchestratorAsync<TResult>(string functionName, string instanceId, object input)
         {
@@ -171,6 +183,74 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             }
 
             return this.CallDurableTaskFunctionAsync<TResult>(functionName, FunctionType.Orchestrator, false, instanceId, null, retryOptions, input);
+        }
+
+        Task<DurableHttpResponse> IDurableOrchestrationContext.CallHttpAsync(HttpMethod method, Uri uri, string content)
+        {
+            DurableHttpRequest req = new DurableHttpRequest(
+                method: method,
+                uri: uri,
+                content: content);
+            return ((IDurableOrchestrationContext)this).CallHttpAsync(req);
+        }
+
+        async Task<DurableHttpResponse> IDurableOrchestrationContext.CallHttpAsync(DurableHttpRequest req)
+        {
+            DurableHttpResponse durableHttpResponse = await this.ScheduleDurableHttpActivityAsync(req);
+
+            HttpStatusCode currStatusCode = durableHttpResponse.StatusCode;
+
+            while (currStatusCode == HttpStatusCode.Accepted && req.AsynchronousPatternEnabled)
+            {
+                Dictionary<string, StringValues> headersDictionary = new Dictionary<string, StringValues>(durableHttpResponse.Headers);
+                DateTime fireAt = default(DateTime);
+                if (headersDictionary.ContainsKey("Retry-After"))
+                {
+                    fireAt = this.InnerContext.CurrentUtcDateTime.AddSeconds(int.Parse(headersDictionary["Retry-After"]));
+                }
+                else
+                {
+                    fireAt = this.InnerContext.CurrentUtcDateTime.AddMilliseconds(this.Config.Options.HttpSettings.DefaultAsyncRequestSleepTimeMilliseconds);
+                }
+
+                await this.InnerContext.CreateTimer(fireAt, CancellationToken.None);
+
+                DurableHttpRequest durableAsyncHttpRequest = this.CreateHttpRequestMessageCopy(req, durableHttpResponse.Headers["Location"]);
+                durableHttpResponse = await this.ScheduleDurableHttpActivityAsync(durableAsyncHttpRequest);
+                currStatusCode = durableHttpResponse.StatusCode;
+            }
+
+            return durableHttpResponse;
+        }
+
+        private async Task<DurableHttpResponse> ScheduleDurableHttpActivityAsync(DurableHttpRequest req)
+        {
+            string serializedRequest = MessagePayloadDataConverter.HttpConverter.Serialize(req);
+
+            // Using JToken to preserve StringValues data in DurableHttpRequest.Headers
+            // because StringValues cannot be deserialized using a default Json Converter
+            JToken durableHttpResponseJson = await this.CallDurableTaskFunctionAsync<JToken>(
+                functionName: HttpOptions.HttpTaskActivityReservedName,
+                functionType: FunctionType.Activity,
+                oneWay: false,
+                instanceId: null,
+                operation: null,
+                retryOptions: null,
+                input: serializedRequest);
+
+            return durableHttpResponseJson.ToObject<DurableHttpResponse>(serializer);
+        }
+
+        private DurableHttpRequest CreateHttpRequestMessageCopy(DurableHttpRequest durableHttpRequest, string locationUri)
+        {
+            DurableHttpRequest newDurableHttpRequest = new DurableHttpRequest(
+                method: HttpMethod.Get,
+                uri: new Uri(locationUri),
+                headers: durableHttpRequest.Headers,
+                content: durableHttpRequest.Content,
+                tokenSource: durableHttpRequest.TokenSource);
+
+            return newDurableHttpRequest;
         }
 
         /// <inheritdoc />
--- a/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
@@ -3,9 +3,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.Net;
+using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.Azure.WebJobs
 {
@@ -73,5 +76,21 @@ namespace Microsoft.Azure.WebJobs
         /// </remarks>
         /// <param name="customStatusObject">The JSON-serializeable value to use as the orchestrator function's custom status.</param>
         void SetCustomStatus(object customStatusObject);
+
+        /// <summary>
+        /// Makes an HTTP call to the specified uri.
+        /// </summary>
+        /// <param name="method">HttpMethod used for api call.</param>
+        /// <param name="uri">uri used to make the HTTP call.</param>
+        /// <param name="content">Content passed in the HTTP request.</param>
+        /// <returns>A <see cref="Task{DurableHttpResponse}"/>Result of the HTTP call.</returns>
+        Task<DurableHttpResponse> CallHttpAsync(HttpMethod method, Uri uri, string content = null);
+
+        /// <summary>
+        /// Makes an HTTP call using the information in the DurableHttpRequest.
+        /// </summary>
+        /// <param name="req">The DurableHttpRequest used to make the HTTP call.</param>
+        /// <returns>A <see cref="Task{DurableHttpResponse}"/>Result of the HTTP call.</returns>
+        Task<DurableHttpResponse> CallHttpAsync(DurableHttpRequest req);
     }
 }
--- a/src/WebJobs.Extensions.DurableTask/DurableHttpClientFactory.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableHttpClientFactory.cs
@@ -1,0 +1,44 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Reflection;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    internal class DurableHttpClientFactory
+    {
+        private HttpClient durableHttpClient;
+
+        public HttpClient GetClient(IDurableHttpMessageHandlerFactory handler)
+        {
+            if (this.durableHttpClient == null)
+            {
+                this.durableHttpClient = new HttpClient(handler.CreateHttpMessageHandler());
+
+                var assembly = typeof(DurableTaskExtension).Assembly;
+                Version assemblyVersion = GetAssemblyVersion(assembly);
+
+                this.durableHttpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue(
+                    assembly.GetName().Name,
+                    assemblyVersion.ToString()));
+            }
+
+            return this.durableHttpClient;
+        }
+
+        private static Version GetAssemblyVersion(Assembly assembly)
+        {
+            var assemblyInfo = FileVersionInfo.GetVersionInfo(assembly.Location);
+            var assemblyVersion = new Version(
+                assemblyInfo.FileMajorPart,
+                assemblyInfo.FileMinorPart,
+                assemblyInfo.FileBuildPart);
+
+            return assemblyVersion;
+        }
+    }
+}
--- a/src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
@@ -105,9 +105,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         private static Dictionary<string, StringValues> CreateDictionaryCopy(IDictionary<string, StringValues> headers)
         {
             Dictionary<string, StringValues> newDictionary = new Dictionary<string, StringValues>(StringComparer.OrdinalIgnoreCase);
-            foreach (KeyValuePair<string, StringValues> pair in headers)
+            if (headers != null)
             {
-                newDictionary.Add(pair.Key, pair.Value);
+                foreach (KeyValuePair<string, StringValues> pair in headers)
+                {
+                    newDictionary.Add(pair.Key, pair.Value);
+                }
             }
 
             return newDictionary;
--- a/src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
@@ -1,0 +1,116 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Runtime.Serialization;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask.Listener;
+using Microsoft.Extensions.Primitives;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    /// <summary>
+    /// Request used to make an HTTP call through Durable Functions.
+    /// </summary>
+    public class DurableHttpRequest
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DurableHttpRequest"/> class.
+        /// </summary>
+        /// <param name="method">Method used for HTTP request.</param>
+        /// <param name="uri">Uri used to make the HTTP request.</param>
+        /// <param name="headers">Headers added to the HTTP request.</param>
+        /// <param name="content">Content added to the body of the HTTP request.</param>
+        /// <param name="tokenSource">AAD authentication attached to the HTTP request.</param>
+        /// <param name="asynchronousPatternEnabled">Specifies whether the DurableHttpRequest should handle the asynchronous pattern.</param>
+        public DurableHttpRequest(
+            HttpMethod method,
+            Uri uri,
+            IDictionary<string, StringValues> headers = null,
+            string content = null,
+            ITokenSource tokenSource = null,
+            bool asynchronousPatternEnabled = true)
+        {
+            this.Method = method;
+            this.Uri = uri;
+            this.Headers = CreateDictionaryCopy(headers);
+            this.Content = content;
+            this.TokenSource = tokenSource;
+            this.AsynchronousPatternEnabled = asynchronousPatternEnabled;
+        }
+
+        internal DurableHttpRequest(JObject jObject)
+        {
+            this.Method = jObject["Method"].ToObject<HttpMethod>();
+            this.Uri = jObject["Uri"].ToObject<Uri>();
+
+            Dictionary<string, StringValues> headerDictStringValues = new Dictionary<string, StringValues>();
+            Dictionary<string, IEnumerable<string>> headersDictEnumerable = jObject["Headers"].ToObject<Dictionary<string, IEnumerable<string>>>();
+            foreach (var header in headersDictEnumerable)
+            {
+                string key = header.Key;
+                string[] headerValues = header.Value.ToArray<string>();
+                StringValues values = new StringValues(headerValues);
+                headerDictStringValues.Add(key, values);
+            }
+
+            this.Headers = headerDictStringValues;
+
+            this.Content = jObject["Content"].Value<string>();
+
+            JsonSerializerSettings serializer = new JsonSerializerSettings();
+            serializer.TypeNameHandling = TypeNameHandling.Auto;
+            string tokenSource = JsonConvert.SerializeObject(jObject["TokenSource"], serializer);
+
+            this.TokenSource = JsonConvert.DeserializeObject<ITokenSource>(tokenSource, serializer);
+            this.AsynchronousPatternEnabled = jObject["AsynchronousPatternEnabled"].Value<bool>();
+        }
+
+        /// <summary>
+        /// HttpMethod used in the HTTP request made by the Durable Function.
+        /// </summary>
+        public HttpMethod Method { get; }
+
+        /// <summary>
+        /// Uri used in the HTTP request made by the Durable Function.
+        /// </summary>
+        public Uri Uri { get; }
+
+        /// <summary>
+        /// Headers passed with the HTTP request made by the Durable Function.
+        /// </summary>
+        public IDictionary<string, StringValues> Headers { get; }
+
+        /// <summary>
+        /// Content passed with the HTTP request made by the Durable Function.
+        /// </summary>
+        public string Content { get; }
+
+        /// <summary>
+        /// Information needed to get a token for a specified service.
+        /// </summary>
+        [JsonProperty(TypeNameHandling = TypeNameHandling.Auto)]
+        public ITokenSource TokenSource { get; }
+
+        /// <summary>
+        /// Specifies whether the Durable HTTP APIs should automatically
+        /// handle the asynchronous HTTP pattern.
+        /// </summary>
+        public bool AsynchronousPatternEnabled { get; } = true;
+
+        private static Dictionary<string, StringValues> CreateDictionaryCopy(IDictionary<string, StringValues> headers)
+        {
+            Dictionary<string, StringValues> newDictionary = new Dictionary<string, StringValues>(StringComparer.OrdinalIgnoreCase);
+            foreach (KeyValuePair<string, StringValues> pair in headers)
+            {
+                newDictionary.Add(pair.Key, pair.Value);
+            }
+
+            return newDictionary;
+        }
+    }
+}
--- a/src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
@@ -109,7 +109,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             {
                 foreach (KeyValuePair<string, StringValues> pair in headers)
                 {
-                    newDictionary.Add(pair.Key, pair.Value);
+                    newDictionary[pair.Key] = pair.Value;
                 }
             }
 
--- a/src/WebJobs.Extensions.DurableTask/DurableHttpResponse.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableHttpResponse.cs
@@ -1,0 +1,94 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask.Listener;
+using Microsoft.Extensions.Primitives;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    /// <summary>
+    /// Response received from the HTTP request made by the Durable Function.
+    /// </summary>
+    // [JsonConverter(typeof(DurableHttpResponseJsonConverter))]
+    public class DurableHttpResponse
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DurableHttpResponse"/> class.
+        /// </summary>
+        /// <param name="statusCode">HTTP Status code returned from the HTTP call.</param>
+        /// <param name="headers">Headers returned from the HTTP call.</param>
+        /// <param name="content">Content returned from the HTTP call.</param>
+        [JsonConstructor]
+        public DurableHttpResponse(
+            HttpStatusCode statusCode,
+            IDictionary<string, StringValues> headers = null,
+            string content = null)
+        {
+            this.StatusCode = statusCode;
+            this.Headers = headers ?? new Dictionary<string, StringValues>(StringComparer.OrdinalIgnoreCase);
+            this.Content = content;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DurableHttpResponse"/> class.
+        /// </summary>
+        /// <param name="jObject">JObject containing information from HttpResponseMessage.</param>
+        public DurableHttpResponse(JObject jObject)
+        {
+            int codeInt = int.Parse(jObject["StatusCode"].Value<string>());
+            HttpStatusCode statusCode = (HttpStatusCode)codeInt;
+
+            Dictionary<string, StringValues> headerDictStringValues = new Dictionary<string, StringValues>(StringComparer.OrdinalIgnoreCase);
+            Dictionary<string, IEnumerable<string>> headersDictEnumerable = jObject["Headers"].ToObject<Dictionary<string, IEnumerable<string>>>();
+            foreach (var header in headersDictEnumerable)
+            {
+                string key = header.Key;
+                string[] headerValues = header.Value.ToArray<string>();
+                StringValues values = new StringValues(headerValues);
+                headerDictStringValues.Add(key, values);
+            }
+
+            this.StatusCode = statusCode;
+            this.Headers = headerDictStringValues;
+            this.Content = jObject["Content"].Value<string>();
+        }
+
+        /// <summary>
+        /// Status code returned from an HTTP request.
+        /// </summary>
+        public HttpStatusCode StatusCode { get; }
+
+        /// <summary>
+        /// Headers in the response from an HTTP request.
+        /// </summary>
+        public IDictionary<string, StringValues> Headers { get; }
+
+        /// <summary>
+        /// Content returned from an HTTP request.
+        /// </summary>
+        public string Content { get; }
+
+        /// <summary>
+        /// Creates a DurableHttpResponse from an HttpResponseMessage.
+        /// </summary>
+        /// <param name="httpResponseMessage">HttpResponseMessage returned from the HTTP call.</param>
+        /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.</returns>
+        public static async Task<DurableHttpResponse> CreateDurableHttpResponseWithHttpResponseMessage(HttpResponseMessage httpResponseMessage)
+        {
+            DurableHttpResponse durableHttpResponse = new DurableHttpResponse(
+                statusCode: httpResponseMessage.StatusCode,
+                headers: TaskHttpActivityShim.CreateStringValuesHeaderDictionary(httpResponseMessage.Headers),
+                content: await httpResponseMessage.Content.ReadAsStringAsync());
+
+            return durableHttpResponse;
+        }
+    }
+}
--- a/src/WebJobs.Extensions.DurableTask/DurableHttpResponseJsonConverter.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableHttpResponseJsonConverter.cs
@@ -1,0 +1,43 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using Microsoft.Extensions.Primitives;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    /// <summary>
+    /// This custom JsonConverter is used to deserialize DurableHttpResponse
+    /// messages and keep the Header's StringValues. StringValues otherwise cannot
+    /// be deserialized without losing their values.
+    /// </summary>
+    internal class DurableHttpResponseJsonConverter : JsonConverter
+    {
+        public override bool CanRead
+        {
+            get { return true; }
+        }
+
+        public override bool CanConvert(Type objectType)
+        {
+            return objectType == typeof(DurableHttpResponse);
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            var jObject = JObject.Load(reader);
+            return new DurableHttpResponse(jObject);
+        }
+
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            throw new NotImplementedException();
+        }
+    }
+}
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -19,6 +19,8 @@ using DurableTask.Core.History;
 using DurableTask.Core.Middleware;
 using Microsoft.AspNetCore.Routing.Template;
 using Microsoft.Azure.WebJobs.Description;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask.Listener;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask.Options;
 using Microsoft.Azure.WebJobs.Host;
 using Microsoft.Azure.WebJobs.Host.Config;
 using Microsoft.Azure.WebJobs.Host.Executors;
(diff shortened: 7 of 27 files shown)


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -173,6 +177,111 @@ Task<TResult> IInterleavingContext.CallSubOrchestratorWithRetryAsync<TResult>(st
...
+                    }
+                }
+            }
+
+            return newHeaders;
+        }
+
+        private static DurableHttpRequest AddLocationHeaderToAsyncRequest(DurableHttpRequest durableHttpRequest, DurableHttpResponse durableHttpResponse)
+        {
+            IDictionary<string, StringValues> durableHttpRequestHeaders = durableHttpResponse.Headers;
```
ConnorMcMahon: The name of the variable is durableHttpRequestHeaders, but the value is the headers of a `DurableHttpResponse`. This is a bit confusing to me. In fact, I don't even know if you need this intermediate variable at all considering you just use it once.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
```diff
@@ -73,5 +76,30 @@ public interface IDurableOrchestrationContext : IInterleavingContext
...
+
+        /// <summary>
+        /// Makes an HTTP call to the specified uri with authentication information
+        /// to get a token.
+        /// </summary>
+        /// <param name="method">HttpMethod used for api call.</param>
+        /// <param name="uri">uri used to make the HTTP call.</param>
+        /// <param name="tokenSource">information to get a token from a specified service.</param>
+        /// <returns>A <see cref="Task{TResult}"/> Result of the HTTP call.</returns>
+        Task<DurableHttpResponse> CallHttpAsync(HttpMethod method, Uri uri, ITokenSource tokenSource);
```
cgillum: Generally speaking, the simplified method overloads should cover the most common cases. In my opinion, this particular overload with an `ITokenSource` and no payload will not be a common case. Rather, I would suggest you have a `CallHttpAsync` method that takes a method, URL, and optional _request content_, defined like this:

```csharp
Task<DurableHttpResponse> CallHttpAsync(HttpMethod method, Uri uri, string content = null);
```
And to be honest I think the first overload (HTTP method, and URL) probably isn't needed either since requests without payloads are only common for GET requests. I think if we just had the proposed one (in addition to the `DurableHttpRequest`) then that will give us the smallest number of methods to maintain while still having very good ease-of-use.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -173,6 +177,111 @@ Task<TResult> IInterleavingContext.CallSubOrchestratorWithRetryAsync<TResult>(st
...
+            string locationValue = durableHttpRequestHeaders["Location"];
+            durableHttpRequest.Uri = new Uri(locationValue);
+            return durableHttpRequest;
+        }
+
+        private DurableHttpRequest CreateNewHttpRequestMessageForAsyncResponse(DurableHttpRequest durableHttpRequest)
```
ConnorMcMahon: When I read this method name, I think that it will take in a DurableHttpResponse object and return a `DurableHttpRequest` object. When I read the body of the method, it looks like it is just cloning the original request. I would change the name to reflect the body in this scenario.

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        private static IDictionary<string, StringValues> CreateStringValuesHeaderDictionary(IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers)
+        {
+            IDictionary<string, StringValues> newHeaders = new Dictionary<string, StringValues>();
+            if (headers != null)
+            {
+                foreach (var header in headers)
+                {
+                    foreach (var headerValue in header.Value)
```
ConnorMcMahon: I think you can entirely remove this inner loop. `StringValues` has a constructor that takes a `string[]`. I think then you can replace this inner loop with `newHeaders[header.Key] = new StringValues(header.Value.ToArray());`.

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            HttpRequestMessage requestMessage = await CreateHttpRequestMessageAsync(rawInput);
+            HttpResponseMessage response = await this.httpClient.SendAsync(requestMessage);
+            DurableHttpResponse durableHttpResponse = await CreateDurableHttpResponseAsync(response);
+
+            string serializedOutput = MessagePayloadDataConverter.HttpConverter.Serialize(durableHttpResponse, true);
+            return serializedOutput;
+        }
+
+        private static async Task<DurableHttpResponse> CreateDurableHttpResponseAsync(HttpResponseMessage response)
```
ConnorMcMahon: NIT: Instead of `CreateDurableHttpResponseAsync`, I would consider naming this `CopyDurableHttpResponseAsync()`.

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -0,0 +1,112 @@
...
+            throw new NotImplementedException();
+        }
+
+        public async override Task<string> RunAsync(TaskContext context, string rawInput)
+        {
+            HttpRequestMessage requestMessage = await CreateHttpRequestMessageAsync(rawInput);
+            HttpResponseMessage response = await this.httpClient.SendAsync(requestMessage);
+            DurableHttpResponse durableHttpResponse = await CreateDurableHttpResponseAsync(response);
+
+            string serializedOutput = MessagePayloadDataConverter.HttpConverter.Serialize(durableHttpResponse, true);
```
ConnorMcMahon: NIT: you can just return this value here instead of assigning it to an intermediate variable.

Also, if you could use the parameter name with the `true` value, as just looking at the code it is unclear what the boolean means.

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -0,0 +1,112 @@
...
+                            newHeaders.Add(header.Key, stringValues);
+                        }
+                    }
+                }
+            }
+
+            return newHeaders;
+        }
+
+        private static async Task<HttpRequestMessage> CreateHttpRequestMessageAsync(string rawInput)
```
ConnorMcMahon: NIT: I would make the name a bit more descriptive here, something like `ReconstructHttpRequestMessageFromSerializedRequestAsync`. Don't love the huge name, but it makes it a bit more clear what is going on from the signature. You can achieve something similar without the `FromSerializedRequest` part if you rename the parameter `rawInput` to `serializedRequest`.

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            JsonSerializerSettings serializerSettings = new JsonSerializerSettings();
```
ConnorMcMahon: In general, if you are using some fixed serializer settings here, it is best to at least make it a class or static variable, so you aren't constantly recreating the same `JsonSerializerSettings` object every time you call the method.

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -0,0 +1,112 @@
...
+            serializerSettings.TypeNameHandling = TypeNameHandling.Auto;
+
+            IList<string> input = JsonConvert.DeserializeObject<List<string>>(rawInput, serializerSettings);
+            string durableHttpRequestString = input.First();
+            DurableHttpRequest durableHttpRequest = JsonConvert.DeserializeObject<DurableHttpRequest>(durableHttpRequestString, serializerSettings);
+
+            // Creating HttpRequestMessage
```
ConnorMcMahon: NIT: This comment doesn't provide a ton of value.

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            return newHeaders;
+        }
+
+        private static async Task<HttpRequestMessage> CreateHttpRequestMessageAsync(string rawInput)
+        {
+            JsonSerializerSettings serializerSettings = new JsonSerializerSettings();
+            serializerSettings.TypeNameHandling = TypeNameHandling.Auto;
+
+            IList<string> input = JsonConvert.DeserializeObject<List<string>>(rawInput, serializerSettings);
```
ConnorMcMahon: Consider adding a comment about why the value of `rawInput` is a `List<string>`, as opposed to just a `DurableHttpRequest`. You may remember the details now, but that is something you are likely to forget in a month or two.

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                foreach (KeyValuePair<string, StringValues> entry in durableHttpRequest.Headers)
+                {
+                    requestMessage.Headers.Add(entry.Key, (IEnumerable<string>)entry.Value);
+                }
+            }
+
+            if (durableHttpRequest.Content != null)
+            {
+                requestMessage.Content = JsonConvert.DeserializeObject<HttpContent>(durableHttpRequest.Content);
```
ConnorMcMahon: We should consider testing various types of HttpContent here, so we know what this does and doesn't work with.

src/WebJobs.Extensions.DurableTask/DurableHttpClientFactory.cs
```diff
@@ -0,0 +1,45 @@
...
+                    assembly.GetName().Name,
+                    assemblyVersion.ToString()));
+            }
+
+            return this.durableHttpClient;
+        }
+
+        private static Version GetAssemblyVersion()
+        {
+            var assembly = typeof(DurableTaskExtension).Assembly;
```
cgillum: This call to `typeof(DurableTaskExtension).Assembly` is redundant to the code from a few lines above. I suggest you pass in the `Assembly` reference as a parameter.

src/WebJobs.Extensions.DurableTask/ManagedIdentityTokenSource.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Runtime.Serialization;
+using System.Threading.Tasks;
+using Microsoft.Azure.Services.AppAuthentication;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    /// <summary>
+    /// Token Source implementation specific to Managed Identity Service.
+    /// </summary>
+    [DataContract]
```
ConnorMcMahon: The requirement for using `[DataContract]` with the ITokenSource for serialization reasons is something we should consider documenting in the `IOrchestrationContext.CallHttpAsync` method comment (or at least on the `ITokenSource` interface documentation), so that users who are making their own token sources know about this requirement.

src/WebJobs.Extensions.DurableTask/Options/HttpOptions.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    /// <summary>
+    /// Used for Durable HTTP functionality.
+    /// </summary>
+    public class HttpOptions
+    {
+        /// <summary>
+        /// Reserved name to know when a TaskActivity should be an HTTP activity.
+        /// </summary>
+        public const string HttpTaskActivityReservedName = "Durable:Http:Async:Activity:Function";
```
ConnorMcMahon: I am not sure if this needs to be public.

src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// Initializes a new instance of the <see cref="DurableHttpRequest"/> class.
+        /// </summary>
+        /// <param name="method">Method used for HTTP request.</param>
+        /// <param name="uri">Uri used to make the HTTP request.</param>
+        public DurableHttpRequest(HttpMethod method, Uri uri)
+        {
+            this.Method = method;
+            this.Uri = uri;
+            this.Headers = new Dictionary<string, StringValues>(StringComparer.OrdinalIgnoreCase);
+            this.TokenSource = null;
```
cgillum: No need to initialize properties to `null`.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -42,6 +42,8 @@
     <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.WebApiCompatShim" Version="2.2.0" />
     <PackageReference Include="Microsoft.AspNetCore.Routing" Version="2.2.0" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-*" PrivateAssets="All" />
```
ConnorMcMahon: We should see if this is actually needed as a part of this PR.

src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        /// <summary>
+        /// Uri used in the HTTP request made by the Durable Function.
+        /// </summary>
+        public Uri Uri { get; set; }
+
+        /// <summary>
+        /// Headers passed with the HTTP request made by the Durable Function.
+        /// </summary>
+        // TODO: Need to change this to IDictionary<string, StringValues>
```
cgillum: We can remove this TODO now.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -6,17 +6,23 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Net;
```
ConnorMcMahon: You can get rid of all of the changes in this file.

test/FunctionsV1/TestDurableHttpRequest.cs
```diff
@@ -0,0 +1,37 @@
...
+
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Runtime.Serialization;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
+{
+    [DataContract]
+    public class TestDurableHttpRequest
```
ConnorMcMahon: You may want to add comments as to why this POCO is required for your tests.

src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
```diff
@@ -0,0 +1,58 @@
...
+        public IDictionary<string, StringValues> Headers { get; set; }
```
cgillum: Let's remove the `set` here. I don't think we want other classes to be able to set this. You can still set it from the constructor even without the `set`. Same with `Uri` and `Method` properties. The less mutable you can make a class, the better (I'm assuming you don't have any cases where you create a `DurableHttpRequest` object and then need to keep manipulating it).

src/WebJobs.Extensions.DurableTask/DurableHttpResponse.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// <summary>
+        /// Status code returned from an HTTP request.
+        /// </summary>
+        public HttpStatusCode StatusCode { get; }
+
+        /// <summary>
+        /// Headers in the response from an HTTP request.
+        /// </summary>
+        // TODO: Need to change this to IDictionary<string, StringValues>
+        public IDictionary<string, StringValues> Headers { get; set; }
```
cgillum: Same comments here about the `TODO` and the `set`.

test/FunctionsV2/DurableHttpTests.cs
```diff
@@ -0,0 +1,958 @@
...
+                var output = status?.Output;
+                DurableHttpResponse response = output.ToObject<DurableHttpResponse>();
+
+                Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+
+                await host.StopAsync();
+            }
+        }
+
+        private static HttpMessageHandler CreateSynchronousHttpMessageHandlerForTestingTokenSource()
```
ConnorMcMahon: NIT: I would replace the word `Create` with `Mock`, so it is clear you are mocking this behavior as to using an actual HttpMessageHandler.

test/FunctionsV2/DurableHttpTests.cs
```diff
@@ -0,0 +1,958 @@
...
+                    acceptedHttpResponseMessageTwo,
+                    acceptedHttpResponseMessageTwo,
+                    acceptedHttpResponseMessageTwo,
+                    okHttpResponseMessage,
+                }).Dequeue);
+
+            return handlerMock.Object;
+        }
+
+        private static HttpMessageHandler CreateSynchronousHttpMessageHandler(HttpResponseMessage httpResponseMessage)
```
ConnorMcMahon: Similar comment about `Create` to `Mock` here.

test/FunctionsV2/DurableHttpTests.cs
```diff
@@ -0,0 +1,958 @@
...
+            var handlerMock = new Mock<HttpMessageHandler>(MockBehavior.Strict);
+            handlerMock
+               .Protected()
+               .Setup<Task<HttpResponseMessage>>("SendAsync", ItExpr.IsAny<HttpRequestMessage>(), ItExpr.IsAny<CancellationToken>())
+               .ReturnsAsync(httpResponseMessage);
+
+            return handlerMock.Object;
+        }
+
+        private static HttpMessageHandler CreateHttpMessageHandlerCheckUserAgent()
```
ConnorMcMahon: Similar comment about `Create` to `Mock` here.

test/FunctionsV2/DurableHttpTests.cs
```diff
@@ -0,0 +1,958 @@
...
+
+            handlerMock
+               .Protected()
+               .Setup<Task<HttpResponseMessage>>("SendAsync", ItExpr.Is<HttpRequestMessage>(req => req.Headers.UserAgent == null), ItExpr.IsAny<CancellationToken>())
+               .ReturnsAsync(forbiddenHttpResponseMessage);
+
+            return handlerMock.Object;
+        }
+
+        private static HttpMessageHandler CreateAsynchronousHttpMessageHandler(HttpResponseMessage acceptedHttpResponseMessage)
```
ConnorMcMahon: Similar comment about `Create` to `Mock` here.

test/FunctionsV2/DurableHttpTests.cs
```diff
@@ -0,0 +1,958 @@
...
+                    acceptedHttpResponseMessage,
+                    acceptedHttpResponseMessage,
+                    acceptedHttpResponseMessage,
+                    okHttpResponseMessage,
+                }).Dequeue);
+
+            return handlerMock.Object;
+        }
+
+        private static HttpMessageHandler CreateAsynchronousHttpMessageHandlerLongRunning(HttpResponseMessage acceptedHttpResponseMessage)
```
ConnorMcMahon: Similar comment here.

test/FunctionsV2/DurableHttpTests.cs
```diff
@@ -0,0 +1,958 @@
...
+                    acceptedHttpResponseMessage,
+                    acceptedHttpResponseMessage,
+                    acceptedHttpResponseMessage,
+                    okHttpResponseMessage,
+                }).Dequeue);
+
+            return handlerMock.Object;
+        }
+
+        private static HttpMessageHandler CreateAsynchronousHttpMessageHandlerWithRetryAfter(HttpResponseMessage acceptedHttpResponseMessage)
```
ConnorMcMahon: Similar comment here.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -173,6 +177,111 @@ Task<TResult> IInterleavingContext.CallSubOrchestratorWithRetryAsync<TResult>(st
...
+                functionName: HttpOptions.HttpTaskActivityReservedName,
+                functionType: FunctionType.Activity,
+                oneWay: false,
+                instanceId: null,
+                operation: null,
+                retryOptions: null,
+                input: serializedRequest);
+
+            JsonSerializer serializer = new JsonSerializer();
+            serializer.Converters.Add(new DurableHttpResponseJsonConverter(typeof(DurableHttpResponse)));
```
cgillum: Instead of manually creating `JsonSerializer` objects like this, can we use the [JsonConverterAttribute](https://www.newtonsoft.com/json/help/html/JsonConverterAttributeClass.htm) on the `DurableHttpResponse` class and the new `MessagePayloadDataConverter.HttpConverter` that you added?

test/FunctionsV2/DurableHttpTests.cs
```diff
@@ -0,0 +1,958 @@
...
+                var output = status?.Output;
+                DurableHttpResponse response = output.ToObject<DurableHttpResponse>();
+
+                Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+
+                await host.StopAsync();
+            }
+        }
+
+        private static HttpMessageHandler CreateAsynchronousHttpMessageHandlerForTestingTokenSource()
```
ConnorMcMahon: Similar comment here.

test/FunctionsV2/TestDurableHttpRequest.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Runtime.Serialization;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
```
ConnorMcMahon: If you move this to common, you shouldn't need a duplicate version of this file in `test/FunctionsV1` and 'test/FunctionsV2`.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -173,6 +177,111 @@ Task<TResult> IInterleavingContext.CallSubOrchestratorWithRetryAsync<TResult>(st
...
+            DurableHttpResponse durableHttpResponse = durableHttpResponseJson.ToObject<DurableHttpResponse>(serializer);
+            return durableHttpResponse;
+        }
+
+        private static IDictionary<string, StringValues> CreateStringValuesHeaderDictionary(IList<KeyValuePair<string, string>> headers)
```
ConnorMcMahon: I don't think this method is used. The one with the other signature is.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -173,6 +177,111 @@ Task<TResult> IInterleavingContext.CallSubOrchestratorWithRetryAsync<TResult>(st
...
+
+                durableHttpResponse = await this.ScheduleDurableHttpActivityFunctionAsync(durableAsyncHttpRequest);
+
+                currStatusCode = durableHttpResponse.StatusCode;
+            }
+
+            return durableHttpResponse;
+        }
+
+        private async Task<DurableHttpResponse> ScheduleDurableHttpActivityFunctionAsync(DurableHttpRequest req)
```
ConnorMcMahon: I would drop `Function` from the name, as technically it is not an activity function, it is just a durable activity.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -173,6 +177,111 @@ Task<TResult> IInterleavingContext.CallSubOrchestratorWithRetryAsync<TResult>(st
...
+        {
+            string serializedRequest = MessagePayloadDataConverter.HttpConverter.Serialize(req);
+            JToken durableHttpResponseJson = await this.CallDurableTaskFunctionAsync<JToken>(
```
ConnorMcMahon: I would leave a comment here as to why you are serializing as `JToken` instead of `DurableHttpRequest`.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -173,6 +177,111 @@ Task<TResult> IInterleavingContext.CallSubOrchestratorWithRetryAsync<TResult>(st
...
+                functionName: HttpOptions.HttpTaskActivityReservedName,
+                functionType: FunctionType.Activity,
+                oneWay: false,
+                instanceId: null,
+                operation: null,
+                retryOptions: null,
+                input: serializedRequest);
+
+            JsonSerializer serializer = new JsonSerializer();
```
ConnorMcMahon: Consider making this a class-level variable to prevent creating one every time this method is called.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -173,6 +177,111 @@ Task<TResult> IInterleavingContext.CallSubOrchestratorWithRetryAsync<TResult>(st
...
+            serializer.Converters.Add(new DurableHttpResponseJsonConverter(typeof(DurableHttpResponse)));
+            DurableHttpResponse durableHttpResponse = durableHttpResponseJson.ToObject<DurableHttpResponse>(serializer);
```
ConnorMcMahon: NIT: return immediately instead of using intermediate value.

src/WebJobs.Extensions.DurableTask/DurableHttpResponse.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// <summary>
+    /// Response received from the HTTP request made by the Durable Function.
+    /// </summary>
+    public class DurableHttpResponse
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DurableHttpResponse"/> class.
+        /// </summary>
+        /// <param name="statusCode">HTTP Status code returned from the HTTP call.</param>
+        public DurableHttpResponse(HttpStatusCode statusCode)
```
cgillum: I suggest you create another constructor that takes `HttpResponseMessage`. I saw that you have logic elsewhere that manually creates `DurableHttpResponse` objects from `HttpResponseMessage` and then sets its properties. This is not ideal because it requires all the properties to be settable. A cleaner design would be to only allow the `DurableHttpresponse` class to manipulate its own internal state.

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -0,0 +1,112 @@
...
+    {
+        private readonly HttpClient httpClient;
+        private readonly DurableTaskExtension config;
+
+        public TaskHttpActivityShim(
+            DurableTaskExtension config,
+            HttpClient httpClientFactory)
+        {
+            this.config = config ?? throw new ArgumentNullException(nameof(config));
+            this.httpClient = httpClientFactory;
```
ConnorMcMahon: Consider doing an `ArgumentNullException` check here as well.

src/WebJobs.Extensions.DurableTask/IDurableHttpMessageHandlerFactory.cs
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Net.Http;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    /// <summary>
+    /// Interface used for testing Durable HTTP.
+    /// </summary>
+    public interface IDurableHttpMessageHandlerFactory
```
ConnorMcMahon: Let's check to see if this interface needs to be public.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -745,6 +762,16 @@ internal void ThrowIfFunctionDoesNotExist(string name, FunctionType functionType
             }
         }
 
+        private static bool IsDurableHttpTask(string functionName)
+        {
+            if (functionName == HttpOptions.HttpTaskActivityReservedName)
```
ConnorMcMahon: You can make this function one line by just doing `return string.Equals(functionName, HttpOptions.HttpTaskActivityReservedName)`.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -314,6 +320,12 @@ void INameVersionObjectManager<TaskActivity>.Add(ObjectCreator<TaskActivity> cre
         /// <returns>An activity shim that delegates execution to an activity function.</returns>
         TaskActivity INameVersionObjectManager<TaskActivity>.GetObject(string name, string version)
         {
+            // checking if the user needs to create a TaskHttpActivityShim
+            if (name == HttpOptions.HttpTaskActivityReservedName)
```
ConnorMcMahon: Replace this check with a call to `IsDurableHttpTask`.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -291,10 +299,8 @@ TaskOrchestration INameVersionObjectManager<TaskOrchestration>.GetObject(string
             {
                 return new TaskEntityShim(this, name);
             }
-            else
```
ConnorMcMahon: Is this change necessary?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -61,12 +63,12 @@ public class DurableTaskExtension :
         private readonly AsyncLock taskHubLock = new AsyncLock();
 
         private readonly bool isOptionsConfigured;
-
```
ConnorMcMahon: NIT: no need for this whitespace.

src/WebJobs.Extensions.DurableTask/DurableHttpResponse.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        /// <summary>
+        /// Status code returned from an HTTP request.
+        /// </summary>
+        public HttpStatusCode StatusCode { get; }
+
+        /// <summary>
+        /// Headers in the response from an HTTP request.
+        /// </summary>
+        // TODO: Need to change this to IDictionary<string, StringValues>
```
ConnorMcMahon: Can get rid of this TODO.

src/WebJobs.Extensions.DurableTask/DurableHttpResponse.cs
cgillum: And it looks like we may also need a constructor for `JObject`. That can replace the logic inside the `DurableHttpResponseJsonConverter` which is directly manipulating `DurableHttpResponse`.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
```diff
@@ -73,5 +76,30 @@ public interface IDurableOrchestrationContext : IInterleavingContext
...
+        Task<DurableHttpResponse> CallHttpAsync(HttpMethod method, Uri uri);
+
+        /// <summary>
+        /// Makes an HTTP call to the specified uri with authentication information
+        /// to get a token.
+        /// </summary>
+        /// <param name="method">HttpMethod used for api call.</param>
+        /// <param name="uri">uri used to make the HTTP call.</param>
+        /// <param name="tokenSource">information to get a token from a specified service.</param>
+        /// <returns>A <see cref="Task{TResult}"/> Result of the HTTP call.</returns>
```
ConnorMcMahon: Should be `Task{DurableHttpResponse}`

src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
ConnorMcMahon: TODO is obsolete.

src/WebJobs.Extensions.DurableTask/DurableHttpResponseJsonConverter.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        public override bool CanRead
+        {
+            get { return true; }
+        }
+
+        public override bool CanConvert(Type objectType)
+        {
+            bool value = this.types.Any(t => t == objectType);
```
ConnorMcMahon: NIT: No need for intermediate variable here.

src/WebJobs.Extensions.DurableTask/DurableHttpResponseJsonConverter.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using Microsoft.Extensions.Primitives;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    internal class DurableHttpResponseJsonConverter : JsonConverter
```
ConnorMcMahon: Consider writing a comment at the top here about how this is meant to get around the StringValues serialization problems.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -173,6 +177,111 @@ Task<TResult> IInterleavingContext.CallSubOrchestratorWithRetryAsync<TResult>(st
...
+            return durableHttpResponse;
+        }
+
+        private static IDictionary<string, StringValues> CreateStringValuesHeaderDictionary(IList<KeyValuePair<string, string>> headers)
+        {
+            IDictionary<string, StringValues> newHeaders = new Dictionary<string, StringValues>();
```
cgillum: Looks like you forgot to add the `StringComparer.OrdinalIgnoreCase` here. To avoid problems like this, we should refactor the code so that `DurableHttpRequest` and `DurableHttpResponse` be responsible for manipulating headers inside their own constructors. See some of my other comments related to this.

src/WebJobs.Extensions.DurableTask/Options/HttpOptions.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        /// <summary>
+        /// Time between the async http requests.
+        /// </summary>
+        public int DefaultAsyncRequestSleepTime { get; set; } = 5000;
```
ConnorMcMahon: We should make it clear this value is in Milliseconds in the name, as well as the documentation comments. Also, we should decide whether `5000` is the right default.

src/WebJobs.Extensions.DurableTask/Options/HttpOptions.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        /// <summary>
+        /// Boolean value specifying if an HTTP Request is async.
```
ConnorMcMahon: This documentation comment is not clear. I would change it to "Specifies whether the Durable HTTP APIs should automatically handle the asynchronous HTTP pattern."

test/FunctionsV2/WebJobs.Extensions.DurableTask.Tests.V2.csproj
```diff
@@ -14,6 +14,11 @@
     <StyleCopTreatErrorsAsWarnings>true</StyleCopTreatErrorsAsWarnings>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|netcoreapp2.0|AnyCPU'">
```
ConnorMcMahon: Not sure if this change is necessary.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -173,6 +177,111 @@ Task<TResult> IInterleavingContext.CallSubOrchestratorWithRetryAsync<TResult>(st
...
+            return ((IDurableOrchestrationContext)this).CallHttpAsync(req);
+        }
+
+        async Task<DurableHttpResponse> IDurableOrchestrationContext.CallHttpAsync(DurableHttpRequest req)
+        {
+            DurableHttpResponse durableHttpResponse = await this.ScheduleDurableHttpActivityFunctionAsync(req);
+
+            HttpStatusCode currStatusCode = durableHttpResponse.StatusCode;
+
+            while (currStatusCode == HttpStatusCode.Accepted && this.Config.Options.HttpSettings.AsynchronousPatternEnabled)
```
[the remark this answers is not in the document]
cgillum: Yes, definitely. Whether a customer wants to use this probably depends on the specific HTTP API they are calling. I suggest adding a property on `DurableHttpRequest` that controls this.

src/WebJobs.Extensions.DurableTask/IDurableHttpMessageHandlerFactory.cs
bachuv: This interface needs to be public because otherwise there's inconsistent accessibility with DurableTaskExtension. DurableTaskExtension is public and has an IDurableHttpMessageHandlerFactory argument in the constructor. Also, the other interfaces included in the DurableTaskExtension constructor are public.

src/WebJobs.Extensions.DurableTask/Options/HttpOptions.cs
bachuv: I changed it to internal.

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -0,0 +1,101 @@
...
+            {
+                foreach (KeyValuePair<string, StringValues> entry in durableHttpRequest.Headers)
+                {
+                    requestMessage.Headers.Add(entry.Key, (IEnumerable<string>)entry.Value);
+                }
+            }
+
+            if (durableHttpRequest.Content != null)
+            {
+                requestMessage.Content = JsonConvert.DeserializeObject<HttpContent>(durableHttpRequest.Content);
```
cgillum: I think it actually needs to be this:
```suggestion
                requestMessage.Content = new StringContent(durableHttpRequest.Content);
```

src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
bachuv: I can take the `set` out for `Method`, but `Uri` and `Headers` are set in DurableOrchestrationContext. `Uri` is changed to set the location header of an asynchronous response (line 244 in DurableOrchestrationContext.cs). `Headers` are set when a copy of the DurableHttpRequest is made (line 250 in DurableOrchestrationContext.cs). Also, when users create DurableHttpRequest messages they set all attributes other than `Method `and `Uri` outside of the constructor.

@cgillum Would it be better to make all of the attributes other then `Method `and `Uri `to be optional arguments in the constructor? This would change how the user creates a DurableHttpRequest.

src/WebJobs.Extensions.DurableTask/DurableHttpResponse.cs
bachuv: Same issues as removing the set from Headers in DurableHttpRequest.

src/WebJobs.Extensions.DurableTask/DurableHttpResponse.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DurableHttpResponse"/> class.
+        /// </summary>
+        /// <param name="jObject">JObject containing information from HttpResponseMessage.</param>
+        public DurableHttpResponse(JObject jObject)
+        {
+            int codeInt = int.Parse(jObject["StatusCode"].Value<string>());
+            HttpStatusCode statusCode = (HttpStatusCode)codeInt;
+
+            Dictionary<string, StringValues> headerDictStringValues = new Dictionary<string, StringValues>();
```
cgillum: Need to use the `StringComparer.OrdinalIgnoreCase` parameter on this dictionary constructor

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            DurableHttpResponse durableHttpResponse = new DurableHttpResponse(response.StatusCode);
+            durableHttpResponse.Headers = CreateStringValuesHeaderDictionary(response.Headers);
+            durableHttpResponse.Content = await response.Content.ReadAsStringAsync();
+            return durableHttpResponse;
+        }
+
+        internal static IDictionary<string, StringValues> CreateStringValuesHeaderDictionary(IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers)
+        {
+            IDictionary<string, StringValues> newHeaders = new Dictionary<string, StringValues>();
```
cgillum: The dictionary constructor should specify `StringComparer.OrdinalIgnoreCase`.

src/WebJobs.Extensions.DurableTask/Options/HttpOptions.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// </summary>
+    public class HttpOptions
+    {
+        /// <summary>
+        /// Reserved name to know when a TaskActivity should be an HTTP activity.
+        /// </summary>
+        internal const string HttpTaskActivityReservedName = "Durable:Http:Async:Activity:Function";
+
+        /// <summary>
+        /// Time between the async http requests.
```
cgillum: I would say "Gets or sets the default number of milliseconds between async HTTP status poll requests."

test/Common/DurableTaskLifeCycleNotificationTest.cs
```diff
@@ -1299,7 +1301,8 @@ public void OrchestrationCustomHelperTypeActivationFailed()
                 wrappedOptions,
                 new LoggerFactory(),
                 new SimpleNameResolver(),
-                new OrchestrationServiceFactory(wrappedOptions, new TestConnectionStringResolver()));
+                new OrchestrationServiceFactory(wrappedOptions, new TestConnectionStringResolver()),
+                new DurableHttpMessageHandlerFactory());
```
cgillum: Do you really need the changes in this file or can they be reverted?

src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
bachuv: I actually ended up removing the set from all of the attributes so the class is immutable and changes can only be made by creating a new DurableHttpRequest.

src/WebJobs.Extensions.DurableTask/DurableHttpResponse.cs
bachuv: I removed the set from each attribute in DurableHttpResponse.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
bachuv: Adding the JsonConverterAttribute made it so serialization had to be specified for `DurableHttpResponse`, not only deserialization and that was complicating the serialization process. Connor and I discussed this and decided to make the `JsonSerializer `a static variable in the interest of time.

src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
```diff
@@ -0,0 +1,104 @@
...
+            HttpMethod method,
+            Uri uri,
+            IDictionary<string, StringValues> headers = null,
+            string content = null,
+            ITokenSource tokenSource = null,
+            bool asynchronousPatternEnabled = true)
+        {
+            this.Method = method;
+            this.Uri = uri;
+            this.Headers = headers ?? new Dictionary<string, StringValues>(StringComparer.OrdinalIgnoreCase);
```
cgillum: It's fine to take the headers as a parameter, but it's dangerous to assign it directly to your `Headers` property because the caller might not have created their dictionary with `StringComparer.OrdinalIgnoreCase`. To avoid problems related to this, it is better to copy the values from the parameter dictionary into a dictionary object you created yourself.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
cgillum: Cool - I think this is the next best solution. Let's go with this.

src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
```diff
@@ -0,0 +1,116 @@
...
+            HttpMethod method,
+            Uri uri,
+            IDictionary<string, StringValues> headers = null,
+            string content = null,
+            ITokenSource tokenSource = null,
+            bool asynchronousPatternEnabled = true)
+        {
+            this.Method = method;
+            this.Uri = uri;
+            this.Headers = CreateDictionaryCopy(headers) ?? new Dictionary<string, StringValues>(StringComparer.OrdinalIgnoreCase);
```
cgillum: It seems to me that you don't actually need the `??` expression here. `CreateDictionaryCopy` will always return a non-null dictionary.

src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
```diff
@@ -0,0 +1,116 @@
...
+        [JsonProperty(TypeNameHandling = TypeNameHandling.Auto)]
+        public ITokenSource TokenSource { get; }
+
+        /// <summary>
+        /// Specifies whether the Durable HTTP APIs should automatically
+        /// handle the asynchronous HTTP pattern.
+        /// </summary>
+        public bool AsynchronousPatternEnabled { get; } = true;
+
+        private static Dictionary<string, StringValues> CreateDictionaryCopy(IDictionary<string, StringValues> headers)
```
cgillum: The `headers` parameter is not used. It looks like the below code is enumerating the new empty dictionary.

bachuv: That's actually a bug, it's supposed to be used while iterating through the headers dictionary. I fixed it and just added another test to check this.

src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            Dictionary<string, StringValues> newDictionary = new Dictionary<string, StringValues>(StringComparer.OrdinalIgnoreCase);
+            foreach (KeyValuePair<string, StringValues> pair in headers)
```
cgillum: I think you need a check for `null` before you try to enumerate `headers`. Otherwise you'll get a `NullReferenceException` here.

PR #867, opened 2019-07-18
Proof of concept for DI in Durable Entities

ConnorMcMahon:
With the work done by @fabiocav, this is relatively trivial. The test code was by far the hardest part, especially since this should only be run in the Functions V2 test suite.

This PR likely will not be merged, as the API is still volatile, but works as a proof of concept.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityContext.cs
@@ -2,11 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
-using System.Collections.Generic;
-using System.Text;
-using Microsoft.Extensions.Logging;
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
+using Microsoft.Azure.WebJobs.Host.Bindings;
 
 namespace Microsoft.Azure.WebJobs
 {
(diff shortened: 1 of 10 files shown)


src/WebJobs.Extensions.DurableTask/EntityScheduler/TypedInvocationExtensions.cs
```diff
@@ -30,8 +30,7 @@ public static class TypedInvocationExtensions
         /// (which is the operation name) and the argument list (which is the operation content, deserialized into
         /// an object array).
         /// </remarks>
-        public static async Task DispatchAsync<T>(this IDurableEntityContext context)
-            where T : new()
+        public static async Task DispatchAsync<T>(this IDurableEntityContext context, params object[] constructorParameters)
```
sebastianburckhardt: It may be helpful to document how these parameters work. In particular in combination with dependency-injected constructor parameters. Is there a ordering requirement?

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityContext.cs
```diff
@@ -44,6 +40,11 @@ public interface IDurableEntityContext
         /// </summary>
         bool IsNewlyConstructed { get; }
 
+        /// <summary>
+        /// Contains function invocation context to assist with dependency injection at Entity construction time.
+        /// </summary>
```
sebastianburckhardt: I think there is a chance users will use this field for their own purposes once they discover it. That's probably fine. However, perhaps the summary should mention that this field is null if not running on NETSTANDARD2_0?

cecilphillip:
@ConnorMcMahon re: Bindings in entity classes, can multiple parameters be passed to `DispatchAsync` ?

Issue #864, opened 2019-07-17
Entities REST API is case sensitive (v2)?


jeffhollan:
This may be ok but just raising as I’m not sure how common case sensitive REST APIs are.  If I have an entity “counter” of class type `Counter` and try to get the state:

This call will fail with a 404 even if instance exists (of `@Counter@test`)
`http://localhost:7071/runtime/webhooks/durabletask/entities/counter/test`

This call succeeds:
`http://localhost:7071/runtime/webhooks/durabletask/entities/Counter/test`

I’m ok if the answer is it’s case sensitive and will stay that way, but figured I’d flag for conversation.

cgillum:
I very much prefer these APIs (and Entity IDs in general) be case-insensitive. @amdeel this would be another good one for you to investigate.

cgillum:
Looking at some documentation and articles on Azure Table Storage, it seems like case insensitive lookups aren't supported. That likely means we'll need to normalize the entity IDs in our code (e.g convert them to lowercase) before saving them to storage. We'll also need to make sure any reflection code we have is also case insensitive.

espray:
According to W3's "HTML and URLs" they should case-sensitive
https://www.w3.org/TR/WD-html40-970708/htmlweb.html

Issue #601, opened 2019-02-05
Passing an invalid instance id into StartNewAsync causes function to be run endlessly.


dantheman999301:
**Describe the bug**
Passing an custom instance id to `DurableOrchestrationClient.StartNewAsync` with characters that are not allowed as Table Storage PartitionKeys or RowKeys causes the function to retry infinitely.

**Investigative information**

- Durable Functions extension version: 1.7.1
- Function App version (1.0 or 2.0): 2.0
- Programming language used: C#

**To Reproduce**

1. Create a function that starts a new Orchestrator, supplying the instance id with a bad character (see https://blogs.msdn.microsoft.com/jmstall/2014/06/12/azure-storage-naming-rules/ for invalid characters)
2. Inside the Orchestrator code, call at least one `CallActivityAsync` or throw and error, don't think it matters which.

e.g.
```
        public static async Task StartOrchestrator(
            [ServiceBusTrigger("transform", Connection = "connectionString")] string fileId,
            [OrchestrationClient] DurableOrchestrationClient starter,
            ILogger logger)
        {
            var existingInstance = await starter.GetStatusAsync(fileId);
            if (existingInstance == null)
            {
                logger.LogInformation("Orchestration starting from Service bus with Id {fileID}", fileId);

                await starter.StartNewAsync("StartProcessing", fileId, fileId);
                logger.LogInformation($"Started orchestration with ID = '{fileId}'.");
            }
        }
```

And basically any of the examples code for Durable Functions should cause it to retry over and over. Easy thing to pass in to cause it is something with a question mark in it.

**Expected behavior**
Ideally, I'd think `StartNewAsync` should throw an error before anything is started.

**Actual behavior**
Gets to the Orchestrator and endlessly retries.

**Known workarounds**
Validate the instance id manually.

Just as a note, I realise that this is essentially user error (and we only found it when we were actually trying to break something else entirely) but thought I'd raise it anyway.

cgillum:
Thanks for opening this issue. It sounds like we need to add a bit more validation for the instance IDs.

PR #873, opened 2019-07-23
EntityId Case Insensitivity Fix and Test

amdeel:
#864

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Bindings/EntityTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/EntityTriggerAttributeBindingProvider.cs
@@ -53,6 +53,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 name = method.GetCustomAttribute<FunctionNameAttribute>()?.Name ?? method.Name;
             }
 
+            name = name.ToLower();
+
             // The entity class name defaults to the method name.
             var entityName = new FunctionName(name);
             this.config.RegisterEntity(entityName, null);
--- a/src/WebJobs.Extensions.DurableTask/Bindings/EntityTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/EntityTriggerAttributeBindingProvider.cs
@@ -53,8 +53,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 name = method.GetCustomAttribute<FunctionNameAttribute>()?.Name ?? method.Name;
             }
 
-            name = name.ToLowerInvariant();
-
             // The entity class name defaults to the method name.
             var entityName = new FunctionName(name);
             this.config.RegisterEntity(entityName, null);
--- a/src/WebJobs.Extensions.DurableTask/Bindings/EntityTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/EntityTriggerAttributeBindingProvider.cs
@@ -53,7 +53,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 name = method.GetCustomAttribute<FunctionNameAttribute>()?.Name ?? method.Name;
             }
 
-            name = name.ToLower();
+            name = name.ToLowerInvariant();
 
             // The entity class name defaults to the method name.
             var entityName = new FunctionName(name);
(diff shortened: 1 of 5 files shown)


cgillum:
Hi @amdeel, thanks for creating this PR! In the future, please add a description that includes a link to the issue your PR resolves. In this case, https://github.com/Azure/azure-functions-durable-extension/issues/601. That allows us to properly link PRs to issues.

src/WebJobs.Extensions.DurableTask/Bindings/EntityTriggerAttributeBindingProvider.cs
```diff
@@ -53,6 +53,8 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
                 name = method.GetCustomAttribute<FunctionNameAttribute>()?.Name ?? method.Name;
             }
 
+            name = name.ToLower();
```
cgillum: Is this `ToLower()` necessary? I think function names are already stored in case-insensitive dictionaries.

src/WebJobs.Extensions.DurableTask/EntityScheduler/EntityId.cs
```diff
@@ -44,7 +44,7 @@ public EntityId(string entityName, string entityKey)
 
         internal static string GetSchedulerIdFromEntityId(EntityId entityId)
         {
-            return $"@{entityId.EntityName}@{entityId.EntityKey}";
+            return $"@{entityId.EntityName}@{entityId.EntityKey}".ToLower();
```
cgillum: I believe we need to use `ToLowerInvariant()` instead of `ToLower()`. More discussion on why here: https://stackoverflow.com/questions/6225808/string-tolower-and-string-tolowerinvariant.

test/Common/TestOrchestrations.cs
```diff
@@ -659,6 +659,22 @@ public static async Task<string> LargeEntity([OrchestrationTrigger] IDurableOrch
...
+            string content = new string('.', 100000);
+            await ctx.CallEntityAsync<int>(entityId, "set", content);
+
+            var result = await ctx.CallEntityAsync<string>(entityId, "get");
+            if (result != content)
+            {
+                return $"fail: wrong entity state";
+            }
+
+            return "ok";
```
cgillum: Since the test you're writing doesn't really care what the entity does, I suggest you just re-use an existing entity function instead of writing a new one.

cgillum:
FYI @sebastianburckhardt - just want to make sure you're okay with the change to make EntityIds case-insensitive.

sebastianburckhardt:
I suggest to apply case-insensitivity only to the entity name, but not the entity key.

Case insensitivity if fine for the entity name since it's an Azure function and they are case-insensitive.

But the interpretation of the entity key is up to the application and assuming case insensitivity oversteps the boundaries a bit IMO. As a programmer, I think I prefer frameworks that make as few assumptions as possible about my data representation. We are already being a bit restrictive by forcing the key to always be a string (as opposed to some typed, general, comparable thing) and case insensitivity makes this worse.

src/WebJobs.Extensions.DurableTask/Bindings/EntityTriggerAttributeBindingProvider.cs
amdeel: This was added after the initial changes broke some tests. We needed to allow for case insensitivity on the entity trigger as well. @ConnorMcMahon

src/WebJobs.Extensions.DurableTask/Bindings/EntityTriggerAttributeBindingProvider.cs
```diff
@@ -53,6 +53,8 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
                 name = method.GetCustomAttribute<FunctionNameAttribute>()?.Name ?? method.Name;
             }
 
+            name = name.ToLowerInvariant();
```
cgillum: I think my question from the first review iteration got missed, but why do we need to lowercase `name` here? I was expecting that we didn't need to do this since we already treat function names as case-insensitive.

ConnorMcMahon: Some of the tests were failing without that line, so it definitely mattered somewhere... Figuring out why might be a good idea, and adding a comment to explain why.

PR #875, opened 2019-07-25
Adding small change to accommodate the way Python worker passes JSON

gled4er:
Hello @cgillum,

We identified that we need small change in the `TaskOrchestrationShim` so we can parse the string (JSON) passed by the Python worker as specified [here](https://github.com/Azure/azure-functions-durable-extension/wiki/Out-of-Proc-Orchestrator-Execution-Schema-Reference#sample-orchestrator-execution-result).

Let us know what you think when you have time.

If you think it is a good fit for Durable for the project we will add unit and integration test to solidify it.

Thank you!

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
@@ -133,7 +133,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             {
                 if (orchestratorInfo.IsOutOfProc)
                 {
-                    var jObj = returnValue as JObject;
+                    var jObj = JObject.Parse(returnValue is string ? returnValue as string : returnValue.ToString());
                     if (jObj != null)
                     {
                         await this.HandleOutOfProcExecution(jObj);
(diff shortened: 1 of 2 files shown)


gled4er:
Hello @cgillum ,

Great point! 

I will work on the modification.

Thank you!

gled4er:
Hello @cgillum,

I improved the code based on your recommendations.

May I ask you to take another look when you have time and let me know what should be fixed next?

Thank you!

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -133,7 +133,20 @@ public override async Task<string> Execute(OrchestrationContext innerContext, st
...
-                    var jObj = returnValue as JObject;
+                    JObject jObj = null;
+
+                    if (returnValue is JObject)
+                    {
+                        jObj = returnValue as JObject;
+                    }
+                    else
+                    {
+                        if (returnValue is string)
```
cgillum: Consider the following implementation as a means of reducing the lines of code (without the comments):

```csharp
// same as before
var jObj = returnValue as JObject;

// new code
if (jObj == null && returnValue is string jsonText)
{
    jObj = JObject.Parse(jsonText);
}
```

gled4er: Thank you very much @cgillum !

Your optimization is great!

Please let me know if we need to change something else.

Thank you!

cgillum:
Also, would you mind cherry-picking your change and submitting another PR to our `v2` branch? The `dev` and `v2` branches have diverged quite a bit so cherry-picks are the best way to keep them in sync.

gled4er:
Thank you very much @cgillum !

I have been missing the work on Durable Functions as well!

Sure, I will add the changes to the `v2` branch.

Thank you very much!

PR #877, opened 2019-07-26
Add precompiled samples

brandonh-msft:
This is to compliment changes being made to docs for https://github.com/MicrosoftDocs/azure-docs/issues/33965

cc @kashimiz @cgillum


brandonh-msft:
hm. looks like we had all the csx ones in the PC ones, but docs didn't necessarily have matches. gimme a sec and i'll re-open this if needed.

PR #878, opened 2019-07-26
fixing 'sampling' host setting to match latest schema

brandonh-msft:
`logging/applicationinsights/sampling` has changed to `logging/applicationinsights/samplingsettings` per the host.json schema.

### DIFF
--- a/samples/precompiled/host.json
+++ b/samples/precompiled/host.json
@@ -2,7 +2,7 @@
   "version": "2.0",
   "logging": {
     "applicationInsights": {
-      "sampling": {
+      "samplingSettings": {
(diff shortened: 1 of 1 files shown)


brandonh-msft:
not sure... [it's documented this way](https://docs.microsoft.com/en-us/azure/azure-functions/functions-host-json#applicationinsights)... 🤷‍♂

cgillum:
Yeah, according to [this](https://docs.microsoft.com/en-us/azure/azure-functions/functions-monitoring#configure-sampling) the name of this property changed between v1 and v2. I'm surprised nobody had caught this until you did. 😅

Issue #861, opened 2019-07-16
Default to a hubname if none provided (v2)


jeffhollan:
Currently if you spin up a new Durable v2 project and try to run without updating the host.json you get the following:

> [7/16/19 9:41:19 PM] A host error has occurred
[7/16/19 9:41:19 PM] Microsoft.Azure.WebJobs.Extensions.DurableTask: A non-empty HubName configuration is required.
Value cannot be null.
Parameter name: provider

Ideally if the host.json is blank we would have intelligent defaults so the project could still run

jeffhollan:
Suspect the same is true for the `storageProvider` property

cgillum:
cc @ConnorMcMahon 

This is by design. Having a default hub name led really difficult to diagnose problems for customers. Basically, any time a customer used the same storage account for two different function apps (whether deployed locally or in Azure) the two apps would steal messages from each other and cause unpredictable behavior.

Another solution we considered was defaulting to the site name. Connor and I were concerned about this too because of the potential to cause issues when using slots - even if we also included the slot name in the default task hub name. I forget what the exact concerns were, but @ConnorMcMahon maybe you remember?

We ultimately decided to just make explicitly specifying a task hub name a requirement in v2 since it's the easiest solution and we can always change our minds later. If you feel strongly about this, we can revisit and try to come up with other solutions.

jeffhollan:
Thanks for the context. Maybe something we could even help solve for on the tooling side?  Not sure.  Pretty much everytime I create a durable v2 app I have to open the docs and copy paste the host.json sample because I don’t remember the schema, so some way to make that easier.  Definitely sympathize with the people attaching different functions to same storage account though if a globally constant default taskhub name.

ConnorMcMahon:
I believe the concern had to do with the slot-swap behavior, where using the slot name no longer worked. There was also the potential to use the site name with the deployment id, but I believe that wasn't accessible on Linux.

jeffhollan:
Can you specify the hub name in the method declaration / function.json?  Maybe in the docs and samples we just switch to something like:

```csharp
public class Counter
{
    [FunctionName(nameof(Counter))]
    public static Task Run([EntityTrigger(HubName = “UniqueHubName”] IDurableEntityContext ctx)
        => ctx.DispatchAsync<Counter>();
}
```

So at least it’s a bit easier for them to know to edit?

cgillum:
[...]

PR #897, opened 2019-08-14
Added default HubName if no hub name is provided

amdeel:
#861

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
+++ b/src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
@@ -13,6 +13,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
     /// </summary>
     public class DurableTaskOptions
     {
+        private string hubName;
+        private bool isDefaultHubName = false;
+
         /// <summary>
         /// Settings used for Durable HTTP functionality.
         /// </summary>
@@ -27,7 +30,24 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// multiple Durable Functions applications from each other, even if they are using the same storage backend.
         /// </remarks>
         /// <value>The name of the default task hub.</value>
-        public string HubName { get; set; }
+        public string HubName
+        {
+            get
+            {
+                if (this.hubName == null)
+                {
+                    this.isDefaultHubName = true;
+                    this.hubName = Environment.GetEnvironmentVariable("WEBSITE_SITE_NAME");
+                }
+
+                return this.hubName;
+            }
+
+            set
+            {
+                this.hubName = value;
+            }
+        }
 
         /// <summary>
         /// The section of configuration related to storage providers.
@@ -169,6 +189,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             if (string.IsNullOrEmpty(this.HubName))
             {
                 throw new InvalidOperationException($"A non-empty {nameof(this.HubName)} configuration is required.");
+            }
+
+            if (Environment.GetEnvironmentVariable("WEBSITE_SLOT_NAME") != "Production" && this.isDefaultHubName)
+            {
+                throw new InvalidOperationException($"Task Hub name must be specified to be a unique value when using slots");
             }
 
             if (this.StorageProvider == null)
--- a/test/Common/DurableTaskEndToEndTests.cs
+++ b/test/Common/DurableTaskEndToEndTests.cs
@@ -3277,6 +3277,80 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
                 argumentException.Message);
         }
 
+        /// <summary>
+        /// End-to-end test which validates that default task hub names are able to be assigned and custom values for tack hub names are allowed/>.
+        /// </summary>
+        [Theory]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        // No specified hubName will assign a default hubName, which will be the site name
+        [InlineData("", "TestSiteName", "Production")]
+        [InlineData("CustomName", "TestSiteName", "Production")]
+        [InlineData("TestSiteName", "TestSiteName", "Test")]
+        public void TaskHubName_HappyPath(string customHubName, string siteName, string slotName)
+        {
+            string currSiteName = Environment.GetEnvironmentVariable("WEBSITE_SITE_NAME");
+            string currSlotName = Environment.GetEnvironmentVariable("WEBSITE_SLOT_NAME");
+
+            try
+            {
+                Environment.SetEnvironmentVariable("WEBSITE_SITE_NAME", "TestSiteName");
+                Environment.SetEnvironmentVariable("WEBSITE_SLOT_NAME", slotName);
+
+                var options = new DurableTaskOptions();
+
+                var hubName = siteName;
+
+                if (customHubName != "")
+                {
+                    hubName = customHubName;
+                    options.HubName = customHubName;
+                }
+
+                var host = TestHelpers.GetJobHost(this.loggerProvider, options, false);
+                Assert.Equal(hubName, options.HubName);
+            }
+            finally
+            {
+                Environment.SetEnvironmentVariable("WEBSITE_SITE_NAME", currSiteName);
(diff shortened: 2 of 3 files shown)


src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -171,6 +194,11 @@ internal void Validate()
                 throw new InvalidOperationException($"A non-empty {nameof(this.HubName)} configuration is required.");
             }
 
+            if (!string.Equals(Environment.GetEnvironmentVariable("WEBSITE_SLOT_NAME"), "Production", StringComparison.InvariantCultureIgnoreCase) && this.isDefaultHubName)
```
cgillum: Two things:
1. In the case where `WEBSITE_SLOT_NAME` is not defined (like when running locally), what is the expected behavior? It seems to me that we will throw even though it's not actually a case where slots are being used.
2. Use `StringComparison.OrdinalIgnoreCase` instead of `StringComparison.InvariantCultureIgnoreCase`. More info [here](https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings#choosing-a-stringcomparison-member-for-your-method-call).

amdeel: I'll fix both of those, thanks! I'll add a boolean method check that returns true if the slot is production or if it doesn't exist to cover that case.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -3277,6 +3277,80 @@ public async Task TaskHubName_Throws_ArgumentException(string taskHubName)
                 argumentException.Message);
         }
 
+        /// <summary>
```
ConnorMcMahon: In general, I think test names along with the test code and the comments within the code should be clear enough that you don't need to add lengthy comments at the method level for tests. This is because when running CI pipelines, you only get the Test name, not the comments associated with them.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -171,6 +194,11 @@ internal void Validate()
                 throw new InvalidOperationException($"A non-empty {nameof(this.HubName)} configuration is required.");
             }
 
+            if (!IsInProductionSlotOrTest() && this.isDefaultHubName)
+            {
+                throw new InvalidOperationException($"Task Hub name must be specified to be a unique value when using slots");
```
ConnorMcMahon: Can you put a TODO here for us to add a link to documentation in this exception? When we make the TODO, it would be a good idea to add a link to a GitHub issue that describes what we need to do to resolve the TODO.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -27,7 +30,27 @@ public class DurableTaskOptions
...
-        public string HubName { get; set; }
+        public string HubName
+        {
+            get
+            {
+                if (this.hubName == null)
+                {
+                    this.isDefaultHubName = true;
+
+                    // "WEBSITE_SITE_NAME" is an environment variable used in production. When testing locally, you can set this
```
ConnorMcMahon: A few comments about this comment:

- In general, we avoid using "you" to describe how customers can configure the code. I would replace that with "customers", or even using the passive voice.
- Be a bit more descriptive than "production", especially since that word is used in another context (production slots) in the same file. I would replace "production" with "Azure Functions infrastructure". 
- `local.json` should be `local.settings.json`

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -193,5 +221,18 @@ internal void Validate()
                 throw new InvalidOperationException($"{nameof(this.MaxConcurrentOrchestratorFunctions)} must be a non-negative integer value.");
             }
         }
+
+        // Returns true if WEBSITE_SLOT_NAME is Production or when WEBSITE_SLOT_NAME is not set.
```
ConnorMcMahon: In general, I am of the belief that non-documentation comments should describe WHY the code is written this way, not WHAT the code this was doing. This comment reads more like a documentation comment, in that it says what the method does. For private methods, the name should ideally be descriptive enough to not require a documentation comment.

I would move the WHY parts of this comment (i.e. non Azure Functions environment likely won't have "WEBSITE_SLOT_NAME" as an environment variable) closer to the code itself, and not put it at the method level.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -3277,6 +3277,80 @@ public async Task TaskHubName_Throws_ArgumentException(string taskHubName)
...
+                Environment.SetEnvironmentVariable("WEBSITE_SLOT_NAME", currSlotName);
+            }
+        }
+
+        /// <summary>
+        /// Test which validates that attempting to use a default task hub name while in a slot other than production will throw an exception <see cref="InvalidOperationException"/>.
+        /// </summary>
+        [Fact]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        public async Task TaskHubName_DefaultNameNonProductionSlot()
```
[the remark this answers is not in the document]
amdeel: Good point! I'll add that!

test/Common/DurableTaskEndToEndTests.cs
amdeel: I agree. Tests should be easy to understand without comments (test are a form of documentation). What is the reason to specifically exclude a short test summary?

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
amdeel: I actually meant this as that documentation-type method summary comment. I can include a comment like you suggested in code instead, though, if you think that would be best.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
amdeel: Sounds good, I'll make those changes!

test/Common/DurableTaskEndToEndTests.cs
ConnorMcMahon: There is nothing necessarily wrong with them, I just generally don't find them helpful most of the time, considering many of the times you get a failing test, you are just given the test name, not the test code (and comments). As long as the comment isn't used as a crunch and the test name is still a good one, then it shouldn't hurt.

amdeel: Just like comments within code, it would be meant to help (new) developers understand the test. Similarly to the test summary, you also don't see the test code or any within-code comments along with a failing test. I personally try to avoid comments unless absolutely necessary, but haven't thought of any reason to avoid summaries.

I do find it important to make sure tests are easy to understand without comments (and summaries). I'll take this opportunity to review my test to make sure it's clear, but leave the summary unless you object.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -27,7 +30,27 @@ public class DurableTaskOptions
...
+        public string HubName
+        {
+            get
+            {
+                if (this.hubName == null)
+                {
+                    this.isDefaultHubName = true;
+
+                    // "WEBSITE_SITE_NAME" is an environment variable used in Azure functions infrastructure. When running locally, this can be
+                    //      specified in local.settings.json file to avoid being defaulted to "TestHubName"
```
ConnorMcMahon: NIT: Can you make the spacing here more consistent?

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -171,6 +194,12 @@ internal void Validate()
                 throw new InvalidOperationException($"A non-empty {nameof(this.HubName)} configuration is required.");
             }
 
+            if (IsInNonProductionSlot() && this.isDefaultHubName)
+            {
+                throw new InvalidOperationException($"Task Hub name must be specified in host.json when using slots. See documentation on Task Hubs for " +
```
ConnorMcMahon: NIT: neither of these strings require the "$", as they have no embedded variables.

Issue #889, opened 2019-08-07
Only reference/load required DTFx dependencies


cgillum:
### Description
The current v2 build of the extension has direct dependencies on `DurableTask.AzureStorage`, `DurableTask.Emulator`, and `DurableTask.Redis`. Soon we will also need to add support for `DurableTask.EventHubs`. This is not ideal. Ideally we only load the dependencies that a particular application actually needs.

### Suggestion
I suggest that instead of having WebJobs.Extensions.DurableTask depend on the DTFx providers, we create separate packages for each provider type we actually want to support. For example, WebJobs.Extensions.DurableTask.EventHubs depends on WebJobs.Extensions.DurableTask and DurableTask.EventHubs. The actual `IOrchestrationService` that we use can be loaded via dependency injection. This means that we have to maintain more nuget packages, but it provides a cleaner decoupling and ensures we only load the code that we actually want to load.

To avoid breaking changes, we should probably have DurableTask.AzureStorage continue to be the default dependency.

cgillum:
Resolved by https://github.com/Azure/azure-functions-durable-extension/pull/974.

PR #899, opened 2019-08-21
Restructure storage dependencies

ConnorMcMahon:
Work in progress for #889 :

- [x] Tests passing
- [x]  Azure storage dependency is moved from core project.
- [ ]  Confirmed that new approach works in EndToEnd scenarios.
- [ ]  Make default nuget package the Azure Storage flavored extension.
- [ ]  Add documentation.
- [ ] Ensure split out packages does not break XML.

This work splits the source code into three separate projects (and likely 3 separate nuget packages as well).

1. WebJobs.Extensions.DurableTask (will likely be ppackaged as Microsoft.Azure.WebJobs.Extensions.DurableTask.Core on Nuget). This package is technically not an extension, but contains 90+% of the business logic for the Durable Functions repo still.

2. Webjobs.Extensions.DurableTask.AzureStorage (this will be the new Microsoft.Azure.WebJobs.Extensions.DurableTask on NuGet). This package should act almost identically to the Durable 2.0 extension today, but with some small host.json changes and removed support for Redis and Emulator as backends.

3. WebJobs.Extensions.DurableTask.CustomStorage (this will be a new nuget package called Microsoft.Azure.WebJobs.Extensions.DurableTask.CustomStorage or something similar). This is a second flavor of the Durable Functions extension; however, it will only work if the customer uses Functions 2.0 and up feature of dependency injection to provide their own storage provider. More about the classes they will provide later.

**Adding new officially supported storage providers**

Looking at the structure for WebJobs.Extensions.DurableTask.AzureStorage should make it fairly straightforward how to accomplish this. The developer will need to inject a new implementation of `IDurabilityProviderFactory` into the runtime, and provide a new subclass of `DurableTaskOptions` that will represent the official JSON that goes in the `durableTask` extension's section of the `host.json`.

**DurabilityProvider**

This class is a superset of `IOrchestrationService`, `IOrchestrationServiceClient`, and some other methods that are not officially supported in `DurableTask.Core`, but are required to have full functionality in Durable Functions. If customers only want support for the methods provided by those interfaces, they can use the existing `DurabilityProvider`, otherwise they will need to override the class, as all additional methods are virtual methods that currently just throw `NotImplementedException`. `AzureStorageDurabilityProvider` is a great reference point for how to do this.

Will add more details soon.


cgillum:
I'm very excited to see this change! I only skimmed the PR very briefly, but can you please add the following information in the PR description?

1. Add a link to the GitHub issue that this PR resolves.
2. Document minimal examples of what the customer needs to configure in their .csproj files for various storage options.

For the examples, it would be helpful to have something like this for customer apps that will reference Emulator, Azure Storage (the default), etc.:

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <!-- ... -->
  <ItemGroup>
    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask.Emulator" Version="x.y.z" />
  </ItemGroup>

</Project>
```

ConnorMcMahon:
Abandoning this PR, as got feedback from Chris that allows me to simplify greatly.

Issue #890, opened 2019-08-07
Remove Entity operations from IDurableOrchestrationClient


cgillum:
### Description
Today we have clients use `IDurableOrchestrationClient` to access both orchestrations and entities. It's a little strange to have entity operations on a client object with "Orchestration" in the name. I suggest we instead move the entity-specific operations into a new interface, `IDurableEntityClient`. In case users want a single client object which handles both, we could also introduce an `IDurableClient` which implements both `IDurableOrchestrationClient` and `IDurableEntityClient`.

We may want to also consider a similar change to the `OrchestrationClient` attribute.

cgillum:
This has been merged and will be part of the v2.0.0-beta2 release.

PR #900, opened 2019-08-22
Added IDurableEntityClient and IDurableClient

amdeel:
#890


amdeel:
@cgillum @ConnorMcMahon 

I had to change a lot of different places that referenced IDurableOrchestrationClient to either IDurableClient or IDurableEntityClient. Some I left as IDurableOrchestrationClient. Chris and Connor, can you make sure that I made the right choices when I did this?

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableClient.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// </summary>
+    public interface IDurableClient : IDurableOrchestrationClient, IDurableEntityClient
+    {
+        /// <summary>
+        /// Gets the name of the task hub configured on this client instance.
+        /// </summary>
+        /// <value>
+        /// The name of the task hub.
+        /// </value>
+        string TaskHubName { get; }
```
amdeel: This gives me a warning because it hides the other interfaces' TaskHubName. I added this in IDurableClient so there wasn't ambiguity when referencing TaskHubName. What is the best way to deal with this?

amdeel:
I will rename "DurableOrchestrationClientBaseTests" and "DurableOrchestrationClientMock" to "DurableClient..."

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableClient.cs
ConnorMcMahon: It looks like using the `new` keyword [can help you get around this.](https://stackoverflow.com/a/16322323/9035640)

src/WebJobs.Extensions.DurableTask/DurableClientAttribute.cs
```diff
@@ -9,12 +9,12 @@
 namespace Microsoft.Azure.WebJobs
 {
     /// <summary>
-    /// Attribute used to bind a function parameter to a <see cref="DurableOrchestrationClient"/> instance.
+    /// Attribute used to bind a function parameter to a <see cref="DurableClient"/> instance.
```
ConnorMcMahon: Considering this documentation is public facing, but `DurableClient` is not, maybe we should reword this. I suggest:

Attribute used to bind a function parameter to an IDurableClient, IDurableOrchestrationClient, or an IDurableEntityClient.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -620,18 +620,18 @@ private static Assembly ResolveAssembly(object sender, ResolveEventArgs args)
         }
 
         /// <summary>
-        /// Gets a <see cref="DurableOrchestrationClient"/> using configuration from a <see cref="OrchestrationClientAttribute"/> instance.
+        /// Gets a <see cref="DurableClient"/> using configuration from a <see cref="DurableClientAttribute"/> instance.
         /// </summary>
         /// <param name="attribute">The attribute containing the client configuration parameters.</param>
-        /// <returns>Returns a <see cref="DurableOrchestrationClient"/> instance. The returned instance may be a cached instance.</returns>
-        protected internal virtual IDurableOrchestrationClient GetClient(OrchestrationClientAttribute attribute)
+        /// <returns>Returns a <see cref="DurableClient"/> instance. The returned instance may be a cached instance.</returns>
```
ConnorMcMahon: NIT: the documentation references the concrete type, but the type signature references the interface. I would make the documentation match the signature..

src/WebJobs.Extensions.DurableTask/EntityScheduler/Proxy/EntityClientProxy.cs
```diff
@@ -1,16 +1,17 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using Microsoft.Azure.WebJobs.Extensions.DurableTask.ContextInterfaces;
```
ConnorMcMahon: I believe we put all of our System `using` statements first, and then it goes in alphabetical order. There should be a warning for this.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -19,7 +19,7 @@ public class DurableTaskOptions
         public HttpOptions HttpSettings { get; set; }
 
         /// <summary>
-        /// Gets or sets default task hub name to be used by all <see cref="DurableOrchestrationClient"/>,
+        /// Gets or sets default task hub name to be used by all <see cref="DurableClient"/>,
```
ConnorMcMahon: Similar comment here about referencing the public facing interfaces instead of internal concrete type.

test/Common/ClientFunctions.cs
```diff
@@ -3,14 +3,15 @@
 
 using System;
 using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask.ContextInterfaces;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
 {
     internal static class ClientFunctions
```
ConnorMcMahon: I would add two more sample Functions here. One for `IDurableOrchestrationClient`, and one for `IDurableEntityClient`. Then we can add a test for each that makes sure that we are properly binding to these interfaces as well.

test/Common/DurableOrchestrationClientMock.cs
```diff
@@ -7,9 +7,9 @@
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
 {
-    internal class DurableOrchestrationClientMock : DurableOrchestrationClient, IDurableOrchestrationClient
+    internal class DurableOrchestrationClientMock : DurableClient, IDurableOrchestrationClient
```
ConnorMcMahon: NIT: I know that this isn't a change you made, but I am not sure why this also needs to implement `IDurableOrchestrationClient`, as it should inherit it automatically from `DurableClient`.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableClient.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.ContextInterfaces
```
ConnorMcMahon: Change this namespace to `Microsoft.Azure.WebJobs`. We do this for all of our binding related types, as then the customers don't need to add the more complicated namespace just to use the binding.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityClient.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.ContextInterfaces
```
ConnorMcMahon: Change this namespace to Microsoft.Azure.WebJobs. We do this for all of our binding related types, as then the customers don't need to add the more complicated namespace just to use the binding.

test/Common/HttpApiHandlerTests.cs
```diff
@@ -11,6 +11,7 @@
 using System.Threading;
 using System.Threading.Tasks;
 using DurableTask.Core;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask.ContextInterfaces;
```
ConnorMcMahon: We can probably remove all of these additional `using` statements once you change the interface namespaces.

PR #910, opened 2019-08-28
Fix the non-deterministic XML documentation file generation

cgillum:
The auto-generated XML documentation file was being updated in unexpected ways whenever someone did a build. This made managing changes annoying because this XML file showed up in PRs in a way that was not expected.

This PR fixes this behavior by auto-generating different XML files, one for each target framework type. Now contributors should only see one of these files change when they make a change to the public APIs.

(Don't bother reviewing the actual diffs of the XML files. The important thing is that we now have two separate files, one for **net461** and one for **netstandard2.0**).

Once merged, I'll cherry-pick this into the v2 branch.


cgillum:
Thanks. We don't yet use this for our documentation, but we will once we transition into docs.microsoft.com. I've been keeping this around anyways to help us keep track of changes that impact the docs or our public surface area in general.

Issue #715, opened 2019-04-18
Implement StartNewOrchestration functionality for actors & orchestrations


sebastianburckhardt:
The current API allows actors and orchestrations to start new orchestrations using fire-and-forget semantics. This is how it works currently:

```
throw new NotImplementedException(); // TODO
```

Obviously we need to either implement this or remove it from the API.

sebastianburckhardt:
After looking at this a little, I think this requires changes to DTFx. Given the tight schedule to the alpha release I vote for removing it from the API for now.

cgillum:
@sebastianburckhardt do you think we still need this even if entities can bind to `IOrchestrationClient` and start orchestrations that way?

I believe we may have discussed this already, but I don't recall what the conclusion was (if any).

sebastianburckhardt:
We need it for the exactly-once guarantee.

It's not particularly difficult to implement (I have a draft in a branch).

We have not made a big deal out of that yet, but I think in the long term the "exactly-once" story is actually important for positioning the Durable Functions model (orchestrations + entities) relative to low-level actor models that usually make no ordering or reliability guarantees for message delivery.

PR #917, opened 2019-08-30
support StartNewOrchestration from within orchestrations and entities

sebastianburckhardt:
As discussed in #715, this PR provides the capability to start new orchestrations from within orchestrations and entities, without waiting for them to complete (fire-and-forget). 

It is based on the corresponding feature recently added to DTFx azure/durabletask#312, and thus depends on using a very recent version of DTFx.

(Do not merge until this DTFx dependency is resolved).

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
@@ -190,6 +190,40 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 isReplay: false);
         }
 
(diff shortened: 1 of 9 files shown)


cgillum:
@ConnorMcMahon when you get a chance, can you publish an updated version of DurableTask.AzureStorage to myget? It seems we need to do that to unblock this PR.

cgillum:
I went ahead and updated myget.org with DurableTask.AzureStorage v1.6.4-preview, which references DurableTask.Core v2.1.3. I've also updated this PR to merge the latest from v2 and also reference the updated nuget packages.

cgillum:
Dependencies are in nuget.org, so I'm going to go ahead and merge this. I've also gone ahead and updated the extension package version to 2.0.0-beta3.

ElanHasson:
Thanks for taking care of this. If always forget and try to use this :)

Issue #775, opened 2019-05-25
[Durable V2] Resolve ILifeCycleNotificationHelper using Dependency Injection


shibayan:
I would like to change LifeCycleNotificationHelper registering to Dependency Injection instead of defining it in host.json.

#### Example
```
public class Startup : FunctionsStartup
{
    public override void Configure(IFunctionsHostBuilder builder)
    {
        builder.Services.AddSingleton<ILifeCycleNotificationHelper, CustomLifeCycleNotification>();
    }
}
```

I want to make changes in v2 because it is a breaking change.

cgillum:
This sounds like a good idea to me!

ConnorMcMahon:
This was actually already release as a part of 2.0.0-beta3.

PR #921, opened 2019-08-31
Resolve custom lifecycle helper using Dependency Injection

shibayan:
Use Dependency Injection to resolve Lifecycle Notification Helper (only Function v2)

Ref #775

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -90,12 +90,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// <param name="nameResolver">The name resolver to use for looking up application settings.</param>
         /// <param name="orchestrationServiceFactory">The factory used to create orchestration service based on the configured storage provider.</param>
         /// <param name="durableHttpMessageHandlerFactory">The HTTP message handler that handles HTTP requests and HTTP responses.</param>
+        /// <param name="lifeCycleNotificationHelper">The lifecycle notification helper used for custom orchestration tracking.</param>
         public DurableTaskExtension(
             IOptions<DurableTaskOptions> options,
             ILoggerFactory loggerFactory,
             INameResolver nameResolver,
             IOrchestrationServiceFactory orchestrationServiceFactory,
-            IDurableHttpMessageHandlerFactory durableHttpMessageHandlerFactory = null)
+            IDurableHttpMessageHandlerFactory durableHttpMessageHandlerFactory = null,
+            ILifeCycleNotificationHelper lifeCycleNotificationHelper = null)
         {
             // Options will be null in Functions v1 runtime - populated later.
             this.Options = options?.Value ?? new DurableTaskOptions();
@@ -110,7 +112,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
             this.TraceHelper = new EndToEndTraceHelper(logger, this.Options.Tracing.TraceReplayEvents);
             this.HttpApiHandler = new HttpApiHandler(this, logger);
-            this.LifeCycleNotificationHelper = this.CreateLifeCycleNotificationHelper();
+            this.LifeCycleNotificationHelper = lifeCycleNotificationHelper ?? this.CreateLifeCycleNotificationHelper();
             this.orchestrationServiceFactory = orchestrationServiceFactory;
             this.isOptionsConfigured = true;
 
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -260,17 +260,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 return new EventGridLifeCycleNotificationHelper(this.Options, this.nameResolver, this.TraceHelper);
             }
 
-            // Second: Custom LifeCycle Helper Type
-            if (!string.IsNullOrEmpty(this.Options.CustomLifeCycleNotificationHelperType))
-            {
-                var notificationType = Type.GetType(this.Options.CustomLifeCycleNotificationHelperType);
-
-                if (notificationType != null && typeof(ILifeCycleNotificationHelper).IsAssignableFrom(notificationType))
-                {
-                    return (ILifeCycleNotificationHelper)Activator.CreateInstance(notificationType);
-                }
-            }
-
             // Fallback: Disable Notification
             return new NullLifeCycleNotificationHelper();
         }
--- a/test/Common/DurableTaskLifeCycleNotificationTest.cs
+++ b/test/Common/DurableTaskLifeCycleNotificationTest.cs
@@ -1344,31 +1344,37 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
 
         [Fact]
         [Trait("Category", PlatformSpecificHelpers.TestCategory)]
-        public void OrchestrationCustomHelperTypeDependencyInjection()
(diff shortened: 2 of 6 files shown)


test/Common/DurableTaskLifeCycleNotificationTest.cs
```diff
@@ -1342,6 +1342,41 @@ public void OrchestrationCustomHelperTypeFallback()
             Assert.IsType<NullLifeCycleNotificationHelper>(lifeCycleNotificationHelper);
         }
 
+        [Fact]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        public async Task OrchestrationCustomHelperTypeDependencyInjection()
```
ConnorMcMahon: This test doesn't quite test that Dependency Injection is happening, as the `MockLifeCycleNotificationHandler` is being handled by the test's dependency injection, not our production dependency injection.

That being said, this test is well structured to test the business logic around our `ILifeCycleNotificationHelper` usage. Instead of using a counter though, I would make a `List<string>` of all of the operations called, so that we can make sure that the order of notification calls (i.e. OrchestratorStarted then OrchestratorCompleted) is correct.

shibayan: If the test is run for v2, it should be resolved from DI. (v1 cannot use DI, it is dependency injection for testing as you indicated.)

This test case is to check if `MockLifeCycleNotificationHelper` is resolved from DI and called, it doesn't consider the calling order.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -90,12 +90,14 @@ public DurableTaskExtension()
...
         /// <param name="durableHttpMessageHandlerFactory">The HTTP message handler that handles HTTP requests and HTTP responses.</param>
+        /// <param name="lifeCycleNotificationHelper">The lifecycle notification helper used for custom orchestration tracking.</param>
         public DurableTaskExtension(
             IOptions<DurableTaskOptions> options,
             ILoggerFactory loggerFactory,
             INameResolver nameResolver,
             IOrchestrationServiceFactory orchestrationServiceFactory,
-            IDurableHttpMessageHandlerFactory durableHttpMessageHandlerFactory = null)
+            IDurableHttpMessageHandlerFactory durableHttpMessageHandlerFactory = null,
+            ILifeCycleNotificationHelper lifeCycleNotificationHelper = null)
```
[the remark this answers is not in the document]
shibayan: Dependency Injection supports constructor default value, so if you don't add a singleton service, the default value of null is passed.

https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-2.1#constructor-injection-behavior

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -90,12 +90,14 @@ public DurableTaskExtension()
...
         /// <param name="orchestrationServiceFactory">The factory used to create orchestration service based on the configured storage provider.</param>
         /// <param name="durableHttpMessageHandlerFactory">The HTTP message handler that handles HTTP requests and HTTP responses.</param>
+        /// <param name="lifeCycleNotificationHelper">The lifecycle notification helper used for custom orchestration tracking.</param>
         public DurableTaskExtension(
             IOptions<DurableTaskOptions> options,
             ILoggerFactory loggerFactory,
             INameResolver nameResolver,
             IOrchestrationServiceFactory orchestrationServiceFactory,
-            IDurableHttpMessageHandlerFactory durableHttpMessageHandlerFactory = null)
+            IDurableHttpMessageHandlerFactory durableHttpMessageHandlerFactory = null,
```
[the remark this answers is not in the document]
shibayan: By agreeing with your opinion, if `EventGridNotificationHelper` itself is solved by Dependency Injection, I think that it can be solved very smartly.
(Event Grid-related settings can be consolidated in one place)

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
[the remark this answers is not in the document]
ConnorMcMahon: I see now why you are handling this with `null` instead of registering the service in `DurableTaskJobHostConfigurationExtensions.AddDurableTask()`. 

It is because the version of `ILifeCycleNotificationHelper` that we use is dependent on the config in `IOptions<DurableTaskOptions>()`. Normally, the approach we would use with this is a factory interface, but considering we already have a method that initializes the correct `ILifeCycleNotificationHelper`, this seems simpler. 

It also allows the customer to directly inject an `ILifeCycleNotificationHelper` without having to create an `ILifeCycleNotificationHelperFactory`. 

Because of this, I would say I prefer your implementation.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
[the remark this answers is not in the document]
ConnorMcMahon: As stated in the other method, I can see why you used your approach, and I think that the use of the default parameter is fine here.

test/Common/DurableTaskLifeCycleNotificationTest.cs
ConnorMcMahon: I see. This test is looking for customer facing dependency injection (for test purposes etc), not that our production code path is properly populating the correct `ILifeCycleNotificationHelper`.  In that case, I think this test is sufficient as is.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -110,7 +112,7 @@ public DurableTaskExtension()
 
             this.TraceHelper = new EndToEndTraceHelper(logger, this.Options.Tracing.TraceReplayEvents);
             this.HttpApiHandler = new HttpApiHandler(this, logger);
-            this.LifeCycleNotificationHelper = this.CreateLifeCycleNotificationHelper();
+            this.LifeCycleNotificationHelper = lifeCycleNotificationHelper ?? this.CreateLifeCycleNotificationHelper();
```
ConnorMcMahon: Now that we support customer facing dependency injections of `ILifeCycleNotificationHelper` for Functions V2, I wonder if we should remove the `DurableOptions.CustomLifeCycleNotificationHelperType` from our .NET Standard implementation, as it makes sense for customers to use DI over our reflection-creation of that type. We will likely need to keep that configuration and logic for Functions V1 since it does not support dependency injection.

shibayan: I was considering removing it, but I was worried that the behavior would change between Function Runtime V1 and V2.

I thought it would be worthwhile to consider custom Life Cycle Notification only available for Function Runtime V2 and removing the existing implementation.

What do you think?

ConnorMcMahon: @cgillum thoughts here? 

It would be a difference in behavior between Functions V1 and V2 that we would have to document it, but I would prefer to push anyone who CAN use DI to use that, instead of our hacky reflection method that we need for Functions V1.

cgillum: Now is the time for us to accept breaking changes (for Durable 2.0), so I would say let's remove the reflection-based extensibility and replace it with the new DI model. I don't see a strong reason to continue supporting it for Functions V1.

shibayan: @ConnorMcMahon @cgillum Thank you for your opinion.
 I will change only to customization using Dependency Injection!

shibayan: I removed old code that uses reflection!

shibayan:
Thank you! I'm looking forward to the beta3 release.

Issue #928, opened 2019-09-09
Entity invocation fails with explicit interface implementations


aaronpowell:
If you implement an interface **explicitly** on the Entity class the `DispatchAsync` will fail as the method(s) aren't found.

Example implementation:

```csharp
public interface IChatRoom {
    DateTime Post(string content);
    List<KeyValuePair<DateTime, string>> Get();
}

public class ChatRoom : IChatRoom {
    public ChatRoom()
    {
        this.ChatEntries = new SortedDictionary<DateTime, string>();
    }

    [JsonProperty("chatEntries")]
    public SortedDictionary<DateTime, string> ChatEntries { get; set; }

    // an operation that adds a message to the chat
    DateTime IChatRoom.Post(string content)
    {
        var timestamp = DateTime.UtcNow;
        this.ChatEntries.Add(timestamp, content);
        return timestamp;
    }

    // an operation that reads all messages in the chat
    List<KeyValuePair<DateTime, string>> IChatRoom.Get()
    {
        return this.ChatEntries.ToList();
    }
}
```

The reason for this is that the method name(s) are `Namespace.InterfaceName.MethodName` on the type implementing it.

I came across this problem when trying to use Durable Entities with F# which **only** supports explicit interface implementations ([ref 1](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/interfaces#calling-interface-methods) and [ref 2](https://fsharpforfunandprofit.com/posts/interfaces/#explicit-and-implicit-interface-implementations)).

Here's an F# sample:

```fsharp
type ITheatersForMovie =
    abstract member Init: Theater array -> unit

type TheatersForMovie() =
    member val Theaters = Array.empty<Theater> with get, set

    [<FunctionName("TheatersForMovie")>]
    member __.Run([<EntityTrigger>] ctx : IDurableEntityContext) =
        ctx.DispatchAsync<TheatersForMovie>()

    interface ITheatersForMovie with
        member this.Init theaters =
            this.Theaters <- theaters
```

The error message shown is similar to the one in #862. 

Digging into the code it seems that the problem is on this line: https://github.com/Azure/azure-functions-durable-extension/blob/v2.0.0-beta2/src/WebJobs.Extensions.DurableTask/EntityScheduler/TypedInvocationExtensions.cs#L37-L42

A workaround is to create a method on the type that matches the name that is on the interface which just calls through to the interface:

```fsharp
type ITheatersForMovie =
[...]

PR #935, opened 2019-09-11
Implementing a fix for #928

aaronpowell:
I extracted the logic for the looking for the method that does the lookup of the member on an entity so that it was easier to test in isolation.

Created some tests for the main scenarios of method discovery (that I could think of).

#928

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/EntityScheduler/TypedInvocationExtensions.cs
+++ b/src/WebJobs.Extensions.DurableTask/EntityScheduler/TypedInvocationExtensions.cs
@@ -1,8 +1,10 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using Microsoft.AspNetCore.JsonPatch.Operations;
 using System;
 using System.Reflection;
+using System.Runtime.InteropServices.ComTypes;
 using System.Threading.Tasks;
 
 namespace Microsoft.Azure.WebJobs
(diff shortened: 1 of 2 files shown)


test/FunctionsV2/EntityMethodDiscoveryTests.cs
```diff
@@ -0,0 +1,79 @@
...
+namespace WebJobs.Extensions.DurableTask.Tests.V2
+{
+    public class EntityMethodDiscoveryTests
+    {
+        private interface ICounter
+        {
+            void Add(int count);
+        }
+
+        [Fact]
```
aaronpowell: I wasn't sure what (if any) `Trait` to put on the tests. Let me know if it's needed and which one.

src/WebJobs.Extensions.DurableTask/EntityScheduler/TypedInvocationExtensions.cs
```diff
@@ -91,5 +84,53 @@ public static async Task DispatchAsync<T>(this IDurableEntityContext context, pa
...
+            {
+                return method;
+            }
+
+            var entityInterface = interfaces[0];
+            var operationName = $"{entityInterface.Namespace}.{TypeNameHierarchy(entityInterface)}.{context.OperationName}";
+
+            return GetMethodByName(operationName);
+
+            MethodInfo GetMethodByName(string name)
```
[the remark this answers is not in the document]
aaronpowell: I'll be honest - it's the first time I'd ever used them! The function didn't seem to make sense outside the context of this function which is why I when "Hey - let's use it!"

test/FunctionsV2/EntityMethodDiscoveryTests.cs
aaronpowell: Added to all tests

test/FunctionsV2/EntityMethodDiscoveryTests.cs
```diff
@@ -0,0 +1,84 @@
+﻿using Microsoft.Azure.WebJobs;
```
cgillum: Can you add our standard copyright header to the top of this file? StyleCop failed the build because this was missing.

aaronpowell: Done

PR #940, opened 2019-09-21
Additional modifications required for durable entities out of process

ConnorMcMahon:
Necessary tweaks to make entities work in out of proc languages, like Durable JS.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
@@ -34,6 +34,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 TaskHubName = client.TaskHubName,
                 CreationUrls = this.config.HttpApiHandler.GetInstanceCreationLinks(),
                 ManagementUrls = this.config.HttpApiHandler.CreateHttpManagementPayload(InstanceIdPlaceholder, attr?.TaskHub, attr?.ConnectionName),
+                BaseUrl = this.config.HttpApiHandler.GetBaseUrl(),
+                RequiredQueryStringParameters = this.config.HttpApiHandler.GetUniversalQueryStrings(),
             };
             return JsonConvert.SerializeObject(payload);
         }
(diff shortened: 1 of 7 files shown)


src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
```diff
@@ -86,6 +88,12 @@ private class OrchestrationClientInputData
 
             [JsonProperty("managementUrls")]
             public HttpManagementPayload ManagementUrls { get; set; }
+
+            [JsonProperty("baseUrl")]
+            public string BaseUrl { get; set; }
```
ConnorMcMahon: Decided to start passing the base url and the required query string parameters, instead of the preconstructed webhook urls. This is a bit more fragile if the webhook urls change, but it has the upside of requiring less work on the extension side in order to make use of new APIs on the JS shim side.

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -196,6 +196,12 @@ private class AsyncAction
 
             [JsonProperty("httpRequest")]
             internal DurableHttpRequest HttpRequest { get; set; }
+
+            [JsonProperty("entity")]
+            internal string EntitySchedulerId { get; set; }
```
ConnorMcMahon: This field is actually redundant, since the entity scheduler id is actually equivalent to the orchestration instance id. 

Technically I believe that `EntityOperation` is also an event behind the scenes, though the that value is not actually the event that gets sent to the orchestration (all of these are translated to the "op" event with the operation name embedded in the body). It might be best to treat it as it's own field here.

sebastianburckhardt: Using separate fields seems fine to me. If we wanted to optimize a little more we could share the field and use a name for that field that reflects both purposes (e.g. `'EventOrOperationName`). Not likely that it would make much of a difference.

ConnorMcMahon:
Just need to test one last time with latest version of Durable JS PR to make sure everything works before merging.

ConnorMcMahon:
I ran the tests locally and everything passed there. It was odd to see 4 non-flakey tests fail on AppVeyor, especially considering the breadth of scope that the tests were across.

That makes me a little nervous, but I am just going to rerun for now to see if we get a green build on rerun.

PR #946, opened 2019-09-24
Overhaul exception handling for entities

sebastianburckhardt:
This PR fixes multiple issues to improve the overall experience with how exceptions are caught, transmitted, and surfaced in entities and the calling orchestrations.

- Inside client, when signaling or reading an entity, throw an exception if there is no entity function with that name.
- Fix serialization settings for exceptions, so reference loops are ignored (this is needed so it does not trip over some AzureStorageExceptions).
- If an operation throws an unserializable exception, wrap the info into a serializable exception, so it can still be sent back to the calling orchestration and observed there
- Catch serialization exceptions, and wrap them into a EntitySchedulerException that states the precise situation in which this happened (serialization/deserialization of input/state)
- Inside entity context, track internal errors and all application errors separately. Surface both of them to both the logging back-end and to the functions host, so they are more visible in the portal and Application Insights.
- if a batch observes an internal error, throw it to DTFx so it aborts the batch and backs off.
- provide special logic for serialization errors that prevent the entity state to be written back. In that case, go back to the previous entity state (before the batch), suppress signals and fire-and-forget from operations in the batch, and send an error message describing the problem back to all calling orchestrations.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
@@ -179,7 +179,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     }
                     catch (Exception e)
                     {
-                        throw new EntitySchedulerException($"failed to construct entity state: {e.Message}", e);
+                        throw new EntitySchedulerException($"Failed to initialize entity state: {e.Message}", e);
                     }
                 }
 
@@ -216,7 +216,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             }
             catch (Exception e)
             {
-                throw new EntitySchedulerException($"failed to deserialize entity state: {e.Message}", e);
+                throw new EntitySchedulerException($"Failed to deserialize entity state: {e.Message}", e);
             }
         }
 
@@ -228,9 +228,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.StateWasAccessed = true;
         }
 
-        internal void Writeback(out ResponseMessage stateSerializationFailedNotification)
+        internal bool TryWriteback(out ResponseMessage serializationErrorMessage)
         {
-            stateSerializationFailedNotification = null;
+            serializationErrorMessage = null;
 
             if (this.StateWasAccessed)
             {
@@ -242,7 +242,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 {
                     // we cannot serialize the entity state - this is an application error.
                     var serializationException = new EntitySchedulerException(
-                        $"failed to serialize state of '{this.FunctionName}' entity: {e.Message}", e);
+                        $"Failed to serialize state of '{this.FunctionName}' entity: {e.Message}", e);
 
                     this.CaptureApplicationError(serializationException);
 
@@ -250,7 +250,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     // is lost, we don't want the calling orchestrations to think everything is o.k.
                     // They should be notified, so we replace all non-error operation results
                     // with an exception result.
-                    stateSerializationFailedNotification = new ResponseMessage()
+                    serializationErrorMessage = new ResponseMessage()
                     {
                         ExceptionType = serializationException.GetType().AssemblyQualifiedName,
                         Result = MessagePayloadDataConverter.ErrorConverter.Serialize(serializationException),
@@ -260,6 +260,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 this.CurrentState = null;
                 this.StateWasAccessed = false;
             }
+
+            return serializationErrorMessage == null;
         }
 
         void IDurableEntityContext.SignalEntity(EntityId entity, string operation, object input)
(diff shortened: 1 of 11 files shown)


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
```diff
@@ -101,7 +169,19 @@ TState IDurableEntityContext.GetState<TState>(Func<TState> initializer)
...
+
+                if (initializer != null)
+                {
+                    try
+                    {
+                        defaultValue = initializer();
+                    }
+                    catch (Exception e)
+                    {
+                        throw new EntitySchedulerException($"failed to construct entity state: {e.Message}", e);
```
cgillum: nit: "failed" should be capitalized. You may also want to prefer "initialize" over "construct" to match the name of the function that threw the exception.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
```diff
@@ -141,11 +228,34 @@ void IDurableEntityContext.SetState(object o)
             this.StateWasAccessed = true;
         }
 
-        internal void Writeback()
+        internal void Writeback(out ResponseMessage stateSerializationFailedNotification)
```
cgillum: nit: consider renaming to `TryWriteback(...)` and having it return `bool`. That follows best practices for .NET API design. A return value of `false` also makes it more clear when to expect the `stateSerializationFailedNotification` to be non-null.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
```diff
@@ -141,11 +228,34 @@ void IDurableEntityContext.SetState(object o)
...
-                this.State.EntityState = MessagePayloadDataConverter.Default.Serialize(this.CurrentState);
+                try
+                {
+                    this.State.EntityState = MessagePayloadDataConverter.Default.Serialize(this.CurrentState);
+                }
+                catch (Exception e)
+                {
+                    // we cannot serialize the entity state - this is an application error.
+                    var serializationException = new EntitySchedulerException(
+                        $"failed to serialize state of '{this.FunctionName}' entity: {e.Message}", e);
```
cgillum: nit: "failed" should be capitalized.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
```diff
@@ -256,22 +366,92 @@ internal void SendEntityMessage(OrchestrationInstance target, string eventName,
...
+                {
+                    Target = target,
+                    EventName = eventName,
+                    EventContent = message,
+                    IsError = isException,
+                });
+            }
+        }
+
+        internal void SendLockMessage(OrchestrationInstance target, string eventName, object message, bool notifyOnFailedWriteback = false)
```
cgillum: The `notifyOnFailedWriteback` parameter is not being used.

src/WebJobs.Extensions.DurableTask/EntityScheduler/RequestMessage.cs
```diff
@@ -78,24 +78,45 @@ internal class RequestMessage
...
+                else
+                {
+                    this.Input = MessagePayloadDataConverter.Default.Serialize(obj);
+                }
             }
-            else
+            catch (Exception e)
             {
-                this.Input = MessagePayloadDataConverter.Default.Serialize(obj);
+                throw new EntitySchedulerException($"failed to serialize input for operation '{this.Operation}': {e.Message}", e);
```
cgillum: nit: "failed" should be capitalized.

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -117,8 +124,15 @@ internal void Rehydrate(string serializedInput)
...
-                // a previous incarnation of this instance called continueAsNew
-                this.context.State = JsonConvert.DeserializeObject<SchedulerState>(serializedInput, MessagePayloadDataConverter.MessageSettings);
+                try
+                {
+                    // a previous incarnation of this instance called continueAsNew
+                    this.context.State = JsonConvert.DeserializeObject<SchedulerState>(serializedInput, MessagePayloadDataConverter.MessageSettings);
+                }
+                catch (Exception e)
+                {
+                    throw new EntitySchedulerException("failed to deserialize entity scheduler state - may be corrupted or wrong version.", e);
```
cgillum: nit: "failed" should be capitalized.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
sebastianburckhardt: ok, I will to that for the other 5 occurrences as well.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
sebastianburckhardt: Makes sense.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
sebastianburckhardt: yes, that's a leftover from copy-paste that should not be there.

PR #949, opened 2019-09-25
Durable Functions Analyzer

amdeel:
Here is the current state of my Durable Functions Analyzer!


DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/DurableFunctionsAnalyzer.csproj
```diff
@@ -0,0 +1,41 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard1.3</TargetFramework>
```
cgillum: We should probably target `netstandard2.0` for this project like we do for the Durable extension. That will make sure there isn't any weird behavior when customers are using packages that target two different netstandard frameworks.

DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/DurableFunctionsAnalyzer.Test/Helpers/CodeFixVerifier.Helper.cs
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        /// <summary>
+        /// Given a document, turn it into a string based on the syntax root
+        /// </summary>
+        /// <param name="document">The Document to be converted to a string</param>
+        /// <returns>A string containing the syntax of the Document after formatting</returns>
+        private static string GetStringFromDocument(Document document)
+        {
+            var simplifiedDoc = Simplifier.ReduceAsync(document, Simplifier.Annotation).Result;
```
ConnorMcMahon: Is it possible to make this method `async`? Calling `.Result` directly can cause some race conditions.

Note this will make all of the callers (and all of their callers up the chain) async as well.

If for some reason that is not possible, the safe way to synchronously call asynchronous methods is this:

`Task.Run(() => _async code here_).GetAwaiter().ConfigureAwait(false).Result`

I'm not 100% sure that is the safe way, but it is something along the lines. Avoid it if at all possible.

DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/SyntaxNodeUtils.cs
```diff
@@ -0,0 +1,178 @@
...
+                    }
+                }
+            }
+            return false;
+        }
+
+        internal static bool TryGetClassSymbol(out INamedTypeSymbol classSymbol, SemanticModel semanticModel)
+        {
+            var classDeclarationSyntax = semanticModel.SyntaxTree.GetRoot().DescendantNodes().OfType<ClassDeclarationSyntax>().Last();
+            if (classDeclarationSyntax != null)
```
ConnorMcMahon: `classDeclarationSyntax` will never be null, as `.Last()` throws an exception if the `IEnumerable` is empty I believe. If you use `.LastOrDefault()`, that will return null if the list is empty, and make this code work better.

DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/SyntaxNodeUtils.cs
```diff
@@ -0,0 +1,178 @@
...
+        public static bool IsInsideOrchestrator(SyntaxNode node)
+        {
+            if (TryGetMethodDeclaration(out SyntaxNode methodDeclaration, node))
+            {
+                var parameterList = methodDeclaration.ChildNodes().Where(x => x.IsKind(SyntaxKind.ParameterList)).First();
+
+                foreach (SyntaxNode parameter in parameterList.ChildNodes())
+                {
+                    var attributeList = parameter.ChildNodes().Where(x => x.IsKind(SyntaxKind.AttributeList));
+                    if (attributeList.Count() >= 1 && attributeList.First().ChildNodes().First().ToString().Equals("OrchestrationTrigger"))
```
ConnorMcMahon: I don't know if we can guarantee that the first attribute is the one that has to have the OrchestratorTrigger. I would imagine most of the time it will be, and if it is not the first, then this will return false negatives over false positives, which is preferable anyways, so i wouldn't highly prioritize that fix.

DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/SyntaxNodeUtils.cs
```diff
@@ -0,0 +1,178 @@
...
+            {
+                returnType = context.SemanticModel.GetTypeInfo((methodDeclaration as MethodDeclarationSyntax).ReturnType).Type;
+                return true;
+            }
+
+            returnType = null;
+            return false;
+        }
+
+        internal static bool TryGetMethodDeclaration(out SyntaxNode methodDeclaration, SyntaxNode node)
```
ConnorMcMahon: NIT: In general, the out parameter of `TryGet` methods go last, and the collection/object you are retrieving the item from goes first.

DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/analyzers/binding/OrchestratorContextAnalyzer.cs
```diff
@@ -0,0 +1,99 @@
...
+                return V1Rule;
+            }
+            else if (version.Equals(DurableVersion.V2))
+            {
+                return V2Rule;
+            }
+            return null;
+        }
+
+        private bool parameterTypeIsCorrectDurableType(SyntaxNode parameterNode)
```
ConnorMcMahon: NIT: Stick to PascalCase for methods in C#.

DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/SyntaxNodeUtils.cs
```diff
@@ -0,0 +1,178 @@
...
+                    classSymbol = classDeclarationSymbol;
+                    return true;
+                }
+            }
+
+            classSymbol = null;
+            return false;
+        }
+
+        internal static bool TryGetFunctionName(out string functionName, SyntaxNode functionAttribute)
```
ConnorMcMahon: This is a note for all of your extension methods. In general, the out parameter of `TryGet` methods go last, and the collection/object you are retrieving the item from goes first.

DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/analyzers/entity/ClassNameAnalyzer.cs
```diff
@@ -0,0 +1,77 @@
...
+                                context.ReportDiagnostic(diagnosticClassName);
+                                context.ReportDiagnostic(diagnosticAttribute);
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        private static bool classNameMatchesFunctionName(INamedTypeSymbol classSymbol, string functionName)
```
ConnorMcMahon: NIT: PascalCase

DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/SyntaxNodeUtils.cs
amdeel: Good catch! I switched this to a for each loop on each AttributeList returned.

DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/Resources.resx
```diff
@@ -0,0 +1,281 @@
...
+    <value>DurableClient attribute must be used with either an IDurableClient, IDurableEntityClient, or an IDurableOrchestrationClient</value>
+  </data>
+  <data name="DurableClientAnalyzerMessageFormat" xml:space="preserve">
+    <value>DurableClient attribute is applied to a '{0}' but must be used with either an IDurableClient, IDurableEntityClient, or an IDurableOrchestrationClient</value>
+  </data>
+  <data name="DurableClientAnalyzerTitle" xml:space="preserve">
+    <value>DurableClient attribute must be used with either an IDurableClient, IDurableEntityClient, or an IDurableOrchestrationClient</value>
+  </data>
+  <data name="EntityClassNameAnalyzerDescription" xml:space="preserve">
+    <value>Entity class name and Entity function name must match</value>
```
cgillum: The second "Entity" should be lowercased. Same feedback for the other messages since "entity" isn't being used as a proper noun.

DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/Resources.resx
```diff
@@ -0,0 +1,281 @@
...
+  </data>
+  <data name="EntityClassNameAnalyzerMessageFormat" xml:space="preserve">
+    <value>Class name ('{0}') and entity function name ('{1}') do not match</value>
```
cgillum: In other places you just use single quotes `'{0}'`, but here you're using both parenthesis and single quotes `('{0}')`. I suggest we just use the former for all of our messages.

DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/Resources.resx
```diff
@@ -0,0 +1,281 @@
...
+    <value>EntityTrigger attribute must be used with an IDurableEntityContext</value>
+  </data>
+  <data name="EntityContextAnalyzerMessageFormat" xml:space="preserve">
+    <value>EntityTrigger attribute is applied to a '{0}' but must be used with an IDurableEntityContext instead</value>
+  </data>
+  <data name="EntityContextAnalyzerTitle" xml:space="preserve">
+    <value>EntityTrigger attribute must be used with an IDurableEntityContext</value>
+  </data>
+  <data name="EntityInterfaceContentAnalyzerDescription" xml:space="preserve">
+    <value>An Entity Interface must only contain methods and must contain at least one method.</value>
```
cgillum: Let's use lowercase "i" for "interface".

DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/Resources.resx
```diff
@@ -0,0 +1,281 @@
...
+  </data>
+  <data name="EntityInterfaceContentAnalyzerMessageFormat" xml:space="preserve">
+    <value>Interface contains content other than methods
```
cgillum: Instead of "content", I suggest you use the word "members".

DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/Resources.resx
```diff
@@ -0,0 +1,281 @@
...
+    <value>Method '{0}' has a return type '{1}'; return type should only be void, Task, or Task&lt;T&gt;</value>
+  </data>
+  <data name="EntityInterfaceReturnTypeAnalyzerTitle" xml:space="preserve">
+    <value>An Entity Interface must only have methods that return void, Task, or Task&lt;T&gt;</value>
+  </data>
+  <data name="EntityStaticAnalyzerDescription" xml:space="preserve">
+    <value>Entity functions must be static</value>
+  </data>
+  <data name="EntityStaticAnalyzerMessageFormat" xml:space="preserve">
+    <value>Function '{0}' is not marked static</value>
```
cgillum: I would clarify this to be "Entity function '{0}' is not marked static."

DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/Resources.resx
```diff
@@ -0,0 +1,281 @@
...
+  </data>
+  <data name="EntityStaticAnalyzerTitle" xml:space="preserve">
+    <value>Entity functions must be static</value>
+  </data>
+  <data name="EnvironmentVariableAnalyzerTitle" xml:space="preserve">
+    <value>Environment Variables must be accessed in a deterministic way inside an orchestrator</value>
```
cgillum: Let's use lowercase "v" in "variables"

DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/Resources.resx
```diff
@@ -0,0 +1,281 @@
...
+  </data>
+  <data name="GuidAnalyzerTitle" xml:space="preserve">
+    <value>Guid calls must be deterministic inside an orchestrator</value>
```
cgillum: I would say "orchestrator function" rather than just "orchestrator". Applies to this message as well as others.

DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/Resources.resx
```diff
@@ -0,0 +1,281 @@
...
+  </data>
+  <data name="IOTypesAnalyzerTitle" xml:space="preserve">
+    <value>I/O operations are not allowed inside an orchestrator</value>
+  </data>
+  <data name="MethodAttributeAnalyzerTitle" xml:space="preserve">
+    <value>Method calls in an orchestrator must be marked [Deterministic] above the method declaration</value>
```
cgillum: I would remove "above the method declaration". The C# compiler allows it to be above or even to the left of the method declaration.

DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/Resources.resx
```diff
@@ -0,0 +1,281 @@
...
+    <value>Activity Function named '{0}' takes a '{1}' but was given a '{2}'</value>
+  </data>
+  <data name="ActivityArgumentAnalyzerTitle" xml:space="preserve">
+    <value>Activity Function call is using the wrong argument type</value>
+  </data>
+  <data name="ActivityNameAnalyzerCloseMessageFormat" xml:space="preserve">
+    <value>Activity function named '{0}' does not exist. Did you mean '{1}'?</value>
+  </data>
+  <data name="ActivityNameAnalyzerDescription" xml:space="preserve">
+    <value>Activity Function call references unknown Activity Function</value>
```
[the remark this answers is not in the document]
amdeel: For the message format:
Yes because {0} is the location of the actual diagnostic. {1} is an activity function that does exist with the most similar name. This is really useful for customers who might be barely off (EX: "ProcessString" vs "ProcessStrings"). This helps them fix their code easily without having to search through all of their function definitions to see which function they might have intended to call.

For the description:
This is a general description without a formatted string

DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/DurableFunctionsAnalyzer/Resources.resx
amdeel: Ok, I'll change it to "on the method declaration." This message will be displayed on the method invocation expression inside the orchestrator function.

cgillum:
I went ahead and made some structural changes based on my original feedback:

- Updated version to v0.1.0
- Restructured project and file locations to match repo convensions
- Fixed DLL name and C# namespaces to match code convensions
- Updated target framework to netstandard2.0
- Updated nuget package dependencies to non-preview versions
- Added strong-name signing

64/83 tests are currently passing, which is the same number from the previous commit.

The latest version can be found [on myget (v0.1.0)](https://www.myget.org/feed/azure-appservice/package/nuget/Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers).

Issue #912, opened 2019-08-29
Support for ITokenSource in Durable HTTP out-of-proc


cgillum:
### Problem
The initial out-of-proc (i.e. JavaScript) Durable HTTP implementation does not support the ITokenSource feature. This means non-.NET orchestrations cannot use the managed identity capabilities of Durable HTTP.

### Additional context
The current out-of-proc schema for starting an HTTP request looks like the following:

```json
{
    "actionType": "CallHttp",
    "httpRequest":
    {
        "method": "POST",
        "uri": "https://example.com",
        "headers": {
            "Content-Type": "application/json",
            "Accept": ["application/json","application/xml"],
            "x-ms-foo": []
        },
        "content": "5"
    }
}
```

### Solution
To support managed identities, we could update the HTTP schema as follows:

```json
{
    "actionType": "CallHttp",
    "httpRequest":
    {
        "method": "POST",
        "uri": "https://example.com",
        "headers": {
            "Content-Type": "application/json",
            "Accept": ["application/json","application/xml"],
            "x-ms-foo": []
        },
        "content": "5",
        "tokenSource": {
            "type": "AzureManagedIdentity",
            "resource": "https://management.core.windows.net"
        }
    }
}
```

The new `tokenSource` field is a JSON object with a `type` field with a value of `AzureManagedIdentity` and a `resource` field that specifies any Azure AD resource (in this case `https://management.core.windows.net`, which corresponds to ARM).

### Future improvements
In the future, we could add more possible values for `type`. For example, `azureTokenStore` for fetching tokens from the [Azure Token Store service](https://github.com/Azure/azure-tokens) (currently in preview). Custom token providers would likely require users to define a function that returns a token.

cgillum:
The work for this has been completed and will be part of the beta3 release.

PR #953, opened 2019-09-30
Durable HTTP Managed Identities for out-of-proc

cgillum:
Resolves https://github.com/Azure/azure-functions-durable-extension/issues/912.

Most of the work is the custom serialization logic to support the proposed out-of-proc schema.

I also included some other small changes to Durable HTTP that I wanted to sneak in before GA.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
@@ -67,9 +67,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         public string Content { get; }
 
         /// <summary>
-        /// Information needed to get a token for a specified service.
+        /// Mechanism for attaching an OAuth token to the request.
         /// </summary>
-        [JsonProperty("tokenSource", TypeNameHandling = TypeNameHandling.Auto)]
+        [JsonProperty("tokenSource")]
+        [JsonConverter(typeof(TokenSourceConverter))]
         public ITokenSource TokenSource { get; }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


src/WebJobs.Extensions.DurableTask/ITokenSource.cs
```diff
@@ -6,9 +6,7 @@
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 {
     /// <summary>
-    /// Provides functionality available to sources of tokens.
-    /// Custom implementation of this interface must include the use
```
ConnorMcMahon: Did you remove the documentation about `[DataContract]`/`[DataMember]`  because it was too much of an implementation detail, or does the use of TokenSourceConverter remove that requirement?

If the latter, we should remove `[DataContract]` and `[DataMember]` from `MockTokenSource` in DurableHttpTests.cs

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -59,7 +59,7 @@
     <PackageReference Include="Microsoft.Azure.DurableTask.Redis" Version="0.1.1-alpha" />
     <PackageReference Include="Microsoft.Azure.DurableTask.Emulator" Version="2.1.3" />
     <PackageReference Include="Microsoft.Azure.DurableTask.AzureStorage" Version="1.6.4" />
-    <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.0.3" />
+    <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.3.1" />
```
ConnorMcMahon: I checked to see that we aren't using a fixed version of this library in V1, and it looks like Functions doesn't have any dependencies on it, so we should be good to use whatever version we want. 

However, once we pick a version for Functions V1, do we have to keep that version consistent to avoid breaking customers?

src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
```diff
@@ -110,5 +111,68 @@ public override bool CanConvert(Type objectType)
                 writer.WriteValue(method.ToString());
             }
         }
+
+        private class TokenSourceConverter : JsonConverter
+        {
+            private enum TokenSourceType
+            {
+                None = 0,
```
ConnorMcMahon: Is this just a placeholder, or do we ever see using this enum value?

src/WebJobs.Extensions.DurableTask/ITokenSource.cs
cgillum: The `TokenSourceConverter` removes this requirement. However, I suspect I'll still need `[DataContract]` and `[DataMember]` on the `MockTokenSource` since it uses standard serialization. It's not a pattern we want others to use though, so I don't want to document it.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
cgillum: Thanks for checking this. I don't expect we'd need to freeze the version we're using. I'm not 100% sure, but my intuition is that V1 customers will have some flexibility to reference other 1.x.x versions. Even so, I'm slightly less worried about breaking customers since they would need to opt-into any updated versions anyways.

src/WebJobs.Extensions.DurableTask/ITokenSource.cs
ConnorMcMahon: Do we not want customers to be able to provide their own customer `ITokenSource` implementations? Or just not ones with state?

src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
cgillum: It's just a placeholder. I recently started a habit adding default enum values like this so that the absence of any value does not automatically get mapped to a real value.

src/WebJobs.Extensions.DurableTask/ITokenSource.cs
cgillum: I guess I hadn't really thought about that. My thinking was that we'd want to come up with some mechanism that works for out-of-proc languages too, but I guess the current model works well-enough for .NET. I would still rather we not include the custom implementation comments here. I think it would be better to put that information in official docs.

Issue #931, opened 2019-09-10
DELETE for single entity rest API


jeffhollan:
Would be nice to have a method supported of `DELETE` that would destroy an instance of an entity.

Somewhat related to this
https://github.com/Azure/azure-functions-durable-extension/issues/930

sebastianburckhardt:
Yes, I think that is a good idea. I think this also begs the question if we should add DeleteEntity/DeleteEntityAsync methods to IDurableEntityClient/IDurableOrchestrationClient.

This could avoid problems where users struggle to discover how to delete entities #932.

cgillum:
I agree - and think we should consider this for the final 2.0 release.

sebastianburckhardt:
I think we need to simplify the current state management further. Currently, the entity state management (get/set state value or object) is separate from the entity lifetime management (entity is created when called, is deleted when `destructOnExit`). 

I think it should all be just "state management". So we would just support (get/set/delete) on the entity state, without the "entity itself" (whatever that means) being subject to management operations.

sebastianburckhardt:
To keep DELETE as simple as possible I think I want to treat it exactly like regular operation called "delete".  Concretely, I plan to implement this as follows:

- calling a DELETE http request on the entity is exactly equivalent to signaling an operation called 'delete'. 

- Inside `DispatchAsync<T>`, when the operation is called `delete` but  `T` does not have a method called `Delete`, act as if there is a method
```csharp
void Delete() 
{
   Entity.Current.DeleteState();
}
```

cgillum:
I like the simplicity, but could this confuse customers who have defined a `delete` operation in their entity already? In that case it seems that a DELETE API call would invoke their operation rather than deleting the entity state.

sebastianburckhardt:
Yes, that is the intent. This can be beneficial, for example to allow users to delete additional state when the entity is deleted. Of course we have to explain it. 

The important point to get across is to that DELETE is not some special management API, but just like any other user-defined operation, and has all the same behaviors (regarding locks, signal vs call, etc). 

BTW, I briefly considered changing the behavior of POST to make it more similar to DELETE: meaning that when you call POST on an entity without specifying the operation name, we invoke an entity operation called 'post'. It is a breaking change though.

Areson:
[...]

PR #957, opened 2019-10-01
Simplify entity state handling

sebastianburckhardt:
In the current implementation, "the existence of state" is somewhat independent of "the existence of an entity". State is created by `ctx.GetState<T>` or `ctx.SetState` (where ctx is the entity context, which can only be accessed while executing an entity operation), and can not be deleted; an entity is created implicitly by any operation, which can be detected after the fact by `ctx.IsFreshlyCreated` and is deleted by calling `ctx.DestructOnExit()` (which also deletes the state). 

*I think this is too confusing* - why not just unify the concepts and the terminology, i.e. consider "entity exists = state exists".

Proposed semantics:

- State is created when an operation calls `ctx.GetState<T>` or `ctx.SetState`
- Existence of State can be tested with `ctx.HasState` 
- State is deleted when an operation calls `ctx.DeleteState` 
  (or via REST API, once we get around to #931) 

This PR does that, and removes `ctx.IsFreshlyCreated` and `ctx.DestructOnExit`.

### DIFF
--- a/samples/v2/entitites-csharp/RideSharing/RideSharing/Entities/UserEntity.cs
+++ b/samples/v2/entitites-csharp/RideSharing/RideSharing/Entities/UserEntity.cs
@@ -34,7 +34,7 @@ namespace RideSharing
         {
             // we initialize the entity explicitly here (instead of relying on the implicit default constructor)
             // so we can set the user id to match the entity key
-            if (context.IsNewlyConstructed)
+            if (!context.HasState)
             {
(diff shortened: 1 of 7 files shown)


cgillum:
Cool. I noticed that this changes some of our public APIs. Would you mind opening a PR to update our public documentation to reflect this change?  You can just click the edit button at the top of the docs page: https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-bindings#entity-trigger

PR #959, opened 2019-10-03
Fix a bug related to using default state initialization + primitives

### LINKED ISSUE #954: Durable entity function keeps firing
### Description
During local development I am experiencing an issue where an entity function is repeatedly firing after being invoked. Initially I thought it might have been something in my code but I've managed to reproduce the issue with this code:

```csharp
namespace MyNamespace
{
    public static class MyClass
    {
        [FunctionName("MyClass")]
        public static void Run([EntityTrigger] IDurableEntityContext ctx)
        {
            var state = ctx.GetState<int>(() => 0);
            state++;
            ctx.SetState(state);
        }
    }
}
```

Which is invoked via the following:

```csharp
var entityId = new EntityId("MyClass", uniqueid);
await entityClient.SignalEntityAsync(entityId, "Evaluate");
```

The function gets called no problems, and then it gets called again, and again and again. In my scenario the HTTP triggered function gets a quick burst of messages which it routes to the same entity instance, then another burst, then a bit of a break and another burst with the traffic to that entity eventually going back to zero within a minute or so.

What I am seeing though is that once I trigger the function against the entity once it keeps of running it again and again. I can't see any exceptions in the console output to make me think it is retrying. And it doesn't appear to stop.

I've proven to myself that it can't be any external traffic by shutting off the source of HttpTriggers so it can't reach my local dev environment (so they aren't new incoming events).

### Expected behavior
I would expect that a call against a durable entity would be called once, and exactly once.

### Actual behavior
Continued calls once triggered.

### Known workarounds
I can stop the replay by resetting the dev storage and deleting all the tables, queues and blobs.

### App Details

- **Durable Functions extension version (e.g. v1.8.3)**: 2.0.0-beta
- **Azure Functions runtime version (1.0 or 2.0)**: 2.0.12673.0
- **Programming language used**: csharp

cgillum:
Fixes #954 

The `GetState<TState>(Func<TState> initializer` breaks when `TState` is not a complex type. For example, consider the following "functional" entity definition:

```csharp
[FunctionName("MyClass")]
public static void RunOperation([EntityTrigger] IDurableEntityContext ctx)
{
    int state = ctx.GetState(() => 0);
    state++;
    ctx.SetState(state);
}
```

The entity gets and saves its state correctly the first time, but subsequent calls to `GetState` fail with a JSON deserialization exception:

```
Microsoft.Azure.WebJobs.EntitySchedulerException
  HResult=0x80131500
  Message=Failed to deserialize entity state: Unexpected initial token 'Integer' when populating object. Expected JSON object or array. Path '', line 1, position 1.
  Source=Microsoft.Azure.WebJobs.Extensions.DurableTask
  StackTrace:
   at Microsoft.Azure.WebJobs.Extensions.DurableTask.DurableEntityContext.GetPopulatedState[TState](TState initialValue, Boolean usedTypeInitializer) in C:\GitHub\azure-functions-durable-extension\src\WebJobs.Extensions.DurableTask\ContextImplementations\DurableEntityContext.cs:line 222
   at Microsoft.Azure.WebJobs.Extensions.DurableTask.DurableEntityContext.Microsoft.Azure.WebJobs.IDurableEntityContext.GetState[TState](Func`1 initializer) in C:\GitHub\azure-functions-durable-extension\src\WebJobs.Extensions.DurableTask\ContextImplementations\DurableEntityContext.cs:line 190
   at VSSample.MyClass.RunOperation(IDurableEntityContext ctx) in C:\GitHub\azure-functions-durable-extension\samples\v2\precompiled\Repro.cs:line 14
   at Microsoft.Azure.WebJobs.Host.Executors.VoidMethodInvoker`2.InvokeAsync(TReflected instance, Object[] arguments) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Executors\VoidMethodInvoker.cs:line 20
   at Microsoft.Azure.WebJobs.Host.Executors.FunctionInvoker`2.<InvokeAsync>d__10.MoveNext() in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Executors\FunctionInvoker.cs:line 52
```

This PR is a simple fix which avoids using the `JsonConvert.PopulateObject` API unless we know that `TState` is a class.

One downside of this fix is that we can't use the `PopulateObject` API for structs. I'm not aware of a way to know whether `TState` represents a JSON object or a JSON primitive value. If we had a way to know for sure whether `TState` represented a JSON object, then we could use `PopulateObject` for that as well. Without this, the behavior for structs is that we will ignore any pre-populated state. It's a bit of a corner case (structs + default initializers), so it might be worth the tradeoff.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
@@ -169,58 +169,86 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             this.ThrowIfInvalidAccess();
 
-            if (!this.StateWasAccessed)
+            if (this.StateWasAccessed)
             {
-                TState defaultValue = default;
+                return (TState)this.CurrentState;
+            }
 
+            TState result;
+
+            if (this.State.EntityExists)
+            {
+                try
+                {
+                    result = JsonConvert.DeserializeObject<TState>(this.State.EntityState);
+                }
+                catch (Exception e)
+                {
+                    throw new EntitySchedulerException($"Failed to deserialize entity state: {e.Message}", e);
+                }
+            }
+            else
+            {
                 if (initializer != null)
                 {
                     try
                     {
-                        defaultValue = initializer();
+                        result = initializer();
                     }
                     catch (Exception e)
                     {
                         throw new EntitySchedulerException($"Failed to initialize entity state: {e.Message}", e);
                     }
                 }
-
-                if (this.State.EntityState != null)
-                {
-                    this.CurrentState = this.GetPopulatedState(defaultValue, initializer != null);
-                }
                 else
                 {
-                    this.CurrentState = defaultValue;
+                    result = default(TState);
                 }
-
-                this.StateWasAccessed = true;
-                this.State.EntityExists = true;
             }
 
-            return (TState)this.CurrentState;
+            this.CurrentState = result;
+            this.StateWasAccessed = true;
+            this.State.EntityExists = true;
+            return result;
         }
 
-        private TState GetPopulatedState<TState>(TState initialValue, bool usedTypeInitializer)
+        internal TState GetStateWithInjectedDependencies<TState>(Func<TState> constructor)
+            where TState : class
         {
+            this.ThrowIfInvalidAccess();
+
+            if (this.StateWasAccessed)
+            {
+                return (TState)this.CurrentState;
+            }
+
+            TState result;
+
             try
             {
-                if (usedTypeInitializer && typeof(TState).IsClass)
-                {
-                    // Only populate serialized state, as some fields may be populated by the initializer
-                    // using dependency injection. Note that we can do this for classes but not for structs.
-                    JsonConvert.PopulateObject(this.State.EntityState, initialValue);
-                    return initialValue;
-                }
-                else
-                {
-                    return JsonConvert.DeserializeObject<TState>(this.State.EntityState);
-                }
+                result = constructor();
             }
             catch (Exception e)
             {
-                throw new EntitySchedulerException($"Failed to deserialize entity state: {e.Message}", e);
+                throw new EntitySchedulerException($"Failed to construct entity state object: {e.Message}", e);
             }
+
+            if (this.State.EntityExists)
+            {
+                try
+                {
+                    JsonConvert.PopulateObject(this.State.EntityState, result);
+                }
+                catch (Exception e)
+                {
+                    throw new EntitySchedulerException($"Failed to populate entity state from JSON: {e.Message}", e);
+                }
+            }
+
+            this.CurrentState = result;
+            this.StateWasAccessed = true;
+            this.State.EntityExists = true;
+            return result;
         }
 
         void IDurableEntityContext.SetState(object o)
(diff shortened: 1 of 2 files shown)


sebastianburckhardt:
I think the problem was introduced when we modified `GetState` to add support for dependency injection with `DispatchAsync` - to support dependency injection, 'Populate' was introduced for constructing objects, rather then the normal Json deserialization, which changes the semantics considerably in all kinds of ways.  

Since dependency injection is really only needed when using the class-based model, my preference would be to return GetState to its original behavior (which does not use populate, but standard json deserialization); and then implement `DispatchAsync` separately (so we would be using  dependency injection and Populate only for code that calls `DispatchAsync`). I think this solves this issue, and provides cleaner semantics for GetState in the untyped model.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
```diff
@@ -169,58 +169,86 @@ TState IDurableEntityContext.GetState<TState>(Func<TState> initializer)
...
-                    // using dependency injection.
-                    JsonConvert.PopulateObject(this.State.EntityState, initialValue);
-                    return initialValue;
+                    JsonConvert.PopulateObject(this.State.EntityState, result);
                 }
-                else
+                catch (Exception e)
                 {
-                    return JsonConvert.DeserializeObject<TState>(this.State.EntityState);
+                    throw new EntitySchedulerException($"Failed to populate entity state from JSON: {e.Message}", e);
```
ConnorMcMahon: Would it be useful to include the type we tried to serialize into in this exception message, or is `TState` included in the stack trace we emit?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
```diff
@@ -169,58 +169,86 @@ TState IDurableEntityContext.GetState<TState>(Func<TState> initializer)
...
             {
-                if (usedTypeInitializer)
+                result = constructor();
+            }
+            catch (Exception e)
+            {
+                throw new EntitySchedulerException($"Failed to construct entity state object: {e.Message}", e);
+            }
+
+            if (this.State.EntityExists)
```
ConnorMcMahon: With this approach, we are essentially marking the entity as existing if the customer invokes the entity via `DispatchAsync()` if they are using dependency injection.

Just double checking that this isn't breaking our entity state/existance semantics here.

sebastianburckhardt: This seems right to me. DispatchAsync() implicitly calls GetState() which always marks the entity as existing. Basically, once the object has been constructed, the entity exists.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
sebastianburckhardt: The error messages usually contain the name of the failed function which is the same as the name of the type (when using the class-based syntax), so I think that is good enough.

cgillum:
@ConnorMcMahon from your perspective are we good to merge this PR?

PR #965, opened 2019-10-08
Fixed Deterministic CodeFixProviders

amdeel:


src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/Entity/DispatchClassNameAnalyzer.cs
```diff
@@ -44,26 +42,15 @@ private void AnalyzeDispatchEntityName(SyntaxNodeAnalysisContext context)
...
-
-                            context.ReportDiagnostic(diagnostic);
-
-                            return;
-                        }
-                        else
-                        {
-                            if (SyntaxNodeUtils.TryGetClassSymbol(expression, context.SemanticModel, out INamedTypeSymbol classSymbol))
+                            var className = classSymbol.Name.ToString();
+                            var identifierName = typeArgumentList.ChildNodes().First();
```
cgillum: `First()` will throw an exception if `ChildNodes()` is an empty list. Do we know that `typeArgumentList.GetChildNodes()` will always return at least one identifier?

test/WebJobs.Extensions.DurableTask.Analyzers.Test/ActivityFunction/ArgumentAnalyzerTests.cs
```diff
@@ -9,7 +9,7 @@
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers.Test.ActivityFunction
 {
-    [TestClass]
+    //[TestClass]
```
cgillum: Did you mean to leave this commented out? If so, could you add a code comment explaining why it's disabled?

src/WebJobs.Extensions.DurableTask.Analyzers/CodefixProviders/Orchestrator/TimerCodeFixProvider.cs
```diff
@@ -51,22 +51,22 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
 
                         if (TryGetCancellationTokenParameter(expression, semanticModel, out string cancellationToken))
```
cgillum: GitHub won't let me add comments further up. I see that we check for `SyntaxNodeUtils.IsInsideOrchestrator` further up on line 44, but do we also need to check if this is inside a method with the `[Deterministic]` attribute?

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/Entity/DispatchClassNameAnalyzer.cs
amdeel: TypeArgumentList will always have a child node. If the type is omitted its called OmittedTypeArgument. I will add a comment to this.

test/WebJobs.Extensions.DurableTask.Analyzers.Test/ActivityFunction/ArgumentAnalyzerTests.cs
amdeel: I can add a comment to this.

src/WebJobs.Extensions.DurableTask.Analyzers/CodefixProviders/Orchestrator/TimerCodeFixProvider.cs
amdeel: This code fix only applies to code inside an orchestrator. Like the DateTime fix, it grabs the context variable name. There is an else if for the deterministic attribute code fix.

amdeel: Also, this code fix only applies for V2. V1 will not have a code fix associated. I noticed that V2 added DurableContextExtension, which added this method overload for CreateTimer that I'm using for the code fix.

amdeel: I will update the message format string for this analyzer for the V1 case so that it mentions the CreateTimer fix.

PR #974, opened 2019-10-15
Utilize Dependency Injection for storage providers

ConnorMcMahon:
Currently we have the following dependencies in our v2 branch.

1. DurableTask.AzureStorage
2. DurableTask.Redis
3. DurableTask.Emulator

Requiring the customers to explicitly take on these dependencies, even if they do not use them, is not a sustainable approach to adding new storage providers. It also forces us to explicitly add support for storage providers in order for customers to use it.

This PR strips out the explicit Redis and Emulator dependencies for customers, and adds support for using Dependency Injection to bring in new backend storage providers for their application. It also provides some helper classes and a new programming model for this, which abstracts away some of the discrepancies between DurableTask.Core and Durable Functions.

**Example storage provider:**

```csharp

using DurableTask.Redis;
using Microsoft.Extensions.Options;

{
    internal class RedisDurabilityProviderFactory : IDurabilityProviderFactory
    {
        private readonly string defaultConnectionName;

        private readonly DurabilityProvider defaultProvider;
        private readonly string defaultHubName;
        private readonly IConnectionStringResolver connectionResolver;

        public RedisDurabilityProviderFactory(IOptions<DurableTaskOptions> options, IConnectionStringResolver connectionStringResolver)
        {
            this.defaultConnectionName = options.Value.StorageProvider["connectionName"] as string;
            string redisConnectionString = connectionStringResolver.Resolve(this.defaultConnectionName);
            this.defaultHubName = options.Value.HubName;
            this.connectionResolver = connectionStringResolver;
            var defaultTaskHubService = new RedisOrchestrationService(new RedisOrchestrationServiceSettings()
            {
                TaskHubName = this.defaultHubName,
                RedisConnectionString = redisConnectionString,
            });

            this.defaultProvider = new DurabilityProvider("Redis", defaultTaskHubService, defaultTaskHubService, this.defaultConnectionName);
        }

        public bool SupportsEntities => false;

        public DurabilityProvider GetDurabilityProvider(DurableClientAttribute attribute)
        {
            if (string.IsNullOrEmpty(attribute.TaskHub) && string.IsNullOrEmpty(attribute.ConnectionName))
            {
                return this.defaultProvider;
            }

            if (string.Equals(attribute.TaskHub, this.defaultHubName) && string.Equals(attribute.ConnectionName, this.defaultConnectionName))
            {
                return this.defaultProvider;
            }

            string redisConnectionString = this.connectionResolver.Resolve(attribute.ConnectionName);
            var redisOrchestartionService = new RedisOrchestrationService(new RedisOrchestrationServiceSettings()
            {
                TaskHubName = attribute.TaskHub,
                RedisConnectionString = redisConnectionString,
            });

            return new DurabilityProvider("Redis", redisOrchestartionService, redisOrchestartionService, attribute.ConnectionName);
        }

        public DurabilityProvider GetDurabilityProvider()
        {
            return this.defaultProvider;
        }
    }
}

```

**Example Startup file:**
```csharp
using DurableFunctionsCustomStorage;
using Microsoft.Azure.Functions.Extensions.DependencyInjection;
using Microsoft.Extensions.DependencyInjection;

[assembly: FunctionsStartup(typeof(MyNamespace.Startup))]

namespace MyNamespace
{
    public class Startup : FunctionsStartup
    {
        public override void Configure(IFunctionsHostBuilder builder)
        {
            builder.Services.AddSingleton<IDurabilityProviderFactory, RedisDurableProviderFactory>();
        }
    }
}
```


src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        /// <inheritdoc/>
+        public async override Task<string> RetrieveSerializedEntityState(string inputState)
+        {
+            // the input was compressed... read it back from blob
+            return await this.serviceClient.DownloadBlobAsync(inputState);
+        }
+
+        /// <inheritdoc/>
+        public async override Task<int> PurgeInstanceHistoryByInstanceId(string instanceId)
```
cgillum: Low-pri/optional: We may want to consider a richer return type. Otherwise it will require a new API if we want to return more info in the future.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -55,7 +57,7 @@ public string HubName
         /// <summary>
         /// The section of configuration related to storage providers.
         /// </summary>
-        public StorageProviderOptions StorageProvider { get; set; }
+        public IDictionary<string, object> Storage { get; set; } = new Dictionary<string, object>();
```
cgillum: This is pretty generic. I suggest we update the documentation comment to describe what this property is storing. One question I have: the property name is "Storage" - does that refer to Azure Storage? If so, we should probably rename this property to `AzureStorage`.  Also, does this mean we will add properties like this for each durability provider we support (e.g. "AzureEventHubs")?

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -172,16 +172,32 @@ internal string GetDebugString()
...
+        }
+
         /// <summary>
         /// A helper method to help retrieve the connection string name for the configured storage provider.
         /// </summary>
         /// <returns>The connection string name for the configured storage provider.</returns>
         public string GetConnectionStringName()
         {
-            return this.StorageProvider.GetConfiguredProvider().ConnectionStringName;
+            if (this.Storage.ContainsKey("connectionStringName"))
```
cgillum: I assume this is specific to the Azure Storage provider. How will the Event Hubs provider specify its connection string? I have many more questions about this, but I think there are some fundamental design pieces that I need to understand first.

ConnorMcMahon: I thought about this a bit more and believe I have addressed this in my latest iteration. I will add clarification in the morning of how this works.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
ConnorMcMahon: Will add examples in the morning.

ConnorMcMahon: The Redis factory shown in the example now shows how different storage providers can utilize this.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
ConnorMcMahon: This is now more clearly handled by the factory/storage provider. The factory provides the connection name to the storage provider, and then the default storage provider instance's connection name is used to build our HTTP API payload, which is the only place this method was used.

test/Common/TestHelpers.cs
```diff
@@ -48,54 +48,40 @@ internal static class TestHelpers
             IDurableHttpMessageHandlerFactory durableHttpMessageHandler = null,
             ILifeCycleNotificationHelper lifeCycleNotificationHelper = null)
         {
-            var durableTaskOptions = new DurableTaskOptions
```
ConnorMcMahon: This can likely be restored to most of it's original logic. This was for the more original complicated design.

test/FunctionsV2/WebJobs.Extensions.DurableTask.Tests.V2.csproj
```diff
@@ -27,11 +27,21 @@
...
+    <PackageReference Include="Microsoft.Azure.DurableTask.Emulator" Version="2.1.3" />
+    <PackageReference Include="Microsoft.Azure.DurableTask.Redis" Version="0.1.1-alpha" />
   </ItemGroup>
 
   <ItemGroup>
     <Compile Include="..\Common\*.cs" />
   </ItemGroup>
+
+  <PropertyGroup Condition="true">
+    <DefineConstants>NET_CORE</DefineConstants>
```
ConnorMcMahon: I'm thinking of renaming this to FUNCTIONS_V2, so it is more explicit.

ConnorMcMahon: Probably wouldn't be a bad idea to do something similar for conditional compilation in the source code as well.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -55,18 +55,17 @@
...
+    <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.0.3" />
   </ItemGroup>
 
   <!-- NuGet Publishing Metadata -->
   <PropertyGroup>
-    <Title>Azure Functions Durable Task Extension</Title>
+    <Title>Azure Functions Durable Task Extension Core</Title>
     <Authors>Microsoft</Authors>
-    <Description>Azure WebJobs SDK Extension for the Durable Task Framework</Description>
+    <Description>Shared logic for the Azure WebJobs SDK Extensions for the Durable Task Framework</Description>
```
ConnorMcMahon: This can be restored to original behavior.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -55,18 +55,17 @@
...
+    <PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="2.1.3" />
     <PackageReference Include="Microsoft.Azure.DurableTask.AzureStorage" Version="1.6.4" />
-    <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.3.1" />
+    <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.0.3" />
   </ItemGroup>
 
   <!-- NuGet Publishing Metadata -->
   <PropertyGroup>
-    <Title>Azure Functions Durable Task Extension</Title>
+    <Title>Azure Functions Durable Task Extension Core</Title>
```
ConnorMcMahon: Need to revert.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -172,16 +172,18 @@ internal string GetDebugString()
             return sb.ToString();
         }
 
-        /// <summary>
-        /// A helper method to help retrieve the connection string name for the configured storage provider.
-        /// </summary>
-        /// <returns>The connection string name for the configured storage provider.</returns>
-        public string GetConnectionStringName()
+        private void AppendStorageProviderValuesToDebugString(StringBuilder sb)
```
ConnorMcMahon: Currently, this only logs what the customer's actually put in the host.json. It may be a better idea to make the factory responsible for logging, as then we get the default values of `AzureStorageOptions` logged as well.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -172,16 +172,18 @@ internal string GetDebugString()
...
+            foreach (var value in this.StorageProvider)
+            {
+                sb.Append(value.Key).Append(": ").Append(value.Value).Append(", ");
+            }
+
+            sb.Append(" }, ");
         }
 
-        internal void Validate()
+        internal virtual void Validate()
```
ConnorMcMahon: Does not need to be virtual anymore.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -295,11 +289,11 @@ async Task<DurableOrchestrationStatus> IDurableOrchestrationClient.GetStatusAsyn
...
             // a better way of handling this
             IList<OrchestrationState> stateList;
-            if (this.client.ServiceClient is AzureStorageOrchestrationService serviceClient)
+            try
             {
-                stateList = await serviceClient.GetOrchestrationStateAsync(instanceId, allExecutions: false, fetchInput: showInput);
+                stateList = await this.durableFunctionsClient.GetOrchestrationStateAsync(instanceId, showHistory, showInput);
             }
-            else
+            catch
```
cgillum: Is there a specific exception we can catch for this?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -396,11 +374,9 @@ private async Task<EntityStateResponse<T>> ReadEntityStateAsync<T>(TaskHubClient
             {
                 string serializedState;
 
-                if (this.client.ServiceClient is AzureStorageOrchestrationService service
-                    && state.Input.StartsWith("http"))
+                if (state.Input.StartsWith("http"))
```
cgillum: The blob compression stuff is specific to the Azure Storage provider, so it seems odd to have logic for handling it at this layer. We may want to consider refactoring how we handle this in a future PR.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -429,105 +405,137 @@ private async Task<EntityStateResponse<T>> ReadEntityStateAsync<T>(TaskHubClient
...
-            var serviceClient = this.client.ServiceClient as AzureStorageOrchestrationService;
-            if (serviceClient == null)
-            {
-                throw new NotSupportedException("Only the Azure Storage state provider is currently supported for the purge instance history feature");
-            }
-
-            AzureStorage.PurgeHistoryResult purgeHistoryResult =
-                await serviceClient.PurgeInstanceHistoryAsync(instanceId);
-            return new PurgeHistoryResult(purgeHistoryResult.InstancesDeleted);
+            int numInstancesDeleted = await this.durableFunctionsClient.PurgeInstanceHistoryByInstanceId(instanceId);
```
cgillum: Related to another one of my comments, would it make more sense to have `PurgeInstanceHistoryByInstanceId` return an instance of `PurgeHistoryResult`?

src/WebJobs.Extensions.DurableTask/DurableOrchestrationStatus.cs
```diff
@@ -2,6 +2,9 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using DurableTask.Core;
```
cgillum: Do we really need all these new `using` statements?

src/WebJobs.Extensions.DurableTask/IDurabilityProviderFactory.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    /// <summary>
+    /// Interface defining methods to build orchestration services and orchestration service clients.
```
cgillum: Should this comment reference the DurabilityProvider type instead of orchestration services and clients?

test/FunctionsV2/PlatformSpecificHelpers.FunctionsV2.cs
```diff
@@ -54,5 +55,37 @@ public static class PlatformSpecificHelpers
 
             return (JobHost)host.Services.GetService<IJobHost>();
         }
+
+        private static IWebJobsBuilder AddCorrectDurableTaskExtension(this IWebJobsBuilder builder, IOptions<DurableTaskOptions> options, string storageProvider)
```
cgillum: "Correct" feels a little odd. It implies that something might be incorrect. I would suggest keeping the previous name.

test/FunctionsV2/PlatformSpecificHelpers.FunctionsV2.cs
```diff
@@ -54,5 +55,37 @@ public static class PlatformSpecificHelpers
...
+            switch (storageProvider)
+            {
+                case TestHelpers.RedisProviderType:
+                    builder.AddRedisDurableTask();
+                    break;
+                case TestHelpers.EmulatorProviderType:
+                    builder.AddEmulatorDurableTask();
+                    break;
+                case TestHelpers.AzureStorageProviderType:
+                    break;
```
cgillum: Let's add a comment here explaining that the Azure Storage provider is built-in.

test/FunctionsV1/WebJobs.Extensions.DurableTask.Tests.V1.csproj
```diff
@@ -32,6 +32,14 @@
   <ItemGroup>
     <Compile Include="..\Common\*.cs" />
   </ItemGroup>
+
+  <PropertyGroup Condition="true">
```
cgillum: I don't believe the `Condition="true"` is needed.

test/FunctionsV2/WebJobs.Extensions.DurableTask.Tests.V2.csproj
```diff
@@ -27,11 +27,17 @@
...
+    <PackageReference Include="Microsoft.Azure.DurableTask.Emulator" Version="2.1.3" />
+    <PackageReference Include="Microsoft.Azure.DurableTask.Redis" Version="0.1.1-alpha" />
   </ItemGroup>
 
   <ItemGroup>
     <Compile Include="..\Common\*.cs" />
   </ItemGroup>
+
+  <ItemGroup>
+    <Compile Remove="..\Common\DurableOptionsConfigurationTests.cs" />
```
cgillum: What is this `Remove` for?

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
```diff
@@ -191,6 +191,15 @@ public interface IDurableOrchestrationClient
...
         Task RewindAsync(string instanceId, string reason);
 
+        /// <summary>
+        /// Gets the status of the specified orchestration instance.
+        /// </summary>
+        /// <param name="instanceId">The ID of the orchestration instance to query.</param>
+        /// <param name="showHistory">Boolean marker for including execution history in the response.</param>
+        /// <param name="showHistoryOutput">Boolean marker for including input and output in the execution history response.</param>
+        /// <returns>Returns a task which completes when the status has been fetched.</returns>
+        Task<DurableOrchestrationStatus> GetStatusAsync(string instanceId, bool showHistory, bool showHistoryOutput);
```
cgillum: I know I've been complaining about extension methods, but if we're using extension methods anyways I wonder if this is a good candidate. I would feel bad for anyone implementing an interface that has multiple overloads for the same method.

src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
```diff
@@ -254,10 +254,9 @@ public virtual Task<IList<OrchestrationState>> GetAllOrchestrationStatesWithFilt
         /// Gets the state of the specified orchestration instance.
         /// </summary>
         /// <param name="instanceId">The ID of the orchestration instance to query.</param>
-        /// <param name="showHistory">Boolean marker for including execution history in the response.</param>
         /// <param name="showInput">If set, fetch and return the input for the orchestration instance.</param>
         /// <returns>Returns a task which completes when the state has been fetched.</returns>
-        public virtual Task<IList<OrchestrationState>> GetOrchestrationStateAsync(string instanceId, bool showHistory, bool showInput = true)
+        public virtual Task<IList<OrchestrationState>> GetOrchestrationStateConfigureInputsAsync(string instanceId, bool showInput)
```
cgillum: Maybe `GetOrchestrationStateWithInputsAsync`?

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
ConnorMcMahon: Unfortunately that doesn't work since extension methods can only access the methods on `IDurableOrchestrationClient`, and this was meant to get around the fact that `Task<DurableOrchestrationStatus> IDurableOrchestrationClient.GetStatusAsync(string instanceId, bool showHistory, bool showHistoryOutput, bool showInput);`  could possibly through a `NotImplementedException`, so we can't use that method in our extension method.

I'm thinking that it would just be better to go back to just silently eat the `showInput` parameter for the small % of customers that use a custom `DurabilityProvider` implementation that does not support the `showInput` parameter.

Issue #944, opened 2019-09-24
Orchestrations with large retry intervals can get stuck


cgillum:
### Description
An orchestrator function that calls a continuously failing activity function with retry options that cause a retry to exceed 7 days can get permanently stuck in a way that is difficult to debug.

The problem is that DurableTask.AzureStorage attempts to enqueue a message with a visibility delay that is too large and the call gets rejected. The extension does not handle this gracefully and the orchestration gets stuck retrying continuously.

The extension should prevent this from happening by enforcing a maximum retry interval that is no greater than 6 days (or use a version of the Azure Storage SDK that allows arbitrary visibility delays).

### Known workarounds
Specify RetryOptions.MaxRetryInterval = TimeSpan.FromDays(6), or some other value that is less than 7 days.

ConnorMcMahon:
This is fixed for V1. Need to get in a PR for V2 before it goes GA.

cgillum:
Fixed with https://github.com/Azure/azure-functions-durable-extension/pull/1006.

PR #977, opened 2019-10-17
Enforce a maximum of 6 day intervals for RetryOptions

ConnorMcMahon:
Our current version of the storage SDK limits our timer functionality to a maximum of 7 days. We already had these checks for timers, but the same limitations apply to our retry intervals. Without enforcing these limits, customers could get permenantly stuck, and it was difficult to debug why. This silently sets the interval to our maximum when the customer attempts to use intervals larger than our maximum.

Addresses issue #944.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Azure.WebJobs
     public sealed class DurableOrchestrationContext : DurableOrchestrationContextBase
     {
         private const string DefaultVersion = "";
-        private const int MaxTimerDurationInDays = 6;
+        internal const int MaxTimerDurationInDays = 6;
 
         private readonly Dictionary<string, Stack> pendingExternalEvents =
             new Dictionary<string, Stack>(StringComparer.OrdinalIgnoreCase);
@@ -421,6 +421,18 @@ namespace Microsoft.Azure.WebJobs
             object input)
         {
             this.ThrowIfInvalidAccess();
+
+            // These check can be removed once the storage provider supports extended timers.
+            // https://github.com/Azure/azure-functions-durable-extension/issues/14
+            if (retryOptions.FirstRetryInterval > TimeSpan.FromDays(MaxTimerDurationInDays))
+            {
+                throw new ArgumentException($"Retry intervals must not exceed {MaxTimerDurationInDays} days.", nameof(retryOptions.FirstRetryInterval));
+            }
+
+            if (retryOptions.MaxRetryInterval > TimeSpan.FromDays(MaxTimerDurationInDays))
+            {
+                throw new ArgumentException($"Retry intervals must not exceed {MaxTimerDurationInDays} days.", nameof(retryOptions.MaxRetryInterval));
+            }
 
             // TODO: Support for versioning
             string version = DefaultVersion;
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
@@ -422,18 +422,6 @@ namespace Microsoft.Azure.WebJobs
         {
             this.ThrowIfInvalidAccess();
 
-            // These check can be removed once the storage provider supports extended timers.
-            // https://github.com/Azure/azure-functions-durable-extension/issues/14
-            if (retryOptions.FirstRetryInterval > TimeSpan.FromDays(MaxTimerDurationInDays))
-            {
-                throw new ArgumentException($"Retry intervals must not exceed {MaxTimerDurationInDays} days.", nameof(retryOptions.FirstRetryInterval));
-            }
-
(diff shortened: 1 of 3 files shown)


ConnorMcMahon:
This unfortunately could break in flight orchestrations that may or may not otherwise execute in a normal fashion. This is because we now throw an exception as soon as we encounter a RetryOptions that has a greater `MaxRetryInterval` than 6 days, where in all prior releases we default that value to `TimeSpan.MaxValue`.

Because of this, it may be better to just log a warning and just edit the value to 6 days instead of throwing an exception message. It's less clear, but it won't break anyone.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -422,6 +422,18 @@ public override void ContinueAsNew(object input, bool preserveUnprocessedEvents)
         {
             this.ThrowIfInvalidAccess();
 
+            // These check can be removed once the storage provider supports extended timers.
```
ConnorMcMahon: Debatably I can remove this comment both here and in the Timers section, as it is highly unlikely that we will ever update the Storage SDK in Durable 1.0.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
```diff
@@ -422,6 +422,18 @@ public override void ContinueAsNew(object input, bool preserveUnprocessedEvents)
...
+            // https://github.com/Azure/azure-functions-durable-extension/issues/14
+            if (retryOptions.FirstRetryInterval > TimeSpan.FromDays(MaxTimerDurationInDays))
+            {
+                throw new ArgumentException($"Retry intervals must not exceed {MaxTimerDurationInDays} days.", nameof(retryOptions.FirstRetryInterval));
+            }
+
+            if (retryOptions.MaxRetryInterval > TimeSpan.FromDays(MaxTimerDurationInDays))
+            {
+                throw new ArgumentException($"Retry intervals must not exceed {MaxTimerDurationInDays} days.", nameof(retryOptions.MaxRetryInterval));
```
cgillum: As you mentioned, this is a breaking change, which we can't allow for Durable 1.x. I think the best path forward is to  just set (overwrite) large retry intervals with our 6 day maximum. In the PR for Durable 2.x, we can throw here and/or throw in the property setters.

src/WebJobs.Extensions.DurableTask/RetryOptions.cs
```diff
@@ -12,6 +12,7 @@ namespace Microsoft.Azure.WebJobs
     public class RetryOptions
     {
         private readonly DurableTaskCore.RetryOptions retryOptions;
+        private readonly TimeSpan maxStorageTimeSpan = TimeSpan.FromDays(DurableOrchestrationContext.MaxTimerDurationInDays);
```
cgillum: Let's make this `static` since it's the same for all instances.

test/FunctionsV1/WebJobs.Extensions.DurableTask.Tests.V1.csproj
```diff
@@ -11,7 +11,7 @@
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="4.19.4" />
     <PackageReference Include="Microsoft.Azure.DurableTask.AzureStorage" Version="1.6.3" />
-    <PackageReference Include="Microsoft.Azure.WebJobs" Version="2.3.0" />
+    <PackageReference Include="Microsoft.Azure.WebJobs" Version="2.2.0" />
```
cgillum: Was this change intentional?

src/WebJobs.Extensions.DurableTask/RetryOptions.cs
```diff
@@ -24,6 +25,7 @@ public class RetryOptions
         public RetryOptions(TimeSpan firstRetryInterval, int maxNumberOfAttempts)
         {
             this.retryOptions = new DurableTaskCore.RetryOptions(firstRetryInterval, maxNumberOfAttempts);
+            this.MaxRetryInterval = TimeSpan.FromDays(DurableOrchestrationContext.MaxTimerDurationInDays);
```
cgillum: Looks like we can reuse your field declared on line 15 here.

test/FunctionsV1/WebJobs.Extensions.DurableTask.Tests.V1.csproj
ConnorMcMahon: Yes, there was a warning in the test project because the src code uses 2.2.0.

ConnorMcMahon: Could go in a separate PR if we want to keep this commit purely what is described in it.

cgillum: Oh - I didn't realize this was the test project. I have no problem with taking this change in the current PR.

ConnorMcMahon:
Need to address this for V2 before release, especially since we would ideally throw an exception if the customer does it (breaking change). Considering this is an Azure Storage specific limitation, was thinking of handling it in PR #974, or in a PR after that is merged.

cgillum:
Oh right - this is storage provider-specific. I guess it's not a simple port then.

PR #982, opened 2019-10-23
Limit the number of actions per orchestration

### LINKED ISSUE #522: Limit the number of actions per orchestration
Performance of orchestrations can drop significantly if an instance has too large of a history. For that reason, we should limit the maximum number of actions a single orchestration instance can make. If more actions are needed, users can break their large orchestrations into sub-orchestrations or use `ContinueAsNew` to limit the size of their history.

amdeel:
Limits the number of actions per orchestration in order to prevent performance from dropping with instances with a large history.

MaxOrchestrationActions is a configurable value that is defaulted to 100,000.

Fixes #522

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -40,7 +40,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private readonly int maxActionCount;
 
-        private static int actionCount;
+        private int actionCount;
 
         private string serializedOutput;
         private string serializedCustomStatus;
@@ -56,7 +56,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         internal DurableOrchestrationContext(DurableTaskExtension config, string functionName)
             : base(config, functionName)
         {
-            actionCount = 0;
+            this.actionCount = 0;
             this.maxActionCount = config.Options.MaxOrchestrationActions;
         }
 
@@ -821,7 +821,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     // Need to round-trip serialization since SendEvent always tries to serialize.
                     string rawInput = events.Dequeue();
                     JToken jsonData = JToken.Parse(rawInput);
-                    this.IncrementActionsOrThrowException();
                     this.InnerContext.SendEvent(instance, eventName, jsonData);
                 }
             }
@@ -1022,13 +1021,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private void IncrementActionsOrThrowException()
         {
-            if (actionCount >= this.maxActionCount)
+            if (this.actionCount >= this.maxActionCount)
             {
-                throw new InvalidOperationException("Maximum amount of orchestration actions (" + this.maxActionCount + ") has been reached. This value can be configured in local.settings.json file as MaxOrchestrationActions.");
+                throw new InvalidOperationException("Maximum amount of orchestration actions (" + this.maxActionCount + ") has been reached. This value can be configured in host.json file as MaxOrchestrationActions.");
             }
             else
             {
-                actionCount++;
+                this.actionCount++;
             }
         }
 
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -38,6 +38,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         private readonly Dictionary<string, Queue<string>> bufferedExternalEvents =
             new Dictionary<string, Queue<string>>(StringComparer.OrdinalIgnoreCase);
 
+        private readonly int maxActionCount;
+
+        private static int actionCount;
+
         private string serializedOutput;
         private string serializedCustomStatus;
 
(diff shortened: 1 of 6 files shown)


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -38,6 +38,10 @@ internal class DurableOrchestrationContext : DurableCommonContext, IDurableOrche
         private readonly Dictionary<string, Queue<string>> bufferedExternalEvents =
             new Dictionary<string, Queue<string>>(StringComparer.OrdinalIgnoreCase);
 
+        private readonly int maxActionCount;
+
+        private static int actionCount;
```
cgillum: Remove `static`. Any code that we add to the `DurableOrchestrationContext` class that can be invoked from an orchestrator function must follow all the determinism rules, and static variables violate those rules. I think the biggest problem in this case would be that multiple orchestrations running concurrently would be updating the same state and cause the exception to be thrown when it shouldn't be.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -1002,9 +1016,22 @@ internal void SendEntityMessage(OrchestrationInstance target, string eventName,
...
 
+            this.IncrementActionsOrThrowException();
             this.InnerContext.SendEvent(target, eventName, eventContent);
         }
 
+        private void IncrementActionsOrThrowException()
+        {
+            if (actionCount >= this.maxActionCount)
+            {
+                throw new InvalidOperationException("Maximum amount of orchestration actions (" + this.maxActionCount + ") has been reached. This value can be configured in local.settings.json file as MaxOrchestrationActions.");
```
cgillum: `local.settings.json` only applies to local development. It does not apply to azure deployments. But shouldn't this actually be configured in host.json?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -418,10 +426,12 @@ string IDurableOrchestrationContext.StartNewOrchestration(string functionName, o
                     System.Diagnostics.Debug.Assert(!oneWay, "The oneWay parameter should not be used for activity functions.");
                     if (retryOptions == null)
                     {
+                        this.IncrementActionsOrThrowException();
```
cgillum: It might be less maintenance to just call this once before the `switch` statement. Maybe after the call to `ThrowIfInvalidAccess()`?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -808,6 +821,7 @@ internal void RescheduleBufferedExternalEvents()
                     // Need to round-trip serialization since SendEvent always tries to serialize.
                     string rawInput = events.Dequeue();
                     JToken jsonData = JToken.Parse(rawInput);
+                    this.IncrementActionsOrThrowException();
```
cgillum: I probably wouldn't count this one since the user doesn't have direct control over this call. It's something that happens in the background without them really knowing about it. Basically, this code is for handling the case where an orchestration receives an external event but doesn't actually process it. This can happen in certain `ContinueAsNew` scenarios. We try to "reschedule" it for the next iteration of the orchestration.

test/Common/TestOrchestrations.cs
```diff
@@ -40,6 +40,14 @@ public static async Task<bool> SayHelloWithActivityWithDeterministicGuid([Orches
             return firstGuid != secondGuid && firstGuid != thirdGuid && secondGuid != thirdGuid;
         }
 
+        public static async Task<string> TwoOrchestratorActivityActions([OrchestrationTrigger] IDurableOrchestrationContext ctx)
+        {
+            string input = ctx.GetInput<string>();
+            string output = await ctx.CallActivityAsync<string>(nameof(TestActivities.Hello), input);
+            string outputTwo = await ctx.CallActivityAsync<string>(nameof(TestActivities.Hello), input);
```
cgillum: As we discussed in person, consider having an orchestration call each of the methods we count _once_, and then test that it fails if we do one more than this.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -3364,14 +3369,16 @@ public async Task MaxOrchestrationAction_MaxReached_OrchestrationFails()
...
 
-                var client = await host.StartOrchestratorAsync(nameof(TestOrchestrations.TwoOrchestratorActivityActions), "", this.output);
+                var counterEntityId = new EntityId("Counter", Guid.NewGuid().ToString());
+
+                var client = await host.StartOrchestratorAsync(orchestratorFunctionNames[0], counterEntityId, this.output);
                 DurableOrchestrationStatus status = await client.WaitForCompletionAsync(this.output);
 
                 Assert.NotNull(status);
                 Assert.Equal(OrchestrationRuntimeStatus.Failed, status.RuntimeStatus);
                 Assert.Equal(
```
cgillum: It seems like this test would pass even if we accidentally threw the exception sometime before the 7th action. To make the test more precise, I suggest you update the orchestration with a `ctx.SetCustomStatus(...)` call right before the expected failure. You can then add an `Assert` here which checks for that custom status in the `status.CustomStatus` field.

amdeel: That assert statement would pass, but the test would only pass if the next assert statement, that checks the error message, passes as well. But I've never used SetCustomStatus before and didn't think about that when I wrote this test. I will add that.

amdeel: Actually, is this necessary? It doesn't change the test from relying on the last assert to compare the output to the error message.

cgillum: You're right. I think the actual (potential) problem I had in mind was slightly different. My worry was actually more that you won't know if the exception is thrown too early. For example, let's say you added one more API call to your orchestration but forgot to update `maxActions` from 7 to 8. The test will still pass, but your last API call will not have been invoked. You could make that kind of problem less likely by checking a custom status which is set only after all the API calls you expected to run have actually run. It's a minor thing, but I think it will make the test a bit more robust.

PR #986, opened 2019-10-24
[Refactor] Generically-typed StartNewAsync() overloads

brandonh-msft:
@cgillum Here's my attempt at a PR to alleviate the issue that others and I have ran in to while using the `StartNewAsync` call in Durable Functions.

Problem:
The following code:
```csharp
var instanceId = await client.StartNewAsync(@"MyOrchestrator", "thisInstanceId");
Debug.Assert(instanceId == @"thisInstanceId");
```

Will throw the assertion. This is because if a user wishes to specify the instanceId for a new orchestration they _must_ give some form of input - `null` or otherwise. While intellisense is the way one would realize what they're doing wrong, it's only after they attempt this and end up getting back an instanceId value other than the one they specified that they realize something's gone awry due to the fact that `string : object` so the `StartNewAsync(string, object)` overload gets hit and executed.

By using generics we can not only prevent this runtime confusion, but also enable an even simpler scenario: `StartNewAsync(@"MyOrchestrator")` which is an added benefit.

Hopefully you find this idea/PR valuable. Thanks! You'll notice I have the target for it as the `v2` branch due to its breaking nature.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextInterfaces/DurableContextExtensions.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextInterfaces/DurableContextExtensions.cs
@@ -397,7 +397,6 @@ namespace Microsoft.Azure.WebJobs
         /// <param name="client">The client object.</param>
         /// <param name="orchestratorFunctionName">The name of the orchestrator function to start.</param>
         /// <param name="instanceId">The ID to use for the new orchestration instance.</param>
-        /// <typeparam name="T">The type of the input value for the orchestrator function.</typeparam>
         /// <returns>A task that completes when the orchestration is started. The task contains the instance id of the started
         /// orchestratation instance.</returns>
         /// <exception cref="ArgumentException">
--- a/src/WebJobs.Extensions.DurableTask/ContextInterfaces/DurableContextExtensions.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextInterfaces/DurableContextExtensions.cs
@@ -396,6 +396,26 @@ namespace Microsoft.Azure.WebJobs
         /// </summary>
         /// <param name="client">The client object.</param>
         /// <param name="orchestratorFunctionName">The name of the orchestrator function to start.</param>
+        /// <param name="instanceId">The ID to use for the new orchestration instance.</param>
+        /// <typeparam name="T">The type of the input value for the orchestrator function.</typeparam>
+        /// <returns>A task that completes when the orchestration is started. The task contains the instance id of the started
+        /// orchestratation instance.</returns>
+        /// <exception cref="ArgumentException">
+        /// The specified function does not exist, is disabled, or is not an orchestrator function.
+        /// </exception>
+        public static Task<string> StartNewAsync(
+            this IDurableOrchestrationClient client,
+            string orchestratorFunctionName,
+            string instanceId)
+        {
+            return client.StartNewAsync<object>(orchestratorFunctionName, instanceId, null);
+        }
+
+        /// <summary>
+        /// Starts a new execution of the specified orchestrator function.
+        /// </summary>
+        /// <param name="client">The client object.</param>
+        /// <param name="orchestratorFunctionName">The name of the orchestrator function to start.</param>
         /// <param name="input">JSON-serializeable input value for the orchestrator function.</param>
         /// <typeparam name="T">The type of the input value for the orchestrator function.</typeparam>
         /// <returns>A task that completes when the orchestration is started. The task contains the instance id of the started
--- a/src/WebJobs.Extensions.DurableTask/ContextInterfaces/DurableContextExtensions.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextInterfaces/DurableContextExtensions.cs
@@ -397,6 +397,26 @@ namespace Microsoft.Azure.WebJobs
         /// <param name="client">The client object.</param>
         /// <param name="orchestratorFunctionName">The name of the orchestrator function to start.</param>
         /// <param name="input">JSON-serializeable input value for the orchestrator function.</param>
+        /// <typeparam name="T">The type of the input value for the orchestrator function.</typeparam>
+        /// <returns>A task that completes when the orchestration is started. The task contains the instance id of the started
+        /// orchestratation instance.</returns>
+        /// <exception cref="ArgumentException">
+        /// The specified function does not exist, is disabled, or is not an orchestrator function.
+        /// </exception>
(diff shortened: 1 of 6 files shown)


cgillum:
Cool! One thing - can you clarify for me how the generic parameter helps? I'm not sure I understand what benefit it provides. I would assume that the example code you provided in your description would still have the same problem - i.e. the C# compiler infers `T` to be `string` in that case?

brandonh-msft:
no because if you tried to pass a `string` for one of the `input` values, you'd _have_ to do `<string>` on the method call or you'd get the "ambiguous reference" error

brandonh-msft:
d'oh you're right in its current state; I forgot an overload. Working on this now! 😊

brandonh-msft:
So to circle back. by not specifying the type parameter, and using the (string, string) overload (function name, instance id) you automatically go to a `null` type param and the instance id specified
if you wanted that to be (function, input) you'd now _have_ to specify `<string>` on the `StartNewAsync` call

src/WebJobs.Extensions.DurableTask/ContextInterfaces/DurableContextExtensions.cs
```diff
@@ -405,11 +406,49 @@ public static Task CallEntityAsync(this IDurableOrchestrationContext context, En
...
+        /// Starts a new execution of the specified orchestrator function.
+        /// </summary>
+        /// <param name="client">The client object.</param>
+        /// <param name="orchestratorFunctionName">The name of the orchestrator function to start.</param>
+        /// <returns>A task that completes when the orchestration is started. The task contains the instance id of the started
+        /// orchestratation instance.</returns>
+        /// <exception cref="ArgumentException">
+        /// The specified function does not exist, is disabled, or is not an orchestrator function.
+        /// </exception>
+        public static Task<string> StartNewAsync(
```
cgillum: nit: I would put this overload at the top of the list since it's the least specific.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
```diff
@@ -117,12 +117,13 @@ public interface IDurableOrchestrationClient
...
         /// <param name="instanceId">The ID to use for the new orchestration instance.</param>
         /// <param name="input">JSON-serializeable input value for the orchestrator function.</param>
+        /// <typeparam name="T">The type of the input value for the orchestrator function.</typeparam>
         /// <returns>A task that completes when the orchestration is started. The task contains the instance id of the started
         /// orchestratation instance.</returns>
         /// <exception cref="ArgumentException">
         /// The specified function does not exist, is disabled, or is not an orchestrator function.
         /// </exception>
-        Task<string> StartNewAsync(string orchestratorFunctionName, string instanceId, object input);
+        Task<string> StartNewAsync<T>(string orchestratorFunctionName, string instanceId, T input);
```
cgillum: Let's consider naming this `TInput` so that it's not confused to be used for the output (CallActivityAsync<T>'s T is for the output whereas this one is for the input).

src/WebJobs.Extensions.DurableTask/ContextInterfaces/DurableContextExtensions.cs
```diff
@@ -396,7 +396,8 @@ public static Task CallEntityAsync(this IDurableOrchestrationContext context, En
         /// </summary>
         /// <param name="client">The client object.</param>
         /// <param name="orchestratorFunctionName">The name of the orchestrator function to start.</param>
-        /// <param name="input">JSON-serializeable input value for the orchestrator function.</param>
+        /// <param name="instanceId">The ID to use for the new orchestration instance.</param>
+        /// <typeparam name="T">The type of the input value for the orchestrator function.</typeparam>
```
ConnorMcMahon: NIT: no type parameter in this method.

test/Common/DurableClientBaseTests.cs
```diff
@@ -31,9 +28,28 @@ public async Task StartNewAsync_is_calling_overload_method()
...
+            var result = await durableOrchestrationClientBaseMock.Object.StartNewAsync(functionName);
+            result.Should().Be(instanceId);
+
+            result = await durableOrchestrationClientBaseMock.Object.StartNewAsync<object>(functionName, null);
+            result.Should().Be(instanceId);
+        }
+
+        [Fact]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        public async Task StartNewAsync_is_calling_overload_method_with_specific_instance_id()
```
ConnorMcMahon: NIT: if this test, and the other test in this file, could be renamed to match the naming scheme of our other tests, that would be great.

cgillum:
Connor's first "nit" isn't really a "nit" - it's causing the build to fail:

>  ContextInterfaces\DurableContextExtensions.cs(400,30): error CS1711: XML comment has a typeparam tag for 'T', but there is no type parameter by that name [C:\projects\azure-functions-durable-extension\src\WebJobs.Extensions.DurableTask\WebJobs.Extensions.DurableTask.csproj]

brandonh-msft:
Weird. Wonder why it was building on my box and in the previous CI? I'll have a look today and push the fix

ConnorMcMahon:
@brandonh-msft, our build is configured to treat warnings as errors when in Release build, which is what our CI uses. It's specifically designed to prevent people from checking in warnings into our main branches. You probably built as debug locally, which is why it built fine.

test/Common/DurableClientBaseTests.cs
brandonh-msft: can you elaborate? I kept the name for one, and added a new one that _seemed_ to follow the same convention....

image

happy to make changes, though!

ConnorMcMahon: I now realize we don't have a firm convention. I have a preference towards Capitalizing the first letter after each `_`, but that is not consistent across our tests, so I won't make you address that in this PR. 

Making it more consistent is for another PR :).

brandonh-msft:
@cgillum point of order on this. When I made these changes and did a build, a couple of XML files were also updated. Do you want those committed? Specifically:
```
modified:   src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask-net471.xml
modified:   src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask-netstandard2.0.xml
```

cgillum:
@brandonh-msft yes, those should also be committed.

brandonh-msft:
closed in favor of PR to `dev`

PR #987, opened 2019-10-24
Added Option to Configure Behavior When Starting a New Orchestrator With an Existing Instance Id

### LINKED ISSUE #367: Add support for atomic StartNewAsync that would execute only if instance is not already running (for singleton orchestrator support)
Could we add an atomic "StartNewAsyncIfNotRunningOrCompletedOrInError" ( name to rework ;-) )? If the instance already exists, keep it running, if not ignore the start request and notify that it was already running.

Currently, we have to check the status, and if not running than start it, which is not concurrency friendly at all (could easily have 2 requests starting the workflow).

I really want something like that for singleton orchestrator scenarios (example: a workflow singleton is started for each user identifier).

Other way would be to have support for Singleton (https://github.com/Azure/azure-functions-host/issues/912) with the consumption plan so that I can safety check the status and call StartNewAsync om the same concurrency safe scope.

amdeel:
Added the ability to configure which states you would like to override an existing orchestrator with the same instanceId.

fixes #367

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -121,8 +121,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 throw new ArgumentException($"Instance ID lengths must not exceed {MaxInstanceIdLength} characters.");
             }
 
+            this.TryGetStatusesNotToOverride(out OrchestrationStatus[] dedupeStatuses);
             Task<OrchestrationInstance> createTask = this.client.CreateOrchestrationInstanceAsync(
-                orchestratorFunctionName, DefaultVersion, instanceId, input);
+                orchestratorFunctionName, DefaultVersion, instanceId, input, null, dedupeStatuses);
 
             this.traceHelper.FunctionScheduled(
                 this.hubName,
@@ -134,6 +135,19 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
             OrchestrationInstance instance = await createTask;
             return instance.InstanceId;
+        }
+
+        private bool TryGetStatusesNotToOverride(out OrchestrationStatus[] dedupeStatuses)
+        {
+            var overridableStates = this.config.Options.OverridableExistingInstanceStates;
+            if (overridableStates == OverridableStates.IfTerminatedFailedOrCompleted)
+            {
+                dedupeStatuses = new OrchestrationStatus[] { OrchestrationStatus.Running, OrchestrationStatus.ContinuedAsNew, OrchestrationStatus.Canceled, OrchestrationStatus.Pending };
+                return true;
+            }
+
+            dedupeStatuses = null;
+            return false;
         }
 
         private static bool IsInvalidCharacter(char c)
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -121,7 +121,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 throw new ArgumentException($"Instance ID lengths must not exceed {MaxInstanceIdLength} characters.");
             }
 
-            this.GetStatusesNotToOverride(out OrchestrationStatus[] dedupeStatuses);
+            var dedupeStatuses = this.GetStatusesNotToOverride();
             Task<OrchestrationInstance> createTask = this.client.CreateOrchestrationInstanceAsync(
                 orchestratorFunctionName, DefaultVersion, instanceId, input, null, dedupeStatuses);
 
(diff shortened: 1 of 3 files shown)


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -136,6 +137,19 @@ async Task<string> IDurableOrchestrationClient.StartNewAsync(string orchestrator
             return instance.InstanceId;
         }
 
+        private bool TryGetStatusesNotToOverride(out OrchestrationStatus[] dedupeStatuses)
+        {
+            var overridableStates = this.config.Options.OverridableExistingInstanceStates;
+            if (overridableStates == OverridableStates.IfTerminatedFailedOrCompleted)
+            {
+                dedupeStatuses = new OrchestrationStatus[] { OrchestrationStatus.Running, OrchestrationStatus.ContinuedAsNew, OrchestrationStatus.Canceled, OrchestrationStatus.Pending };
```
cgillum: I suggest formatting this so that each entry is on its own line. It makes reading the code a bit easier (and comparing diffs). Like this:

```csharp
dedupeStatuses = new OrchestrationStatus[]
{
    OrchestrationStatus.Running,
    OrchestrationStatus.ContinuedAsNew,
    OrchestrationStatus.Pending,
};
```

Also, I don't think we need the `Canceled` status check here (DTFx doesn't actually use this status for anything).

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -136,6 +137,19 @@ async Task<string> IDurableOrchestrationClient.StartNewAsync(string orchestrator
             return instance.InstanceId;
         }
 
+        private bool TryGetStatusesNotToOverride(out OrchestrationStatus[] dedupeStatuses)
```
ConnorMcMahon: I wouldn't follow the `Try*` pattern unless you are actually going to check if the request succeeded or not.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -128,6 +128,11 @@ public string HubName
         /// </value>
         public int ExtendedSessionIdleTimeoutInSeconds { get; set; } = 30;
 
+        /// <summary>
+        ///  States that will override an existing orchestrator when attempting to start a new orchestrator with the same instance Id.
+        /// </summary>
+        public OverridableStates OverridableExistingInstanceStates { get; set; } = OverridableStates.AnyState;
```
ConnorMcMahon: For V2, I think we want to default to `IfTerminatedFailedOrCompleted`.

cgillum: The default should actually be `OverridableStates.IfTerminatedFailedOrCompleted`.

src/WebJobs.Extensions.DurableTask/OverridableStates.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        /// <summary>
+        /// Option to start a new orchestrator instance with an existing instnace Id when the existing
+        /// instance is in any state.
+        /// </summary>
+        AnyState,
+
+        /// <summary>
+        /// Option to only start a new orchestrator instance with an existing instance Id when the existing
+        /// instance is in a termincated, failed, or completed state.
```
cgillum: termincated --> terminated

src/WebJobs.Extensions.DurableTask/OverridableStates.cs
```diff
@@ -0,0 +1,24 @@
...
+        /// </summary>
+        IfTerminatedFailedOrCompleted,
```
cgillum: An alternative name to consider: `NonRunningStates`.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -136,6 +137,22 @@ async Task<string> IDurableOrchestrationClient.StartNewAsync(string orchestrator
             return instance.InstanceId;
         }
 
+        private void GetStatusesNotToOverride(out OrchestrationStatus[] dedupeStatuses)
```
ConnorMcMahon: Can we make this return `dedupeStatuses` instead of using an `out` parameter?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -136,6 +137,22 @@ async Task<string> IDurableOrchestrationClient.StartNewAsync(string orchestrator
...
+            {
+                dedupeStatuses = new OrchestrationStatus[]
+                {
+                    OrchestrationStatus.Running,
+                    OrchestrationStatus.ContinuedAsNew,
+                    OrchestrationStatus.Pending,
+                };
+            }
+
+            dedupeStatuses = null;
```
ConnorMcMahon: While the Azure Storage implementation properly handles the null array case, it is possible that others will not. It might be a bit safer to use an empty array instead of null.

amdeel: Good point!

cgillum:
@amdeel Do we also need a PR to set `ThrowExceptionOnInvalidDedupeStatus` to `true` by default for Durable Functions?

PR #990, opened 2019-10-25
Support for scheduled signals

sebastianburckhardt:
Adding support for specifying a delivery time for scheduled signals.

The idea is that anyone who can signal an operation (whether it is a client, an orchestration, or an entity), can specify an intended delivery time. When the signal reaches the entity, and the time has not yet arrived, the operation is durably buffered. 

This feature can also be used for timer/reminder type functionality, as entities can signal an operation to themselves.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -206,7 +206,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// <inheritdoc />
         Task<TResult> IDurableOrchestrationContext.CallSubOrchestratorAsync<TResult>(string functionName, string instanceId, object input)
         {
-            return this.CallDurableTaskFunctionAsync<TResult>(functionName, FunctionType.Orchestrator, false, instanceId, null, null, input);
+            return this.CallDurableTaskFunctionAsync<TResult>(functionName, FunctionType.Orchestrator, false, instanceId, null, null, input, null);
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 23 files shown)


sebastianburckhardt:
With the latest fixes to DTFx (see azure/durabletask#339) all the tests are passing now.

sebastianburckhardt:
I have second thoughts on this implementation. It is still failing some tests, and it will perform inefficiently in some situations. I now believe we should go for a much simpler implementation that does not use DTFx timers but somehow tells the storage backend directly that a message has a delay. The current backend can do a much better job at managing such messages than DTFx, very simply using visibility delay.

sebastianburckhardt:
Good to go! As discussed, the whole mechanism is simpler now.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -1009,25 +1026,40 @@ private void ThrowIfInvalidAccess()
...
-                    this.InstanceId,
-                    this.ExecutionId,
-                    target.InstanceId,
-                    eventName,
-                    eventContent);
+                this.InstanceId,
+                this.ExecutionId,
+                target.InstanceId,
+                eventName,
+                eventContent);
```
cgillum: Looks like the indentation got reverted here.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -1009,25 +1026,40 @@ private void ThrowIfInvalidAccess()
...
             if (eventContent is RequestMessage requestMessage)
             {
-                this.MessageSorter.LabelOutgoingMessage(
-                    requestMessage,
-                    target.InstanceId,
-                    this.InnerContext.CurrentUtcDateTime,
-                    TimeSpan.FromMinutes(this.Config.Options.EntityMessageReorderWindowInMinutes));
+                if (requestMessage.ScheduledTime.HasValue)
+                {
+                    eventName = $"op@{requestMessage.ScheduledTime.Value:o}";
```
cgillum: It looks like this event name generation logic has been duplicated into at least three different files (DurableClient.cs, DurableEntityContext.cs, and DurableOrchestrationContext.cs). Is there any way we can consolidate it into some common method?

sebastianburckhardt: Yes. From a code maintenance perspective it probably makes sense to create a static class that consolidates all the event naming logic in  a single place.

PR #991, opened 2019-10-25
[Refactor] Generically-typed StartNewAsync() overloads

brandonh-msft:
Duplicate of #986 but rebased on to `dev` since v2 got merged in

---

Problem:
The following code:
```csharp
var instanceId = await client.StartNewAsync(@"MyOrchestrator", "thisInstanceId");
Debug.Assert(instanceId == @"thisInstanceId");
```

Will throw the assertion. This is because if a user wishes to specify the instanceId for a new orchestration they _must_ give some form of input - `null` or otherwise. While intellisense is the way one would realize what they're doing wrong, it's only after they attempt this and end up getting back an instanceId value other than the one they specified that they realize something's gone awry due to the fact that `string : object` so the `StartNewAsync(string, object)` overload gets hit and executed.

By using generics we can not only prevent this runtime confusion, but also enable an even simpler scenario: `StartNewAsync(@"MyOrchestrator")` which is an added benefit.

Hopefully you find this idea/PR valuable. Thanks! You'll notice I have the target for it as the `v2` branch due to its breaking nature.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -101,7 +101,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         }
 
         /// <inheritdoc />
-        async Task<string> IDurableOrchestrationClient.StartNewAsync(string orchestratorFunctionName, string instanceId, object input)
+        async Task<string> IDurableOrchestrationClient.StartNewAsync<T>(string orchestratorFunctionName, string instanceId, T input)
         {
             this.config.ThrowIfFunctionDoesNotExist(orchestratorFunctionName, FunctionType.Orchestrator);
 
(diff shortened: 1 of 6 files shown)


brandonh-msft:
is something wrong with my branch name? perhaps the inclusion of `off` or `dev` is masked by the Azure pipelines bot?

cgillum:
@amdeel any ideas on why this pipeline won't trigger for the `dev` branch? I updated the azure-pipelines.yaml file as shown [here](https://github.com/Azure/azure-functions-durable-extension/blob/9393d46d3a6356e896e6eda97b9ead9eacb4e363/azure-pipelines.yml#L1-L3) to work with both `v2` and `dev`, but either that isn't enough or I did it wrong. :(

amdeel:
@cgillum Did you add that yaml file to the dev branch?

amdeel:
@cgillum Figured it out. Our PR override settings needed updated to include dev.

brandonh-msft:
looks like there's something wrong w/ the pipeline configs? 
image

brandonh-msft:
🎉 

thanks!

PR #992, opened 2019-10-25
Added ReplaySafeLogger that will only log when not replaying

### LINKED ISSUE #834: Default to log only if not replaying
It's become very common to see people only log if the current evaluation is not during a replay.  Today most people decorate each log statement with a single-line if statement.  As this becomes the norm, in v2 we should consider making this the default behavior.  It would need to work across all languages, and would need to still have some way for users to opt *in* to the behavior of logging on replays as well.

amdeel:
Added CreateReplaySafeLogger to IDurableOrchestrationContext, returning a replay-safe logger that will only log when the orchestrator is not replaying.

fixes #834

Previously, when using ILogger in an orchestrator function, log calls would execute every time the orchestrator replayed. To avoid this, the log call could be wrapped in a call to the context to make sure that line had not executed before. This would have to be done around every log call inside the orchestrator function.

This PR adds the ability to add one line of code that returns an ILogger that is replay safe and can be used throughout the orchestrator function. This means it will automatically check if the line of code is replaying and eliminates the need to wrap log calls.
```csharp
            [OrchestrationTrigger] IDurableOrchestrationContext context, ILogger log)
        {
		log = context.CreateReplaySafeLogger(log);
	...
```
Here's an example outlining the difference between the old and new way to make replay safe logger calls in an orchestrator function.

Old:
```csharp
	[OrchestrationTrigger] IDurableOrchestrationContext context, ILogger log)
{

	if (!context.IsReplaying)
	{
		log.logInformation("About to call first activity");
	}
	
	if (!context.IsReplaying)
	{
		log.logInformation("About to call second activity");
	}
	
	if (!context.IsReplaying)
	{
		log.logInformation("About to call third activity");
	}
	
	outputs.Add(await context.CallActivityAsync<string>("E1_SayHello_DirectInput", "London"));

	return outputs;
}

```
		
Old:
```csharp
	[OrchestrationTrigger] IDurableOrchestrationContext context, ILogger log)
{

	if (!context.IsReplaying) log.LogInformation("About to call first activity.");
	if (!context.IsReplaying) log.LogInformation("About to call second activity.");
	if (!context.IsReplaying) log.LogInformation("About to call third activity.");
	outputs.Add(await context.CallActivityAsync<string>("E1_SayHello_DirectInput", "London"));

	return outputs;
}
```

New:
```csharp
	[OrchestrationTrigger] IDurableOrchestrationContext context, ILogger log)
{
	log = context.CreateReplaySafeLogger(log);
	
	log.logInformation("About to call first activity");
	log.logInformation("About to call second activity");
	log.logInformation("About to call third activity");
	outputs.Add(await context.CallActivityAsync<string>("E1_SayHello_DirectInput", "London"));

	return outputs;
}
```


src/WebJobs.Extensions.DurableTask/ReplaySafeLogger.cs
```diff
@@ -0,0 +1,42 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    internal class ReplaySafeLogger : ILogger
+    {
+        private IDurableOrchestrationContext context;
+        private ILogger logger;
+
+        private ReplaySafeLogger()
```
ConnorMcMahon: We shouldn't need this constructor.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
```diff
@@ -339,5 +340,13 @@ public interface IDurableOrchestrationContext
...
         /// <returns>The instance id of the new orchestration.</returns>
         string StartNewOrchestration(string functionName, object input, string instanceId = null);
+
+        /// <summary>
+        /// Returns an instance of ILogger that is replay safe, ensuring the logger logs only when the orchestrator
+        /// is not replaying that line of code.
+        /// </summary>
+        /// <param name="logger">An instance of ILogger.</param>
+        /// <returns>An instance of a replay safe ILogger.</returns>
+        ILogger CreateReplaySafeLogger(ILogger logger);
```
ConnorMcMahon: Instead of adding this to the interface and implementing it directly on `DurableOrchestrationContext`, let's make it an extension method. You can put it in `DurableContextExtensions`. This way, the customer gets access to our implementation of `CreateReplaySafeLogger()` even if they aren't using the `DurableOrchestrationContext` implementation. Otherwise they will need to mock that behavior themselves.

ConnorMcMahon:
We could also use a test for this. You could have an orchestration that uses our safe replay logger, and pass in an ILogger implementation that counts how often it actually get's written to. If that count exceeds how many times we log on the last replay, that means that it isn't replay safe.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -14,6 +14,7 @@
 using DurableTask.Core;
 using DurableTask.Core.Exceptions;
 using Microsoft.Azure.WebJobs.Extensions.DurableTask.Options;
+using Microsoft.Extensions.Logging;
```
ConnorMcMahon: NIT: can likely get rid of this using statement now.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
```diff
@@ -7,6 +7,7 @@
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+using Microsoft.Extensions.Logging;
```
ConnorMcMahon: Same nit here.

src/WebJobs.Extensions.DurableTask/ReplaySafeLogger.cs
```diff
@@ -0,0 +1,38 @@
...
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs
+{
+    internal class ReplaySafeLogger : ILogger
+    {
+        private IDurableOrchestrationContext context;
+        private ILogger logger;
```
cgillum: Let's make these two fields `readonly`.

cgillum:
@ConnorMcMahon do you think this PR is good to be merged?

PR #1006, opened 2019-10-31
Add Azure Storage time delay exceptions

ConnorMcMahon:
Ensure we never schedule a queue message longer than 6 days when using the Azure Storage durability provider.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -30,13 +30,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
     {
         public const string DefaultVersion = "";
 
-        private const int MaxTimerDurationInDays = 6;
-
         private readonly Dictionary<string, Stack> pendingExternalEvents =
             new Dictionary<string, Stack>(StringComparer.OrdinalIgnoreCase);
 
         private readonly Dictionary<string, Queue<string>> bufferedExternalEvents =
             new Dictionary<string, Queue<string>>(StringComparer.OrdinalIgnoreCase);
+
+        private readonly DurabilityProvider durabilityProvider;
 
         private string serializedOutput;
         private string serializedCustomStatus;
@@ -49,9 +49,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private MessageSorter messageSorter;
 
-        internal DurableOrchestrationContext(DurableTaskExtension config, string functionName)
+        internal DurableOrchestrationContext(DurableTaskExtension config, DurabilityProvider durabilityProvider, string functionName)
             : base(config, functionName)
         {
+            this.durabilityProvider = durabilityProvider;
         }
 
         internal string ParentInstanceId { get; set; }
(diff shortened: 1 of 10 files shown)


src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
```diff
@@ -374,5 +374,17 @@ public Task PurgeOrchestrationHistoryAsync(DateTime thresholdDateTimeUtc, Orches
...
             return this.GetOrchestrationServiceClient().PurgeOrchestrationHistoryAsync(thresholdDateTimeUtc, timeRangeFilterType);
         }
+
+        /// <summary>
+        /// Uses durability provider specific logic to verify whether a timespan for a timer
+        /// or a retry interval is allowed by the provider.
+        /// </summary>
+        /// <param name="timespan">The timespan for a timer or retry interval.</param>
+        /// <returns>A boolean indicating whether the time interval is valid and an error message if it is invalid.</returns>
+        public virtual (bool, string) CheckTimeInterval(TimeSpan timespan)
```
cgillum: I would prefer we not use Tuples in our public APIs. If we need to return two values, let's either create a new type or use a return value and an `out` parameter. There is some interesting discussion of this topic [here](https://stackoverflow.com/questions/3017352/is-using-net-4-0-tuples-in-my-c-sharp-code-a-poor-design-decision).

cgillum: I also suggest we consider a name like `ValidateTimerExpiration` to make it more clear what scenarios the timespan applies to.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -401,6 +402,19 @@ string IDurableOrchestrationContext.StartNewOrchestration(string functionName, o
         {
             this.ThrowIfInvalidAccess();
 
+            (bool isValidTimespan, string errorMessage) = this.durabilityProvider.CheckTimeInterval(retryOptions.MaxRetryInterval);
+
+            if (!isValidTimespan)
+            {
+                throw new ArgumentException($"Invalid max retry interval duration: {errorMessage}");
```
cgillum: Consider using a simpler pattern for your exception messages, like this:

```csharp
throw new ArgumentException(errorMessage, nameof(retryOptions.MaxRetryInterval);
```

The resulting exception will be:

```
System.ArgumentException: The Azure Storage provider supports a maximum of 6 days for time-based delays
Parameter name: MaxRetryInterval
```

This would be easier to maintain while still providing the right level of detail.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -2923,6 +2923,110 @@ public async Task DurableEntity_EntityNameCaseInsensitivity(bool extendedSession
...
+
+                var invalidFireAtTime = DateTime.UtcNow.AddDays(7);
+
+                var client = await host.StartOrchestratorAsync(orchestrationFunctionName, invalidFireAtTime, this.output);
+
+                var status = await client.WaitForCompletionAsync(this.output);
+
+                Assert.Equal(OrchestrationRuntimeStatus.Completed, status.RuntimeStatus);
+
+                Assert.False((bool)status.Output);
```
cgillum: Consider having the orchestration fail with an exception, and then checking the output for some expected part of the exception message, like invalid parameter name. That would ensure we don't have false positives caused by unrelated `ArgumentException`s.

src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
ConnorMcMahon: I didn't want to necessarily use TImer in the name, as it also applies to timeouts and retry intervals. Maybe `ValidateDelayTime`, but if you don't like that, I think that `ValidateTimerExpiration` still works.

cgillum: I mentioned Timers since our time constraints are only needed for scenarios where timers are used. But I guess one could argue that this may be too specific to the Azure storage provider. In any case, I like the new name you chose. :)

src/WebJobs.Extensions.DurableTask/RetryOptions.cs
```diff
@@ -13,6 +13,10 @@ public class RetryOptions
     {
         private readonly DurableTaskCore.RetryOptions retryOptions;
 
+        // Would like to make this durability provider specific, but since this is a customer
+        // facing type, that is difficult.
+        private static readonly TimeSpan defaultMaxRetryinterval = TimeSpan.FromDays(6);
```
cgillum: nit: this should be a capital "D".

PR #1007, opened 2019-10-31
Updated xmldoc generation to work with docs.microsoft.com

cgillum:
In order for the docs to show up correctly in the [Azure API browser](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.webjobs.extensions.durabletask?view=azure-dotnet-preview), the generated XML name file must match the assembly name exactly.


cgillum:
@ConnorMcMahon thanks for the review. What I'm being told is that the API browser support for Azure SDKs is pretty limited right now. We're able to have APIs for a "legacy" version, a "current" version, and a "preview" version. I'm not sure what their plans are for supporting more than this.

PR #1009, opened 2019-10-31
Runtime scale implementation for Durable extension

gzuber:
Resolves https://github.com/Azure/azure-functions-host/issues/4957

This is a mostly-done PR. I'm in the process of adding unit tests. Creating this PR early to speed up the feedback process.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Listener/DurableTaskListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/DurableTaskListener.cs
@@ -2,25 +2,42 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using DurableTask.AzureStorage.Monitoring;
 using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Azure.WebJobs.Host.Listeners;
+#if !FUNCTIONS_V1
+using Microsoft.Azure.WebJobs.Host.Scale;
+#endif
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 {
+#if !FUNCTIONS_V1
+    internal sealed class DurableTaskListener : IListener, IScaleMonitorProvider
+#else
     internal sealed class DurableTaskListener : IListener
+#endif
     {
         private readonly DurableTaskExtension config;
+        private readonly string functionId;
         private readonly FunctionName functionName;
         private readonly ITriggeredFunctionExecutor executor;
         private readonly FunctionType functionType;
+        private readonly string storageConnectionString;
+#if !FUNCTIONS_V1
+        private readonly Lazy<DurableTaskScaleMonitor> scaleMonitor;
+#endif
 
         public DurableTaskListener(
             DurableTaskExtension config,
+            string functionId,
             FunctionName functionName,
             ITriggeredFunctionExecutor executor,
-            FunctionType functionType)
+            FunctionType functionType,
+            string storageConnectionString)
         {
             this.config = config ?? throw new ArgumentNullException(nameof(config));
             this.executor = executor ?? throw new ArgumentNullException(nameof(executor));
@@ -30,8 +47,18 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 throw new ArgumentNullException(nameof(functionName));
             }
 
+            this.functionId = functionId;
             this.functionName = functionName;
             this.functionType = functionType;
+            this.storageConnectionString = storageConnectionString;
+#if !FUNCTIONS_V1
+            this.scaleMonitor = new Lazy<DurableTaskScaleMonitor>(() => new DurableTaskScaleMonitor(
+                                                                                        this.functionId,
+                                                                                        this.functionName,
+                                                                                        this.config.Options.HubName,
+                                                                                        this.storageConnectionString,
+                                                                                        this.config.TraceHelper));
+#endif
         }
 
         public Task StartAsync(CancellationToken cancellationToken)
@@ -68,5 +95,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         public void Dispose()
         {
         }
+
+#if !FUNCTIONS_V1
+        public IScaleMonitor GetMonitor()
+        {
+            return this.scaleMonitor.Value;
+        }
+#endif
     }
 }
--- a/src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
@@ -1,0 +1,187 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#if !FUNCTIONS_V1
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Net.WebSockets;
+using System.Text;
+using System.Threading.Tasks;
+using DurableTask.AzureStorage.Monitoring;
+using Dynamitey.DynamicObjects;
+using Microsoft.Azure.WebJobs.Host.Scale;
+using Microsoft.WindowsAzure.Storage;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    internal sealed class DurableTaskScaleMonitor : IScaleMonitor<DurableTaskTriggerMetrics>
+    {
+        private readonly string functionId;
+        private readonly FunctionName functionName;
+        private readonly string hubName;
+        private readonly string storageConnectionString;
+        private readonly EndToEndTraceHelper traceHelper;
+        private readonly ScaleMonitorDescriptor scaleMonitorDescriptor;
+
+        private DisconnectedPerformanceMonitor performanceMonitor;
+
+        public DurableTaskScaleMonitor(
+            string functionId,
+            FunctionName functionName,
+            string hubName,
+            string storageConnectionString,
+            EndToEndTraceHelper traceHelper,
+            DisconnectedPerformanceMonitor performanceMonitor = null)
+        {
+            this.functionId = functionId;
+            this.functionName = functionName;
+            this.hubName = hubName;
+            this.storageConnectionString = storageConnectionString;
+            this.performanceMonitor = performanceMonitor;
+            this.traceHelper = traceHelper;
+            this.scaleMonitorDescriptor = new ScaleMonitorDescriptor($"{this.functionId}-DurableTaskTrigger-{this.hubName}".ToLower());
+        }
+
+        public ScaleMonitorDescriptor Descriptor
+        {
+            get
+            {
+                return this.scaleMonitorDescriptor;
+            }
+        }
+
+        private DisconnectedPerformanceMonitor GetPerformanceMonitor()
+        {
+            if (this.performanceMonitor == null)
+            {
+                if (this.storageConnectionString == null)
+                {
+                    throw new ArgumentNullException(nameof(this.storageConnectionString));
+                }
+
+                this.performanceMonitor = new DisconnectedPerformanceMonitor(this.storageConnectionString, this.hubName);
+            }
+
+            return this.performanceMonitor;
+        }
+
+        async Task<ScaleMetrics> IScaleMonitor.GetMetricsAsync()
+        {
+            return await this.GetMetricsAsync();
+        }
+
+        public async Task<DurableTaskTriggerMetrics> GetMetricsAsync()
+        {
+            DurableTaskTriggerMetrics metrics = new DurableTaskTriggerMetrics();
+
+            // Durable stores its own metrics, so we just collect them here
+            PerformanceHeartbeat heartbeat = null;
+            try
+            {
+                DisconnectedPerformanceMonitor performanceMonitor = this.GetPerformanceMonitor();
+                heartbeat = await performanceMonitor.PulseAsync();
+            }
+            catch (StorageException e)
+            {
+                this.traceHelper.ExtensionWarningEvent(this.hubName, this.functionName.Name, string.Empty, e.ToString());
+            }
+
+            if (heartbeat != null)
+            {
+                metrics.PartitionCount = heartbeat.PartitionCount;
+                metrics.ControlQueueLengths = JsonConvert.SerializeObject(heartbeat.ControlQueueLengths);
+                metrics.ControlQueueLatencies = JsonConvert.SerializeObject(heartbeat.ControlQueueLatencies);
+                metrics.WorkItemQueueLength = heartbeat.WorkItemQueueLength;
+                if (heartbeat.WorkItemQueueLatency != null)
+                {
+                    metrics.WorkItemQueueLatency = heartbeat.WorkItemQueueLatency.ToString();
+                }
+            }
+
+            return metrics;
+        }
+
+        ScaleStatus IScaleMonitor.GetScaleStatus(ScaleStatusContext context)
+        {
+            return this.GetScaleStatusCore(context.WorkerCount, context.Metrics?.Cast<DurableTaskTriggerMetrics>().ToArray());
+        }
+
+        public ScaleStatus GetScaleStatus(ScaleStatusContext<DurableTaskTriggerMetrics> context)
+        {
+            return this.GetScaleStatusCore(context.WorkerCount, context.Metrics?.ToArray());
+        }
+
+        private ScaleStatus GetScaleStatusCore(int workerCount, DurableTaskTriggerMetrics[] metrics)
+        {
+            var scaleStatus = new ScaleStatus() { Vote = ScaleVote.None };
+            if (metrics == null)
+            {
+                return scaleStatus;
+            }
+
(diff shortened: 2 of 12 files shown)


src/WebJobs.Extensions.DurableTask/Listener/DurableTaskListener.cs
```diff
@@ -2,25 +2,42 @@
...
+using DurableTask.AzureStorage.Monitoring;
 using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Azure.WebJobs.Host.Listeners;
+#if NETSTANDARD2_0
+using Microsoft.Azure.WebJobs.Host.Scale;
+#endif
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 {
+#if !FUNCTIONS_V1
```
mathewc: I wonder if using a partial class would make this cleaner? Is it that the same code is being compiled for both v1 and v2, rather than having separate branches for v1/v2?

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
```diff
@@ -0,0 +1,149 @@
...
+                var webException = e.InnerException as WebException;
+                if (webException != null && webException.Status == WebExceptionStatus.ProtocolError)
+                {
+                    string statusCode = ((HttpWebResponse)webException.Response).StatusCode.ToString();
+                    string statusDesc = ((HttpWebResponse)webException.Response).StatusDescription;
+                    error = $"Durable Functions trigger error during PulseAsync: {statusCode}: {statusDesc}";
+                }
+                else if (webException != null && webException.Status == WebExceptionStatus.NameResolutionFailure)
+                {
+                    error = $"Durable Functions trigger storage error during PulseAsync: {webException.Message}";
```
mathewc: How do we know this is a storage error?

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
```diff
@@ -0,0 +1,149 @@
...
+                    error = e.ToString();
+                }
+
+                this.traceHelper.ExtensionWarningEvent(this.hubName, this.functionName.Name, string.Empty, error);
+            }
+
+            if (heartbeat != null)
+            {
+                metrics.PartitionCount = heartbeat.PartitionCount;
+                metrics.ControlQueueLengths = JsonConvert.SerializeObject(heartbeat.ControlQueueLengths);
```
mathewc: What are the upper bounds for the size of these array properties (queue lengths and latencies)? Ultimately these will be stored in table storage as an entity property, and the limit for a string property is 64K (doc [here](https://docs.microsoft.com/en-us/rest/api/storageservices/understanding-the-table-service-data-model#property-types)).

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTriggerMetrics.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    internal class DurableTaskTriggerMetrics : ScaleMetrics
+    {
+        /// <summary>
+        /// The number of partitions in the task hub.
+        /// </summary>
+        public int PartitionCount { get; set; }
+
+        /// <summary>
+        /// The number of messages across control queues.
```
mathewc: What will be the format of this. Can you include an example in a remarks section please? For the latencies property as well

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTriggerMetrics.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// </summary>
+        public string ControlQueueLatencies { get; set; }
+
+        /// <summary>
+        /// The number of messages in the work-item queue.
+        /// </summary>
+        public int WorkItemQueueLength { get; set; }
+
+        /// <summary>
+        /// A trend describing the latency of messages in the work-item queue
```
ConnorMcMahon: Unclear from the comments how this double matches a trend. Is it some fractional value between -1 and 1, is it an average of many integers, etc.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTriggerMetrics.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        /// <summary>
+        /// The number of messages across control queues.
+        /// </summary>
+        public string ControlQueueLengths { get; set; }
+
+        /// <summary>
+        /// The latency of messages across control queues.
+        /// </summary>
+        public string ControlQueueLatencies { get; set; }
```
ConnorMcMahon: These strings seem to map to `List<int>`s. Is there a reason we have to store them as strings here?

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskListener.cs
```diff
@@ -30,8 +47,18 @@ internal sealed class DurableTaskListener : IListener
                 throw new ArgumentNullException(nameof(functionName));
             }
 
+            this.functionId = functionId;
             this.functionName = functionName;
             this.functionType = functionType;
+            this.storageConnectionString = storageConnectionString;
+#if !FUNCTIONS_V1
+            this.scaleMonitor = new Lazy<DurableTaskScaleMonitor>(() => new DurableTaskScaleMonitor(
```
ConnorMcMahon: Make sure to handle the case where `storageConnectionString` is null, (i.e. when a customer is using a non AzureStorage provider).

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskListener.cs
```diff
@@ -2,25 +2,42 @@
...
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using DurableTask.AzureStorage.Monitoring;
 using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Azure.WebJobs.Host.Listeners;
+#if NETSTANDARD2_0
```
ConnorMcMahon: nit: Use !FUNCTIONS_V1

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
gzuber: The storage error message I took from `ScaleController` and added the "during PulseAsync", I'm not sure how that determination is made other than it being of type `StorageException`. My guess is that the `NameResolutionFailure` means the storage account wasn't found? 

The "trigger error" I just realized I added the "error" part so I've fixed it to match `ScaleController`.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
gzuber: That's a good question. @cgillum is there any documentation on the max number of control queues? I believe I found [some code](https://github.com/Azure/durabletask/blob/master/src/DurableTask.AzureStorage/AzureStorageOrchestrationService.cs#L572) that indicates # of control queues is correlated with # of partitions?

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTriggerMetrics.cs
gzuber: Yes -- I've included some examples in the comments (per Mathew) that should help. We have to serialize and deserialize these values because they're being stored in an Azure Storage table that can't handle complex objects

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTriggerMetrics.cs
gzuber: This is actually left over from PerformanceHeartbeat, so I'm removing it -- it's not necessary for MakeScaleRecommendation

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskListener.cs
gzuber: I believe that case is actually covered, but would love a second opinion. StorageConnectionString only needs to be valid if they've enabled runtime scaling. The scale monitor will be created no matter what, but the `DisconnectedPerformanceMonitor` won't be created unless runtime scaling is enabled -- at which point it will throw an exception if the connection string is null. So a customer that's using non-azure storage won't be affected unless they enable runtime scaling.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
mathewc: What's the maximum number of partitions? For both of these, just figure out what the max string length would be and ensure it fits

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTriggerMetrics.cs
mathewc: Yeah, it's unfortunate we have to convert to string here. The underlying storage for metrics is pluggable, but the default impl. is Azure Tables, and it only supports simple properties (code [here](https://github.com/Azure/azure-functions-host/blob/643c21f882218283a894dc29d78ab4088e073de7/src/WebJobs.Script.WebHost/Scale/TableEntityConverter.cs#L76)). There isn't any built in conversions from unsupported types.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
cgillum: We discussed this over IM, but the max partition count is 16, as documented [here](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-perf-and-scale#orchestrator-scale-out).

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTriggerMetrics.cs
cgillum: FWIW, if you're serializing `DurableTaskTriggerMetrics` as JSON before inserting the properties into Azure Storage, then you can use serialization attributes to separate the class properties from the actual JSON data contract. For example, this is a potential workaround that I used to see the WCF team do all the time, and it works with Json.NET:

```csharp
[DataContract]
class DurableTaskTriggerMetrics : ScaleMetrics
{
    [DataMember(Name = "ControlQueueLatencies")]
    private string controlQueueLatenciesAsString;

    [DataMember]
    public int PartitionCount { get; set; }

    [DataMember]
    public string ControlQueueLengths { get; set; }

    public List<int> ControlQueueLatencies { get; set; }
    
    [OnSerializing]
    private void OnSerializing(StreamingContext ctx)
    {
        this.controlQueueLatenciesAsString = JsonConvert.SerializeObject(this.ControlQueueLatencies);
    }
    
    [OnDeserializing]
    private void OnDeserializing(StreamingContext ctx)
    {
        this.ControlQueueLatencies = JsonConvert.DeserializeObject<List<int>>(this.controlQueueLatenciesAsString);
    }
}
```

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
```diff
@@ -0,0 +1,180 @@
...
+            FunctionName functionName,
+            string hubName,
+            string storageConnectionString,
+            EndToEndTraceHelper traceHelper,
+            DisconnectedPerformanceMonitor performanceMonitor = null)
+        {
+            this.functionId = functionId;
+            this.functionName = functionName;
+            this.hubName = hubName;
+            this.storageConnectionString = storageConnectionString;
```
cgillum: We've already established that in certain scenarios that the connection string could be null here. If that's the case, should we go ahead and throw at this point? If I'm reading this code correctly, an app could end up into a continuous failure loop if we allow a null connection string to be set here.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
```diff
@@ -0,0 +1,180 @@
...
+            }
+            catch (StorageException e)
+            {
+                string errorMsg = string.Empty;
+                var webException = e.InnerException as WebException;
+                if (webException != null && webException.Status == WebExceptionStatus.ProtocolError)
+                {
+                    string statusCode = ((HttpWebResponse)webException.Response)?.StatusCode.ToString();
+                    string statusDesc = ((HttpWebResponse)webException.Response)?.StatusDescription;
+                    errorMsg = $"Durable Functions trigger status during PulseAsync: {statusCode}: {statusDesc}";
```
cgillum: I'm a little concerned that we might be missing some important information compared to just doing `e.ToString()`. Was there a particular reason why you decided to have different handling for different `WebExceptionStatus` values?

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
```diff
@@ -0,0 +1,180 @@
...
+                    scaleStatus.Vote = ScaleVote.None;
+                    break;
+            }
+
+            this.traceHelper.ExtensionInformationalEvent(
+                                    this.hubName,
+                                    string.Empty,
+                                    this.functionName.Name,
+                                    $"Durable Functions Trigger Scale Decision: {scaleStatus.Vote.ToString()}, Reason: {scaleRecommendation?.Reason}",
+                                    writeToUserLogs: false);
```
cgillum: Any particular reason for these to *not* go in user logs? Is there a middle ground where we write to user logs for just `ScaleVote.ScaleOut` and `ScaleVote.ScaleIn`?

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
gzuber: So `DurableTaskScaleMonitor` will always be created for any function app. If we were to throw here then any app that doesn't use Azure Storage -- even if they have runtime scale **disabled** -- would fail. That's why the `DisconnectedPerformanceMonitor` is only created when needed during `GetMetricsAsync` or `GetScaleStatus` -- because those functions will only be called if the user has runtime scaling enabled. I think it might be worth it to explicitly throw is the connection string is null, but it would have to be when the performance monitor is created.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
gzuber: Yeah I'm open to changing this error handling. This if/else was copied straight from the durable ScaleController logic, so my intent was to make the error codes here look as similar to ScaleController errors as possible.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
gzuber: No, no strong reason. I can make that change.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
gzuber: According to [these]() docs, Azure Storage tables use UTF-16. So event if all 16 possible control queues were the max int value, it should be bigger than ~1288 bytes. The latencies max value shouldn't be bigger than ~1536 bytes assuming the timespan doesn't get bigger than 1000's of days

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
cgillum: My reading of the code is that this constructor is only called by `IScaleMonitor.GetMonitor()`. Are you saying we call this interface method even if runtime scaling is disabled?

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
cgillum: I don't recall exactly why the scale controller did it this way, but I would feel better if we just did `e.ToString()`.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
gzuber: Correct. The constructor is only called by `IScaleMonitor.GetMonitor()`, but that is called on start up [here](https://github.com/Azure/azure-webjobs-sdk/blob/7b97fd9e3b5b00ef5808eeda4439768cb947f530/src/Microsoft.Azure.WebJobs.Host/Listeners/HostListenerFactory.cs#L107). After construction, `GetMetricsAsync()` and `GetScaleStatus()` will only be called if runtime scaling is enabled [here](https://github.com/Azure/azure-functions-host/blob/3d13bf10712edceb521d6b25b7f92b1ba32a17b2/src/WebJobs.Script/Scale/FunctionsScaleMonitorService.cs#L47)

cgillum: Got it, thanks for the explanation. I think Connor and I can probably take a shot at refactoring this later as we add support for alternate backends.

Issue #976, opened 2019-10-17
Invalid Table Name on App Service deploy


jsacapdev:
### Description

If my durable function is deployed to an app service i get this error:

```
The function runtime is unable to start. Microsoft.Azure.WebJobs.Extensions.DurableTask: Task hub name 'dash-breaks-this' should contain only alphanumeric characters excluding '-' and have length up to 50. Microsoft.WindowsAzure.Storage: Invalid table name. Check MSDN for more information about valid table naming.
```

### Expected behavior

I should see the function app is running page.

### Actual behavior

I get the error above. Also trying to hit the app service I get:

`Function host is not running.`

### Relevant source code snippets

See the following respository for the code, and also how the resources were created (using the AZ CLI)

https://github.com/jsacapdev/dashbreaks

### Known workarounds

This seems to indicate it was a problem before and it was fixed:

https://github.com/Azure/azure-functions-durable-extension/issues/505

### App Details

See repository above.

### Screenshots

![image](https://user-images.githubusercontent.com/10151920/67029372-128ef500-f105-11e9-8c74-999d50d1c49e.png)

### If deployed to Azure

See repository for how to create.

cgillum:
@ConnorMcMahon can you take a look at this? I suspect it's caused by the new logic which derives the task hub name from the site name. We'll need to do some sanitization to avoid this kind of problem.

@jsacapdev You should be able to work around this by defining a task hub name in your host.json file:

```json
{
  "version": "2.0",
  "extensions": {
    "durableTask": {
      "hubName": "MyTaskHub"
    }
  }
}
```

rouafi:
@cgillum hi, thanks for the workaround but I am still not able to make it work.
Any idea ? 
My function is using the package : Microsoft.Azure.WebJobs.Extensions.DurableTask version : 2.0.0-beta3

cgillum:
Fixed with https://github.com/Azure/azure-functions-durable-extension/pull/1010

PR #1010, opened 2019-10-31
Sanitize the default task hub names for Azure Storage

ConnorMcMahon:
In a previous release, we started using the site name as the default
task hub value. However, it is possible for the site name not to be a
valid task hub name for Azure Storage, due to having '-' characters and
not meeting the length requirements. We now ensure that no matter the
sitename, the customer will have a valid task hub.

Addresses issue #976

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
+++ b/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
@@ -27,8 +27,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             JsonConvert.PopulateObject(JsonConvert.SerializeObject(this.options.StorageProvider), this.azureStorageOptions);
 
             this.azureStorageOptions.Validate();
-            if (this.options.HubName != null // Need to access HubName once to populate IsDefaultHubName
-                && !this.options.IsDefaultHubName)
+            if (!this.options.IsDefaultHubName())
             {
                 this.azureStorageOptions.ValidateHubName(this.options.HubName);
             }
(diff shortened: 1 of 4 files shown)


src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
```diff
@@ -27,7 +27,15 @@ internal class AzureStorageDurabilityProviderFactory : IDurabilityProviderFactor
             JsonConvert.PopulateObject(JsonConvert.SerializeObject(this.options.StorageProvider), this.azureStorageOptions);
 
             this.azureStorageOptions.Validate();
-            this.azureStorageOptions.ValidateHubName(this.options.HubName);
+            if (this.options.HubName != null // Need to access HubName once to populate IsDefaultHubName
```
ConnorMcMahon: I don't love this. I may need to rework this logic a little bit.

src/WebJobs.Extensions.DurableTask/Options/AzureStorageOptions.cs
```diff
@@ -113,9 +119,37 @@ public void ValidateHubName(string hubName)
...
+
+        internal bool IsSanitizedHubName(string hubName, out string sanitizedHubName)
+        {
+            sanitizedHubName = new string(hubName.ToCharArray()
+                                .Where(char.IsLetterOrDigit)
+                                .Take(MaxTaskHubNameSize)
+                                .ToArray());
+            if (sanitizedHubName.Length < MinTaskHubNameSize)
+            {
+                sanitizedHubName = sanitizedHubName + new string(TaskHubPadding, MinTaskHubNameSize - sanitizedHubName.Length);
```
cgillum: Can you summarize what your strategy is here? I was expecting something that simply stripped dashes and perhaps truncated the name.

```csharp
string siteName = "my-site-1";
string hubName = siteName.Replace("-", "");
hubName = hubName.Substring(0, Math.Min(hubName.Length, 45)) + "Hub";
// result is mysite1Hub
```
I'm a bit confused by the padding logic.

ConnorMcMahon: The padding is for too short of names (i.e. someone has a site-name like like "ab" or "a-b"). I do like your notion of just appending "Hub" instead of just appending `a`'s to the end.

PR #1014, opened 2019-11-01
Throws an exception on invalid dedupe status when starting a new orchestration

amdeel:
Starting a new orchestration with an existing instance id now has configurable behavior (#987 ). This PR changes the behavior to throw an exception when attempting to start a new orchestration in an invalid dedupe status.

This PR also adds tests for each of the configurable OverridableStates to test this behavior.

Issue related to this change:
#367

### DIFF
--- a/test/Common/DurableTaskEndToEndTests.cs
+++ b/test/Common/DurableTaskEndToEndTests.cs
@@ -3496,7 +3496,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
         [MemberData(nameof(TestDataGenerator.GetExtendedSessionAndFullFeaturedStorageProviderOptions), MemberType = typeof(TestDataGenerator))]
         public async Task Dedupe_Default_NotRunning_ThrowsException(bool extendedSessions, string storageProvider)
         {
-           var instanceId = "OverridableStatesTest";
+           var instanceId = "OverridableStatesDefaultTest";
 
            using (JobHost host = TestHelpers.GetJobHost(
                 this.loggerProvider,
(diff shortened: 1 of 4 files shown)


test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -3491,6 +3491,102 @@ public async Task MaxOrchestrationAction_MaxReached_OrchestrationFails()
             }
         }
 
+        [Theory]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory + "_BVT")]
+        [MemberData(nameof(TestDataGenerator.GetExtendedSessionAndFullFeaturedStorageProviderOptions), MemberType = typeof(TestDataGenerator))]
+        public async Task DedupeStates_NotRunningStates_ThrowsException(bool extendedSessions, string storageProvider)
+        {
+            DurableTaskOptions options = new DurableTaskOptions();
+            options.OverridableExistingInstanceStates = OverridableStates.NonRunningStates;
```
cgillum: This is the default value, right? If so, can we remove this line? That way this test will fail if we accidentally change that default in the future.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -3491,6 +3491,102 @@ public async Task MaxOrchestrationAction_MaxReached_OrchestrationFails()
             }
         }
 
+        [Theory]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory + "_BVT")]
```
cgillum: You can remove the `+ "_BVT"` from these test traits.

PR #1024, opened 2019-11-04
Analyzer: FunctionReturnTypeAnalyzer wont flag code when expected return type is Task

### LINKED ISSUE #998: Calling activity function w/out awaiting causes DF0110 warning
### Description
The following code snippet results in a DF0110 warning about an invalid return type.

```csharp
Task[] tasks = new Task[count];
for (int i = 0; i < tasks.Length; i++)
{
    tasks[i] = context.CallActivityAsync("SayHello", i.ToString("0000"));
}

await Task.WhenAll(tasks);
```

### Expected behavior
This should not generate a warning. `Task` should always be a valid return value for `CallActivityAsync`, or any of the async Durable APIs.

### Known workarounds
The problem goes away if I change `CallActivityAsync("SayHello", ...)` to `CallActivityAsync<string>("SayHello", ...)` (in other words, if I provide a type parameter).

### App Details

- **Analyzer version**: 0.1.1
- **Durable functions extension version**: 2.0.0-beta3

amdeel:
Fixed activity FunctionReturnTypeAnalyzer to accept the expected return type to be Task regardless of activity function definition return type.

resolves #998

### DIFF
--- a/src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/FunctionReturnTypeAnalyzer.cs
+++ b/src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/FunctionReturnTypeAnalyzer.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers
                 var functionDefinition = availableFunctions.Where(x => x.FunctionName == node.Name).SingleOrDefault();
                 if (functionDefinition != null)
                 {
-
+                    // Functions can always return Task, regardless of function definition return type
                     if (functionDefinition.ReturnType != node.ExpectedReturnType &&
                         (node.ExpectedReturnType != "System.Threading.Tasks.Task"))
                     {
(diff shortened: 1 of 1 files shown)


src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/FunctionReturnTypeAnalyzer.cs
```diff
@@ -29,7 +29,7 @@ public void ReportProblems(CompilationAnalysisContext cac, IEnumerable<ActivityF
                 {
 
                     if (functionDefinition.ReturnType != node.ExpectedReturnType &&
-                        !(functionDefinition.ReturnType == "System.Void" && node.ExpectedReturnType == "System.Threading.Tasks.Task"))
+                        (node.ExpectedReturnType != "System.Threading.Tasks.Task"))
```
cgillum: Could you add a comment briefly explaining what this logic is looking for?

cgillum: Also, should we add a unit test to cover this case?

PR #1031, opened 2019-11-07
Add support for delete operation in DispatchAsync and in Http API

sebastianburckhardt:
As discussed in #931, this PR adds a bit of convenience for implementing deletion of entities. 

- One can now issue a DELETE request to the HTTP API, with the effect of signaling an operation called 'delete' on the entity.

- For the class-based syntax, one can call `delete` on an entity with class `T` even if `T` does not have a delete method. In that case `DispatchAsync<T>` simply calls `Entity.Current.Delete()`.


cgillum:
Were there any more changes planned for this PR or is it ready to be merged?

sebastianburckhardt:
It is ready.

Issue #951, opened 2019-09-26
Activity functions cannot be manually triggered from the Azure Portal


cgillum:
### Description
Activity functions cannot be invoked from the Azure Portal. Unlike orchestrator functions, there should be no technical reason we can't support this.

### Expected behavior
Manually invoking activity functions in the Azure Portal should work in the same way that HTTP or queue trigger functions can be manually invoked.

### Actual behavior
The following error was experienced when trying to invoke the activity trigger function:

> Microsoft.Azure.WebJobs.Extensions.DurableTask: Unable to cast object of type 'System.String' to type 'Microsoft.Azure.WebJobs.DurableActivityContext'.

PR #1058, opened 2019-11-18
Allows activity functions to be manually triggered in Azure Portal

ConnorMcMahon:
Addresses issue #951

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
@@ -108,7 +108,19 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
             public Task<ITriggerData> BindAsync(object value, ValueBindingContext context)
             {
-                var activityContext = (DurableActivityContext)value;
+                if (!(value is DurableActivityContext activityContext))
+                {
+                    if (!(value is string serializedInput))
+                    {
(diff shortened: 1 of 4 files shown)


src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
```diff
@@ -108,7 +108,19 @@ private class ActivityTriggerBinding : ITriggerBinding
...
+                {
+                    if (!(value is string serializedInput))
+                    {
+                        throw new InvalidOperationException($"Cannot execute an Activity Trigger without a {nameof(DurableActivityContext)} or a {nameof(String)} that represents the serialized input.");
+                    }
+
+                    // Durable functions expects input as a JArray with one element.
+                    serializedInput = $"[{serializedInput}]";
+
+                    activityContext = new DurableActivityContext($"azure-portal-execution-{Guid.NewGuid().ToString()}", serializedInput);
```
cgillum: I'm not so sure about using `azure-portal-` in the string here since this could theoretically apply to other scenarios as well. I do recommend we add a comment explaining that we added this logic primarily for Azure portal manual invocation, however.

ConnorMcMahon: I was unsure too. I thought customers may want to be able to filter out portal requests, but as you said, I am not positive Azure Portal is the only scenario this lights up (it technically is any request to the `/admin/{functionName}` endpoint I believe).

PR #1075, opened 2019-11-28
Simplify Durable Entity operation method discovery for implicit interface implementations

BrunoZell:
This simplifies the code introduced in #935.

It's a small change, but useful in the long run:
- Simplifies and optimizes code
- Making code more safe as we don't rely on the details of the `Type.FullName` string (e.g. how nested types are represented)
- Handles type names edge cases we don't know of yet or other .NET languages introduce later on.

Instead of searching for the correct method on the concrete type itself after one failed attempt, we look for the method with the desired name directly on the interface type, thus having a `MethodInfo` related to the implemented interface. The invokation later on will succeed as long as the durable entity object implements the interface (which is should).


cgillum:
Thanks for submitting this PR! I like the simplification.

cgillum:
Awesome. I will go ahead and merge this. It will go out with our v2.1.0 release.

PR #1085, opened 2019-12-03
Update HttpApiHandler to work with AllowSynchronousIO = false

### LINKED ISSUE #1039: Synchronous deserialization of request content in HandleStartOrchestratorRequestAsync results in InvalidOperationException.
### Description

Calling the Durable Functions HTTP API to [start an orchestration](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-http-api#start-orchestration) when running in a 3.0-preview (3.0.1740) Azure Functions Host results in an `InvalidOperationException` from Kestrel due to changes in .NET Core 3.0 regarding synchronous operations:

```json
{
    "Message": "Something went wrong while processing your request",
    "ExceptionMessage": "Synchronous operations are disallowed. Call ReadAsync or set AllowSynchronousIO to true instead.",
    "ExceptionType": "System.InvalidOperationException",
    "StackTrace": "   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpRequestStream.Read(Byte[] buffer, Int32 offset, Int32 count)\n   at System.Net.Http.DelegatingStream.Read(Byte[] buffer, Int32 offset, Int32 count)\n   at System.IO.StreamReader.ReadBuffer(Span`1 userBuffer, Boolean& readToUserBuffer)\n   at System.IO.StreamReader.ReadSpan(Span`1 buffer)\n   at System.IO.StreamReader.Read(Char[] buffer, Int32 index, Int32 count)\n   at Newtonsoft.Json.JsonTextReader.ReadData(Boolean append, Int32 charsRequired)\n   at Newtonsoft.Json.JsonTextReader.ParseValue()\n   at Newtonsoft.Json.JsonReader.ReadAndMoveToContent()\n   at Newtonsoft.Json.JsonReader.ReadForType(JsonContract contract, Boolean hasConverter)\n   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)\n   at Newtonsoft.Json.JsonSerializer.DeserializeInternal(JsonReader reader, Type objectType)\n   at Newtonsoft.Json.JsonSerializer.Deserialize(JsonReader reader, Type objectType)\n   at Newtonsoft.Json.JsonSerializer.Deserialize[T](JsonReader reader)\n   at Microsoft.Azure.WebJobs.Extensions.DurableTask.HttpApiHandler.HandleStartOrchestratorRequestAsync(HttpRequestMessage request, String functionName, String instanceId) in d:\\a\\r1\\a\\azure-functions-durable-extension\\src\\WebJobs.Extensions.DurableTask\\HttpApiHandler.cs:line 590\n   at Microsoft.Azure.WebJobs.Extensions.DurableTask.HttpApiHandler.HandleRequestAsync(HttpRequestMessage request) in d:\\a\\r1\\a\\azure-functions-durable-extension\\src\\WebJobs.Extensions.DurableTask\\HttpApiHandler.cs:line 200"
}
```

The synchronous read appears to be [here](https://github.com/Azure/azure-functions-durable-extension/blob/a9657173791f288d58098781b5a61a0a571bfde4/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs#L590).

See the [Azure Functions Host issue](https://github.com/Azure/azure-functions-host/issues/4876) regarding this breaking change in .NET Core 3.0.

### Expected behavior

The read of the JSON request body should be async so that Kestrel's happy.

### Actual behavior

The unhandled exception results in a `500` being returned from the HTTP API, as described above.

### Known workarounds

Presumably setting `AllowSynchronousIO` to true in appsettings.json should work around the issue (at the cost to throughput), but I haven't been able to get that to work in the context of Azure Functions.

### App Details

- **Durable Functions extension version (e.g. v1.8.3)**: `2.0.0`
- **Azure Functions runtime version (1.0 or 2.0)**: `3.0.1740`
- **Programming language used**: [Rust](https://github.com/peterhuene/azure-functions-rs) 😄

anthonychu:
- Functions V3 now sets AllowSynchronousIO = false
- JsonSerializer.Deserialize<T> calls Read on stream (throws exception)

Fixes #1039

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
+++ b/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
@@ -582,7 +582,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 object input = null;
                 if (request.Content?.Headers?.ContentLength > 0)
                 {
-                    var json = await request.Content.ReadAsStringAsync();
+                    string json = await request.Content.ReadAsStringAsync();
                     input = JsonConvert.DeserializeObject(json, MessagePayloadDataConverter.MessageSettings);
                 }
 
--- a/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
+++ b/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
@@ -579,7 +579,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 IDurableOrchestrationClient client = this.GetClient(request);
 
                 object input = null;
-                if (request.Content?.Headers?.ContentLength > 0)
+                if (request.Content?.Headers?.ContentLength != 0)
                 {
                     string json = await request.Content.ReadAsStringAsync();
                     input = JsonConvert.DeserializeObject(json, MessagePayloadDataConverter.MessageSettings);
--- a/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
+++ b/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.Diagnostics;
-using System.IO;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
--- a/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
+++ b/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
@@ -580,15 +580,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 IDurableOrchestrationClient client = this.GetClient(request);
 
                 object input = null;
-                if (request.Content != null)
+                if (request.Content?.Headers?.ContentLength > 0)
                 {
-                    using (Stream s = await request.Content.ReadAsStreamAsync())
-                    using (StreamReader sr = new StreamReader(s))
-                    using (JsonReader reader = new JsonTextReader(sr))
-                    {
-                        JsonSerializer serializer = JsonSerializer.Create(MessagePayloadDataConverter.MessageSettings);
-                        input = serializer.Deserialize<object>(reader);
-                    }
+                    var json = await request.Content.ReadAsStringAsync();
+                    input = JsonConvert.DeserializeObject(json, MessagePayloadDataConverter.MessageSettings);
                 }
 
                 string id = await client.StartNewAsync(functionName, instanceId, input);


anthonychu:
This might also work to avoid allocating the string. But `JObject.LoadAsync()` doesn't offer any any control over the serializer settings.

```csharp
object input = null;
if (request.Content?.Headers?.ContentLength > 0)
{
    using (Stream s = await request.Content.ReadAsStreamAsync())
    using (StreamReader sr = new StreamReader(s))
    using (JsonReader reader = new JsonTextReader(sr))
    {
        JsonSerializer serializer = JsonSerializer.Create(MessagePayloadDataConverter.MessageSettings);
        input = await JObject.LoadAsync(reader);
    }
}
```

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -580,15 +579,10 @@ private static int GetIntQueryParameterValue(NameValueCollection queryStringName
                 IDurableOrchestrationClient client = this.GetClient(request);
 
                 object input = null;
-                if (request.Content != null)
+                if (request.Content?.Headers?.ContentLength > 0)
```
ConnorMcMahon: The end result of 'request.Content?.Headers?.ContentLength' is `long?`, meaning we could still get a NullReferenceException here. You could fix this logic with `(request.Content?.Headers?.ContentLength  ?? 0 ) > 0`, though that is getting kind of ugly.

In general, I am not sure I like the new dependency on ContentLength. While almost all clients will appropriately assign a ContentLength header when sending a body, I am not sure if it is technically required in the HTTP spec. In the case where a client is forgetting to send the ContentLength header, we would now no longer reading their input. Instead, consider checking that `json` below is not null or empty.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -580,15 +579,10 @@ private static int GetIntQueryParameterValue(NameValueCollection queryStringName
...
                 {
-                    using (Stream s = await request.Content.ReadAsStreamAsync())
-                    using (StreamReader sr = new StreamReader(s))
-                    using (JsonReader reader = new JsonTextReader(sr))
-                    {
-                        JsonSerializer serializer = JsonSerializer.Create(MessagePayloadDataConverter.MessageSettings);
-                        input = serializer.Deserialize<object>(reader);
-                    }
+                    string json = await request.Content.ReadAsStringAsync();
+                    input = JsonConvert.DeserializeObject(json, MessagePayloadDataConverter.MessageSettings);
```
ConnorMcMahon: I don't think we should pivot to `JObject.LoadAsync()`. As you mentioned, we have no way to utilize our `JsonSerializationSettings` then. I'm also not sure how well the type coercion works under the hood from `JObject` to the customer's specified types in their code (I'm assuming well, but I prefer the notion of leaving it an `object`).

What we really want is `serializer.DeserializeAsync()`, but that only exists in net core 3.0, not even in .net standard 2.1. Regardless, at the version we are stuck with, I highly doubt we will get access to it. 

In general, the main downside here is the [memory usage](https://www.newtonsoft.com/json/help/html/Performance.htm). Since this is only used in orchestration starts, and in general customers will hit large IO costs when using objects > 64 KB anyways (due to having to store the message in blobs instead of queues), I am not too concerned about this currently. We can always revisit this later if it becomes an issue.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
cgillum: I don't think there is actually a risk of a null-ref here. Operators like `>` work fine with nullable value types. However, there is a risk with specifically checking for `ContentLength > 0`. I suggest instead that you compare `ContentLength != 0` to account for the missing `Content-Length` header.

ConnorMcMahon: Huh, today I learned!

anthonychu: Yeah I had to test it out when I was writing it because it looked odd. Most surprising was `request.Content?.Headers?.ContentLength == 0` is false when there's no value.

PR #1092, opened 2019-12-05
Add azure-pipelines.yml for v1 branch

ConnorMcMahon:


cgillum:
Looks like there were some test failures?

ConnorMcMahon:
The test failures are all with Functions V1, and seem to be some odd build related errors (i.e. dlls not being loaded into the correct locations). Running the tests locally works fine.

azure-pipelines.yml
```diff
@@ -0,0 +1,114 @@
...
+      buildConfiguration: 'Debug'
+
+    steps:
+    - task: NuGetToolInstaller@0
+      inputs:
+        versionSpec: '5.3.0'
+
+    - task: NuGetCommand@2
+      inputs:
+        restoreSolution: '$(solution)'
```
amdeel: Why did you have to change this from the DotNetCoreCLI task restore?

ConnorMcMahon: I was getting those weird assembly mismatch issues when using the DotNetCoreCli. Changing to this got rid of those errors. I'm also hestitant to use the DotNetCoreCli command for restoring all of our projects, since some are full framework.

PR #1098, opened 2019-12-10
Added ListEntitiesAsync Api

### LINKED ISSUE #997: Allow to get all available Durable Entities
I'd like to have a way to get the list of all existing Durable Entities. I can think of different scenarios where this could be useful. e.g. 

- An application with multiple users who create entities, and the app administrator would like to know what entities have been created, 
- The user not necessarily remembers the identifier of all entities that have been created. 

I'd like to have this available as part via the admin API and the client within the programming model. Additionally, some kind of paging might be useful. 

- **API**: `GET /runtime/webhooks/durabletask/entities/` Similar to https://github.com/Azure/azure-functions-durable-extension/issues/733 but instead of returning the state of a particular entity, return an array of identifiers of the available entities. Also, similar to https://docs.microsoft.com/en-gb/azure/azure-functions/durable/durable-functions-http-api#get-all-instances-status but instead of returning orchestrations, to return entities. 
- **Via the client**, Similar to https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-instance-management#query-all-instances, but instead of returning orchestrations, to return entities. 

cc: @cgillum, as discussed previously via Twitter

amdeel:
resolves #997

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityClient.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
@@ -39,5 +40,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// <param name="connectionName">The name of the connection string associated with <paramref name="taskHubName"/>.</param>
         /// <returns>a response containing the current state of the entity.</returns>
         Task<EntityStateResponse<T>> ReadEntityStateAsync<T>(EntityId entityId, string taskHubName = null, string connectionName = null);
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="query"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        Task<EntityQueryResult> ListEntitiesAsync(EntityQuery query, CancellationToken cancellationToken)
     }
 }
--- a/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityClient.cs
@@ -1,8 +1,8 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
-using System.Threading;
 using System;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
--- a/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityClient.cs
@@ -41,7 +41,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// <returns>a response containing the current state of the entity.</returns>
         Task<EntityStateResponse<T>> ReadEntityStateAsync<T>(EntityId entityId, string taskHubName = null, string connectionName = null);
 
-        /// /// <summary>
+        /// <summary>
         /// Gets the status of all entity instances with paging that match the specified query conditions.
         /// </summary>
         /// <param name="query">Return entity instances that match the specified query conditions.</param>
--- a/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityClient.cs
@@ -45,8 +45,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// Gets the status of all entity instances with paging that match the specified query conditions.
         /// </summary>
         /// <param name="query">Return entity instances that match the specified query conditions.</param>
-        /// <param name="cancellationToken">Cancellation token that can be used to cancel the status query operation.</param>
-        /// <returns>Returns each page of entity status for all instances and continuation token of next page.</returns>
+        /// <param name="cancellationToken">Cancellation token that can be used to cancel the query operation.</param>
+        /// <returns>Returns a page of entity instances and a continuation token for fetching the next page.</returns>
         Task<EntityQueryResult> ListEntitiesAsync(EntityQuery query, CancellationToken cancellationToken);
     }
 }
--- a/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityClient.cs
@@ -41,12 +41,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// <returns>a response containing the current state of the entity.</returns>
         Task<EntityStateResponse<T>> ReadEntityStateAsync<T>(EntityId entityId, string taskHubName = null, string connectionName = null);
 
-        /// <summary>
-        /// 
+        /// /// <summary>
+        /// Gets the status of all entity instances with paging that match the specified query conditions.
         /// </summary>
-        /// <param name="query"></param>
-        /// <param name="cancellationToken"></param>
-        /// <returns></returns>
-        Task<EntityQueryResult> ListEntitiesAsync(EntityQuery query, CancellationToken cancellationToken)
+        /// <param name="query">Return entity instances that match the specified query conditions.</param>
+        /// <param name="cancellationToken">Cancellation token that can be used to cancel the status query operation.</param>
+        /// <returns>Returns each page of entity status for all instances and continuation token of next page.</returns>
+        Task<EntityQueryResult> ListEntitiesAsync(EntityQuery query, CancellationToken cancellationToken);
     }
 }
--- a/src/WebJobs.Extensions.DurableTask/EntityQuery.cs
+++ b/src/WebJobs.Extensions.DurableTask/EntityQuery.cs
@@ -1,36 +1,39 @@
-﻿using System;
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
 using System.Collections.Generic;
 using System.Text;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 {
     /// <summary>
-    /// 
+    /// Query condition for searching the status of entity instances.
     /// </summary>
     public class EntityQuery
     {
         /// <summary>
-        /// 
+        /// Return entity instances associated with this entity name.
         /// </summary>
         public string EntityName { get; set; }
 
         /// <summary>
-        /// 
+        /// Return entity instances which had operations after this DateTime.
         /// </summary>
         public DateTime LastOperationFrom { get; set; } = DateTime.MinValue;
 
         /// <summary>
-        /// 
+        /// Return entity instances which had operations before this DateTime.
         /// </summary>
         public DateTime LastOperationTo { get; set; } = DateTime.MaxValue;
 
         /// <summary>
-        /// 
+        /// Number of records per one request. The default value is 100.
         /// </summary>
         public int PageSize { get; set; } = 100;
 
         /// <summary>
-        /// 
+        /// ContinuationToken of the pager.
         /// </summary>
         public string ContinuationToken { get; set; }
     }
--- a/src/WebJobs.Extensions.DurableTask/EntityQuery.cs
+++ b/src/WebJobs.Extensions.DurableTask/EntityQuery.cs
@@ -36,5 +36,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// ContinuationToken of the pager.
         /// </summary>
         public string ContinuationToken { get; set; }
+
+        /// <summary>
+        /// If true, the query will get the input of the entity.
+        /// </summary>
+        public bool FetchInput { get; set; } = false;
     }
 }
--- a/src/WebJobs.Extensions.DurableTask/EntityQuery.cs
+++ b/src/WebJobs.Extensions.DurableTask/EntityQuery.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         public string ContinuationToken { get; set; }
 
         /// <summary>
-        /// Determines whether the query will include the input of the entity.
+        /// Determines whether the query will include the state of the entity.
         /// </summary>
         public bool FetchState { get; set; } = false;
     }
--- a/src/WebJobs.Extensions.DurableTask/EntityQuery.cs
+++ b/src/WebJobs.Extensions.DurableTask/EntityQuery.cs
@@ -38,8 +38,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         public string ContinuationToken { get; set; }
 
         /// <summary>
-        /// If true, the query will get the input of the entity.
+        /// Determines whether the query will include the input of the entity.
         /// </summary>
-        public bool FetchInput { get; set; } = false;
+        public bool FetchState { get; set; } = false;
     }
 }
--- a/src/WebJobs.Extensions.DurableTask/EntityQuery.cs
+++ b/src/WebJobs.Extensions.DurableTask/EntityQuery.cs
@@ -1,0 +1,37 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class EntityQuery
+    {
+        /// <summary>
+        /// 
+        /// </summary>
+        public string EntityName { get; set; }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public DateTime LastOperationFrom { get; set; } = DateTime.MinValue;
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public DateTime LastOperationTo { get; set; } = DateTime.MaxValue;
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public int PageSize { get; set; } = 100;
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public string ContinuationToken { get; set; }
+    }
+}
--- a/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
+++ b/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
@@ -214,11 +214,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                         string entityName = (string)routeValues[EntityNameRouteParameter];
                         string entityKey = (string)routeValues[EntityKeyRouteParameter];
                         EntityId entityId = new EntityId(entityName, entityKey);
-                        if (entityKey != null && request.Method == HttpMethod.Get)
+                        if (!string.IsNullOrEmpty(entityKey) && request.Method == HttpMethod.Get)
                         {
                             return await this.HandleGetEntityRequestAsync(request, entityId);
                         }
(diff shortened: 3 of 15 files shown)


src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -212,10 +214,14 @@ public async Task<HttpResponseMessage> HandleRequestAsync(HttpRequestMessage req
                         string entityName = (string)routeValues[EntityNameRouteParameter];
                         string entityKey = (string)routeValues[EntityKeyRouteParameter];
                         EntityId entityId = new EntityId(entityName, entityKey);
-                        if (request.Method == HttpMethod.Get)
+                        if (!string.IsNullOrEmpty(entityKey) && request.Method == HttpMethod.Get)
```
cgillum: Unfortunately this means it won't be possible to query a single entity with an empty `entityKey` value (which is something we intentionally support). I'm not sure the best way to fix this.  Maybe if we get a request for `GET /entities/$` we could treat the `$` internally as the empty string? I don't think we need to block on this for now, but something to consider. If we don't fix it, we should add a TODO comment explaining that we need to revisit this.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -212,10 +214,14 @@ public async Task<HttpResponseMessage> HandleRequestAsync(HttpRequestMessage req
...
                         EntityId entityId = new EntityId(entityName, entityKey);
-                        if (request.Method == HttpMethod.Get)
+                        if (!string.IsNullOrEmpty(entityKey) && request.Method == HttpMethod.Get)
                         {
                             return await this.HandleGetEntityRequestAsync(request, entityId);
                         }
+                        else if (request.Method == HttpMethod.Get)
+                        {
+                            return await this.HandleListEntitiesRequestAsync(request, entityName);
+                        }
```
cgillum: I suggest we structure this a bit differently:
```csharp
else if (request.Method == HttpMethod.Get)
{
    // TODO: Need a convention for handling the case for a single entity with an empty entity key
    if (!string.IsNullOrEmpty(entityKey))
    {
        return await this.HandleGetEntityRequestAsync(request, entityId);
    }
    else
    {
        return await this.HandleListEntitiesRequestAsync(request, entityName);
    }
}
```
This is a slightly safer way to implement the exact same logic. It's a little easier to understand and it is less likely to be accidentally broken by future if/else condition changes.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -367,6 +373,46 @@ public async Task<HttpResponseMessage> HandleRequestAsync(HttpRequestMessage req
...
         }
 
+        private async Task<HttpResponseMessage> HandleListEntitiesRequestAsync(
+            HttpRequestMessage request, string entityName)
+        {
+            IDurableEntityClient client = this.GetClient(request);
+            var queryNameValuePairs = request.GetQueryNameValuePairs();
+            var lastOperationTimeFrom = GetDateTimeQueryParameterValue(queryNameValuePairs, LastOperationTimeFrom, default(DateTime));
+            var lastOperationTimeTo = GetDateTimeQueryParameterValue(queryNameValuePairs, LastOperationTimeTo, default(DateTime));
+            var pageSize = GetIntQueryParameterValue(queryNameValuePairs, PageSizeParameter);
```
cgillum: What happens if no pageSize is specified by the caller? I think we should default to 100 in that case, but I didn't see any logic which checks for this.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -507,6 +553,16 @@ private StatusResponsePayload ConvertFrom(DurableOrchestrationStatus status)
             };
         }
 
+        private StatusResponsePayload ConvertFrom(DurableEntityStatus status)
```
cgillum: Make this method `static` since it doesn't access any instance members.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -507,6 +553,16 @@ private StatusResponsePayload ConvertFrom(DurableOrchestrationStatus status)
...
         }
 
+        private StatusResponsePayload ConvertFrom(DurableEntityStatus status)
+        {
+            return new StatusResponsePayload
+            {
+                InstanceId = status.EntityId.ToString(),
+                Input = status.State,
+                LastUpdatedTime = status.LastOperationTime.ToString("s") + "Z",
+            };
```
cgillum: I don't think `StatusResponsePayload` is the right thing to return. That class is for orchestration instances, and fields like `InstanceId`, `Input`, and `LastUpdatedTime` won't make sense to customers when they are querying for entities. Rather, we should be returning something closer to `DurableEntityStatus` which has field names that will make much more sense for entities.

One change I would make to `DurableEntityStatus`, however, is that instead of trying to serialize an `EntityId` property, we should have separate string fields for the entity key and the entity name. I expect the result should look something like this:

```json
[{
  "entityName": "cat",
  "entityId": "whiskers",
  "state": {
    "age": 2,
    "status": "hungry"
  },
  "lastOperationTime": "2019-12-11T06:30:03Z"
}]
```

A simple way to do this would be to return a hand-crafted `JObject` from this method instead of a `StatusResponsePayload`. Alternatively, you could modify `DurableEntityStatus` to serialize itself into JSON this way - which has the nice (but not required) benefit of allowing customers to easily serialize these objects in their code the same way we serialize them.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -367,6 +373,46 @@ public async Task<HttpResponseMessage> HandleRequestAsync(HttpRequestMessage req
...
+            var query = new EntityQuery()
+            {
+                EntityName = entityName,
+                LastOperationFrom = lastOperationTimeFrom,
+                LastOperationTo = lastOperationTimeTo,
+                PageSize = pageSize,
+                ContinuationToken = continuationToken,
+            };
+
+            var context = await client.ListEntitiesAsync(query, CancellationToken.None);
```
cgillum: Instead of `context`, consider naming this variable `result`.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -50,6 +50,8 @@ internal class HttpApiHandler
         private const string RuntimeStatusParameter = "runtimeStatus";
         private const string PageSizeParameter = "top";
         private const string ReturnInternalServerErrorOnFailure = "returnInternalServerErrorOnFailure";
+        private const string LastOperationTimeFrom = "lastOperationTimeFrom";
+        private const string LastOperationTimeTo = " lastOperationTimeTo";
```
cgillum: It looks like we're missing the `fetchState` query string parameter. This flag determines whether or not to return the entity state. This will be important since state can be potentially large and customers will definitely want a way to avoid paying the overhead of returning the state of all fetched entities (it may be very expensive in terms of CPU and network bandwidth). If not specified, we should assume `false` as the default value for `fetchState`.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -367,6 +373,46 @@ public async Task<HttpResponseMessage> HandleRequestAsync(HttpRequestMessage req
...
+                LastOperationTo = lastOperationTimeTo,
+                PageSize = pageSize,
+                ContinuationToken = continuationToken,
+            };
+
+            var context = await client.ListEntitiesAsync(query, CancellationToken.None);
+            var statusForAllInstances = context.Entities.ToList();
+            var nextContinuationToken = context.ContinuationToken;
+
+            var results = new List<StatusResponsePayload>(statusForAllInstances.Count);
```
cgillum: Instead of `results`, I suggest naming this `responseArrayPayload`.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
amdeel: My understanding was that you required an entityKey for that, that's why you pass in an EntityId, which throws an exception when you try to initialize one without an entityKey.

cgillum: It throws for `null` but not for the empty string.

test/Common/HttpApiHandlerTests.cs
```diff
@@ -913,6 +913,75 @@ public async Task GetEntity_Returns_State_Or_HTTP_404(bool hasKey, bool exists)
...
+
+            var responseMessage = await httpApiHandler.HandleRequestAsync(requestMessage);
+            Assert.Equal(HttpStatusCode.OK, responseMessage.StatusCode);
+            Assert.Equal("YYYY-YYYYYYYY-YYYYYYYYYYYY", responseMessage.Headers.GetValues("x-ms-continuation-token").FirstOrDefault());
+            var actual = JsonConvert.DeserializeObject<IList<DurableEntityStatus>>(await responseMessage.Content.ReadAsStringAsync());
+            clientMock.Verify(x => x.ListEntitiesAsync(It.IsAny<EntityQuery>(), It.IsAny<CancellationToken>()));
+            Assert.Equal("DoThis", actual[0].EntityId.EntityName);
+            Assert.Equal("one", actual[0].EntityId.EntityKey);
+            Assert.Equal("DoThat", actual[1].EntityId.EntityName);
+            Assert.Equal("02", actual[1].EntityId.EntityKey);
```
cgillum: Is this test passing? There are a couple things that I would expect wouldn't work, but maybe I'm misunderstanding:
* It's not clear to me why Line 976 would see an `x-ms-continuation-token` that returns `YYYY-...`.
* I would expect in the case where you're querying for a specific entity (`hasName = true`) that only one comes back. This test always seems to assert that both come back.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -3457,6 +3457,62 @@ public async Task GetStatus_WithCondition(bool extendedSessions, string storageP
             }
         }
 
+        [Theory(Skip = "Azure Storage fails due to container deletion")]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        [MemberData(nameof(TestDataGenerator.GetExtendedSessionAndFullFeaturedStorageProviderOptions), MemberType = typeof(TestDataGenerator))]
+        public async Task DurableEntity_GetStatus_WithCondition(bool extendedSessions, string storageProvider, bool fetchInput)
+        {
+            var taskHubName = "GetStatus1";
```
cgillum: Usually we name the task hub after the name of the test, so `nameof(DurableEntity_GetStatus_WithCondition)`. Is there a particular reason you didn't do that here?

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -3457,6 +3457,62 @@ public async Task GetStatus_WithCondition(bool extendedSessions, string storageP
...
+                if (fetchInput)
+                {
+                    Assert.NotNull(result.Entities.First());
+                }
+                else
+                {
+                    Assert.Null(result.Entities.First());
+                }
+
+                await host.StopAsync();
```
cgillum: I suggest also testing a few cases where none of the entities match the query (wrong EntityName, wrong time range, etc.).

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -521,25 +590,47 @@ private static IEnumerable<T> GetIEnumerableQueryParameterValue<T>(NameValueColl
...
             }
 
-            return results;
+            if (results.Any())
+            {
+                collection = results;
+                return true;
+            }
+
+            collection = null;
```
cgillum: I think this is actually an internal breaking change that could result in null-reference exceptions elsewhere. Previously this API never returned `null`, but now it can. I suggest you instead set `collection` to `Enumerable.Empty<T>()`.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -690,6 +781,11 @@ private static int GetIntQueryParameterValue(NameValueCollection queryStringName
         {
             IDurableEntityClient client = this.GetClient(request);
 
+            if (entityId.EntityKey.Equals("$"))
```
cgillum: Let's add a comment explaining why we're doing this - i.e. "$" means to return the default entity, which is the one with no entity key.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -322,43 +333,44 @@ public async Task<HttpResponseMessage> HandleRequestAsync(HttpRequestMessage req
...
-            var createdTimeFrom = GetDateTimeQueryParameterValue(queryNameValuePairs, CreatedTimeFromParameter, default(DateTime));
-            var createdTimeTo = GetDateTimeQueryParameterValue(queryNameValuePairs, CreatedTimeToParameter, default(DateTime));
-            var runtimeStatus = GetIEnumerableQueryParameterValue<OrchestrationRuntimeStatus>(queryNameValuePairs, RuntimeStatusParameter);
-            var pageSize = GetIntQueryParameterValue(queryNameValuePairs, PageSizeParameter);
 
-            var continuationToken = "";
-            if (request.Headers.TryGetValues("x-ms-continuation-token", out var headerValues))
+            var condition = new OrchestrationStatusQueryCondition();
+
+            if (TryGetDateTimeQueryParameterValue(queryNameValuePairs, CreatedTimeFromParameter, out DateTime createdTimeFrom))
```
ConnorMcMahon: In general, I feel the move from `Get*Value()` method with defaults to `TryGet*Value()` methods made the code more clunky, without gaining much. We should consider moving back to the old approach.

cgillum: I actually suggested that Amanda do it this way. Otherwise we're always overwriting the default values in the `OrchestrationStatusQueryCondition` object with some other defaulting behavio, which isn't ideal IMO. I see the `TryGet` approach as minimizing duplication.

test/Common/HttpApiHandlerTests.cs
amdeel: You're right about the YYYY thing, I left that in from another test accidentally.

This is just a unit test to make sure that ListEntitiesAsync is being called, DoThis/DoThat should always be returned.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityClient.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System.Threading;
```
cgillum: nit: `using` statements should be sorted.

src/WebJobs.Extensions.DurableTask/OrchestrationStatusQueryCondition.cs
```diff
@@ -40,5 +55,15 @@ public class OrchestrationStatusQueryCondition
...
+
+        /// <summary>
+        /// Return orchestration instances that have this instance id prefix.
+        /// </summary>
+        public string InstanceIdPrefix { get; set; }
+
+        /// <summary>
+        /// Determines whether the query will include the input of the orchestration.
+        /// </summary>
+        public bool FetchState { get; set; } = false;
```
cgillum: Since this is the orchestration query condition, this should be named `FetchInput`.

src/WebJobs.Extensions.DurableTask/EntityQuery.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// </summary>
+        public int PageSize { get; set; } = 100;
+
+        /// <summary>
+        /// ContinuationToken of the pager.
+        /// </summary>
+        public string ContinuationToken { get; set; }
+
+        /// <summary>
+        /// Determines whether the query will include the input of the entity.
```
cgillum: Should be "_state_ of the entity" rather than "_input_ of the entity" since entities don't have a concept of "inputs".

test/Common/HttpApiHandlerTests.cs
cgillum: > This is just a unit test to make sure that ListEntitiesAsync is being called, DoThis/DoThat should always be returned.

In that case, I suggest you remove the `hasName` and `entityName` variables from this test code. The URL you are generating does not match the behavior of your mocked HTTP API, which is confusing and may mislead people.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -406,7 +441,7 @@ public async Task<HttpResponseMessage> HandleRequestAsync(HttpRequestMessage req
 
             if (TryGetBooleanQueryParameterValue(queryNameValuePairs, FetchInputParameter, out bool fetchInput))
```
cgillum: This parameter name should be `fetchState` and not `fetchInput`. We should also update the local variable to reflect this.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -3652,10 +3652,18 @@ public async Task DurableEntity_ListEntitiesAsync_FetchInput(bool fetchInput, st
                 EntityName = "aaab",
                 LastOperationFrom = yesterday,
                 LastOperationTo = tomorrow,
-                FetchInput = fetchInput,
+                FetchState = fetchInput,
```
cgillum: Let's also rename this local variable to `fetchState`. I suggest we also update the name of the test method to be `DurableEntity_ListEntitiesAsync_FetchState`.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -3703,14 +3719,23 @@ public async Task DurableEntity_ListEntitiesAsync_NoResults(bool fetchInput, str
                 EntityName = "noResult",
                 LastOperationFrom = yesterday,
                 LastOperationTo = tomorrow,
+                FetchState = fetchInput,
```
cgillum: Let's rename `fetchInput` to `fetchState` here too.

test/Common/HttpApiHandlerTests.cs
```diff
@@ -919,65 +921,68 @@ public async Task GetEntity_Returns_State_Or_HTTP_404(bool hasKey, bool exists)
...
+                uriBuilder.Path += $"/entities/";
+            }
 
             var lastOperationTimeFrom = new DateTime(2018, 3, 10, 10, 1, 0, DateTimeKind.Utc);
             var lastOperationTimeTo = new DateTime(2018, 3, 10, 10, 23, 59, DateTimeKind.Utc);
 
-            var getStatusRequestUriBuilder = new UriBuilder(TestConstants.NotificationUrl);
-            getStatusRequestUriBuilder.Path += $"/Entities/";
-            getStatusRequestUriBuilder.Query = $"lastOperationTimeFrom={WebUtility.UrlEncode(lastOperationTimeFrom.ToString())}&lastOperationTimeTo={WebUtility.UrlEncode(lastOperationTimeTo.ToString())}&fetchInput=false&top=100";
+            uriBuilder.Query = $"lastOperationTimeFrom={WebUtility.UrlEncode(lastOperationTimeFrom.ToString())}&lastOperationTimeTo={WebUtility.UrlEncode(lastOperationTimeTo.ToString())}&fetchInput=false&top=100";
```
cgillum: Per my other feedback, `fetchInput=false` will need to be changed to `fetchState=false` here in the query string.

test/Common/HttpApiHandlerTests.cs
```diff
@@ -919,65 +921,68 @@ public async Task GetEntity_Returns_State_Or_HTTP_404(bool hasKey, bool exists)
...
             clientMock
                     .Setup(x => x.ListEntitiesAsync(It.IsAny<EntityQuery>(), It.IsAny<CancellationToken>()))
-                    .Returns(Task.FromResult(result));
+                    .Returns(Task.FromResult(mockResult));
 
-            var httpApiHandler = new ExtendedHttpApiHandler(clientMock.Object);
+            // Build Uri
+            var uriBuilder = new UriBuilder(TestConstants.NotificationUrl);
+
+            if (hasName)
```
cgillum: This is much better! Can we rename this from `hasName` to `useNameFilter`? That will make it clear that this is not about whether an entity has a name (they both do), but rather it's about whether we're using the entity name filter.

cgillum:
I started debugging the test and found at least one more issue. I'll submit an update to this PR shortly.

cgillum:
OK, this will take a bit longer than I thought. It looks like there may be a problem with the query generation. I'm not sure if the problem is in the extension code or in the DTFx code. I'll need to study the code a bit more.

PR #1100, opened 2019-12-11
Allow Azure Storage timers longer than 7 days for Functions V2

### LINKED ISSUE #14: Allow timers to last longer than 7 days

ConnorMcMahon:
In Functions V2, we now use version 9.x of the Azure Storage SDK, which
removed the maximum visibility delay on queues. We can now support
indefinite timers.

Resolves #14

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
@@ -120,11 +120,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         public override bool ValidateDelayTime(TimeSpan timespan, out string errorMessage)
         {
+#if FUNCTIONS_V1
             if (timespan > MaxTimerDuration)
             {
                 errorMessage = $"The Azure Storage provider supports a maximum of {MaxTimerDuration.TotalDays} days for time-based delays";
                 return false;
             }
+#endif
 
             return base.ValidateDelayTime(timespan, out errorMessage);
         }
--- a/test/Common/DurableTaskEndToEndTests.cs
+++ b/test/Common/DurableTaskEndToEndTests.cs
@@ -1083,6 +1083,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
             }
         }
 
+#if FUNCTIONS_V1
         /// <summary>
         /// End-to-end test which validates correct exceptions for invalid timeout values.
         /// </summary>
@@ -1110,6 +1111,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
                 await host.StopAsync();
             }
         }
+#endif
 
         /// <summary>
         /// End-to-end test which validates that orchestrations run concurrently of each other (up to 100 by default).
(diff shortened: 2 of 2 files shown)


src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
```diff
@@ -120,11 +120,13 @@ public async override Task<OrchestrationStatusQueryResult> GetOrchestrationState
 
         public override bool ValidateDelayTime(TimeSpan timespan, out string errorMessage)
```
cgillum: This code path gets called for everything that involves creating a durable timer, right?

ConnorMcMahon: Yes. The only other place this limitation was coded was in the default max retry interval in RetryOptions. Because that was a customer-facing value, and we don't have access to the durability provider at the time of creation, we had to choose a universal default for the constructor. That being said, this method is called on the various `TimeSpan` fields of the RetryOptions object when the object is passed into the `DurableOrchestrationContext` methods.

ConnorMcMahon:
Made old tests V1 only, still need to write V2 tests.

ConnorMcMahon:
I'm abandoning this PR, as it turns out that `initialVisibilityDelay` in the Storage SDK 9.x release is still stuck at no more than 6.23:59:59, meaning this enhancement was less straightforward than we initially hoped. This will likely need to be punted for the next release.

PR #1101, opened 2019-12-11
Added SerializerSettingsFactory to allow customers to provide their own JsonSerializerSettings

### LINKED ISSUE #65: Allow configuration of JsonSerializerSettings in DurableOrchestrationContext and DurableActivityContext
This will enable for complex entities to be passed around (for example: polymorphism/discriminator converter).
Thanks!

amdeel:
resolves #65


src/WebJobs.Extensions.DurableTask/EntityScheduler/RequestMessage.cs
```diff
@@ -12,6 +12,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
     /// </summary>
     internal class RequestMessage
     {
+        private MessagePayloadDataConverter dataConverter;
```
cgillum: Let's make this `readonly`.

src/WebJobs.Extensions.DurableTask/EntityScheduler/ResponseMessage.cs
```diff
@@ -9,6 +9,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 {
     internal class ResponseMessage
     {
+        private MessagePayloadDataConverter dataConverter;
```
cgillum: Let's make this `readonly`.

src/WebJobs.Extensions.DurableTask/MessagePayloadDataConverter.cs
```diff
@@ -13,31 +13,68 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
...
     {
-        // The default JsonDataConverter for DTFx includes type information in JSON objects. This causes issues
-        // because the type information generated from C# scripts cannot be understood by DTFx. For this reason, explicitly
-        // configure the JsonDataConverter to not include CLR type information. This is also safer from a security perspective.
-        internal static readonly JsonSerializerSettings MessageSettings = new JsonSerializerSettings
-        {
-            TypeNameHandling = TypeNameHandling.None,
-        };
-
         private static readonly JsonSerializerSettings ErrorSettings = new JsonSerializerSettings
```
cgillum: Since we only use this when the customer doesn't supply their own, let's only initialize it inside of the new `ErrorConverter` property you added below.

src/WebJobs.Extensions.DurableTask/MessagePayloadDataConverter.cs
```diff
@@ -13,31 +13,68 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
...
+            this.MessageSettings = serializerSettingsFactory.CreateJsonSerializerSettings();
+        }
 
-        public MessagePayloadDataConverter(JsonSerializerSettings settings)
+        private MessagePayloadDataConverter(JsonSerializerSettings settings)
             : base(settings)
         {
         }
 
+        internal JsonSerializerSettings MessageSettings { get; private set; }
```
cgillum: Can we remove the `private set` here?

src/WebJobs.Extensions.DurableTask/MessagePayloadDataConverter.cs
```diff
@@ -13,31 +13,68 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
...
 
-        // Default singleton instances
-        public static readonly MessagePayloadDataConverter Default = new MessagePayloadDataConverter(MessageSettings);
-        public static readonly MessagePayloadDataConverter ErrorConverter = new MessagePayloadDataConverter(ErrorSettings);
-        public static readonly JsonSerializer DefaultSerializer = JsonSerializer.Create(MessageSettings);
+        private MessagePayloadDataConverter messageConverter;
+        private MessagePayloadDataConverter errorConverter;
+        private JsonSerializer messageSerializer;
+
+        public MessagePayloadDataConverter(ISerializerSettingsFactory serializerSettingsFactory)
```
cgillum: The other constructor called into the base constructor and passed in the appropriate settings. Don't we need to do that here as well? Otherwise it looks like it will use some other settings that it generates itself.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -89,15 +89,17 @@ public DurableTaskExtension()
...
         public DurableTaskExtension(
             IOptions<DurableTaskOptions> options,
             ILoggerFactory loggerFactory,
             INameResolver nameResolver,
             IDurabilityProviderFactory orchestrationServiceFactory,
             IDurableHttpMessageHandlerFactory durableHttpMessageHandlerFactory = null,
             ILifeCycleNotificationHelper lifeCycleNotificationHelper = null,
-            ISerializerSettingsFactory serializerSettingsFactory = null)
+            IMessageSerializerSettingsFactory messageSerializerSettingsFactory = null,
+            IErrorSerializerSettingsFactory errorSerializerSettingsFactory = null)
```
cgillum: This works, but had you considered an alternate design where we have a single interface that defines two methods instead of having two interfaces that each define a single method?

amdeel: Yeah, I considered that. But if a customer is implementing the interface they'll need to implement both methods and I wanted to avoid that.

src/WebJobs.Extensions.DurableTask/EntityScheduler/RequestMessage.cs
```diff
@@ -81,7 +88,7 @@ public void SetInput(object obj)
                 }
                 else
                 {
-                    this.Input = MessagePayloadDataConverter.Default.Serialize(obj);
+                    this.Input = this.dataConverter.MessageConverter.Serialize(obj);
                 }
```
cgillum: Shouldn't we be using this data converter in the two `GetInput` methods below? @sebastianburckhardt

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -3720,6 +3721,46 @@ public async Task TaskHubName_DefaultNameNonProductionSlot_ThrowsException()
...
+        public async Task CustomISerializerSettingsFactory()
+        {
+            string[] orchestratorFunctionNames =
+            {
+                nameof(this.CustomISerializerSettingsFactory),
+            };
+
+            using (var host = TestHelpers.GetJobHost(
+                this.loggerProvider,
+                nameof(this.ReplaySafeLogger_LogsOnlyOnce),
```
cgillum:
```suggestion
                nameof(this.CustomISerializerSettingsFactory),
```

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -3720,6 +3721,46 @@ public async Task TaskHubName_DefaultNameNonProductionSlot_ThrowsException()
...
+                };
+
+                var client = await host.StartOrchestratorAsync(orchestratorFunctionNames[0], inputWithEnum, this.output);
+                await client.WaitForCompletionAsync(this.output);
+                var status = client.GetStatusAsync();
+
+                Assert.NotNull(status);
+
+                var expectedResult = "Value2";
+                Assert.Contains(expectedResult, status.Result.Output.ToString());
```
cgillum: Just to make sure this is working the way we expect, I would suggest we run two orchestrations: one without your customizations and one with your customizations. That way we won't get a false positive if we accidentally change the default behavior for enum serialization in the future (which we don't want to do since that is a breaking change).

test/Common/TestOrchestrations.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.Azure.WebJobs.Host.TestCommon;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
+using static Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests.DurableTaskEndToEndTests;
```
cgillum: Not used?

src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask-net471.xml
```diff
@@ -508,10 +508,23 @@
             <param name="timeout">Total allowed timeout for output from the durable function. The default value is 10 seconds.</param>
             <returns>An HTTP response which may include a 202 and location header or a 200 with the durable function output in the response body.</returns>
         </member>
+<<<<<<< HEAD
```
cgillum: These merge conflicts need to be resolved

src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask.xml
```diff
@@ -475,11 +475,24 @@
             <param name="operationName">The name of the operation.</param>
             <returns>A task representing the completion of the operation on the entity.</returns>
         </member>
+<<<<<<< HEAD:src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask-netstandard2.0.xml
```
cgillum: These merge conflicts also need to be resolved.

src/WebJobs.Extensions.DurableTask/EntityScheduler/RequestMessage.cs
sebastianburckhardt: Yes, I would think so also. Not sure why we are not.

src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask-net471.xml
ConnorMcMahon: @amdeel, the easiest way to resolve these is to just rebuild the project once all of your *.cs merge conflicts are resolved.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
ConnorMcMahon: We could use one interface assuming we did the following:

```csharp
public interface ISerializerSettingsFactory {
    JsonSerializerSettings GetMessageSerializerSettings();
    JsonSerializerSettings GetErrorSerializerSettings();
}
```

```csharp
public class DefaultSerializerSettingsFactory {
    public virtual JsonSerializerSettings GetMessageSerializerSettings() {
         // Default implementation here.
    }
    public virtual JsonSerializerSettings GetErrorSerializerSettings() {
         // Default implementation here.
    }
}
```

Then customers could just make a subclass of DefaultSerializerSettingsFactory, and only override the method they want to.

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -18,10 +18,12 @@ internal class TaskOrchestrationShim : TaskCommonShim
     {
         private readonly DurableOrchestrationContext context;
         private readonly OutOfProcOrchestrationShim outOfProcShim;
+        private readonly DurableTaskExtension config;
```
ConnorMcMahon: Instead of passing in the whole extension config, can you just pass in the dataconverter?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
amdeel: Good idea, I hadn't thought of that option!

I'm going to leave the current implementation for consistency reasons though, unless anyone objects.

test/Common/TestOrchestrations.cs
amdeel: I added an orchestration that uses the ComplexType that is defined in that file. Should I pull it out into its own file?

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -3937,9 +3937,45 @@ public async Task CustomISerializerSettingsFactory()
...
+                nameof(TestOrchestrations.ComplexTypeOrchestrator),
+            };
 
-                var expectedResult = "Value2";
-                Assert.Contains(expectedResult, status.Result.Output.ToString());
+            using (var host = TestHelpers.GetJobHost(
+                this.loggerProvider,
+                nameof(this.DefaultIMessageSerializerSettingsFactory),
+                true,
+                serializerSettings: new DefaultEnumSettings()))
```
cgillum: I think this should be `null` instead of using a new settings object. If we accidentally changed the default, then this test would not catch it since default settings are duplicated in `DefaultEnumSettings`.

amdeel: "Default..." and "Custom..." are meant to be opposites of each other, so we dont have to rely on the default settings to test for custom settings. My "Default..." test doesn't actually use the default settings, it specifies its own. It just happens to match the current default settings.

cgillum: > My "Default..." test doesn't actually use the default settings, it specifies its own. It just happens to match the current default settings.

Right, this is exactly my concern. I don't think there is value in testing something that _matches_ the default settings. Instead I think we want to test the real default settings.

amdeel: I think there is value in having more than one test that tests different/opposite custom settings to be sure that the default settings didn't change to the same settings in my custom test. But I can add a real default test as well.

src/WebJobs.Extensions.DurableTask/EntityScheduler/RequestMessage.cs
```diff
@@ -12,13 +12,25 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
     /// </summary>
     internal class RequestMessage
     {
-        private readonly MessagePayloadDataConverter dataConverter;
+        // This no-arg constructor is used when a RequestMessage is deserialized.
+        public RequestMessage() { }
```
cgillum: I was looking at this issue as well. I have an alternate fix that I can create as a PR if you're interested. It might be a bit simpler.

pb-qs:
Can you provide a correct example of how to set the Json Serializer settings now please?

I have looked at how your test does it and come up with the following which appears to work, but I worry that it's hacky and at risk of not working when updates are shipped?

````
class CustomMessageSerializer : IMessageSerializerSettingsFactory
{
    public JsonSerializerSettings CreateJsonSerializerSettings() => new JsonSerializerSettings
    {
        TypeNameHandling = TypeNameHandling.None,
        Converters = new[] { new StringEnumConverter() }
    };
}
````

Then in Main() of my WebJob (netcore 3.1) 
````
var builder = new HostBuilder()
    .ConfigureServices(services => 
        services.AddSingleton<IMessageSerializerSettingsFactory, CustomMessageSerializer>());
````

Issue #1117, opened 2019-12-18
Issues with list entities API


cgillum:
### Description
There are a few issues with the newly introduced ListEntitiesAsync API found during manual testing.

* Specifying no parameters results in an internal `ArgumentException` because of a DateTime parsing issue in DurableTask.AzureStorage.
* Listing entities without an entity name filter attempts to also list orchestrations, which causes `ArgumentOutOfRangeException` when trying to parse orchestration instance IDs.
* The returned state data is always a string, regardless of the actual JSON format.

I'll add more note here as I continue my testing. We'll need to get this fixed before the v2.1.0 release can go out.

FYI @amdeel

PR #1119, opened 2019-12-19
Last minute updates for v2.1.0 release

cgillum:
- Resolves https://github.com/Azure/azure-functions-durable-extension/issues/1117
- Updated samples to point to v2.1.0

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
+++ b/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
@@ -337,42 +337,44 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             IDurableOrchestrationClient client = this.GetClient(request);
             var queryNameValuePairs = request.GetQueryNameValuePairs();
-            var createdTimeFrom = GetDateTimeQueryParameterValue(queryNameValuePairs, CreatedTimeFromParameter, default(DateTime));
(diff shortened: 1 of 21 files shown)


src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -340,10 +340,14 @@ public async Task<HttpResponseMessage> HandleRequestAsync(HttpRequestMessage req
             var createdTimeFrom = GetDateTimeQueryParameterValue(queryNameValuePairs, CreatedTimeFromParameter, default(DateTime));
             var createdTimeTo = GetDateTimeQueryParameterValue(queryNameValuePairs, CreatedTimeToParameter, default(DateTime));
             var runtimeStatus = GetIEnumerableQueryParameterValue<OrchestrationRuntimeStatus>(queryNameValuePairs, RuntimeStatusParameter);
-            var pageSize = GetIntQueryParameterValue(queryNameValuePairs, PageSizeParameter);
+
+            if (!TryGetIntQueryParameterValue(queryNameValuePairs, PageSizeParameter, out int pageSize))
+            {
+                pageSize = 100;
```
amdeel: This is the reason your tests are failing. The previous behavior assigned this to 0. We can change the default to 100, but if you notice below, it's calling a different GetStatusAsync. We'll have to fix the unit tests. I intentionally left this change out because I didn't want to hold up the release.

cgillum: Yep you are right. My next commit fixed one of the tests to do the right thing. The other tests still need to be modified, however.

PR #1120, opened 2019-12-19
Fixed GetStatusAsync calls in HttpApiHandlerTests

amdeel:


src/WebJobs.Extensions.DurableTask/OrchestrationStatusQueryCondition.cs
```diff
@@ -72,6 +72,6 @@ internal OrchestrationStatusQueryCondition(EntityQuery entityQuery)
         /// <summary>
         /// Determines whether the query will include the input of the orchestration.
         /// </summary>
-        public bool FetchInput { get; set; } = false;
+        public bool ShowInput { get; set; } = true;
```
cgillum: From a code perspective, `FetchInput` is more accurate. `Show` implies that it's there but we're just hiding it. `Fetch` indicates that we're not even attempting to load it from storage.  Also, since this is a public API, isn't this a breaking change?

amdeel: I used show because that's consistent with the showInput query string parameter. This is not a breaking change because I added this property in my ListEntitiesAsync PR.

amdeel: And true is the default, which is consistent with the behavior before this property was added.

cgillum: Ah, right, I forgot that the `FetchInput` property never existed before. The point about consistency with the query string parameter is well-taken.

PR #1121, opened 2019-12-19
Update azure-pipelines.yml to run on all branches

amdeel:
Previously, tests did not run on PRs that use a base branch other than dev. This fix allows tests to run on all PRs on this repository. There will be no change in behavior for outside contributors.


amdeel:
@cgillum Exactly!

PR #1125, opened 2019-12-26
Durable Client Factory (DurableOrchestrationClient outside Azure Functions )

davidrevoledo:
Ref  : https://github.com/Azure/azure-functions-durable-extension/issues/161

This is intended to init azure durable functions outside of an azure function orchestrators host.
It could have extensive applications to init durable functions from different applications event from other azure functions applications.

Usability is straight forward for a client. Let's take an Asp.Net Core application.

To configure the application just call AddDurableTask, this will use service collection instead of webjob extension builder. 
```C#
        public void ConfigureServices(IServiceCollection services)
        {
            services.AddMvc().SetCompatibilityVersion(CompatibilityVersion.Version_2_2);
            services.AddDurableTask();
        }
```

To create the client just inject `IDurableClientFactory` under the hood this will create a client using the same approach Orchestrator host is doing to avoid breaking changes, we could configure hub name or the connection string for the storage account
```C#
private readonly IDurableClientFactory _clientFactory;

        public SampleController(IDurableClientFactory clientFactory)
        {
            _clientFactory = clientFactory;
        }

        [HttpGet]
        public async Task<IActionResult> Get()
        {
            var client = _clientFactory.CreateClient(new DurableClientOptions
            {
                IsExternalClient = true
            });

            var instanceId = await client.StartNewAsync("Chaining");
            return Ok(instanceId);
        }
```

The implementation on the orchestrator host is exactly the same without any client needed to start the flow.
```C#
  [FunctionName("Chaining")]
        public static async Task<object> Chaining(
        {
            var x = await context.CallActivityAsync<object>("F1", null);
            var y = await context.CallActivityAsync<object>("F2", x);
            var z = await context.CallActivityAsync<object>("F3", y);
            return await context.CallActivityAsync<object>("F4", z);
        }
```

Let me know your thoughts and will complete a few unit tests required for the external client flag to avoid the validation of know function names. 
https://github.com/Azure/azure-functions-durable-extension/pull/1125/files#diff-b4be2ec3bf4e13899ce8445628aa091aR109 

Any suggestion to control that validation is welcome.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -38,23 +38,35 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         private readonly DurableTaskExtension config;
         private readonly DurableClientAttribute attribute; // for rehydrating a Client after a webhook
         private readonly MessagePayloadDataConverter messageDataConverter;
+        private readonly DurableTaskOptions durableTaskOptions;
+
+        internal DurableClient(
+            DurabilityProvider serviceClient,
+            HttpApiHandler httpHandler,
+            DurableClientAttribute attribute,
+            MessagePayloadDataConverter messageDataConverter,
+            EndToEndTraceHelper traceHelper,
+            DurableTaskOptions durableTaskOptions)
+        {
+            this.messageDataConverter = messageDataConverter;
+
+            this.client = new TaskHubClient(serviceClient, this.messageDataConverter);
+            this.durabilityProvider = serviceClient;
+            this.traceHelper = traceHelper;
+            this.httpApiHandler = httpHandler;
+            this.durableTaskOptions = durableTaskOptions;
+            this.hubName = attribute.TaskHub ?? this.durableTaskOptions.HubName;
+            this.attribute = attribute;
+        }
 
         internal DurableClient(
             DurabilityProvider serviceClient,
             DurableTaskExtension config,
             HttpApiHandler httpHandler,
             DurableClientAttribute attribute)
+            : this(serviceClient, httpHandler, attribute, config.MessageDataConverter, config.TraceHelper, config.Options)
         {
-            this.config = config ?? throw new ArgumentNullException(nameof(config));
-
-            this.messageDataConverter = config.MessageDataConverter;
-
-            this.client = new TaskHubClient(serviceClient, this.messageDataConverter);
-            this.durabilityProvider = serviceClient;
-            this.traceHelper = config.TraceHelper;
-            this.httpApiHandler = httpHandler;
-            this.hubName = attribute.TaskHub ?? config.Options.HubName;
-            this.attribute = attribute;
+            this.config = config;
         }
 
         public string TaskHubName => this.hubName;
@@ -110,7 +122,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// <inheritdoc />
         async Task<string> IDurableOrchestrationClient.StartNewAsync<T>(string orchestratorFunctionName, string instanceId, T input)
         {
-            if (this.ClientReferencesCurrentApp(this))
+            if (!this.attribute.ExternalClient && this.ClientReferencesCurrentApp(this))
             {
                 this.config.ThrowIfFunctionDoesNotExist(orchestratorFunctionName, FunctionType.Orchestrator);
             }
@@ -151,7 +163,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private OrchestrationStatus[] GetStatusesNotToOverride()
         {
-            var overridableStates = this.config.Options.OverridableExistingInstanceStates;
+            var overridableStates = this.durableTaskOptions.OverridableExistingInstanceStates;
             if (overridableStates == OverridableStates.NonRunningStates)
             {
                 return new OrchestrationStatus[]
@@ -317,7 +329,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private bool TaskHubMatchesCurrentApp(DurableClient client)
         {
-            var taskHubName = this.config.Options.HubName;
+            var taskHubName = this.durableTaskOptions.HubName;
             return client.TaskHubName.Equals(taskHubName);
         }
 
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClientFactory.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClientFactory.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using Microsoft.Azure.WebJobs.Extensions.DurableTask.Options;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.ContextImplementations
 {
@@ -11,14 +12,17 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.ContextImplementations
     public class DurableClientFactory : IDurableClientFactory
     {
         private readonly DurableTaskExtension durableTaskConfig;
+        private readonly DurableClientOptions defaultDurableClientOptions;
 
         /// <summary>
         ///     Initializes a new instance of the <see cref="DurableClientFactory"/> class.
         /// </summary>
         /// <param name="durableTaskConfig">Configuration for the Durable Functions extension.</param>
-        public DurableClientFactory(DurableTaskExtension durableTaskConfig)
+        /// <param name="defaultDurableClientOptions">Default Options to Build Durable Clients.</param>
+        public DurableClientFactory(DurableTaskExtension durableTaskConfig, IOptions<DurableClientOptions> defaultDurableClientOptions)
         {
             this.durableTaskConfig = durableTaskConfig;
+            this.defaultDurableClientOptions = defaultDurableClientOptions.Value;
         }
 
         /// <summary>
@@ -34,6 +38,15 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.ContextImplementations
             }
 
             return this.durableTaskConfig.GetClient(new DurableClientAttribute(durableClientOptions));
+        }
+
+        /// <summary>
+        /// Gets a <see cref="IDurableClient"/> using configuration from a <see cref="DurableClientOptions"/> instance.
+        /// </summary>
+        /// <returns>Returns a <see cref="IDurableClient"/> instance. The returned instance may be a cached instance.</returns>
+        public IDurableClient CreateClient()
+        {
+            return this.CreateClient(this.defaultDurableClientOptions);
         }
     }
 }
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClientFactory.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClientFactory.cs
@@ -1,0 +1,39 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.Azure.WebJobs.Extensions.DurableTask.Options;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.ContextImplementations
+{
+    /// <summary>
+    ///     Factory class to create Durable Client to start works outside an azure function context.
+    /// </summary>
+    public class DurableClientFactory : IDurableClientFactory
+    {
+        private readonly DurableTaskExtension durableTaskConfig;
+
+        /// <summary>
+        ///     Initializes a new instance of the <see cref="DurableClientFactory"/> class.
+        /// </summary>
+        /// <param name="durableTaskConfig">Configuration for the Durable Functions extension.</param>
+        public DurableClientFactory(DurableTaskExtension durableTaskConfig)
+        {
+            this.durableTaskConfig = durableTaskConfig;
+        }
+
+        /// <summary>
+        /// Gets a <see cref="IDurableClient"/> using configuration from a <see cref="DurableClientOptions"/> instance.
+        /// </summary>
+        /// <param name="durableClientOptions">options containing the client configuration parameters.</param>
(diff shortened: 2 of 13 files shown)


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClientFactory.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            this.durableTaskConfig = durableTaskConfig;
+        }
+
+        /// <summary>
+        /// Gets a <see cref="IDurableClient"/> using configuration from a <see cref="DurableClientOptions"/> instance.
+        /// </summary>
+        /// <param name="durableClientOptions">options containing the client configuration parameters.</param>
+        /// <returns>Returns a <see cref="IDurableClient"/> instance. The returned instance may be a cached instance.</returns>
+        public IDurableClient CreateClient(DurableClientOptions durableClientOptions)
```
davidrevoledo: It also could be CreateExternalClient and remove the Boolean property and set always true in the attribute to avoid this overhead, I thought it also can be used inside an azure function somewhere

cgillum:
I like this idea of having a first-class way to initialize an `IDurableClient` from an ASP.NET Core web app! I would love to get thoughts from @ConnorMcMahon and @anthonychu on this PR. Personally I would try to simplify a few things and fix a few namespaces, but the high-level approach seems reasonable to me.

It might be useful to see a few more examples, such as the ASP.NET Core setup required to use a non-default task hub name and Azure Storage account.

davidrevoledo:
Fair enough ! @cgillum  will close it.

anthonychu:
This is great. Thanks!

I wonder if we should publish this as a separate package that references `Microsoft.Azure.WebJobs.Extensions.DurableTask` and move the ASP.NET specific stuff like `public static IServiceCollection AddDurableTask(this IServiceCollection serviceCollection)` to it.

davidrevoledo:
@cgillum ok will do also I think it might be a good idea allow configure IOption<DurableClientOption> to directly configure it in the appsetting.json.

What are you namespace suggestions to clean up the code ? I've only created the factory under `ContextImplementations` namespace and configuration provider at the same level the webjob configuration provider is, but I can do both changes and create a few mores samples as well.

 - Asp.Net Application using non default connections.
 - Console App starting a durable function.
 - Azure Function starting a durable function hosted in a separated Azure Function Instance.

davidrevoledo:
@anthonychu great thanks ! :) 

 you mean something like `Microsoft.Extensions.DependencyInjection.DurableExtensions` ?

anthonychu:
Maybe. I haven’t thought this far ahead 🙂 but if we head down this path, we should figure out the right package name to publish under.

davidrevoledo:
Sorry guys I was out for the holiday season of new year, will address this during the week to get a new version more friendly with @cgillum suggestions.

davidrevoledo:
@cgillum @anthonychu 

I've tried to use a connectionstring instead for configuration but to do that we should changes several providers so I still think the best way is keep configure the connection name and read it from configurations.

With changes I made it's easy to configure in the asp.net app and we can call create client w/o params as well have the flexibility to call different clients if any need to reach different durable functions.


```C#
 public class Startup
    {
        public Startup(IConfiguration configuration)
        {
            Configuration = configuration;
        }

        public IConfiguration Configuration { get; }

        public void ConfigureServices(IServiceCollection services)
        {
            services.AddMvc().SetCompatibilityVersion(CompatibilityVersion.Version_2_2);

            services
                .AddDurableTask(options =>
                {
                    options.ConnectionName = "Default";
                    options.TaskHub = "name";
                });
        }

        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
        {
            if (env.IsDevelopment())
            {
                app.UseDeveloperExceptionPage();
            }
            else
            {
                app.UseHsts();
            }

            app.UseHttpsRedirection();
            app.UseMvc();
        }
    }
``` 

then to use is extremely straight forward.
``` C#
 [Route("api/[controller]")]
    [ApiController]
    public class ValuesController : ControllerBase
    {
        private readonly IDurableClientFactory _clientFactory;

        public ValuesController(IDurableClientFactory clientFactory)
        {
            _clientFactory = clientFactory;
        }


        [HttpGet]
        public async Task<IActionResult> Get()
        {
            var client = _clientFactory.CreateClient();
            var instanceId = await client.StartNewAsync("Chaining");
            return Ok(instanceId);
        }
    }
``` 
and if any one want to reach a specific azure durable function who is not the default configured in the startup (it will be only for complicated and long structures) the code have the flexibility to call like this.

``` C#
 [HttpGet]
        public async Task<IActionResult> Get()
        {
            var client = _clientFactory.CreateClient(new DurableClientOptions
            {
                ConnectionName = "InvoicingAppStorage",
                TaskHub = "InvoicingHub"
            });
            var instanceId = await client.StartNewAsync("Chaining");
            return Ok(instanceId);
        }
```

Finally I agree with @anthonychu to move:
 - AddDurableTask
 - StandardConnectionStringProvider (we can rename it if so)
 - DurableClientFactory

outside of this project and create some like `Microsoft.Extensions.DependencyInjection.DurableExtensions` to be clear that it works using Microsoft DependencyInjection it's because is dangerous using this and manually using the same package inside an azure functions, if so the connection string resolver will be `WebJobsConnectionStringProvider` and this won't work as excepted, however if we move out this implementation to a different package is clear enough how to use in a console app, asp.net core app or even in an azure function that it'll use Configurations to read the connection.

Let me know your thoughts, I'm very open to do any changes you think it will be necessary and looking forward to collaborate more often in the future :).

davidrevoledo:
any thoughts ?

ConnorMcMahon:
@davidrevoledo 

Sorry for the delay here. I am going to try to review this in the next day or two.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -106,7 +106,10 @@ async Task<IActionResult> IDurableOrchestrationClient.WaitForCompletionOrCreateC
         /// <inheritdoc />
         async Task<string> IDurableOrchestrationClient.StartNewAsync<T>(string orchestratorFunctionName, string instanceId, T input)
         {
-            this.config.ThrowIfFunctionDoesNotExist(orchestratorFunctionName, FunctionType.Orchestrator);
+            if (!this.attribute.ExternalClient)
```
ConnorMcMahon: We will have to do something similar in `SignalEntityAsyncInternal`.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -106,7 +106,10 @@ async Task<IActionResult> IDurableOrchestrationClient.WaitForCompletionOrCreateC
...
+            {
+                this.config.ThrowIfFunctionDoesNotExist(orchestratorFunctionName, FunctionType.Orchestrator);
```
ConnorMcMahon: @cgillum, is there any reason for us to have this check at all? A `DurableClient` can be created with a different task hub already, and there is no guarantee that the app that is calling this `DurableClient` will have the same functions as the app with the TaskHub the client is referencing.

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
```diff
@@ -40,10 +42,48 @@ public static IWebJobsBuilder AddDurableTask(this IWebJobsBuilder builder)
             serviceCollection.TryAddSingleton<IDurabilityProviderFactory, AzureStorageDurabilityProviderFactory>();
             serviceCollection.TryAddSingleton<IMessageSerializerSettingsFactory, MessageSerializerSettingsFactory>();
             serviceCollection.TryAddSingleton<IErrorSerializerSettingsFactory, ErrorSerializerSettingsFactory>();
+            serviceCollection.TryAddSingleton<IDurableClientFactory, DurableClientFactory>();
```
ConnorMcMahon: I don't think we are currently using `IDurableClientFactory` in the Durable Functions extension, so we shouldn't add it to this extension method.

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
```diff
@@ -40,10 +42,48 @@ public static IWebJobsBuilder AddDurableTask(this IWebJobsBuilder builder)
...
+            return serviceCollection;
+        }
+
+        /// <summary>
+        /// Adds the Durable Task extension to the provided <see cref="IServiceCollection"/>.
+        /// </summary>
+        /// <param name="serviceCollection">The <see cref="IServiceCollection"/> to configure.</param>
+        /// <param name="optionsBuilder">Populate default configurations of <see cref="DurableClientOptions"/> to create Durable Clients.</param>
+        /// <returns>Returns the provided <see cref="IServiceCollection"/>.</returns>
+        public static IServiceCollection AddDurableTask(this IServiceCollection serviceCollection, Action<DurableClientOptions> optionsBuilder)
```
ConnorMcMahon: I think this method should be `AddDurableTaskFactory()`, and honestly might be better served in it's own namespace/package.

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
```diff
@@ -40,10 +42,48 @@ public static IWebJobsBuilder AddDurableTask(this IWebJobsBuilder builder)
...
+        /// <param name="serviceCollection">The <see cref="IServiceCollection"/> to configure.</param>
+        /// <returns>Returns the provided <see cref="IServiceCollection"/>.</returns>
+        public static IServiceCollection AddDurableTask(this IServiceCollection serviceCollection)
+        {
+            if (serviceCollection == null)
+            {
+                throw new ArgumentNullException(nameof(serviceCollection));
+            }
+
+            serviceCollection.TryAddSingleton<DurableTaskExtension>();
```
ConnorMcMahon: We should probably only add the services needed for `DurableClientFactory`

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClientFactory.cs
```diff
@@ -0,0 +1,52 @@
...
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.ContextImplementations
+{
+    /// <summary>
+    ///     Factory class to create Durable Client to start works outside an azure function context.
+    /// </summary>
+    public class DurableClientFactory : IDurableClientFactory
+    {
+        private readonly DurableTaskExtension durableTaskConfig;
```
ConnorMcMahon: Pulling in `DurableTaskExtension` adds in a whole bevy of things that probably aren't needed to create clients (including starting a `TaskHubWorker`, which will start actively pulling orchestration and activity items off of the task hub it is listening for).

Instead, consider refactoring `DurableClient` to take only the things it needs from `DurableTaskExtension` (which is largely just the `GetClient()` method).

davidrevoledo:
@ConnorMcMahon  thanks for your feedback I'll do the proper changes do you want I create the separete project to put the IOC and ClientFactory as suggested ? any recommendation for naming it ?

cgillum:
Personally I'd prefer to _not_ create a separate package. I think the reason to have a separate package is if that package contained additional dependencies that don't make sense in the core Microsoft.Azure.WebJobs.Extensions.DurableTask. It doesn't look like you're adding any new dependencies, so I vote that we keep this all in the same package.

cgillum:
Hey @davidrevoledo were you still interested in completing this PR? I was hoping to include this in our v2.3.0 release, but we're wrapping that up now and I see there are some merge conflicts that need to be resolved. I'd be happy to include this in our v2.4.0 release if you're up to finishing it.

davidrevoledo:
sure thing @cgillum , I've been so busy sorry to forget this PR, I've resolved the conflict but let me check if I've changed all the mentioned points in here.

One thing to let you know is that I needed to call durable functions from a asp.net core api so I've been using this for a while, will be great to delete the internal code and having from the library itself :D

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
davidrevoledo: @ConnorMcMahon this will allow users to start an external call, do you suggest they having to add them manually?

davidrevoledo: The idea was to be able to start the client from anywhere including an azure durable functions, some one may need to start an orchestration in another hub or another azure function and this will be the same usage as it would be called from a console app or an asp.net core app. what do you think ?

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
ConnorMcMahon: As discussed elsewhere in this review, we will keep it in this namespace/package.

There are questions to be brought up about other function apps that don't want to use Durable except for the external client and how we support that, as the functions runtime will automatically start up an instance of the extension, but that can be addressed later.

@bachuv, we may still want to rename this before the v2.4.0 release.

ConnorMcMahon:
@davidrevoledo So sorry for all of the delays getting this work in. We are going to merge in now for our v2.4.0 release, and I know that we will have a lot of very excited customers to use this. We will make sure to give you credit for these new features in our next release.

@bachuv is going to help us drive this to production, with some samples of usage in a console app as well as some tests and validation. Let us know if you are interested in being involved as a reviewer in those PRs!

davidrevoledo:
Not a problem ! @ConnorMcMahon  honoured to contribute, is part of my commitment as an Azure MVP 
of will be ready in case something needs to be addressed @bachuv I was using it for too long (embedded in a project code)  however it was the old implementation.

Ready here for anything you might need !

olitomlinson:
Will this also allow me to inject `IDurableEntity` ?

ConnorMcMahon:
@olitomlinson, to clarify, do you mean `IDurableEntityClient`? 

This PR does not explicitly support that, but the good news is that the `IDurableClient` is by definition also a `IDurableEntityClient` by definition, so you can just cast it as soon as you get the object returned.

olitomlinson:
Sorry yes, I do mean ‘IDurableEntityClient’ - 

I’m having to use the Entity HTTP API from ASP.NET project to signal and read entity state which isn’t terribly easy, so being able to use the durable entity client will be brilliant!

Issue #1123, opened 2019-12-21
Provide better error message when customers use invalid storage accounts


cgillum:
### Description
A customer tried to use a premium storage account with their Durable Function and got the following exception:

> Microsoft.WindowsAzure.Storage.StorageException : No such host is known ---> System.Net.Http.HttpRequestException : No such host is known ---> System.Net.Sockets.SocketException : No such host is known

In this case, they were doing something unsupported but we should have an error message that better helps them identify the root cause.  For example, an error message that includes the text "Make sure you are using a storage account type that supports tables, queues, and blobs".

The better error message would need to be added to the DurableTask.AzureStorage dependency.

PR #1126, opened 2019-12-26
Validate Storage Exception

davidrevoledo:
Ref: https://github.com/Azure/azure-functions-durable-extension/issues/1123 

It could be a validation in `DurableTask.AzureStorage` but we could also catch a problem here there is no more info we can get from the exception at this level so maybe other error message could be more accurate. 

If this doesn't make any sense here feel free to close it. :) 
Just trying to improve the error we are returning here when a premium account is being used.


cgillum:
While I appreciate the effort that goes into creating a PR, I'm not really comfortable with this approach. One thing I'm uncomfortable with is the try/catch around context.GetWebHookHandler() because one wouldn't expect this API to reach into Azure Storage for this first time (even if it does, it seems likely that this could change over time since there isn't a logical connection between these two things). As you suggested, I think it would be better to make a change in DurableTask.AzureStorage.

davidrevoledo:
Fair enough @cgillum  will close it.

PR #1129, opened 2019-12-30
Analyzer: Fixed TimerCodeFixProvider

amdeel:


src/WebJobs.Extensions.DurableTask.Analyzers/CodefixProviders/Orchestrator/TimerCodeFixProvider.cs
```diff
@@ -118,50 +135,82 @@ private bool TryGetCancellationTokenParameter(SyntaxNode expression, SemanticMod
...
+            if (nodes.FirstOrDefault() != null)
+            {
+                return true;
             }
 
-            typeReference = null;
             return false;
         }
 
+        private bool TryGetTypeName(SemanticModel semanticModel, SyntaxNode identifier, out string typeName)
```
cgillum: It looks like these private methods should all be `static`.

src/WebJobs.Extensions.DurableTask.Analyzers/CodefixProviders/Orchestrator/TimerCodeFixProvider.cs
```diff
@@ -118,50 +135,82 @@ private bool TryGetCancellationTokenParameter(SyntaxNode expression, SemanticMod
...
+            typeName = null;
+            return false;
+        }
+        
+        private string GetTypeWithoutNamespace(string type)
+        {
+            var splitString = type.Split('.');
+            var length = splitString.Length;
+            return splitString[length - 1];
+        }
```
cgillum: This works, but it's more efficient to use `type.LastIndexOf('.')` instead of `type.Split('.')`. String splitting is generally more expensive because it allocates multiple string objects in memory. I've even seen cases where this has resulted in massive performance problems because doing it too frequently can cause the .NET Garbage Collector to get really busy. That's probably not going to be an issue here, but preferring `string.IndexOf` and `string.LastIndexOf` instead of `string.Split` is a good habit to get into.

Issue #472, opened 2018-10-12
ObjectDisposedException when function host shuts down


cgillum:
Exception and call stack:

```
System.ObjectDisposedException: Cannot access a disposed object.
Object name: 'LoggerFactory'.
   at Microsoft.Extensions.Logging.LoggerFactory.CreateLogger(String categoryName)
   at Microsoft.Azure.WebJobs.Script.WebHost.ScriptLoggerFactory.CreateLogger(String categoryName) in C:\azure-webjobs-sdk-script\src\WebJobs.Script.WebHost\Diagnostics\ScriptLoggerFactory.cs:line 26
   at Microsoft.Azure.WebJobs.Host.Loggers.FunctionInstanceLogger.Log(LogLevel level, FunctionDescriptor descriptor, Guid functionId, String message, Exception exception) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Loggers\FunctionInstanceLogger.cs:line 49
   at Microsoft.Azure.WebJobs.Host.Loggers.FunctionInstanceLogger.LogFunctionCompletedAsync(FunctionCompletedMessage message, CancellationToken cancellationToken) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Loggers\FunctionInstanceLogger.cs:line 36
   at Microsoft.Azure.WebJobs.Host.Executors.FunctionExecutor.TryExecuteAsync(IFunctionInstance functionInstance, CancellationToken cancellationToken) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Executors\FunctionExecutor.cs:line 136
   at Microsoft.Azure.WebJobs.Host.Executors.TriggeredFunctionExecutor`1.TryExecuteAsync(TriggeredFunctionData input, CancellationToken cancellationToken) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Executors\TriggeredFunctionExecutor.cs:line 54
   at Microsoft.Azure.WebJobs.Extensions.DurableTask.TaskActivityShim.RunAsync(TaskContext context, String rawInput)
```

Here is the kusto query where this issue was discovered:

```
FunctionsLogs 
| where PreciseTimeStamp between(datetime(2018-10-10 14:47) .. datetime(2018-10-10 14:50))
[...]

PR #1132, opened 2019-12-31
Tracing improvements

cgillum:
Some tracing fixes to make certain types of issues easier to debug.

- More tracing for task hub worker start/stop
- Trace extension configuration as a JSON blob
- Report errors when WebJobs fails to execute a function

Related to https://github.com/Azure/azure-functions-durable-extension/issues/472

Pending tasks:
- [x] Add unit test to ensure the host properly shuts down at the end of the test
- [x] Add unit test to ensure that configuration is logged as expected on host startup
- [x] ~~Remove NotificationUrl from DurableTaskOptions.~~ (too much work, out of scope)
- [x] Manually test HubName trace behavior using Functions core tools

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -169,19 +169,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         }
 #endif
 
-        /// <summary>
-        /// Gets the base URI for handling HTTP webhooks.
-        /// </summary>
-        /// <remarks>
-        /// This property is intended to be used only in self-hosting scenarios. When
-        /// hosting in the Azure Functions runtime, this property is set automatically
-        /// when the extension is initialized.
-        /// </remarks>
-        /// <value>
-        /// The base URI for handling HTTP webhooks.
-        /// </value>
-        public Uri NotificationUrl { get; set; }
-
         internal DurableTaskOptions Options { get; }
 
         internal HttpApiHandler HttpApiHandler { get; private set; }
@@ -217,10 +204,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.Options.Validate();
 
             // For 202 support
-            if (this.NotificationUrl == null)
+            if (this.Options.NotificationUrl == null)
             {
 #pragma warning disable CS0618 // Type or member is obsolete
-                this.NotificationUrl = context.GetWebhookHandler();
+                this.Options.NotificationUrl = context.GetWebhookHandler();
 #pragma warning restore CS0618 // Type or member is obsolete
             }
 
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -104,6 +104,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             // Options will be null in Functions v1 runtime - populated later.
             this.Options = options?.Value ?? new DurableTaskOptions();
             this.nameResolver = nameResolver ?? throw new ArgumentNullException(nameof(nameResolver));
+            this.ResolveAppSettingOptions();
 
             if (loggerFactory == null)
             {
@@ -194,12 +195,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 this.InitializeForFunctionsV1(context);
             }
 
-            if (this.nameResolver.TryResolveWholeString(this.Options.HubName, out string taskHubName))
-            {
-                // use the resolved task hub name
-                this.Options.HubName = taskHubName;
-            }
-
             // Throw if any of the configured options are invalid
             this.Options.Validate();
 
@@ -247,16 +242,36 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.taskHubWorker.AddOrchestrationDispatcherMiddleware(this.OrchestrationMiddleware);
         }
 
+        private void ResolveAppSettingOptions()
+        {
+            if (this.Options == null)
+            {
+                throw new InvalidOperationException($"{nameof(this.Options)} must be set before resolving app settings.");
+            }
+
+            if (this.nameResolver == null)
+            {
+                throw new InvalidOperationException($"{nameof(this.nameResolver)} must be set before resolving app settings.");
+            }
+
+            if (this.nameResolver.TryResolveWholeString(this.Options.HubName, out string taskHubName))
+            {
+                // use the resolved task hub name
+                this.Options.HubName = taskHubName;
+            }
+        }
+
         private void InitializeForFunctionsV1(ExtensionConfigContext context)
         {
 #if FUNCTIONS_V1
             context.ApplyConfig(this.Options, "DurableTask");
+            this.nameResolver = context.Config.NameResolver;
+            this.ResolveAppSettingOptions();
             ILogger logger = context.Config.LoggerFactory.CreateLogger(LoggerCategoryName);
             this.TraceHelper = new EndToEndTraceHelper(logger, this.Options.Tracing.TraceReplayEvents);
             this.connectionStringResolver = new WebJobsConnectionStringProvider();
             this.durabilityProviderFactory = new AzureStorageDurabilityProviderFactory(new OptionsWrapper<DurableTaskOptions>(this.Options), this.connectionStringResolver);
             this.defaultDurabilityProvider = this.durabilityProviderFactory.GetDurabilityProvider();
-            this.nameResolver = context.Config.NameResolver;
             this.LifeCycleNotificationHelper = this.CreateLifeCycleNotificationHelper();
             this.HttpApiHandler = new HttpApiHandler(this, logger);
 #endif
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -169,6 +169,19 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         }
 #endif
 
+        /// <summary>
+        /// Gets the base URI for handling HTTP webhooks.
+        /// </summary>
+        /// <remarks>
+        /// This property is intended to be used only in self-hosting scenarios. When
+        /// hosting in the Azure Functions runtime, this property is set automatically
+        /// when the extension is initialized.
+        /// </remarks>
+        /// <value>
+        /// The base URI for handling HTTP webhooks.
+        /// </value>
+        public Uri NotificationUrl { get; set; }
+
         internal DurableTaskOptions Options { get; }
 
         internal HttpApiHandler HttpApiHandler { get; private set; }
@@ -204,10 +217,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.Options.Validate();
 
             // For 202 support
-            if (this.Options.NotificationUrl == null)
+            if (this.NotificationUrl == null)
             {
 #pragma warning disable CS0618 // Type or member is obsolete
-                this.Options.NotificationUrl = context.GetWebhookHandler();
+                this.NotificationUrl = context.GetWebhookHandler();
 #pragma warning restore CS0618 // Type or member is obsolete
             }
 
@@ -892,9 +905,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 // If there is no listener for a function (e.g. it's a [NoAutomaticTrigger] function) then we don't
                 // need to wait for it to be deregistered before shutting down the task hub worker.
                 if (this.isTaskHubWorkerStarted &&
-                    this.knownOrchestrators.Values.Count(info => !info.IsDeregistered && info.Executor != null) == 0 &&
-                    this.knownActivities.Values.Count(info => !info.IsDeregistered && info.Executor != null) == 0 &&
-                    this.knownEntities.Values.Count(info => !info.IsDeregistered && info.Executor != null) == 0)
+                    !this.knownOrchestrators.Values.Any(info => info.HasActiveListener) &&
+                    !this.knownActivities.Values.Any(info => info.HasActiveListener) &&
+                    !this.knownEntities.Values.Any(info => info.HasActiveListener))
                 {
                     bool isGracefulStop = this.Options.UseGracefulShutdown;
 
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Net.Http;
 using System.Reflection;
@@ -12,8 +13,6 @@ using System.Threading;
 using System.Threading.Tasks;
 using DurableTask.AzureStorage;
 using DurableTask.Core;
-using DurableTask.Core.Common;
-using DurableTask.Core.Exceptions;
 using DurableTask.Core.History;
 using DurableTask.Core.Middleware;
 using Microsoft.Azure.WebJobs.Description;
@@ -265,12 +264,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private void TraceConfigurationSettings()
         {
-            this.TraceHelper.ExtensionInformationalEvent(
-                this.Options.HubName,
-                instanceId: string.Empty,
-                functionName: string.Empty,
-                message: $"Initializing extension with the following settings: {this.Options.GetDebugString()}",
-                writeToUserLogs: true);
+            this.Options.TraceConfiguration(this.TraceHelper);
         }
 
         private ILifeCycleNotificationHelper CreateLifeCycleNotificationHelper()
@@ -864,8 +858,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                             message: "Starting task hub worker",
                             writeToUserLogs: true);
 
+                        Stopwatch sw = Stopwatch.StartNew();
                         await this.defaultDurabilityProvider.CreateIfNotExistsAsync();
                         await this.taskHubWorker.StartAsync();
+
+                        this.TraceHelper.ExtensionInformationalEvent(
+                            this.Options.HubName,
+                            instanceId: string.Empty,
+                            functionName: string.Empty,
+                            message: $"Task hub worker started. Latency: {sw.Elapsed}",
+                            writeToUserLogs: true);
 
                         // Enable flowing exception information from activities
                         // to the parent orchestration code.
@@ -888,15 +890,26 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     this.knownOrchestrators.Values.Count(info => !info.IsDeregistered) == 0 &&
                     this.knownActivities.Values.Count(info => !info.IsDeregistered) == 0)
                 {
+                    bool isGracefulStop = this.Options.UseGracefulShutdown;
+
                     this.TraceHelper.ExtensionInformationalEvent(
                         this.Options.HubName,
                         instanceId: string.Empty,
                         functionName: string.Empty,
-                        message: "Stopping task hub worker",
+                        message: $"Stopping task hub worker. IsGracefulStop: {isGracefulStop}",
                         writeToUserLogs: true);
 
-                    await this.taskHubWorker.StopAsync(isForced: !this.Options.UseGracefulShutdown);
+                    Stopwatch sw = Stopwatch.StartNew();
+                    await this.taskHubWorker.StopAsync(isForced: !isGracefulStop);
                     this.isTaskHubWorkerStarted = false;
+
+                    this.TraceHelper.ExtensionInformationalEvent(
+                        this.Options.HubName,
+                        instanceId: string.Empty,
+                        functionName: string.Empty,
+                        message: $"Task hub worker stopped. IsGracefulStop: {isGracefulStop}. Latency: {sw.Elapsed}",
+                        writeToUserLogs: true);
+
                     return true;
                 }
             }
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -60,7 +60,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         private readonly bool isOptionsConfigured;
         private IDurabilityProviderFactory durabilityProviderFactory;
         private INameResolver nameResolver;
-        private DurabilityProvider durabilityProvider;
+        private DurabilityProvider defaultDurabilityProvider;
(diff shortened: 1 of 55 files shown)


src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -763,7 +757,7 @@ internal void RegisterEntity(FunctionName entityFunction, RegisteredFunctionInfo
         internal void DeregisterEntity(FunctionName entityFunction)
         {
             RegisteredFunctionInfo existing;
-            if (this.knownOrchestrators.TryGetValue(entityFunction, out existing) && !existing.IsDeregistered)
+            if (this.knownEntities.TryGetValue(entityFunction, out existing) && !existing.IsDeregistered)
```
cgillum: This was a bug where we weren't correctly deregistering entities.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -710,15 +704,15 @@ internal void RegisterActivity(FunctionName activityFunction, ITriggeredFunction
...
             {
+                var info = new RegisteredFunctionInfo(executor, isOutOfProc: false);
+                this.knownActivities[activityFunction] = info;
+
                 this.TraceHelper.ExtensionInformationalEvent(
                     this.Options.HubName,
                     instanceId: string.Empty,
                     functionName: activityFunction.Name,
-                    message: $"Registering activity function named {activityFunction}.",
+                    message: $"Registered activity function named {activityFunction}.",
```
cgillum: These changes were simply to make it easier for me to filter trace messages by making them all consistent (starting messages with "Registered" instead of having some of them start with "Registering"). There is no functional change here.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -886,17 +888,29 @@ internal async Task<bool> StopTaskHubWorkerIfIdleAsync()
             {
                 if (this.isTaskHubWorkerStarted &&
                     this.knownOrchestrators.Values.Count(info => !info.IsDeregistered) == 0 &&
-                    this.knownActivities.Values.Count(info => !info.IsDeregistered) == 0)
+                    this.knownActivities.Values.Count(info => !info.IsDeregistered) == 0 &&
+                    this.knownEntities.Values.Count(info => !info.IsDeregistered) == 0)
```
cgillum: The other bug I found was that we weren't making sure all entity functions were deregistered before shutting down the task hub worker. I'm not aware of any issues this has caused anyone, but theoretically it could cause us to shut down too early.

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -726,6 +726,22 @@ public void ExtensionWarningEvent(string hubName, string functionName, string in
...
+                hubName,
+                LocalAppName,
+                LocalSlotName,
+                configurationJsonString,
+                ExtensionVersion);
+
+            this.logger.LogInformation(
+                "Durable extension configuration loaded: {details}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}.",
+                configurationJsonString, hubName, LocalAppName, LocalSlotName, ExtensionVersion);
+        }
```
cgillum: I created a new event specifically for tracing configuration to make it easier to search just for these events in Kusto.

src/WebJobs.Extensions.DurableTask/Listener/TaskActivityShim.cs
```diff
@@ -54,7 +54,28 @@ public override async Task<string> RunAsync(TaskContext context, string rawInput
                 functionType: FunctionType.Activity,
                 isReplay: false);
 
-            FunctionResult result = await this.executor.TryExecuteAsync(triggerInput, CancellationToken.None);
+            FunctionResult result;
+            try
+            {
+                result = await this.executor.TryExecuteAsync(triggerInput, CancellationToken.None);
```
cgillum: Previously we did not catch or trace exceptions thrown by `TryExecuteAsync`, which is why the ObjectDisposedException problem was so hard to diagnose. When a function fails with an exception, this still returns by with a `FunctionResult.Succeeded` value set to `false`. Now we look for both cases. An exception means that something unexpected failed in WebJobs.

src/WebJobs.Extensions.DurableTask/Listener/TaskActivityShim.cs
```diff
@@ -54,7 +54,28 @@ public override async Task<string> RunAsync(TaskContext context, string rawInput
...
+            {
+                this.config.TraceHelper.FunctionFailed(
+                    this.config.Options.HubName,
+                    this.activityName,
+                    instanceId,
+                    $"An internal error occurred while attempting to execute this function: " + e,
+                    functionType: FunctionType.Activity,
+                    isReplay: false);
+
+                // TODO: Throw a different exception that can be used to abort the orchestration rather than failing it.
```
ConnorMcMahon: This new exception has to be defined in DurableTask.Core right?

test/Common/TestActivities.cs
```diff
@@ -90,7 +90,7 @@ public static string BindToPOCO([ActivityTrigger] PlainOldClrObject poco)
             return poco.Foo;
         }
 
-        // Mark as no automatic trigger to allow usage of output wrapper.
+        // Mark as no automatic trigger to allow usage of addition parameters not associated with bindings (i.e. outputWrapper).
```
ConnorMcMahon: typo: addition->additional

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -35,20 +38,26 @@ public string HubName
...
                 }
 
-                return this.hubName;
+                return this.resolvedHubName;
             }
 
             set
             {
-                this.hubName = value;
+                if (this.originalHubName == null)
```
ConnorMcMahon: This approach won't quite work for original hub name if I am not mistaken. When diagnosing the initial bug we had with the new default hub name logic, we realized that the way that the Microsoft libraries for populating Options objects from JSON involves calling the `get` method BEFORE the `set` method. That means that original hub name will always be set to the empty string in real world scenarios.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -163,50 +172,40 @@ public void SetDefaultHubName(string hubName)
...
+            JObject configurationJson = JObject.FromObject(
+                this,
+                new JsonSerializer
+                {
+                    Converters = { new StringEnumConverter() },
+                    ContractResolver = new CamelCasePropertyNamesContractResolver(),
+                });
+
+            // Don't trace the notification URL query string since it may contain secrets.
+            // This is the only property which we expect to contain secrets. Everything else should be *names*
```
ConnorMcMahon: I'm concerned that this may not always be true, and we could eventually start logging secrets without realizing it. I'm not sure of a safe/more discoverable way of handling this. We could  implicitly treat everything as secrets unless it is tagged otherwise, but that has it's own pains.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -163,50 +172,40 @@ public void SetDefaultHubName(string hubName)
...
-        }
-
-        private void AppendStorageProviderValuesToDebugString(StringBuilder sb)
-        {
-            sb.Append(nameof(this.StorageProvider)).Append(": { ");
-            foreach (var value in this.StorageProvider)
+            // At this stage the task hub name is expected to have been resolved. However, we want to know
+            // what the original value was in addition to the resolved value, so we're updating the JSON
+            // blob property to use the original, unresolved value.
+            if (configurationJson.TryGetValue(nameof(this.HubName), StringComparison.OrdinalIgnoreCase, out _))
```
ConnorMcMahon: Wouldn't `this.HubName`, and hence its value in the configuration JSON, always have a value? Or is this call meant to force using the default value?

src/WebJobs.Extensions.DurableTask/Listener/TaskActivityShim.cs
cgillum: Right. I've drafted a proposal for this here: https://github.com/Azure/azure-functions-durable-extension/issues/1133.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -886,10 +886,15 @@ internal async Task<bool> StopTaskHubWorkerIfIdleAsync()
...
+                // Wait to shut down the task hub worker until all function listeners have been shut down.
+                // We set IsDeregistered to true when Stop() is called on the listener
+                // We set Executor to a non-null value only if there is a listener associated with this function.
+                // If there is no listener for a function (e.g. it's a [NoAutomaticTrigger] function) then we don't
+                // need to wait for it to be deregistered before shutting down the task hub worker.
                 if (this.isTaskHubWorkerStarted &&
-                    this.knownOrchestrators.Values.Count(info => !info.IsDeregistered) == 0 &&
-                    this.knownActivities.Values.Count(info => !info.IsDeregistered) == 0 &&
-                    this.knownEntities.Values.Count(info => !info.IsDeregistered) == 0)
+                    this.knownOrchestrators.Values.Count(info => !info.IsDeregistered && info.Executor != null) == 0 &&
```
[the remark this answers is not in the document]
cgillum: Good call. In fact, we probably don't need most of these comments if we factor this into another property like you suggested.

src/WebJobs.Extensions.DurableTask/Listener/TaskActivityShim.cs
cgillum: Agreed on both points.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
cgillum: OK, let me revisit this then. Is there a recommended way to reproduce that behavior? For example, will I need to manually run this in the functions host?

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
cgillum: If this data is populated by host.json then it should never have secrets. It actually doesn't even make sense for `NotificationUrl` to be here. I guess I could solve this by removing it.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
cgillum: I'm not sure I understand your question. This code does not reference the value of either `this.HubName` or the default value (`this.defaultHubName`). The intent is to log whatever un-resolved value was specified in host.json (including the name of the app settings from which the task hub is derived, if applicable).

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
ConnorMcMahon: I just created a nupkg and consumed it in a VS Durable Functions project while debugging that project.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
ConnorMcMahon: Can we safely remove this? That seems like a breaking change.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
ConnorMcMahon: I guess my point is that `configurationJson` will always have a value for `HubName`, as it is just parsing `this` which will always give you a value for `DurableTaskOptions.HubName`. Because of that, I don't really see the value of the `if` statment since it will always be true. You can just always perform the assignment of `HubName` to `this.originalHubName`.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
ConnorMcMahon: Is that safe to do? Seems like a breaking change.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
ConnorMcMahon: I just create a nupkg locally, and then consume that in a VS Durable Functions project with a debugger to step through it.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
cgillum: It is a breaking change, but I would argue there is no valid scenario for referencing this property. The documentation even mentions that this is for internal use.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
cgillum: Oh, you're asking why am I am using `TryGetValue` for `HubName` since that property should always be defined. It's just out of habit in writing defensive code. It didn't seem harmful or misleading to keep it.

test/Common/TestHelpers.cs
```diff
@@ -85,32 +86,6 @@ internal static class TestHelpers
...
-                    KeySettingName = eventGridKeySettingName,
-                    TopicEndpoint = eventGridTopicEndpoint,
-                    PublishEventTypes = eventGridPublishEventTypes,
-                },
-            };
-            options.HttpSettings = new HttpOptions()
-            {
-                DefaultAsyncRequestSleepTimeMilliseconds = httpAsyncSleepTime,
-            };
-            options.NotificationUrl = notificationUrl;
```
cgillum: When i removed the `NotificationUrl` property, I noticed that there was a bunch of duplicated code here. Probably some accidental copy/paste error. I removed the duplicate lines of code.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
cgillum: I went down the path of moving `NotificationUrl` from `DurableTaskOptions` to `DurableTaskExtension` in a previous commit. I was almost successful except that getting a final test passing would require some additional DI interfaces, which felt way out of scope for this PR. Perhaps I can consider it for a different release.

cgillum:
Full testing of this PR is currently blocked on https://github.com/Azure/azure-functions-durable-extension/pull/1134.

Also, I just realized that my current approach of serializing `DurableTaskOptions` as JSON is not sufficient. The problem is that `DurableTaskOptions.StorageProvider` is a dictionary and not the actual "options" object (e.g. `AzureStorageOptions`), so the traces don't show any of the storage provider default values. I'll need to come up with a way to ensure we get those default values traced.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -163,50 +171,38 @@ public void SetDefaultHubName(string hubName)
...
-            sb.Append(nameof(this.UseGracefulShutdown)).Append(": ").Append(this.UseGracefulShutdown).Append(", ");
-            if (this.ExtendedSessionsEnabled)
-            {
-                sb.Append(nameof(this.ExtendedSessionIdleTimeoutInSeconds)).Append(": ").Append(this.ExtendedSessionIdleTimeoutInSeconds).Append(", ");
-            }
-
-            if (this.NotificationUrl != null)
+            // Clone the options to avoid making changes to the original.
+            // We make updates to the clone rather than to JSON to ensure we're updating what we think we're updating.
+            DurableTaskOptions clone = JObject.FromObject(this).ToObject<DurableTaskOptions>();
```
cgillum: @ConnorMcMahon just a heads-up on this change. Through manual testing I discovered that I wasn't correctly overwriting existing config values. I was instead adding new values (different capitalization) to the trace output. To avoid issues like this, I decided to switch from modifying JSON data to modifying the Options object directly (or rather, a clone of the options object) and then serializing the updated version. That way this tracing code never needs to be concerned with the details of JSON serialization.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -61,7 +60,7 @@ public class DurableTaskExtension :
         private readonly bool isOptionsConfigured;
         private IDurabilityProviderFactory durabilityProviderFactory;
         private INameResolver nameResolver;
-        private DurabilityProvider defaultDurabilityProvider;
+        private DurabilityProvider durabilityProvider;
```
cgillum: @ConnorMcMahon I hope you don't mind, but I decided to rename `defaultDurabilityProvider` to just `durabilityProvider`. I initially misinterpreted what this variable represented when I was investigating how to trace durability provider settings. Typically when I see that something is "default" then there must be some "non-default" version of it elsewhere. I looked for a non-default provider but could find one. What I eventually realized is that there is no "non-default" so I thought it would be more appropriate to remove the "default" prefix in this variable name to avoid future confusion.

ConnorMcMahon: I don't have a huge complaint about it. The non-default `DurabilityProvider` in this case is the `DurabilityProvider` created in `DurableTask.Extension.GetClient()`, i.e. when the customer provides a different task hub name and/or connection string. I do think that distinction was moderately helpful, but I could definitely buy the argument that it adds more confusion than it helps, especially since those non-default `DurabilityProviders` are only referenced on a couple of lines.

cgillum: Oh, I see. So the way you were using it actually was in agreement with my definition of "default". I just failed to notice/remember the `GetClient()` case where a different instance gets created.

With that additional context, I think I'll roll this back to the old name. Thanks for letting me know about this.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -35,20 +38,25 @@ public string HubName
...
                 }
 
-                return this.hubName;
+                return this.resolvedHubName;
             }
 
             set
             {
-                this.hubName = value;
+                if (this.originalHubName == null)
```
ConnorMcMahon: This technically won't work when `SetDefaultHubName()` is called, as it calls `HubName.set`. We could edit that method to set `this.resolvedHubName` directly, so that it doesn't accidentally set `this.originalHubName` in the process.

test/Common/TestHelpers.cs
```diff
@@ -91,31 +92,6 @@ internal static class TestHelpers
             options.MaxConcurrentActivityFunctions = 200;
             options.NotificationHandler = eventGridNotificationHandler;
 
-            options.HubName = GetTaskHubNameFromTestName(testName, enableExtendedSessions);
```
ConnorMcMahon: It's wild to me that we never caught all of this duplicate code!

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -163,50 +171,38 @@ public void SetDefaultHubName(string hubName)
...
+                clone.NotificationUrl = new Uri(clone.NotificationUrl.GetLeftPart(UriPartial.Path));
             }
 
-            sb.Append(nameof(this.Notifications)).Append(": { ");
-            this.Notifications.AddToDebugString(sb);
-            sb.Append(" }, ");
+            // At this stage the task hub name is expected to have been resolved. However, we want to know
+            // what the original value was in addition to the resolved value, so we're updating the JSON
+            // blob property to use the original, unresolved value.
+            clone.HubName = this.originalHubName;
```
ConnorMcMahon: Do we ever log the resolved value now?

cgillum: Yes, all traces include a separate column for the resolved host name.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -899,19 +901,36 @@ internal async Task<bool> StopTaskHubWorkerIfIdleAsync()
         {
             using (await this.taskHubLock.AcquireAsync())
             {
+                // Wait to shut down the task hub worker until all function listeners have been shut down.
```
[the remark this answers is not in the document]
cgillum: I thought I removed this. It might have come back as part of me reverting some other changes. I'll update it.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
ConnorMcMahon: Duh. My bad.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
cgillum: I'm testing this using func.exe and I'm never seeing `SetDefaultHubName()` getting called. In what conditions will it be called?

ConnorMcMahon: It happens when we need to sanitize the default name, i.e. when the customer's site name has `-` characters. 

Essentially you can test this scenario by setting the Environment variable WEBSITE_SITE_NAME to something like "invalid-site-name" and don't provide a TaskHub.

cgillum: I tested this and was able to successfully trigger the `SetDefaultHubName()` code path and it works in a way that I think is fairly reasonable:

```json
{
  "httpSettings": null,
  "hubName": "Sample-Hub-VS",
  "storageProvider": {
    
  },
  "tracing": {
    "traceInputsAndOutputs": false,
    "traceReplayEvents": false
  },
  "notifications": {
    "eventGrid": null
  },
  "maxConcurrentActivityFunctions": 80,
  "maxConcurrentOrchestratorFunctions": 80,
  "notificationUrl": "http://localhost:7071/runtime/webhooks/durabletask",
  "extendedSessionsEnabled": false,
  "extendedSessionIdleTimeoutInSeconds": 30,
  "maxOrchestrationActions": 100000,
  "overridableExistingInstanceStates": "NonRunningStates",
  "entityMessageReorderWindowInMinutes": 30,
  "useGracefulShutdown": false
}
```

Basically, it set `hubName` to `Sample-Hub-VS`, which was the value I set for `WEBSITE_HOST_NAME`. I guess the alternative could have been that it logs the empty string since nothing was explicitly specified in host.json, though I wonder if this is actually more informative from a debugging perspective. Thoughts?

ConnorMcMahon: It's probably fine. I have a slight preference for the `""` or `null`, as that is a bit more explicit that the customer didn't provide a `hubName`. That being said, we can easily infer that if the `hubName` matches their site name.

cgillum: I can't think of a simple way to achieve this without further refactoring how we manage hub name. If they don't set it in host.json, the `set` method never gets called until after we do the resolution. At that point this code thinks the resolved value is the original. If you don't mind, I think for the sake of getting this release done we go ahead with what we have here.

PR #1134, opened 2020-01-01
Fix TaskHub validation

### LINKED ISSUE #1122: Unable to load hubName from appsettings for durable function
### Description
According to the documentation found at https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-task-hubs#hostjson-functions-20-1, it should be possible to use a value from the appsettings as the hubname in the **host.json** file. However, after updating **Microsoft.Azure.WebJobs.Extensions.DurableTask** to **v2.1.0**, an exception is now thrown during startup.

### Expected behavior
When using **"%MyTaskHub%"** as the value of the **"hubName"** property in the host.json file, I'm expecting the runtime to look for a key matching **"MyTaskHub"** in the appsettings and use its value as the hub.

### Actual behavior
When the function app starts up, I receive an exception:
>System.Private.CoreLib: Exception has been thrown by the target of an invocation. Microsoft.Azure.WebJobs.Extensions.DurableTask: Task hub name '%MyTaskHub%' should contain only alphanumeric characters excluding '-' and have length up to 45. Microsoft.WindowsAzure.Storage: Invalid container name. Check MSDN for more information about valid container naming.
Value cannot be null. (Parameter 'provider')

Instead of pasting the source code of my actual functions app which is a bit more complex, I've managed to replicate the issue in a simpler form using the Visual Studio functions app template. Below are steps to replicate:

1. Open Visual Studio 2019.
2. Create a new **empty** functions app using **Azure Functions v3 (.NET Core)** and select **Storage Emulator** as the storage account.
3. Once the project has been scaffolded, right-click on the project and add **New Azure Function**. Select **Durable Functions Orchestration** as the type.
4. Update the **Microsoft.NET.Sdk.Functions** NuGet package to **3.0.2**.
5. Update the **Microsoft.Azure.WebJobs.Extensions.DurableTask** NuGet package to **2.1.0**.
6. If you try and build the project at this point, you'll get errors since the types have changed. Update the function class with the correct types to resolve the build errors.
7. Open **host.json** and replace the contents with `{
  "version": "2.0",
  "extensions": {
    "durableTask": {
      "hubName": "%MyTaskHub%"
    }
  }
}`. This snippet is taken directly from the documentation (link at the top of this issue).
8. Open **local.settings.json** and add `"MyTaskHub" : "samplehubname"` to the **"Values"** section. Again, this snippet is taken directly from the documentation.
9. Open the function class and update `[DurableClient]` to `[DurableClient(TaskHub = "%MyTaskHub%")]`.
10. Debug the functions app and during startup, the exception described above is thrown.

### Known workarounds
The only way I've managed to get it working is to manually make sure the host.json **"hubName"** matches the value of **"MyTaskHub"** in local.settings.json. You can verify this by changing the value of **"hubName"** in host.json to **"samplehubname"** after following the replication steps above. After doing so, debug the project and it should at least run correctly.

This workaround isn't ideal as it means I would need to update the value everytime instead of the value coming from the appsettings.

### App Details
- **AzureFunctionsVersion: v3**
- **TargetFramework: netcoreapp3.0**
- **Microsoft.NET.Sdk.Functions: v3.0.2**
- **Microsoft.Azure.WebJobs.Extensions.DurableTask: v2.1.0**

ConnorMcMahon:
TaskHub validation needs to occur after the TaskHub has been resolved.
To achieve this, we resolve app settings earlier in the pipeline, as
well as waiting to validate the TaskHub until after finishing
constructing AzureStorageDurabilityProviderFactory.

Resolves #1122

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
+++ b/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
@@ -40,9 +40,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         }
 
         // This method should not be called before the app settings are resolved into the options.
-        // Because of this, we wait to validate the options until righ before building a durability provider, rather
+        // Because of this, we wait to validate the options until right before building a durability provider, rather
         // than in the Factory constructor.
-        private void EnsureOptionsAreValid()
+        private void EnsureInitialized()
         {
             if (!this.hasValidatedOptions)
             {
@@ -62,7 +62,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         public DurabilityProvider GetDurabilityProvider()
         {
-            this.EnsureOptionsAreValid();
+            this.EnsureInitialized();
             if (this.defaultStorageProvider == null)
             {
                 var defaultService = new AzureStorageOrchestrationService(this.defaultSettings);
@@ -74,7 +74,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         public DurabilityProvider GetDurabilityProvider(DurableClientAttribute attribute)
         {
-            this.EnsureOptionsAreValid();
+            this.EnsureInitialized();
             return this.GetAzureStorageStorageProvider(attribute);
         }
 
--- a/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
+++ b/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
@@ -17,7 +17,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         private AzureStorageDurabilityProvider defaultStorageProvider;
 
         // Must wait to get settings until we have validated taskhub name.
-        private bool haveValidatedOptions;
+        private bool hasValidatedOptions;
         private AzureStorageOrchestrationServiceSettings defaultSettings;
 
         public AzureStorageDurabilityProviderFactory(
@@ -30,7 +30,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
             this.azureStorageOptions.Validate();
 
-            this.connectionStringResolver = connectionStringResolver;
+            this.connectionStringResolver = connectionStringResolver ?? throw new ArgumentNullException(nameof(connectionStringResolver));
             this.defaultConnectionName = this.azureStorageOptions.ConnectionStringName ?? ConnectionStringNames.Storage;
         }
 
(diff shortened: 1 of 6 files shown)


ConnorMcMahon:
I'd recommend using the checkbox that hides whitespace changes.

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
```diff
@@ -1,150 +1,164 @@
...
+            IOptions<DurableTaskOptions> options,
+            IConnectionStringResolver connectionStringResolver)
+        {
+            this.options = options.Value;
+            this.azureStorageOptions = new AzureStorageOptions();
+            JsonConvert.PopulateObject(JsonConvert.SerializeObject(this.options.StorageProvider), this.azureStorageOptions);
+
+            this.azureStorageOptions.Validate();
+
+            this.connectionStringResolver = connectionStringResolver;
```
cgillum: I suggest checking for null since this is a public constructor and other methods depend on this being non-null.
```suggestion
            this.connectionStringResolver = connectionStringResolver ?? throw new ArgumentNullException(nameof(connectionStringResolver));
```

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
```diff
@@ -1,150 +1,164 @@
...
+    internal class AzureStorageDurabilityProviderFactory : IDurabilityProviderFactory
+    {
+        private readonly DurableTaskOptions options;
+        private readonly AzureStorageOptions azureStorageOptions;
+        private readonly IConnectionStringResolver connectionStringResolver;
+        private readonly string defaultConnectionName;
+        private AzureStorageDurabilityProvider defaultStorageProvider;
+
+        // Must wait to get settings until we have validated taskhub name.
+        private bool haveValidatedOptions;
```
cgillum: Small naming nit:
```suggestion
        private bool hasValidatedOptions;
```

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
```diff
@@ -1,150 +1,167 @@
...
-            {
-                ConnectionString = resolvedStorageConnectionString,
-            };
-        }
-    }
- }
+        }
+
+        // This method should not be called before the app settings are resolved into the options.
+        // Because of this, we wait to validate the options until righ before building a durability provider, rather
```
cgillum: righ --> right

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
```diff
@@ -1,150 +1,167 @@
...
+        // than in the Factory constructor.
+        private void EnsureOptionsAreValid()
```
cgillum: I almost wonder if you should name this `Initialize()` or `EnsureIsInitialized()`. It kinda seems like that's what this method is doing - not just validating.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -4026,6 +4027,63 @@ public async Task TaskHubName_DefaultNameNonProductionSlot_ThrowsException()
...
+        public async Task TaskHubName_AppSettingReference_ValidTaskHub_UsesResolvedTaskHub()
+        {
+            string taskHubSettingName = "TaskHubName";
+            string taskHubName = "ValidTaskHub";
+            DurableTaskOptions durableTaskOptions = new DurableTaskOptions();
+            durableTaskOptions.HubName = $"%{taskHubSettingName}%";
+
+            var mockNameResolver = new Mock<INameResolver>();
+            mockNameResolver.Setup(res => res.Resolve(taskHubSettingName))
+                .Returns(taskHubName);
```
cgillum: Instead of mocking, it might be more straightforward to use `SimpleNameResolver`.

PR #1135, opened 2020-01-02
Task hub fix with lazy

salv85:
TaskHub validation needs to occur after the TaskHub has been resolved and after the costruction of the AzureStorageDurabilityProviderFactory.
To achieve this, we use lazy objects to get the configuration only when we need it, always after the initializations.
Improved slot recognition and "development environment" recognition
Improved default task hub name assignation.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
+++ b/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
@@ -13,8 +13,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         private readonly DurableTaskOptions options;
         private readonly AzureStorageOptions azureStorageOptions;
         private readonly IConnectionStringResolver connectionStringResolver;
-        private readonly AzureStorageOrchestrationServiceSettings defaultSettings;
         private readonly string defaultConnectionName;
+        private readonly Lazy<AzureStorageOrchestrationServiceSettings> defaultSettings;
         private AzureStorageDurabilityProvider defaultStorageProvider;
 
         public AzureStorageDurabilityProviderFactory(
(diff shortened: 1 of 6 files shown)


src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
```diff
@@ -1,150 +1,143 @@
...
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    internal class AzureStorageDurabilityProviderFactory : IDurabilityProviderFactory
+    {
+        private readonly DurableTaskOptions options;
+        private readonly AzureStorageOptions azureStorageOptions;
+        private readonly IConnectionStringResolver connectionStringResolver;
+        private readonly string defaultConnectionName;
+        private readonly Lazy<AzureStorageOrchestrationServiceSettings> defaultSettings;
```
ConnorMcMahon: I think the approach with `Lazy` is actually just a more graceful version of what I tried to do in my PR. I might change my approach to utilize this.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -61,7 +59,7 @@ public class DurableTaskExtension :
         private readonly bool isOptionsConfigured;
         private IDurabilityProviderFactory durabilityProviderFactory;
         private INameResolver nameResolver;
-        private DurabilityProvider defaultDurabilityProvider;
+        private Lazy<DurabilityProvider> defaultDurabilityProvider;
```
ConnorMcMahon: I'm not sure this second layer of `Lazy` is better or worse than my approach of changing the order of when we resolve app settings into the options vs when we get our durability provider. 

Your approach is probably a bit less brittle than mine, but I am afraid it would be harder to debug if it broke for some reason. Mine is a bit more explicit in the intention that we need to resolve app settings before getting our durability provider.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -209,16 +214,62 @@ private void AppendStorageProviderValuesToDebugString(StringBuilder sb)
             sb.Append(" }, ");
         }
 
+        internal void ValidateHubName(string hubName)
```
ConnorMcMahon: So this logic actually used to be here, but I moved it out. The reason is that this validation is very Azure Storage specific, so I needed to move it into a place where it won't affect other durability providers in the future. I would like to keep this in the `AzureStorageOptions`.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -237,15 +288,20 @@ internal void Validate()
...
+            return Environment.GetEnvironmentVariable("WEBSITE_SLOT_NAME") == null;
         }
 
         private static bool IsInNonProductionSlot()
         {
             var slotName = Environment.GetEnvironmentVariable("WEBSITE_SLOT_NAME");
 
             // slotName can be null in a test environment
-            if (slotName != null && !string.Equals(slotName, "Production", StringComparison.OrdinalIgnoreCase))
+            if (slotName == null || !string.Equals(slotName, "Production", StringComparison.OrdinalIgnoreCase))
```
ConnorMcMahon: I get the desire to split out the case where it is a development environment, but I'm not certain that this implementation made it any clearer, as this method now returns `true` if the environment is dev environment, and the method name no longer really makes sense, as development environments are not in non-production slots.

salv85:
@ConnorMcMahon of course you may use the pieces of code you like.

ConnorMcMahon:
@salv85, after taking another look at my code, I didn't use the Lazy approach as it caused some complications when keeping the validation logic in the Azure Storage specific code segments.

Thanks for submitting the PR! It was good to see how you tackled the problem, and made me more critically examine my own approach. We are always happy to see PRs from the community.

PR #1136, opened 2020-01-02
Analyzer: Fixed MethodAttributeAnalyzer to allow local functions inside an orchestrator

### LINKED ISSUE #1045: Consider enhancing the analyzer to process local (inline) functions
The durable functions analyzer does not seem to analyze local (inline) functions.

amdeel:
Fixed bug that caused functions defined in an orchestrator to be flagged unless the orchestrator had the Deterministic attribute. Also updated the message appearing on the diagnostic to be more clear.

resolves #1045

### DIFF
--- a/src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/Orchestrator/MethodAttributeAnalyzer.cs
+++ b/src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/Orchestrator/MethodAttributeAnalyzer.cs
@@ -16,7 +16,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers
         public const string DiagnosticId = "DF0107";
 
         private static readonly LocalizableString Title = new LocalizableResourceString(nameof(Resources.MethodAttributeAnalyzerTitle), Resources.ResourceManager, typeof(Resources));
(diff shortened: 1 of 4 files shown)


src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/Orchestrator/MethodAttributeAnalyzer.cs
```diff
@@ -53,7 +53,7 @@ private void AnalyzeMethod(SyntaxNodeAnalysisContext context)
 
             var declaration = syntaxReference.GetSyntax(context.CancellationToken);
 
-            if (SyntaxNodeUtils.IsMarkedDeterministic(declaration))
+            if (SyntaxNodeUtils.IsMarkedDeterministic(declaration) || SyntaxNodeUtils.IsInsideOrchestrator(declaration))
```
ConnorMcMahon: Just to clarify, IsInsideOrchestrator() will return `true` for a local function that is nested inside of an orchestrator function, right?

amdeel: Correct. In the case that this is true, it will not create the diagnostic. So when there is a local function inside an orchestrator, it will already be analyzed and there is no need for the Deterministic attribute. This is a temporary fix to allow this case to work smoothly until I refactor this analyzer to not look for Deterministic attributes at all.

PR #1139, opened 2020-01-03
Gracefully handle ObjectDisposedException on host shutdown

cgillum:
Fixes https://github.com/Azure/azure-functions-durable-extension/issues/472.

This PR currently has a lot of commits because it is waiting for multiple other PRs to be merged.

### Summary
This PR improves upon https://github.com/Azure/azure-functions-durable-extension/pull/1132 by throwing a `SessionAbortedException` inside `TaskActivityShim` if the WebJobs APIs throw any exception (such as `ObjectDisposedException`).

As part of https://github.com/Azure/durabletask/pull/359, the `SessionAbortedException` now is capable of aborting activity execution. Instead of the activity going into a "failed" state, the execution is abandoned and restarted. The restart will happen on an alternate, healthy instance of the functions runtime and will complete successfully.

### Validation
I used the following orchestration code to reproduce the `ObjectDisposedException` and validate that this code change works as intended:

```csharp
[FunctionName("ODE_Orchestration")]
public static async Task Run(
{
    await context.CallActivityAsync("ODE_Activity", null);
}

[FunctionName("ODE_Activity")]
public static void SayHello([ActivityTrigger] IDurableActivityContext context)
{
    Thread.Sleep(TimeSpan.FromSeconds(60));
}
```

**Repro steps**
This can be done using the Functions core tools on a local machine:

1. Start the orchestration above
2. Wait 5 seconds and then make a change to `host.json` to trigger a function host restart
3. Wait for the activity function to complete
4. Observe the task failure log with `ObjectDisposedException` in the message
5. Wait until the activity function restarts and then completes
6. Verify that the orchestration completed successfully

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Listener/TaskActivityShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/TaskActivityShim.cs
@@ -65,15 +65,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     this.config.Options.HubName,
                     this.activityName,
                     instanceId,
-                    $"An internal error occurred while attempting to execute this function: " + e,
+                    $"An internal error occurred while attempting to execute this function. The execution will be aborted and retried. Details: {e}",
                     functionType: FunctionType.Activity,
                     isReplay: false);
 
-                // TODO: Throw a different exception that can be used to abort the orchestration rather than failing it.
-                //       https://github.com/Azure/azure-functions-durable-extension/issues/472
-                throw new TaskFailureException(
-                    $"An internal error occurred while attempting to execute '{this.activityName}'.",
-                    Utils.SerializeCause(e, this.config.DataConverter.ErrorConverter));
+                // This will abort the execution and cause the message to go back onto the queue for re-processing
+                throw new SessionAbortedException(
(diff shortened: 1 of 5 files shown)


src/WebJobs.Extensions.DurableTask/Listener/TaskActivityShim.cs
```diff
@@ -54,7 +54,26 @@ public override async Task<string> RunAsync(TaskContext context, string rawInput
...
+                this.config.TraceHelper.FunctionFailed(
+                    this.config.Options.HubName,
+                    this.activityName,
+                    instanceId,
+                    $"An internal error occurred while attempting to execute this function. The execution will be aborted and retried. Details: {e}",
+                    functionType: FunctionType.Activity,
+                    isReplay: false);
+
+                // This will abort the execution and cause the message to go back onto the queue for re-processing
+                throw new SessionAbortedException(
```
ConnorMcMahon: Is this the only change unique to this PR?

cgillum: This commit is the only change https://github.com/Azure/azure-functions-durable-extension/pull/1139/commits/a2d5460a01951c36bc4ae3c37e5ccfe0d0893aff

PR #1142, opened 2020-01-06
Change FunctionAborted trace from error to warning

cgillum:
There are two problems with the current trace experience for the newly added function aborted logic:

1. It's an error, but it should actually be a warning since it's a transient issue.
2. We trace `FunctionFailed` but it didn't actually "fail".

I'm adding a new `FunctionAborted` trace instead of reusing `FunctionFailed`. This is both to improve clarity but also to avoid counting problems. For example, it would be odd to have a function that both failed and completed, and could cause existing Kusto or AI queries to double-count function executions in some cases.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
+++ b/src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
@@ -382,6 +382,20 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.WriteEvent(223, TaskHub, AppName, SlotName, Details, ExtensionVersion);
         }
 
+        [Event(224, Level = EventLevel.Warning)]
+        public void FunctionAborted(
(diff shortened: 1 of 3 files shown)


src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
```diff
@@ -382,6 +382,20 @@ private EtwEventSource()
             this.WriteEvent(223, TaskHub, AppName, SlotName, Details, ExtensionVersion);
         }
 
+        [Event(224, Level = EventLevel.Warning)]
```
ConnorMcMahon: We are probably good, but as a part of ANT87 we should consider adding more of a buffer to these events.

cgillum: Agreed.

PR #1149, opened 2020-01-08
Analyzer: Removed need for Deterministic attribute

amdeel:
The orchestrator analyzers will now run on methods defined in customer source code that are being used inside an orchestrator without including the Deterministic attribute on the method declaration.

When a method used inside an orchestrator is not deterministic, the method invocation will be flagged, as well as the non-deterministic code inside the method itself. This also works with indirect calls.


src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/Orchestrator/MethodAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
+﻿using Microsoft.CodeAnalysis;
```
cgillum: For all the new files in this PR, can you add the copyright header to the tops of the files?

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/Orchestrator/DateTimeAnalyzer.cs
```diff
@@ -20,43 +20,36 @@ public class DateTimeAnalyzer : DiagnosticAnalyzer
...
-        private static void AnalyzeIdentifierDateTimeNow(SyntaxNodeAnalysisContext context)
-        {
-            var identifierName = context.Node as IdentifierNameSyntax;
-            if (identifierName != null)
+            foreach (SyntaxNode descendant in method.DescendantNodes())
             {
-                var identifierText = identifierName.Identifier.ValueText;
-                if (identifierText == "Now" || identifierText == "UtcNow" || identifierText == "Today")
+                var identifierName = descendant as IdentifierNameSyntax;
+                if (identifierName != null)
```
cgillum: Tip: you can save a line of code by doing this instead if doing a separate `as` operation + `null` check:
```csharp
if (descendant is IdentifierNameSyntax identifierName)
{
    ...
}
```

I see other analyzers use the same pattern.

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/Orchestrator/DateTimeAnalyzer.cs
```diff
@@ -20,43 +20,36 @@ public class DateTimeAnalyzer : DiagnosticAnalyzer
...
-                        return;
-                    }
+                        var memberAccessExpression = identifierName.Parent;
+                        var memberSymbol = semanticModel.GetSymbolInfo(memberAccessExpression).Symbol;
 
-                    if (!SyntaxNodeUtils.IsInsideOrchestrator(identifierName) && !SyntaxNodeUtils.IsMarkedDeterministic(identifierName))
-                    {
-                        return;
-                    }
+                        if (memberSymbol != null && memberSymbol.ToString().StartsWith("System.DateTime"))
```
cgillum: Are you using `.StartsWith("System.DateTime")` to account for both `System.DateTime` and `System.DateTimeOffset`? If so, I recommend adding a comment to make it clear that you're doing this intentionally.

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/Orchestrator/DateTimeAnalyzer.cs
```diff
@@ -20,43 +20,36 @@ public class DateTimeAnalyzer : DiagnosticAnalyzer
         private const string Category = SupportedCategories.Orchestrator;
         public const DiagnosticSeverity severity = DiagnosticSeverity.Warning;
 
-        private static DiagnosticDescriptor Rule = new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category, severity, isEnabledByDefault: true, description: Description);
+        public static DiagnosticDescriptor Rule = new DiagnosticDescriptor(DiagnosticId, Title, MessageFormat, Category, severity, isEnabledByDefault: true, description: Description);
```
cgillum: Can this be made `readonly`? Same question for the cases that I see in your other analyzer classes.

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/Orchestrator/EnvironmentVariableAnalyzer.cs
```diff
@@ -20,44 +21,37 @@ public class EnvironmentVariableAnalyzer: DiagnosticAnalyzer
         private const string Category = SupportedCategories.Orchestrator;
         public const DiagnosticSeverity severity = DiagnosticSeverity.Warning;
```
cgillum: General C# convention is to make constant variable names capitalized - i.e. `Severity` instead of `severity`. Same thing for this file and other files.

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/Orchestrator/MethodAnalyzer.cs
```diff
@@ -0,0 +1,108 @@
...
+            foreach (MethodInformation method in methodInformationList)
+            {
+                var declaration = method.Declaration;
+
+                if (DateTimeAnalyzer.RegisterDiagnostic(declaration, context, semanticModel) |
+                    EnvironmentVariableAnalyzer.RegisterDiagnostic(declaration, context, semanticModel) |
+                    GuidAnalyzer.RegisterDiagnostic(declaration, context, semanticModel) |
+                    IOTypesAnalyzer.RegisterDiagnostic(declaration, context, semanticModel) |
+                    ThreadTaskAnalyzer.RegisterDiagnostic(declaration, context, semanticModel) |
+                    TimerAnalyzer.RegisterDiagnostic(declaration, context, semanticModel) |
```
cgillum: Are you intentionally use `|` instead of `||` here?  It seems like there could be a negative performance impact when using `|` instead of `||` as your logical OR operator if you don't actually need to.

amdeel: Yes, I intentionally use | instead of ||, and included a test for it. Here, I want all of the "analyzers" to run the code, the bool tells me if at least one produced a diagnostic.

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/Orchestrator/DateTimeAnalyzer.cs
amdeel: I suppose that makes sense if I don't have to cast it many times throughout my methods, which I don't think I'd need to do for these.

amdeel: Oh cool, I didn't realize I could do it this way.

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/Orchestrator/MethodAnalyzer.cs
cgillum: Cool. I suggest adding a comment explaining that this is intentional. I've never actually seen anyone do this before in C# so it might not be obvious to future readers of this code (for example, I could see someone changing it not knowing it was done intentionally).

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/Orchestrator/TimerAnalyzer.cs
```diff
@@ -21,56 +21,53 @@ public class TimerAnalyzer : DiagnosticAnalyzer
...
+        private static DurableVersion version;
 
-        public override void Initialize(AnalysisContext context)
+        internal static bool RegisterDiagnostic(SyntaxNode method, CompilationAnalysisContext context, SemanticModel semanticModel)
         {
-            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
-            context.EnableConcurrentExecution();
-            context.RegisterSyntaxNodeAction(AnalyzeIdentifierTask, SyntaxKind.IdentifierName);
-            context.RegisterSyntaxNodeAction(AnalyzeIdentifierThread, SyntaxKind.IdentifierName);
+            return (AnalyzeIdentifierTask(method, context, semanticModel) |
```
cgillum: Same comment about adding a code comment describing your intentional use of the `|` operator here.

src/WebJobs.Extensions.DurableTask.Analyzers/MethodInformation.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers
+{
+    internal class MethodInformation
```
cgillum: It looks like this class is only used by `MethodAnalyzer`. If so, I suggest you make it a private inner class of `MethodAnalyzer` rather than making it a top-level class. One less file in the repo and it makes the expected usage more well-defined. :)

PR #1166, opened 2020-01-15
Local HTTP server for out-of-proc

cgillum:
# Problem
Today the JS client makes API calls like `startNew`, `getStatus`, etc. by making HTTP calls to the function app's webhook handler. This was convenient at the time but has some serious limitations:

* **High latency**: These HTTP calls involve DNS lookups and go outside the hosting VM and through the Azure load balancer.
* **Unreliable**: All the layers in between the client and the server make the calls fragile. Cold-start also seems to create occasional HTTP 503 problems.
* **Not scalable**: The VMs running the client functions have limited SNAT allocation and outbound TCP connection limits, and these API calls consume these limited resources.
* **Incompatible with external auth**: These APIs don't work at all when Easy Auth is protecting access to the function app.

### Proposed solution - Local HTTP endpoint
The idea is to create a local communication channel between the language worker and the durable extension to avoid the expensive HTTP roundtrips that we have today. Originally we considered gRPC but more recently we decided to use vanilla HTTP over localhost because of the significantly lower engineering cost. gRPC might make a comeback in the future, but leveraging the rich bindings work done by the Functions core team.

The idea is that the extension will open a local HTTP endpoint, separate from the HTTP endpoint exposed by the functions host:

```
http://127.0.0.1:17071/durabletask/
```

This endpoint is unauthenticated but will not be accessible outside of the local VM/container/sandbox. It will directly forward any requests it receives to the internal HTTP handler that manages the existing set of [HTTP APIs](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-http-api).

The expected, relative benefits include:

* **Low latency**: No network calls outside of the loopback adapter and no cold-starts to deal with.
* **Reliable**: The endpoint will always be available.
* **Scalable**: Does not consume any SNAT or outbound TCP connection resources.
* **Works with external auth**: All communication is internal, so it cannot be interfered with by external gateways or features like Easy Auth.

This PR does not require any changes to the Durable JS npm module. The URLs used by that module are already passed as parameters by the extension, and this PR simply passes internal endpoint URLs instead of external endpoint URLs.

### Validation tasks
- [x] New unit tests ensuring that we can pass the internal URLs to an out-of-proc client function.
- [x] New integration tests showing that the internal URLs are accessible and function as expected. 
- [x] Tracing
- [x] Manually validate that local JS development works as expected and new traces appear
- [x] Manually validate that Azure deployments work as expected
- [x] Stress test

Resolves https://github.com/Azure/azure-functions-durable-js/issues/89
Resolves https://github.com/Azure/azure-functions-durable-js/issues/62


anthonychu:
Looking good! Wonder if it might be better to pick an unused port instead of hardcoding it. The host does it like this: https://github.com/Azure/azure-functions-host/blob/15fb05f849aa96023eae310ab78dd81114b450b9/src/WebJobs.Script/Workers/Http/Configuration/HttpWorkerOptionsSetup.cs#L52

cgillum:
@anthonychu thanks for the pointer! I wasn't sure about how to search for available ports and that would definitely be a nice improvement to avoid the possibility of conflicts. Perhaps we could add that as a separate PR? (been working on this so long and I'm trying to get this off my plate 😅)

anthonychu:
Yep sounds good. Just wanted to point this out in case for some reason that port is already in use. Should be super rare.

src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
```diff
@@ -37,6 +37,14 @@ public string DurableOrchestrationClientToString(IDurableOrchestrationClient cli
                 BaseUrl = this.config.HttpApiHandler.GetBaseUrl(),
                 RequiredQueryStringParameters = this.config.HttpApiHandler.GetUniversalQueryStrings(),
             };
+
+            if (this.config.Options.LocalRpcEndpointEnabled != false)
+            {
+                // If this field is missing, the out-of-proc SDK is expected to either fail
```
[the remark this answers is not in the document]
cgillum: I agree we should pick a standard for this, and we can document that standard. I added this comment because we already know that JS is going to do one thing (revert to old behavior) whereas Python will do something else (fail). The only reason to have JS revert to the old behavior is for backwards compatibility with older versions of the extension. New languages should always expect that the RPC endpoint will be available.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -241,8 +252,45 @@ void IExtensionConfigProvider.Initialize(ExtensionConfigContext context)
...
+                // Default behavior is to only start the local RPC server for non-.NET function languages.
+                // We'll enable it if it's non-.NET or if the FUNCTIONS_WORKER_RUNTIME value isn't present.
+                string functionsWorkerRuntime = this.nameResolver.Resolve("FUNCTIONS_WORKER_RUNTIME");
+                shouldEnable = !string.Equals(functionsWorkerRuntime, "dotnet", StringComparison.OrdinalIgnoreCase);
+            }
+
+            // This is only applicable to Functions 2.0 and above
+            if (shouldEnable == true)
+            {
+                this.HttpApiHandler.StartLocalHttpServerAsync().GetAwaiter().GetResult();
```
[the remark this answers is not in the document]
cgillum: From my experience, this is the best way to handle it. It ensures that if an exception is thrown, we get the actual exception we care about rather than an AggregateException, which is what `.Wait()` or `.Result` would throw if there was a failure.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -193,9 +212,21 @@ public async Task<HttpResponseMessage> HandleRequestAsync(HttpRequestMessage req
...
+                else if (this.config.Options.NotificationUrl != null)
+                {
+                    basePath = this.config.Options.NotificationUrl.AbsolutePath;
+                }
+                else
+                {
+                    throw new InvalidOperationException($"Don't know how to handle request to {request.RequestUri}.");
+                }
+
+                string path = "/" + request.RequestUri.AbsolutePath.Substring(basePath.Length).Trim('/');
```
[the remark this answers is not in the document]
cgillum: That assumes there isn't a `/` in the beginning. I'm doing this intentionally to account for all cases for `/` appearing at the beginning or end.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -878,6 +909,8 @@ internal string GetBaseUrl()
             return hostUrl + notificationUri.AbsolutePath.TrimEnd('/');
         }
 
+        internal string GetRpcBaseUrl() => InternalRpcUri.OriginalString;
```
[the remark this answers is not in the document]
cgillum: I'm not sure I understand the benefit of the TryGet pattern here. We will always return a value.

src/WebJobs.Extensions.DurableTask/IHostLifetime.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    /// <summary>
+    /// Service interface for signaling the extension when the host is starting up or shutting down.
+    /// </summary>
+    /// <remarks>
+    /// This interface is expected to be used as an injected service.
+    /// </remarks>
+    public interface IHostLifetime
```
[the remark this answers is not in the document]
cgillum: Hmm...I know there was another interface for this but it had a different name. Any suggestions on how to best avoid confusion? I considered IFunctionHostLifetime but that could still be confused with something that's defined by the Functions runtime.

src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
```diff
@@ -0,0 +1,137 @@
...
+            this.extension = extension ?? throw new ArgumentNullException(nameof(extension));
+            this.handler = handler ?? throw new ArgumentNullException(nameof(handler));
+            this.localWebHost = this.CreateWebHost(listenUri);
+        }
+
+        public bool IsListening { get; private set; }
+
+        public void Dispose()
+        {
+            this.localWebHost?.Dispose();
```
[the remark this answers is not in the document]
cgillum: Because we will initialize `localWebHost` to `null` if this is running in Functions v1.

One could argue that we shouldn't even go down this code path at all for Functions v1, but my concern is that would make the end-to-end code even harder to reason about due to the additional conditional compilation.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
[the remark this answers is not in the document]
ConnorMcMahon: I definitely see your point.

My thinking was along the line of making it more clear when to use the result of this method. For instance, the only current caller of this method wraps the call with a check that `this.config.Options.LocalRpcEndpointEnabled != false`, because that is really the only case where we need this URL. The TryGet convention makes it a bit more clear to to callers that it might not make sense to use the result from this method, even if it always returns a non-null value.

src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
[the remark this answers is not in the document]
ConnorMcMahon: What if in `this.CreateWebHost()` we just create a dummy `IWebHost` that no-ops on all of it's interface methods? It would make the only place we need conditional compilation that method, and reduce the likelihood of null reference exceptions.

src/WebJobs.Extensions.DurableTask/IHostLifetime.cs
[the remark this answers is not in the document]
ConnorMcMahon: Is there a reason we even have this custom interface? It seems like our only implementation is just a wrapper for `IApplicationLifetime`, and I don't immediately see the gain in that wrapper.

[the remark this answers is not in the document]
cgillum: It was to avoid the complexity of introducing an interface that is available only in `netstandard2.0`. Having my own interface essentially reduced the number of scope of `#if` statements I needed throughout the code (I think test code was also implicated in this).

src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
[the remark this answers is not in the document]
cgillum: That's a good suggestion. Let me look into that.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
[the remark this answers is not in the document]
cgillum: Ah, actually yes I agree this would be cleaner. I think I missed this because I only saw a small snippet of the calling code and not the lines around it. Let me look into this.

src/WebJobs.Extensions.DurableTask/IHostLifetime.cs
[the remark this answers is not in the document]
ConnorMcMahon: I see... that makes sense. Can you add this detail to the summary section? Maybe consider a name like `ICrossFunctionsVersionAppLifeTime` to indicate why we have a new interface?

In general, this is less of an issue in Visual Studio, because I could have dove into the interface and saw that it was a custom interface, not the .NET Standard interface.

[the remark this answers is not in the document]
cgillum: I'll rename it to `IApplicationLifetimeWrapper`. I don't want to make it more descriptive than that because it could mislead people into thinking it has more complex semantics than the underlying `IApplicationLifetime`.

src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
```diff
@@ -0,0 +1,153 @@
...
+        public void Dispose() => this.localWebHost.Dispose();
+
+        public async Task StartAsync()
+        {
+            if (this.IsListening == true)
+            {
+                throw new InvalidOperationException("The local HTTP listener has already started.");
+            }
+
+#if !FUNCTIONS_V1
```
ConnorMcMahon: We no longer need conditional compilation here now that `this.localWebHost() is a `NoOpWebHost`.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -984,6 +1017,51 @@ private void ThrowIfWebhooksNotConfigured()
...
+
+            // The app owner explicitly disabled the local RPC endpoint.
+            rpcBaseUrl = null;
+            return false;
+        }
+
+#if !FUNCTIONS_V1
+        internal async Task StartLocalHttpServerAsync()
+        {
+            if (!this.localHttpListener.IsListening)
```
ConnorMcMahon: Why can't we do these checks and logging inside of `LocalHttpListener.Start/StopAsync()` and make those methods idempotent. Then we don't need to do the conditional compilation guards around this method or in DurableTaskExtension.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -57,6 +57,7 @@
     <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.14" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.WebApiCompatShim" Version="2.2.0" />
     <PackageReference Include="Microsoft.AspNetCore.Routing" Version="2.2.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel" Version="2.2.0" />
```
ConnorMcMahon: I have a slight concern that this may cause a conflict in Functions V3, as they may have changed their internal dependency on this package since they are on .NET Core 3.1.

test/FunctionsV2/OutOfProcTests.cs
```diff
@@ -122,5 +138,209 @@ public async Task CallHttpActionOrchestrationWithManagedIdentity()
...
+                //   "rpcBaseUrl": "http://127.0.0.1:17071/durabletask/" (or null)
+                // }
+
+                Assert.True(outerJson.TryGetValue("taskHubName", out JToken taskHubName));
+                Assert.StartsWith(nameof(this.BindToDurableClientAsString), (string)taskHubName);
+
+                Func<JToken, string, string[], Uri> commonUriValidation = (
+                    JToken json,
+                    string fieldName,
+                    string[] requiredSegments) =>
```
ConnorMcMahon: Not quite sure what we gain from this requiredSegments approach instead of just constructing the expected URI. The URIs seem deterministic, so it seems like we should use those instead of just checking if the path contains some segments.

test/FunctionsV2/OutOfProcTests.cs
```diff
@@ -122,5 +138,209 @@ public async Task CallHttpActionOrchestrationWithManagedIdentity()
...
+                await host.StartAsync();
+
+                // Check to see whether the local RPC endpoint has been opened
+                IPGlobalProperties ipGlobalProperties = IPGlobalProperties.GetIPGlobalProperties();
+                IPEndPoint[] endpoints = ipGlobalProperties.GetActiveTcpListeners();
+
+                const string LocalRcpAddress = "127.0.0.1:17071";
+                if (enabledExpected)
+                {
+                    Assert.Contains(LocalRcpAddress, endpoints.Select(ep => ep.ToString()));
```
ConnorMcMahon: To be a bit more robust, we should check that this endpoint is not already open before the test. If it is, is it possible for a test to be marked as inconclusive?

test/FunctionsV2/OutOfProcTests.cs
```diff
@@ -122,5 +138,209 @@ public async Task CallHttpActionOrchestrationWithManagedIdentity()
...
+                localRpcEndpointEnabled: true,
+                notificationUrl: null))
+            {
+                await host.StartAsync();
+
+                using (var client = new WebClient())
+                {
+                    string jsonString = client.DownloadString("http://localhost:17071/durabletask/instances");
+
+                    // The data doesn't matter, but the result should be an array
```
ConnorMcMahon: This comment is just meant to indicate this test does not care about the data because this test is just making sure the request is actually going through, right? Presumably the schema of this data is expected to match the webhook call.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
cgillum: I had the same fear, but then I realized this was the latest version of Kestrel, so they couldn't have a newer dependency than this. :)

ConnorMcMahon: Good to know! 

If I remember from my previous porting to .NET Core 3.0, they started shipping a lot of these packages as a part of .NET Core, so you no longer explicitly reference the package.

That being said, it is still probably a good idea to do a final end-to-end test with both Functions V2 and Functions V3.

src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
cgillum: It's still needed because `IWebHost.StartAsync()` and `IWebHost.StopAsync()` are not defined in `net471`.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
cgillum: I've actually verified both already (Functions 3.0 locally and Functions 2.0 on Azure), so I think we're good.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
cgillum: That's possibly true, but I'm afraid of accidentally breaking v1 with this code path, even though I've gone through all the trouble of no-op'ing it. I think at this point I'd prefer to avoid additional churn and will just leave this as-is for now.

test/FunctionsV2/OutOfProcTests.cs
cgillum: Might have been an artifact from a previous iteration. I'll see if I can clean it up.

test/FunctionsV2/OutOfProcTests.cs
cgillum: I suppose, but if it was already opened then the test would fail when we do `host.StartAsync()`.

cgillum: It also looks like xunit doesn't allow you to mark a test as inconclusive from code.

src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
ConnorMcMahon: That is annoying. We should probably remove the Start/StopAsync() methods on `NoOpWebHost`, since they are functionally useless due to the differences in IWebHost in netstandard2.0 and net471.

test/FunctionsV2/OutOfProcTests.cs
ConnorMcMahon: Fair enough. Let's just leave it as is.

test/FunctionsV2/OutOfProcTests.cs
cgillum: Yes, this is why I'm doing `JArray.Parse(jsonString)`. I'm expecting the result to be an array and this will throw if it's not an array. In theory it should be an empty array, but whether it is empty or not doesn't really matter from the POV of this test.

test/FunctionsV2/OutOfProcTests.cs
cgillum: I simplified the structure of this test quite a bit, but constructing the full URLs were going to be more work than I'm willing to do at this point. I'm okay with revisiting this next time we need to update this code if we think stricter validation will improve the quality of this test.

PR #1168, opened 2020-01-15
Deprecated non-paged GetStatusAsync APIs

### LINKED ISSUE #1116: Orchestration query API should default to 100 instances
In cases where there are more than 100 instances in the instance store, this request can take a huge amount of time to complete by default. We should set a reasonable default of 100 instances (which matches the new ListEntities API).

### LINKED ISSUE #1147: Deprecate non-paged IDurableOrchestrationClient.GetStatusAsync APIs
We have too many redundant overloads of `IDurableOrchestrationClient.GetStatusAsync()`. Some of them also don't support paging, which can result in really slow, really expensive queries.

We should just have a single `GetStatusAsync()` API. Even better if we can rename it to `ListInstancesAsync()` (or something similar), which is a more descriptive name. Basically, this overload would be preferred:
```csharp
Task<OrchestrationStatusQueryResult> ListInstancesAsync(
    OrchestrationStatusQueryCondition filter,
    CancellationToken cancellationToken);
```
We can't make a breaking change, though, so other overloads need to be preserved. However we should decorate them with the `[Obsolete]` attribute so that developers know not to take a dependency on those APIs. In a Durable 3.0 release we can delete the deprecated APIs.

We could update the overloads to all take paging parameters, but it would be cleaner (I think) to just have a single flexible overload.

**Additional context**
Related to https://github.com/Azure/azure-functions-durable-extension/issues/1116.

amdeel:
resolves #1147 
resolves #1116

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -378,10 +378,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private OrchestrationStatusQueryCondition CreateConditionFromParameters(DateTime? createdTimeFrom, DateTime? createdTimeTo, IEnumerable<OrchestrationRuntimeStatus> runtimeStatus)
         {
-            var condition = new OrchestrationStatusQueryCondition
-            {
-                PageSize = int.MaxValue,
-            };
+            var condition = new OrchestrationStatusQueryCondition();
 
             if (createdTimeFrom != null)
             {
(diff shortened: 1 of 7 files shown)


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -358,28 +358,44 @@ async Task<DurableOrchestrationStatus> IDurableOrchestrationClient.GetStatusAsyn
...
-            }
-
-            return results;
+            return await this.GetAllStatusHelper(null, null, null, cancellationToken);
         }
 
         /// <inheritdoc />
         async Task<IList<DurableOrchestrationStatus>> IDurableOrchestrationClient.GetStatusAsync(DateTime createdTimeFrom, DateTime? createdTimeTo, IEnumerable<OrchestrationRuntimeStatus> runtimeStatus, CancellationToken cancellationToken)
         {
-            IList<OrchestrationState> states = await this.DurabilityProvider.GetAllOrchestrationStatesWithFilters(createdTimeFrom, createdTimeTo, runtimeStatus, cancellationToken);
```
cgillum: Does your change also allow us to deprecate/remove `GetAllOrchestrationStatesWithFilters`?

amdeel: I didn't notice this before, but yes, it does.

cgillum:
It looks like the AppVeyor CI is failing because of build warnings. @amdeel can you update the PR to fix the build warnings? Once that's done, I'm good to sign off.

@ConnorMcMahon it would be good to have you review this as well, especially now that we're touching the durability provider abstract base class.

Issue #1152, opened 2020-01-09
Return message of Durable Entity is lost after the first time


salv85:
### Description
Return message of Durable Entity is lost after the first time.

To replicate this behaviour:

1. Send one message to the client function like this:
```json
{
	"id" : 1,
	"orchestratorName": "TestOrchestrator",
	"sampleText": "Hello World 1"
}
```
2. Wait for completion 

3. Send one or more new messages to the client function like this (do not change the orchestrator name):
```json
{
	"id" : 2,
	"orchestratorName": "TestOrchestrator",
	"sampleText": "Hello World 2"
}
```
4. The orchestrator won't receive the second message

Logs reports this:
```
Function 'TestOrchestrator (Orchestrator)' saved a '802d29f8-2b26-5d60-9197-1d34253271f6' event to an in-memory queue. State: ExternalEventDropped. HubName: TestTaskHub. AppName: . SlotName: . ExtensionVersion: 2.1.0. SequenceNumber: 220.
```

### Expected behavior
_CallEntityAsync_ returns the message every time.

### Actual behavior
_CallEntityAsync_ returns the message only the first time.

### Relevant source code snippets
```csharp
using System;
using System.Collections.Generic;
using System.Linq;
using System.Net.Http;
using System.Threading.Tasks;
using Microsoft.Azure.Functions.Extensions.DependencyInjection;
using Microsoft.Azure.WebJobs;
using Microsoft.Azure.WebJobs.Extensions.DurableTask;
using Microsoft.Azure.WebJobs.Extensions.Http;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Logging;
using Microsoft.Rest;
using Newtonsoft.Json;

[assembly: FunctionsStartup(typeof(TestFunction.Startup))]

namespace TestFunction
{
    public class Startup : FunctionsStartup
    {
        public override void Configure(IFunctionsHostBuilder builder)
        {
            if (builder == null) throw new ArgumentNullException(nameof(builder));
            builder.Services.AddLogging();
        }
    }

    public class ClientModel
    {
        public int Id { get; set; }
        public string OrchestratorName { get; set; }
        public string SampleText { get; set; }
    }

    public class OrchestratorStatus
    {
        public string SampleText { get; set; }
        public bool? Action1Completed { get; set; }
        public int? RandomNumber { get; set; }
        public bool? Action2Completed { get; set; }
    }

    public class ClientFunction
    {
[...]

PR #1169, opened 2020-01-16
Fix message sorter

sebastianburckhardt:
Previously, for messages sent to entities, the execution id of the originating orchestration was not being tracked, only the instance id. This can lead to problems with the message sorter logic, as discussed in #1152. I added

- tracking and tracing of execution id for messages sent from orchestrations to entities
- include this execution id in events representing responses
- add a testcase where an orchestration sends multiple messages separated by continueAsNew
- modify the message sorter so it starts over when receiving a message from a different execution id

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
@@ -554,7 +554,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             string hubName,
             string functionName,
             string instanceId,
-            string requestingInstance,
+            string requestingInstanceId,
+            string requestingExecutionId,
             string requestId,
             bool isReplay)
         {
@@ -566,7 +567,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 LocalSlotName,
                 functionName,
                 instanceId,
-                requestingInstance,
+                requestingInstanceId,
+                requestingExecutionId,
                 requestId,
                 FunctionType.Entity.ToString(),
                 ExtensionVersion,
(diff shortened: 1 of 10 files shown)


src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -575,8 +577,8 @@ public void ExtensionWarningEvent(string hubName, string functionName, string in
             if (this.ShouldLogEvent(isReplay: isReplay))
             {
                 this.logger.LogInformation(
-                    "{instanceId}: Function '{functionName} ({functionType})' granted lock to request {requestId} by instance {requestingInstance}. State: {state}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
-                    instanceId, functionName, functionType, requestId, requestingInstance, FunctionState.LockAcquired, hubName,
+                    "{instanceId}: Function '{functionName} ({functionType})' granted lock to request {requestId} by instance {requestingInstanceId} {requestingExecutionId}. State: {state}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
```
cgillum: Do we want some text in between {requestingInstanceId} and {requestingExecutionId} in the log message?

src/WebJobs.Extensions.DurableTask/EntityScheduler/RequestMessage.cs
```diff
@@ -43,6 +43,12 @@ internal class RequestMessage
         [JsonProperty(PropertyName = "parent", DefaultValueHandling = DefaultValueHandling.Ignore)]
         public string ParentInstanceId { get; set; }
 
+        /// <summary>
+        /// The parent instance that called this operation.
+        /// </summary>
+        [JsonProperty(PropertyName = "parentEx", DefaultValueHandling = DefaultValueHandling.Ignore)]
```
cgillum: "Ex" as a suffix isn't super clear. I would suggest at least "parentExecution". I'm not too worried about the extra 7 characters.

src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
```diff
@@ -301,20 +301,21 @@ private EtwEventSource()
...
-            string RequestingInstance,
+            string RequestingInstanceId,
+            string RequestingExecutionId,
             string RequestId,
             string FunctionType,
             string ExtensionVersion,
             bool IsReplay)
         {
-            this.WriteEvent(219, TaskHub, AppName, SlotName, FunctionName, InstanceId, RequestingInstance, RequestId, FunctionType, ExtensionVersion, IsReplay);
+            this.WriteEvent(219, TaskHub, AppName, SlotName, FunctionName, InstanceId, RequestingInstanceId, RequestingExecutionId, RequestId, FunctionType, ExtensionVersion, IsReplay);
```
cgillum: Since `RequestingExecutionId` can be null, we need to change this to `RequestingExecutionId ?? ""`. I think it's better to do that here rather than try to remember doing it at the point in which this method gets called.

sebastianburckhardt: sure.

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
sebastianburckhardt: ok.

src/WebJobs.Extensions.DurableTask/EntityScheduler/RequestMessage.cs
sebastianburckhardt: Sure.

PR #1171, opened 2020-01-18
Handle OutOfProc exceptions properly

ConnorMcMahon:
In order to prevent non-determinism errors when out-of-proc Durable orchestrations throw an exception, we need to read state from the RpcException. We can then properly replay the actions performed by the out-of-proc orchestration before throwing the exception.


src/WebJobs.Extensions.DurableTask/Listener/TaskActivityShim.cs
```diff
@@ -61,6 +61,13 @@ public override async Task<string> RunAsync(TaskContext context, string rawInput
                 // without the outer FunctionInvocationException.
                 Exception exceptionToReport = StripFunctionInvocationException(result.Exception);
 
+                // If the exception looks like an RPC exception, wrap it in an OutOfProcessException
```
ConnorMcMahon: This comment is out of date.

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -80,7 +80,7 @@ public override async Task<string> Execute(OrchestrationContext innerContext, st
                     this.context.IsReplaying));
             }
 
-            object returnValue;
+            object returnValue = null;
```
cgillum: I'm a little concerned that you have to assign this to `null` here. I usually try to make an effort to avoid this to reduce the chances of `NullReferenceExceptions` and to make correctness easier to reason about. I'm wondering if you can refactor the `catch` block a bit to avoid needing to do this. The `TryExtract...` pattern I suggested elsewhere might make this easier.

src/WebJobs.Extensions.DurableTask/Listener/TaskActivityShim.cs
```diff
@@ -61,6 +61,13 @@ public override async Task<string> RunAsync(TaskContext context, string rawInput
...
+                // to strip unhelpful information from message.
+                if (OutOfProcExceptionHelpers.IsRpcException(exceptionToReport))
```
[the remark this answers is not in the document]
ConnorMcMahon: We can name the method differently. The reason I named it `IsRpcException` is because it is a best effort to catch the exception type `RpcException`, which is thrown by the Functions runtime. Eventually, this check could be replaced with a check of `exceptionToReport is RpcException`.

That being said, that is kind of an implementation detail, and I can definitely see the merit in renaming this `IsOutOfProcException` to hide that detail.

src/WebJobs.Extensions.DurableTask/OutOfProcExceptionHelpers.cs
```diff
@@ -0,0 +1,82 @@
...
+                return outOfProcMessage.Substring(jsonStart, jsonLength);
+            }
+        }
+
+        private static string GetFullOutOfProcMessage(Exception ex)
+        {
+            string rpcExceptionMessage = ex.Message;
+
+            int messageStart = rpcExceptionMessage.IndexOf(MessageLabel) + MessageLabel.Length;
+            int messageEnd = rpcExceptionMessage.IndexOf(StackTraceLabel);
```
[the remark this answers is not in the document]
ConnorMcMahon: Following the `Try*` pattern, it is now more obvious that these values will never be negative.

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -178,6 +131,65 @@ public override async Task<string> Execute(OrchestrationContext innerContext, st
...
+                {
+                    throw new InvalidOperationException("The WebJobs runtime returned a invocation task that does not support return values!");
+                }
+            }
+            catch (Exception e)
+            {
+                if (OutOfProcExceptionHelpers.TryExtractOutOfProcStateJson(e.InnerException, out string returnValue)
+                    && !string.IsNullOrEmpty(returnValue))
+                {
+                    return returnValue;
```
cgillum: If we get here, does it mean the orchestrator function failed with an unhandled error? If so, don't we still need to trace `FunctionFailed` and send the `OrchestratorFailed` lifecycle event?

ConnorMcMahon: Hmm probably. At least part of this is handled in `HandleOutOfProcExecution()` in the below code:

```csharp
if (!string.IsNullOrEmpty(execution.Error))
{
    var orchestrationException = new OrchestrationFailureException(
    $"Orchestrator function '{this.context.Name}' failed: {execution.Error}");
    this.context.OrchestrationException = ExceptionDispatchInfo.Capture(orchestrationException);
    throw orchestrationException;
}
```

That being said, the above code block is missing some logging and the lifecycle event. I will factor out the rest of this `catch` block into a method that can be called from within `HandleOutOfProcExecution()`.

ConnorMcMahon: I have completely refactored how I handle this, due to needing the `Exception` thrown here to be available in `HandleOutOfProcExecution`. Would love some feedback on this new approach.

PR #1177, opened 2020-01-24
Make ControlQueueBufferThreshold configurable for Durable 1.x

ConnorMcMahon:
This buffer size appears to impact throughput in a substantial way, so
we are allowing customers to configure this in their host.json to try
and optimize throughput.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
@@ -42,11 +42,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// <summary>
         /// Gets or set the number of control queue messages that can be buffered in memory
         /// at a time, at which point the dispatcher will wait before dequeuing any additional
-        /// messages. The default is 64.
+        /// messages. The default is 256. The maximum value is 1000.
         /// </summary>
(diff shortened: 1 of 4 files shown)


ConnorMcMahon:
Should probably update the documentation for this once 1.8.5 releases.

src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
```diff
@@ -39,6 +39,15 @@ public class DurableTaskOptions
...
         public int ControlQueueBatchSize { get; set; } = 32;
 
+        /// <summary>
+        /// Gets or set the number of control queue messages that can be buffered in memory
+        /// at a time, at which point the dispatcher will wait before dequeuing any additional
+        /// messages. The default is 64.
+        /// </summary>
+        /// <remarks>This has historically always been fixed to 64, but increasing it may increase
+        /// throughput.</remarks>
+        public int ControlQueueBufferThreshold { get; set; } = 64;
```
cgillum: In V2 this is defaulted to 256. I suggest we do the same here. Also, consider copy/pasting my remarks there here: https://github.com/Azure/azure-functions-durable-extension/blob/ee7bc391671188268c005ba94aef4fa3e0dae44a/src/WebJobs.Extensions.DurableTask/Options/AzureStorageOptions.cs#L54-L65

PR #1182, opened 2020-01-28
Analyzer: Changed entity analyzers and added extra checks for test projects

amdeel:
Continued work to slowly transition all analyzers to compilation analyzers (joke).

Changed Entity ClassName, Dispatch, and Static analyzers to be compilation analyzers. They now more accurately analyze code.

Entity analyzer changes:
*ClassName now checks if a FunctionName above an EntityTrigger method matches any class in existence.
*Dispatch checks that the class it's being used with matches the FunctionName it's used within.
*Static is now called from the dispatch analyzer. All the methods that get analyzed by the dispatch analyzer get stored in a list. If that method has a FunctionName attribute matching the class it's contained it (is inside an entity class) it will flag the method name if it is not marked static.

I got rid of the EntityClassName CodeFixProvider. Instead, I added in the messageFormat description string a suggestion for the closest class name string matching what they typed in their FunctionName attribute.

I also changed some analyzers to check if it's being used in a Function (has the FuncitonName attribute). This is to prevent extra warnings on test code.


src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/EntityInterface/InterfaceAnalyzer.cs
```diff
@@ -44,21 +55,18 @@ public override void Initialize(AnalysisContext context)
...
+                ParameterAnalyzer.ReportProblems(context, entityInterface);
+                EntityInterfaceReturnTypeAnalyzer.ReportProblems(context, entityInterface);
             }
         }
 
         public void FindEntityCalls(SyntaxNodeAnalysisContext context)
         {
-            if (context.Node is MemberAccessExpressionSyntax expression)
+            var expression = context.Node as MemberAccessExpressionSyntax;
+            if (expression != null && SyntaxNodeUtils.IsInsideFunction(expression))
```
ConnorMcMahon: You can simplify this back to 

`if (context.Node is MemberAccessExpressionSyntax expression && SyntaxNodeUtils.IsInsideFunction(expression))`

src/WebJobs.Extensions.DurableTask.Analyzers/SyntaxNodeUtils.cs
```diff
@@ -24,7 +25,12 @@ public static DurableVersion GetDurableVersion(SemanticModel semanticModel)
             version = versionTwoInterface != null ? DurableVersion.V2 : DurableVersion.V1;
             return (DurableVersion)version;
         }
-        
+
+        public static string GetClosestString(string name, IEnumerable<string> availableNames)
+        {
+            return availableNames.OrderBy(x => x.LevenshteinDistance(name)).First();
```
ConnorMcMahon: What behavior do we want to have if `availableNames` is empty? Right now it will throw an exception.

PR #1187, opened 2020-01-29
Update HttpApiHandler to work with AllowSynchronousIO = false (#1085)

### LINKED ISSUE #1178: Fix Functions V3 async deserialization in Durable extension V1
- [x] Port [this fix](https://github.com/Azure/azure-functions-durable-extension/pull/1085) to V1. Anyone using bundles V1 (which will be the default for some time) will still be using Durable V1. 
- [x] Release an updated extension bundle that contains this fix.

ConnorMcMahon:
Functions V3 now uses .NET Core 3.1. Starting in .NET Core 3.0, AllowSynchronousIO defaults to false. Therefore, we needed to change our JSON deserialization logic in the Start Orchestration API to avoid synchronous stream reads

Fixes #1178

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
+++ b/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.Diagnostics;
-using System.IO;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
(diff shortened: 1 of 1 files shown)


src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -483,15 +482,10 @@ private static int GetIntQueryParameterValue(NameValueCollection queryStringName
                 DurableOrchestrationClientBase client = this.GetClient(request);
 
                 object input = null;
-                if (request.Content != null)
+                if (request.Content?.Headers?.ContentLength != 0)
```
cgillum: This is actually not fully correct. It should be this (I've made this change in another PR):
```suggestion
                if (request.Content != null && request.Content.Headers.ContentLength != 0)
```

PR #1197, opened 2020-02-03
Update WaitForExternalEvent to use CancellationTokens

### LINKED ISSUE #1041: WaitForExternalEvent uncancellable timer fix was dropped from v2.0
Opening a new bug to track this; original bug was #816.

@ConnorMcMahon / @cgillum Sadly, it looks like my original change for #816 was completely blown away shortly after getting merged into dev, and none of it ended up getting included in the v2 release. ☹️

As best as I can tell, it looks like the fault lies with 780749d3a3fbbc648b04e71e2d17afc1efe789d4; there were tons of merge conflicts, but no work appears to have been done to actually resolve most of them, instead just ignoring them.

Trying to replay that merge locally yields the following:
```
You have unmerged paths.
  (fix conflicts and run "git commit")
  (use "git merge --abort" to abort the merge)

Changes to be committed:

        modified:   .github/ISSUE_TEMPLATE/bug_report.md
        new file:   .github/ISSUE_TEMPLATE/feature_request.md
        new file:   .github/ISSUE_TEMPLATE/new-release-template.md
        modified:   samples/v1/javascript/package-lock.json
        modified:   samples/v1/javascript/package.json

Unmerged paths:
  (use "git add/rm <file>..." as appropriate to mark resolution)

        both modified:   samples/v1/precompiled/VSSample.csproj
        deleted by us:   src/WebJobs.Extensions.DurableTask/DurableOrchestrationContext.cs
        deleted by us:   src/WebJobs.Extensions.DurableTask/DurableOrchestrationContextBase.cs
        both modified:   src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
        deleted by us:   src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
        both modified:   src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
        both modified:   src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask.xml
        both modified:   src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
        both modified:   test/Common/DurableTaskEndToEndTests.cs
        both modified:   test/Common/TestHelpers.cs
        both modified:   test/Common/TestOrchestrations.cs
        both modified:   test/FunctionsV1/WebJobs.Extensions.DurableTask.Tests.V1.csproj
```

What's the right plan of action here?  Apologies for not noticing this sooner, I thought you were going to attempt a v1.8.4 release prior to shipping v2.0. ☹️

AtOMiCNebula:
Rebased original fix from #816 that was lost in the run-up to v2.0.

Previously, `WaitForExternalEvent` when invoked with a specified timeout value would create a Timer instance that was uncancelable by the surrounding orchestration.  This could have the practical effect of making orchestrations hang until either the event was received (causing the timer to be cancelled) or the timeout was reached (because the timer expired).

This change adds overloads for `TimeSpan`-accepting overloads of `WaitForExternalEvent` that allow providing a `CancellationToken`, which the calling orchestration can cancel if it no longer intends to wait for the event to be raised.

Fixes #1041

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
@@ -223,12 +223,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// </remarks>
         /// <param name="name">The name of the event to wait for.</param>
         /// <param name="timeout">The duration after which to throw a TimeoutException.</param>
+        /// <param name="cancelToken">The <c>CancellationToken</c> to use for cancelling <paramref name="timeout"/>'s internal timer.</param>
         /// <typeparam name="T">Any serializeable type that represents the JSON event payload.</typeparam>
         /// <returns>A durable task that completes when the external event is received.</returns>
         /// <exception cref="TimeoutException">
         /// The external event was not received before the timeout expired.
         /// </exception>
-        Task<T> WaitForExternalEvent<T>(string name, TimeSpan timeout);
+        Task<T> WaitForExternalEvent<T>(string name, TimeSpan timeout, CancellationToken cancelToken);
 
         /// <summary>
         /// Waits asynchronously for an event to be raised with name <paramref name="name"/> and returns the event data.
@@ -240,10 +241,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// <param name="name">The name of the event to wait for.</param>
         /// <param name="timeout">The duration after which to return the value in the <paramref name="defaultValue"/> parameter.</param>
         /// <param name="defaultValue">The default value to return if the timeout expires before the external event is received.</param>
+        /// <param name="cancelToken">The <c>CancellationToken</c> to use for cancelling <paramref name="timeout"/>'s internal timer.</param>
         /// <typeparam name="T">Any serializeable type that represents the JSON event payload.</typeparam>
         /// <returns>A durable task that completes when the external event is received, or returns the value of <paramref name="defaultValue"/>
         /// if the timeout expires.</returns>
-        Task<T> WaitForExternalEvent<T>(string name, TimeSpan timeout, T defaultValue);
+        Task<T> WaitForExternalEvent<T>(string name, TimeSpan timeout, T defaultValue, CancellationToken cancelToken);
 
         /// <summary>
         /// Acquires one or more locks, for the specified entities.
--- a/test/Common/DurableTaskEndToEndTests.cs
+++ b/test/Common/DurableTaskEndToEndTests.cs
@@ -1114,6 +1114,36 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
         }
 
         /// <summary>
+        /// End-to-end test which validates a CancellationToken-providing overload of WaitForExternalEvent.
+        /// </summary>
+        [Fact]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        public async Task WaitForExternalEventWithCancellationToken()
+        {
+            var orchestratorFunctionNames = new[] { nameof(TestOrchestrations.ApprovalWithCancellationToken) };
+            var extendedSessions = false;
+            using (ITestHost host = TestHelpers.GetJobHost(
+                this.loggerProvider,
+                nameof(this.WaitForExternalEventWithCancellationToken),
+                extendedSessions))
+            {
+                await host.StartAsync();
+
+                var timeout = TimeSpan.FromSeconds(10);
+                var client = await host.StartOrchestratorAsync(orchestratorFunctionNames[0], timeout, this.output);
+                await client.WaitForStartupAsync(this.output);
+
+                await client.RaiseEventAsync("approval", this.output);
+
+                var status = await client.WaitForCompletionAsync(this.output);
+                Assert.Equal(OrchestrationRuntimeStatus.Completed, status?.RuntimeStatus);
+                Assert.InRange(status.LastUpdatedTime - status.CreatedTime, TimeSpan.Zero, timeout);
+
+                await host.StopAsync();
+            }
+        }
+
+        /// <summary>
         /// End-to-end test which validates that orchestrations run concurrently of each other (up to 100 by default).
         /// </summary>
         [Theory]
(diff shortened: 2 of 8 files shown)


src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
```diff
@@ -223,12 +223,13 @@ public interface IDurableOrchestrationContext
...
         /// <param name="name">The name of the event to wait for.</param>
         /// <param name="timeout">The duration after which to throw a TimeoutException.</param>
+        /// <param name="cancelToken">The <c>CancellationToken</c> to use for cancelling <paramref name="timeout"/>'s internal timer.</param>
         /// <typeparam name="T">Any serializeable type that represents the JSON event payload.</typeparam>
         /// <returns>A durable task that completes when the external event is received.</returns>
         /// <exception cref="TimeoutException">
         /// The external event was not received before the timeout expired.
         /// </exception>
-        Task<T> WaitForExternalEvent<T>(string name, TimeSpan timeout);
+        Task<T> WaitForExternalEvent<T>(string name, TimeSpan timeout, CancellationToken cancelToken);
```
ConnorMcMahon: I think we need to include a default parameter here to prevent this from being a breaking change in production code.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
```diff
@@ -240,10 +241,11 @@ public interface IDurableOrchestrationContext
         /// <param name="name">The name of the event to wait for.</param>
         /// <param name="timeout">The duration after which to return the value in the <paramref name="defaultValue"/> parameter.</param>
         /// <param name="defaultValue">The default value to return if the timeout expires before the external event is received.</param>
+        /// <param name="cancelToken">The <c>CancellationToken</c> to use for cancelling <paramref name="timeout"/>'s internal timer.</param>
         /// <typeparam name="T">Any serializeable type that represents the JSON event payload.</typeparam>
         /// <returns>A durable task that completes when the external event is received, or returns the value of <paramref name="defaultValue"/>
         /// if the timeout expires.</returns>
-        Task<T> WaitForExternalEvent<T>(string name, TimeSpan timeout, T defaultValue);
+        Task<T> WaitForExternalEvent<T>(string name, TimeSpan timeout, T defaultValue, CancellationToken cancelToken);
```
ConnorMcMahon: Same here.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
AtOMiCNebula: Are the extension methods I included insufficient?

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
AtOMiCNebula: See earlier comment.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
ConnorMcMahon: Ah you are totally right. The main concern now is that this change will break tests that were mocking this method, as Moq is unable to mock an extension method. I think this is fine as long as we call it out in the release notes.

@cgillum, thoughts?

cgillum: I'm okay with this change since it only impacts tests, but we'll definitely need to call it out as a breaking change in the release notes.

AtOMiCNebula: Yep, don't think there's any way around that one.  I also ran into this very issue with one of the built-in unit tests.  In my experience, issues like these are easy to recover from. :smile:

AtOMiCNebula: Is there anything I can add to help make sure this gets into release notes, or is that something you manage outside of the repo?  I don't see an in-repo changelog, so I assume the latter, but am happy to include something extra here if it'd help.

ConnorMcMahon: We generally just look at the change log from the previous release and construct the release notes from that, though I am not opposed to maintaining something as we merge PRs. It would simplify the work we have to do at release time.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -1113,6 +1113,36 @@ public async Task WaitForExternalEventWithTooLargeTimeout()
...
+        /// <summary>
+        /// End-to-end test which validates a CancellationToken-providing overload of WaitForExternalEvent.
+        /// </summary>
+        [Fact]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        public async Task WaitForExternalEventWithCancellationToken()
+        {
+            var orchestratorFunctionNames = new[] { nameof(TestOrchestrations.ApprovalWithCancellationToken) };
+            var extendedSessions = false;
+            using (JobHost host = TestHelpers.GetJobHost(
```
ConnorMcMahon: It looks like since you have written your code, `TestHelpers.GetJobHost()` returns an `ITestHost` instead of a `JobHost`. This is causing our CI pipeline to fail.

AtOMiCNebula: Whoops!  Resolved. :)

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
AtOMiCNebula: In that case, I don't want to start a new process for you all, so I'll let the normal process continue.  Happy to adapt in case you decide to change this for the future!

ConnorMcMahon:
Will merge this for 2.2.0, as we want any test-related breaking changes to go in minor releases.

AtOMiCNebula:
Sounds good!  I noticed there were merge conflicts, which I just took care of (net471.xml renamed back to net461.xml); no other changes otherwise.

PR #1204, opened 2020-02-06
Added configurable OverridableStates

### LINKED ISSUE #1203: Starting an orchestration that is already started
DF 1.8.4 
Consumption plan
UK South

---

Not entirely sure whats happened here but it looks like an orchestration was allowed to start again while it was already running. I didn't think this was possible, could you take a look at what happened?

Instance started : `2020-02-01T00:01:48.284742Z`
instance started again : `2020-02-01T00:02:20.1960877Z`

Looking at the history there are 2 execution IDs logged `8191956ead1545dc831ae7567dcb63d1`
: `6ea0b8c622f9411fa761ff9e22eedfae`

---

I'm believe that the orchestration was started twice because we start the orchestration in a `ServiceBusTrigger` function, I believe the service bus message failed to Ack, so the message was retried and the code to start orchestration was executed again. 

However it is my understanding that you can't start an orchestration that is already in a `running` state? I expected it would throw an exception or just silently proceed without actually starting a new instance.

amdeel:
Like in v2.0.0, added OverridableStates, a configurable option to allow users to specify which states they would like to override an orchestrator instance if an existing instanceId exists.

Defaults to AnyState to avoid changing behavior.

Resolves #1203 

Related to https://github.com/Azure/azure-functions-durable-extension/pull/987 and https://github.com/Azure/azure-functions-durable-extension/pull/1014

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
@@ -92,7 +92,7 @@ namespace Microsoft.Azure.WebJobs
 
             var dedupeStatuses = this.GetStatusesNotToOverride();
             Task<OrchestrationInstance> createTask = this.client.CreateOrchestrationInstanceAsync(
-            orchestratorFunctionName, DefaultVersion, instanceId, input, null, dedupeStatuses);
+                orchestratorFunctionName, DefaultVersion, instanceId, input, null, dedupeStatuses);
 
             this.traceHelper.FunctionScheduled(
                 this.hubName,
--- a/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
@@ -90,8 +90,9 @@ namespace Microsoft.Azure.WebJobs
                 throw new ArgumentException($"Instance ID lengths must not exceed {MaxInstanceIdLength} characters.");
             }
 
+            var dedupeStatuses = this.GetStatusesNotToOverride();
             Task<OrchestrationInstance> createTask = this.client.CreateOrchestrationInstanceAsync(
-                orchestratorFunctionName, DefaultVersion, instanceId, input);
+            orchestratorFunctionName, DefaultVersion, instanceId, input, null, dedupeStatuses);
 
             this.traceHelper.FunctionScheduled(
                 this.hubName,
@@ -103,6 +104,22 @@ namespace Microsoft.Azure.WebJobs
 
             OrchestrationInstance instance = await createTask;
             return instance.InstanceId;
+        }
+
+        private OrchestrationStatus[] GetStatusesNotToOverride()
+        {
+            var overridableStates = this.config.Options.OverridableExistingInstanceStates;
+            if (overridableStates == OverridableStates.NonRunningStates)
+            {
+                return new OrchestrationStatus[]
+                {
+                    OrchestrationStatus.Running,
+                    OrchestrationStatus.ContinuedAsNew,
+                    OrchestrationStatus.Pending,
+                };
+            }
+
+            return new OrchestrationStatus[0];
         }
 
         /// <inheritdoc />
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
@@ -302,6 +302,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// </remarks>
         public bool? LocalRpcEndpointEnabled { get; set; }
 
+        /// <summary>
+        ///  States that will override an existing orchestrator when attempting to start a new orchestrator with the same instance Id.
+        /// </summary>
+        public OverridableStates OverridableExistingInstanceStates { get; set; } = OverridableStates.AnyState;
+
         internal string GetDebugString()
         {
             var sb = new StringBuilder(4096);
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
@@ -311,6 +311,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             var sb = new StringBuilder(4096);
             sb.AppendLine("Initializing extension with the following settings:");
+            sb.Append(nameof(this.HubName)).Append(": ").Append(this.HubName).Append(", ");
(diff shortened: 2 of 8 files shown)


amdeel:
I still need to add the setting to throw an exception on invalid dedupe status and a test.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -90,8 +90,9 @@ public override async Task<string> StartNewAsync(string orchestratorFunctionName
                 throw new ArgumentException($"Instance ID lengths must not exceed {MaxInstanceIdLength} characters.");
             }
 
+            var dedupeStatuses = this.GetStatusesNotToOverride();
             Task<OrchestrationInstance> createTask = this.client.CreateOrchestrationInstanceAsync(
-                orchestratorFunctionName, DefaultVersion, instanceId, input);
+            orchestratorFunctionName, DefaultVersion, instanceId, input, null, dedupeStatuses);
```
ConnorMcMahon: nit: I don't love the spacing change here. I would recommend putting each parameter on it's own line, at 1+ tab depth. That also gives you room to label parameters that might not be clear (i.e. `null` and `dedupeStatuses` if you rename it).

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -90,8 +90,9 @@ public override async Task<string> StartNewAsync(string orchestratorFunctionName
                 throw new ArgumentException($"Instance ID lengths must not exceed {MaxInstanceIdLength} characters.");
             }
 
+            var dedupeStatuses = this.GetStatusesNotToOverride();
```
ConnorMcMahon: `dedupestatuses` is the parameter name for `CreateOrchestrationInstanceAsync()`, but it is not a clear representation of what the variable means in this context. Maybe `nonOverwriteableStatuses`

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -2659,6 +2659,95 @@ public async Task TaskHubName_Throws_ArgumentException(string taskHubName)
...
+                await client.RaiseEventAsync("operation", "incr", this.output);
+                await client.WaitForCustomStatusAsync(waitTimeout, this.output, 1);
+
+                // Make sure it's still running and didn't complete early (or fail).
+                var status = await client.GetStatusAsync();
+                Assert.True(
+                    status?.RuntimeStatus == OrchestrationRuntimeStatus.Running ||
+                    status?.RuntimeStatus == OrchestrationRuntimeStatus.ContinuedAsNew);
+
+                await host.StartOrchestratorAsync(nameof(TestOrchestrations.Counter), initialValue, this.output, instanceId: instanceId);
```
ConnorMcMahon: I'm a bit confused as to what this is testing? Is it just that we don't throw an exception?

We should probably check that we overwrote the previous orchestration with this new one.

src/WebJobs.Extensions.DurableTask/DurableOrchestrationClient.cs
```diff
@@ -105,6 +106,22 @@ public override async Task<string> StartNewAsync(string orchestratorFunctionName
             return instance.InstanceId;
         }
 
+        private OrchestrationStatus[] GetStatusesNotToOverride()
+        {
+            var overridableStates = this.config.Options.OverridableExistingInstanceStates;
+            if (overridableStates == OverridableStates.NonRunningStates)
```
ConnorMcMahon: I generally don't love using `if` statements with enums, instead I try to use `switch` statements. I find it a bit easier to tell when I have exhausted all of the enumerations at this point. In fact, there are even [some analyzers](https://roslyn-analyzers.readthedocs.io/en/latest/analyzers-info/enums/switch-on-enum-must-handle-all-cases.html#switch-on-enum-must-handle-all-cases) out in the wild that will give you warnings/exceptions if you add a new case to your enum without explicitly handling it in a switch statement.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -2659,6 +2659,95 @@ public async Task TaskHubName_Throws_ArgumentException(string taskHubName)
...
+
+                // Perform some operations
+                await client.RaiseEventAsync("operation", "incr", this.output);
+                await client.WaitForCustomStatusAsync(waitTimeout, this.output, 1);
+
+                // Make sure it's still running and didn't complete early (or fail).
+                var status = await client.GetStatusAsync();
+                Assert.True(
+                    status?.RuntimeStatus == OrchestrationRuntimeStatus.Running ||
+                    status?.RuntimeStatus == OrchestrationRuntimeStatus.ContinuedAsNew);
```
ConnorMcMahon: The rest of the test assumes that the status here is running, not continued as new, so I'm not sure if we need the second half of this predicate.

src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
```diff
@@ -302,6 +302,11 @@ public class DurableTaskOptions
         /// </remarks>
         public bool? LocalRpcEndpointEnabled { get; set; }
 
+        /// <summary>
+        ///  States that will override an existing orchestrator when attempting to start a new orchestrator with the same instance Id.
+        /// </summary>
+        public OverridableStates OverridableExistingInstanceStates { get; set; } = OverridableStates.AnyState;
```
ConnorMcMahon: We should almost certainly account for this setting in `GetDebugString()`.

src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
```diff
@@ -344,8 +351,16 @@ internal string GetDebugString()
...
+            {
+                sb.Append(nameof(this.LocalRpcEndpointEnabled)).Append(": ").Append(this.LocalRpcEndpointEnabled).Append(", ");
+            }
+
             sb.Append(nameof(this.MaxQueuePollingInterval)).Append(": ").Append(this.MaxQueuePollingInterval).Append(", ");
-            sb.Append(nameof(this.LogReplayEvents)).Append(": ").Append(this.LogReplayEvents);
+            sb.Append(nameof(this.LogReplayEvents)).Append(": ").Append(this.LogReplayEvents).Append(", ");
+            sb.Append(nameof(this.OverridableExistingInstanceStates)).Append(": ").Append(this.OverridableExistingInstanceStates).Append(", ");
+            sb.Append(nameof(this.TraceInputsAndOutputs)).Append(": ").Append(this.TraceInputsAndOutputs).Append(", ");
+            sb.Append(nameof(this.CustomLifeCycleNotificationHelperType)).Append(": ").Append(this.CustomLifeCycleNotificationHelperType);
```
ConnorMcMahon: We may want to wrap this in a null check.

PR #1217, opened 2020-02-14
Exclude x-functions-key header from async polling request in CallHttp()

### LINKED ISSUE #1141: Using the built in HTTP activity with x-functions-key header breaks polling functionality
### Description
When using `IDurableOrchestrationContext.CallHttpAsync` to call another Azure Function with the custom header `x-functions-key` set to the default key, polling will fail with 403s.

The `Location` header that's received from the first call, contains the correct `code` param (that equals the `durabletask_extension` key), and the full URL is accessible by manually opening the page in the browser. It seems that the custom `x-functions-key` is also used during polling. Therefore the HTTP request ends up with 2 keys (both as a query param and header), and this breaks the polling because it seems that the one in the header takes priority.

### Expected behavior
The polling should not fail with a 403.
This means that one of the following options are available:
- The header `x-functions-key` should not be sent again when polling
- The key in the query param takes priority
- Both keys in the query param and header are checked for validity

### Actual behavior
As mentioned above, even if the key in the URL is valid, it seems that the header takes priority if the header is set. This causes the request to fail with a 403.

```csharp
// Given that context is a IDurableOrchestrationContext
var data = new { SomeData = "here" };
var uri = "https://azurefunction.azurewebsites.net/api/HttpTrigger";
var headers = new Dictionary<string, StringValues> { ["x-functions-key"] = "default-key-here" };
var request = new DurableHttpRequest(HttpMethod.Post, uri, headers, JsonConvert.SerializeObject(data));
var response = await context.CallHttpAsync(request);
```

### Known workarounds
Instead of using the header, the key can also be directly injected into the URL as query param.
While this is definitely doable, I do expect the header to work as well.

### App Details
- **Durable Functions extension version**: v2.1.0
- **Azure Functions runtime version**: 3.0

### If deployed to Azure
- **Orchestration instance ID(s)**: 10c18cb1f7334b38aeb9cae446fa6fa9 (caller), 1c4c46139f624dadaae19186b0cdfe7e (callee)

ConnorMcMahon:
In the case where CallHTTP() is called on a Durable orchestration
implementing the server-side asynchronous HTTP pattern, the
x-functions-key used to trigger the orchestration does not have
sufficient permissions to poll the status of that orchestration. Our
solution is to strip this header when copying headers from the initial
request to the polling request.

Resolves #1141

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -245,16 +245,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                         StringComparer.OrdinalIgnoreCase);
 
                 DateTime fireAt = default(DateTime);
-                if (headersDictionary.ContainsKey("Retry-After"))
+                if (headersDictionary.TryGetValue("Retry-After", out StringValues retryAfter))
                 {
-                    fireAt = this.InnerContext
-                                .CurrentUtcDateTime
-                                .AddSeconds(int.Parse(headersDictionary["Retry-After"]));
+                    fireAt = this.InnerContext.CurrentUtcDateTime
+                                .AddSeconds(int.Parse(retryAfter));
                 }
                 else
                 {
-                    fireAt = this.InnerContext
-                                .CurrentUtcDateTime
+                    fireAt = this.InnerContext.CurrentUtcDateTime
                                 .AddMilliseconds(this.Config.Options.HttpSettings.DefaultAsyncRequestSleepTimeMilliseconds);
                 }
 
@@ -298,10 +296,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             // Do not copy over the x-functions-key header, as in many cases, the
             // functions key used for the initial request will be a Function-level key
             // and the status endpoint requires a master key.
-            if (newDurableHttpRequest.Headers.ContainsKey("x-functions-key"))
-            {
-                newDurableHttpRequest.Headers.Remove("x-functions-key");
-            }
+            newDurableHttpRequest.Headers.Remove("x-functions-key");
 
             return newDurableHttpRequest;
         }
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -240,21 +240,30 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
             while (currStatusCode == HttpStatusCode.Accepted && req.AsynchronousPatternEnabled)
             {
-                Dictionary<string, StringValues> headersDictionary = new Dictionary<string, StringValues>(durableHttpResponse.Headers);
+                var headersDictionary = new Dictionary<string, StringValues>(
+                        durableHttpResponse.Headers,
+                        StringComparer.OrdinalIgnoreCase);
+
                 DateTime fireAt = default(DateTime);
                 if (headersDictionary.ContainsKey("Retry-After"))
                 {
-                    fireAt = this.InnerContext.CurrentUtcDateTime.AddSeconds(int.Parse(headersDictionary["Retry-After"]));
+                    fireAt = this.InnerContext
+                                .CurrentUtcDateTime
+                                .AddSeconds(int.Parse(headersDictionary["Retry-After"]));
                 }
                 else
                 {
-                    fireAt = this.InnerContext.CurrentUtcDateTime.AddMilliseconds(this.Config.Options.HttpSettings.DefaultAsyncRequestSleepTimeMilliseconds);
+                    fireAt = this.InnerContext
+                                .CurrentUtcDateTime
+                                .AddMilliseconds(this.Config.Options.HttpSettings.DefaultAsyncRequestSleepTimeMilliseconds);
                 }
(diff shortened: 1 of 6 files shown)


test/Common/DurableHttpTests.cs
```diff
@@ -754,7 +754,7 @@ public async Task DurableHttpAsync_AsynchronousAPI_LongRunning(string storagePro
                 nameof(this.DurableHttpAsync_AsynchronousAPI_LongRunning),
                 enableExtendedSessions: false,
                 storageProviderType: storageProvider,
-                httpAsyncSleepTime: 10000,
+                httpAsyncSleepTime: 1000,
```
ConnorMcMahon: This is just to speed up the test.

test/Common/DurableHttpTests.cs
```diff
@@ -525,7 +525,7 @@ public async Task DurableHttpAsync_OneHeaderAndContentTest(string storageProvide
         public async Task DurableHttpAsync_AsynchronousAPI_RetryAfterTest(string storageProvider)
         {
             Dictionary<string, string> testHeaders = new Dictionary<string, string>();
-            testHeaders.Add("Retry-After", "20");
+            testHeaders.Add("Retry-After", "3");
```
ConnorMcMahon: This is to speed up the test.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -240,21 +240,30 @@ async Task<DurableHttpResponse> IDurableOrchestrationContext.CallHttpAsync(Durab
...
+                var headersDictionary = new Dictionary<string, StringValues>(
+                        durableHttpResponse.Headers,
+                        StringComparer.OrdinalIgnoreCase);
+
                 DateTime fireAt = default(DateTime);
                 if (headersDictionary.ContainsKey("Retry-After"))
                 {
-                    fireAt = this.InnerContext.CurrentUtcDateTime.AddSeconds(int.Parse(headersDictionary["Retry-After"]));
+                    fireAt = this.InnerContext
+                                .CurrentUtcDateTime
```
cgillum: nit: Could we keep .CurrentUtcDateTime on the first line? It's a little weird to have hanging property references like this.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -240,21 +240,30 @@ async Task<DurableHttpResponse> IDurableOrchestrationContext.CallHttpAsync(Durab
 
             while (currStatusCode == HttpStatusCode.Accepted && req.AsynchronousPatternEnabled)
             {
-                Dictionary<string, StringValues> headersDictionary = new Dictionary<string, StringValues>(durableHttpResponse.Headers);
+                var headersDictionary = new Dictionary<string, StringValues>(
+                        durableHttpResponse.Headers,
+                        StringComparer.OrdinalIgnoreCase);
+
                 DateTime fireAt = default(DateTime);
                 if (headersDictionary.ContainsKey("Retry-After"))
```
cgillum: Shouldn't we be using the `TryGetValue` pattern here to avoid double dictionary lookups?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -286,6 +295,14 @@ private DurableHttpRequest CreateHttpRequestMessageCopy(DurableHttpRequest durab
                 content: durableHttpRequest.Content,
                 tokenSource: durableHttpRequest.TokenSource);
 
+            // Do not copy over the x-functions-key header, as in many cases, the
+            // functions key used for the initial request will be a Function-level key
+            // and the status endpoint requires a master key.
+            if (newDurableHttpRequest.Headers.ContainsKey("x-functions-key"))
```
cgillum: Do we need to consider the `code` query string parameter as well?

ConnorMcMahon: We don't copy over query string parameters, unlike headers.

ConnorMcMahon: I think it is debatable we should even send any headers from the initial request to the polling request, but that is a breaking change at this point.

cgillum: Agreed. If you think we should remove it in a future version, then I suggest we open an issue to track doing so.

test/Common/DurableHttpTests.cs
```diff
@@ -790,7 +790,7 @@ public async Task DurableHttpAsync_AsynchronousAPI_MultipleAsyncCalls(string sto
 
             using (ITestHost host = TestHelpers.GetJobHost(
                 this.loggerProvider,
-                nameof(this.DurableHttpAsync_AsynchronousAPI_MultipleAsyncCalls),
+                nameof(this.DurableHttpAsync_AsynchronousAPI_MultipleAsyncCalls).Substring(0, 40),
```
cgillum: Alternatively we could reduce the name of the test, e.g. `DurableHttp_AsyncAPI_MultipleCalls`.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
ConnorMcMahon: #1225

PR #1218, opened 2020-02-14
Ensure default task hub names are compatible with Azure Tables

### LINKED ISSUE #1210: If taskhub name not provided, and site name starts with a number, default task hub is invalid
### Description
Azure Durable Function after deployment shows following error message that didn't help to troubleshoot it.

"The function runtime is unable to start. Microsoft.WindowsAzure.Storage: Invalid table name."
Table name - didn't not specified in an exception test. 

### Expected behavior
Application is up and running

### Actual behavior
Error message

```csharp
// insert code snippet here
```

### Known workarounds
Unknown

### App Details

    <PackageReference Include="Microsoft.ApplicationInsights" Version="2.12.1" />
    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="2.1.1" />
    <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="3.0.3" />
- C#

### Screenshots
image

Storage account exist and table exist
image
I even try to explicitly specify connection name in a host file, but it didn't help 
image
Application itself is up and running 
image
Account names specified by VS 
image

If I will directly call http trigger- I will receive 
image

### If deployed to Azure
It deployed, but don't want to disclose

ConnorMcMahon:
Azure Table storage requires names that do not begin with a number. The
current default task hub name does not account for this. We now prepend
't' to the front of the default task hub if it would start with a
number.

Resolves #1210

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Options/AzureStorageOptions.cs
+++ b/src/WebJobs.Extensions.DurableTask/Options/AzureStorageOptions.cs
@@ -155,7 +155,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
             if (!validHubNameCharacters.Any())
             {
-                sanitizedHubName = TaskHubPadding;
+                sanitizedHubName = "DefaultTaskHub";
                 return false;
             }
 
--- a/src/WebJobs.Extensions.DurableTask/Options/AzureStorageOptions.cs
+++ b/src/WebJobs.Extensions.DurableTask/Options/AzureStorageOptions.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.Serialization;
 using Microsoft.WindowsAzure.Storage;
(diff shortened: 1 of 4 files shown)


src/WebJobs.Extensions.DurableTask/Options/AzureStorageOptions.cs
```diff
@@ -144,15 +145,31 @@ public void ValidateHubName(string hubName)
 
         private static string GetTaskHubErrorString(string hubName)
         {
-            return $"Task hub name '{hubName}' should contain only alphanumeric characters and have length between {MinTaskHubNameSize} and {MaxTaskHubNameSize}.";
+            return $"Task hub name '{hubName}' should contain only alphanumeric characters, start with a letter, and have length between {MinTaskHubNameSize} and {MaxTaskHubNameSize}.";
         }
 
         internal bool IsSanitizedHubName(string hubName, out string sanitizedHubName)
```
ConnorMcMahon: These changes would also need to be ported to the Web App scale controller, as well as Functions V2 runtime (at least until the scale controller fix roles out).

src/WebJobs.Extensions.DurableTask/Options/AzureStorageOptions.cs
```diff
@@ -144,15 +145,31 @@ public void ValidateHubName(string hubName)
...
         {
-            sanitizedHubName = new string(hubName.ToCharArray()
-                                .Where(char.IsLetterOrDigit)
+            // Only alphanumeric characters are valid.
+            var validHubNameCharacters = hubName.ToCharArray().Where(char.IsLetterOrDigit);
+
+            if (!validHubNameCharacters.Any())
+            {
+                sanitizedHubName = TaskHubPadding;
```
cgillum: What is this condition about? If there are no valid characters, we name the task hub "Hub"? I find your use of the `TaskHubPadding` constant here really confusing since you're not using it to pad anything.

ConnorMcMahon: That's fair.

I think in the case where there are no valid characters (i.e. someone names their site `------------`), we should just default to something simple like "DefaultTaskHub".

PR #1229, opened 2020-02-19
Add proper storage provider logging

### LINKED ISSUE #1137: Durable 2.x does not log default storage provider configuration
### Description
In the 1.x versions of the Durable Extension, there was a trace that showed the full configuration of the extension, including the settings specific to DurableTask.AzureStorage. The config values were traced regardless of whether anything was explicitly configured in host.json.

However, this appears to have changed in Durable Functions 2.x now that we have the new extensible storage provider model. It seems we now only trace values that are explicitly configured in the `storageProvider` section of host.json.

I noticed this when testing my fix for https://github.com/Azure/azure-functions-durable-extension/pull/1132.

### Expected behavior
Durable Functions 2.x should trace default storage provider settings in addition to explicitly configured settings. This behavior is important to help with CRI investigations.

### Actual behavior
Durable Functions 2.x only traces user-specific storage provider settings. Default settings are not traced.

ConnorMcMahon:
Default values for Storage Providers are currently not emitted, which
makes servicability much harder. This change adds a virtual method to
DurabilityProvider that allows the default provider to provide its
configuration in JSON form for us to log. Resolves #1137

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
@@ -10,6 +10,9 @@ using DurableTask.AzureStorage;
 using DurableTask.AzureStorage.Tracking;
 using DurableTask.Core;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+using Newtonsoft.Json.Linq;
+using Newtonsoft.Json.Serialization;
 using AzureStorage = DurableTask.AzureStorage;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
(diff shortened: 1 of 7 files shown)


cgillum:
Can you share an example of what the new output will look like?

ConnorMcMahon:
Here is an example:

```
{
  "httpSettings": {
    "defaultAsyncRequestSleepTimeMilliseconds": 500
  },
  "hubName": "ValidateExtensionLifecycleLogsV2",
  "storageProvider": {
    "connectionStringName": null,
    "controlQueueBatchSize": 32,
    "partitionCount": 4,
    "controlQueueBufferThreshold": 256,
    "controlQueueVisibilityTimeout": "00:05:00",
    "workItemQueueVisibilityTimeout": "00:05:00",
    "trackingStoreConnectionStringName": null,
    "trackingStoreNamePrefix": null,
    "fetchLargeMessagesAutomatically": true,
    "maxQueuePollingInterval": "00:00:30"
  },
  "tracing": {
    "traceInputsAndOutputs": true,
    "traceReplayEvents": true
  },
  "notifications": {
    "eventGrid": {
      "topicEndpoint": null,
      "keySettingName": null,
      "publishRetryCount": 0,
      "publishRetryInterval": "00:05:00",
      "publishRetryHttpStatus": null,
      "publishEventTypes": null
    }
  },
  "maxConcurrentActivityFunctions": 200,
  "maxConcurrentOrchestratorFunctions": 200,
  "notificationUrl": null,
  "localRpcEndpointEnabled": false,
  "extendedSessionsEnabled": false,
  "extendedSessionIdleTimeoutInSeconds": 30,
  "maxOrchestrationActions": 100000,
  "overridableExistingInstanceStates": "NonRunningStates",
  "entityMessageReorderWindowInMinutes": 30,
  "useGracefulShutdown": false
}
```

It essentially works exactly as you would expect, we just swap out the `Dictionary<string,object>` JSON representation that only includes values set in host.json, with the JSON representation of `AzureStorageOptions`, which includes set _and_ default values

Issue #1223, opened 2020-02-18
Add setting to configure whether entity state on an exception


anthonychu:
(This may only apply to C#, need to verify)

From the discussion in #1219, entity state is persisted even when there's an exception. This behavior may be counterintuitive.

Propose adding a new setting (in code or config) to configure whether entity state is persisted on an exception. Possibly make this the default in future versions of Durable Entities.

sebastianburckhardt:
To keep things consistent, I think the behavior should be the same for all languages.

To support rollback in the implementation, what we need to do is make sure we keep a serialized snapshot of the entity state prior to performing *any* operation. Unfortunately it is not possible to accurately predict or detect whether an operation modifies the state, so we have to be conservative.

We already pay the cost of creating such a snapshot at the end of each operation batch, so we have a snapshot of the entity state before we start the first operation in a batch. However, we are not currently doing that in between the operations of a batch. 

We would have to add logic for taking such snapshots, for catching exceptions during an operation, for rolling back to the pre-operation snapshot, and then continuing with the next operation in the batch. This logic needs to be implemented separately for each language (e.g. the batch processing inside the javascript would need to call `JSON.stringify` or similar in between operations so it can restore the state of an operation that throws an exception).

sebastianburckhardt:
If we provide this "transactional entity state" guarantee, we could even go one step further and roll back entity states of all participating entities when a critical section throws an exception!

sebastianburckhardt:
Any suggestions on how one would configure this option? Perhaps a `[RollbackOnExceptions]` attribute somewhere? Not sure what would be a good place for it.

msajid:
I think it should be a default behavior. Developers should opt in for non-rollback version using some kind of attribute or config

sebastianburckhardt:
Yeah, I think I am warming up to that idea. Rollback is bad for the performance with large entity states with lots of batching but those aren't our main use cases at the moment, and they already don't perform well with the current state management approach so we will probably address those issues using some alternative state configuration options anyway.

sebastianburckhardt:
At least on the C# side this is actually a pretty easy change and in fact simplifies the current implementation. We can get this done pretty quickly I think.

anthonychu:
Great! I just wonder if this is going to change behavior for existing apps. It wouldn't affect deployed apps until they update the extension and redeploy. @cgillum Thoughts on whether this is a breaking change or a bug fix?

[...]

PR #1231, opened 2020-02-21
Implement state rollback for exceptions inside entity operations

sebastianburckhardt:
This is the feature we discussed in #1223. It is configurable as an option `RollbackEntityStateOnExceptions` which is true by default.

With rollback enabled, exceptions inside entity operations cause the state of the entity to be rolled back to what it was before the operation. This rollback also undoes creation and deletion of state, and discards any signals sent by the operation.  

To make this possible, the entity state is serialized after each individual operation that accesses the state, which makes it possible to restore that state if the next operation fails. If rollback is disabled (which is the previous behavior), the entity state is not serialized after each individual operation, but only after an entire batch of operations completes, which can be faster.

This PR also includes tests for this feature, both for class-based and function-based entities.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
@@ -32,7 +32,17 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.shim = shim;
         }
 
-        internal bool StateWasAccessed { get; set; }
+        // The last serialized checkpoint of the entity state is always stored in
+        // the fields this.State.EntityExists (a boolean) and this.State.EntityState (a string).
+        // The current state is determined by this.CurrentStateAccess and this.CurrentState.
+        internal enum StateAccess
+        {
+            NotAccessed, // current state is same as last checkpoint in this.State
+            Accessed, // current state is stored in this.CurrentState
+            Deleted, // current state is deleted
+        }
+
+        internal StateAccess CurrentStateAccess { get; set; }
 
         internal object CurrentState { get; set; }
 
@@ -74,9 +84,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             get
             {
                 this.ThrowIfInvalidAccess();
-                return this.State.EntityExists;
+                switch (this.CurrentStateAccess)
+                {
+                    case StateAccess.Accessed: return true;
+                    case StateAccess.Deleted: return false;
+                    default: return this.State.EntityExists;
+                }
             }
         }
+
+        internal int OutboxPosition => this.outbox.Count;
 
 #if !FUNCTIONS_V1
         public FunctionBindingContext FunctionBindingContext { get; set; }
@@ -144,14 +161,23 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             }
         }
 
+        public void Rollback(int outboxPositionBeforeOperation)
+        {
+            // We discard the current state, which means we go back to the last serialized one
+            this.CurrentStateAccess = StateAccess.NotAccessed;
+            this.CurrentState = null;
+
+            // we also roll back the list of outgoing messages,
+            // so any signals sent by this operation are discarded.
+            this.outbox.RemoveRange(outboxPositionBeforeOperation, this.outbox.Count - outboxPositionBeforeOperation);
+        }
+
         void IDurableEntityContext.DeleteState()
         {
             this.ThrowIfInvalidAccess();
 
-            this.StateWasAccessed = false;
+            this.CurrentStateAccess = StateAccess.Deleted;
             this.CurrentState = null;
-            this.State.EntityExists = false;
-            this.State.EntityState = null;
         }
 
         TInput IDurableEntityContext.GetInput<TInput>()
@@ -170,14 +196,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             this.ThrowIfInvalidAccess();
 
-            if (this.StateWasAccessed)
+            if (this.CurrentStateAccess == StateAccess.Accessed)
             {
                 return (TState)this.CurrentState;
             }
 
             TState result;
 
-            if (this.State.EntityExists)
+            if (this.State.EntityExists && this.CurrentStateAccess != StateAccess.Deleted)
             {
                 try
                 {
@@ -207,9 +233,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 }
             }
 
+            this.CurrentStateAccess = StateAccess.Accessed;
             this.CurrentState = result;
-            this.StateWasAccessed = true;
-            this.State.EntityExists = true;
             return result;
         }
 
@@ -218,7 +243,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             this.ThrowIfInvalidAccess();
 
-            if (this.StateWasAccessed)
+            if (this.CurrentStateAccess == StateAccess.Accessed)
             {
                 return (TState)this.CurrentState;
             }
@@ -234,7 +259,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 throw new EntitySchedulerException($"Failed to construct entity state object: {e.Message}", e);
             }
 
-            if (this.State.EntityExists)
+            if (this.State.EntityExists
+                && this.CurrentStateAccess != StateAccess.Deleted)
             {
                 try
                 {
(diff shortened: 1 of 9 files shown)


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
```diff
@@ -144,14 +158,19 @@ public bool ErrorsPresent(out string description)
             }
         }
 
+        public void Rollback(int position)
+        {
+            this.CurrentStateAccess = StateAccess.NotAccessed;
+            this.CurrentState = null;
+            this.outbox.RemoveRange(position, this.outbox.Count - position);
```
cgillum: It might be worth adding comments here explaining why this works as a means of rolling back the operation(s). My assumption based on our conversations is that setting `CurrentState` to `null` basically causes us to not write back the updates to storage. Is that accurate? And I assume removing items from the outbox effectively discards side effects such as sending signals to other entities?

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -182,6 +182,16 @@ public string HubName
...
         public bool UseGracefulShutdown { get; set; } = false;
 
+        /// <summary>
+        /// Whether exceptions inside entity operations cause the state of the entity to be rolled back. The rollback undoes all
+        /// internal effects of an operation (sent signals, and state creation, deletion, or modification).
+        /// However, it does not roll back external effects (such as I/O that was performed).
+        /// If true, the entity state is serialized after each individual operation that accessed the state, to create a snapshot
+        /// from which to restore the state if the next operation fails. If false, the entity state
+        /// is not serialized after each individual operation, but only after an entire batch of operations completes.
+        /// </summary>
```
cgillum: The `<summary>` tag should contain a short, ideally one sentence description of what this property does. The longer explanation should be separated out into a `<remarks>` section.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
sebastianburckhardt: Yes, your assumptions are correct. I can add some comments there. Essentially, `this.CurrentState` (which can be an object) is the current state of the entity in memory, and gets modified while an operation executes, while `this.State.EntityState` (which is a string) represents the state of the entity before the operation begins.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
sebastianburckhardt: ok.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
```diff
@@ -234,21 +252,21 @@ internal TState GetStateWithInjectedDependencies<TState>(Func<TState> constructo
...
             }
 
-            if (this.State.EntityExists)
+            if (this.State.EntityExists
+                && this.CurrentStateAccess != StateAccess.Deleted)
             {
                 try
                 {
-                    JsonConvert.PopulateObject(this.State.EntityState, result);
+                    JsonConvert.PopulateObject(this.State.EntityState, result, this.dataConverter.MessageSettings);
```
sebastianburckhardt: Hmm, not sure if this should be there. Are we meant to use MessageSettings for entity state serialization?

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -298,8 +315,28 @@ private async Task ProcessOperationRequestAsync(RequestMessage request)
...
+                if (!operationFailed)
+                {
+                    if (!this.context.TryWriteback(out var errorResponseMessage))
+                    {
+                        // state serialization failed; create error response and roll back.
+                        this.context.CurrentOperationResponse = errorResponseMessage;
+                        operationFailed = true;
+                    }
+                }
+                else
```
sebastianburckhardt: This is not quite right I think, we want the rollback to happen if the serialization fails. I should also add a test that catches this.

test/Common/TestOrchestrations.cs
```diff
@@ -730,6 +732,98 @@ public static async Task<string> StringStoreWithCreateDelete([OrchestrationTrigg
...
+            {
+                if (isClassBased)
+                {
+                    await Assert.ThrowsAsync<TargetInvocationException>(() => entity.SetThenThrow(333));
+                }
+                else
+                {
+                    await Assert.ThrowsAsync<TestEntityClasses.FaultyEntity.SerializableKaboom>(() => entity.SetThenThrow(333));
+                }
+            }
```
sebastianburckhardt: this needs another case to handle v1 correctly, since v1 throws a FunctionFailedException

test/Common/TestOrchestrations.cs
```diff
@@ -730,6 +732,98 @@ public static async Task<string> StringStoreWithCreateDelete([OrchestrationTrigg
...
+        {
+            // construct entity id and proxy
+            var entityId = ctx.GetInput<EntityId>();
+            var entity = ctx.CreateEntityProxy<TestEntityClasses.IFaultyEntity>(entityId);
+            bool isClassBased = entityId.EntityName == "ClassBasedFaultyEntity".ToLowerInvariant();
+            async Task ExpectException(Task t)
+            {
+                if (isClassBased)
+                {
+                    await Assert.ThrowsAsync<TargetInvocationException>(() => entity.SetThenThrow(333));
```
sebastianburckhardt: should be '`'

test/Common/TestOrchestrations.cs
```diff
@@ -730,6 +732,98 @@ public static async Task<string> StringStoreWithCreateDelete([OrchestrationTrigg
...
+                }
+                else
+                {
+                    await Assert.ThrowsAsync<TestEntityClasses.FaultyEntity.SerializableKaboom>(() => entity.SetThenThrow(333));
```
sebastianburckhardt: should be `t`

ConnorMcMahon:
@sebastianburckhardt, what is the status of this PR. Will it be ready in the next couple of days?

sebastianburckhardt:
It's ready for review.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -182,6 +182,19 @@ public string HubName
...
+        /// <remarks>
+        /// The rollback undoes all internal effects of an operation
+        /// (sent signals, and state creation, deletion, or modification).
+        /// However, it does not roll back external effects (such as I/O that was performed).
+        /// This setting can affect serialization overhead: if true, the entity state is serialized
+        /// after each individual operation. If false, the entity state is serialized
+        /// only after an entire batch of operations completes.
+        /// </remarks>
+        public bool RollbackEntityStateOnExceptions { get; set; } = true;
```
ConnorMcMahon: can we change this to false for `2.2.0`?

sebastianburckhardt: yes, will do that.

sebastianburckhardt: I am having second thoughts ... the behavior of our current implementation (without rollback) is kind of buggy: I just found a bug as I was adding tests like you suggested, this bug means that serialization errors were not properly reported in an operation

So enabling this by default could help remove some user confusion and leave us with a better thought out and tested design. It's not just about the state rollback... it's also about reporting serialization exceptions promptly for an operation, as opposed to only after the entire batch is done.

I think keeping this to true as default provides overall better experience for our users - I would hate to get more issues asking about weird behavior that we already fixed.

test/Common/TestOrchestrations.cs
ConnorMcMahon: Is this comment still valid? From what I can tell, we are catching a FunctionFailedException above?

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
ConnorMcMahon: Are we sure that the bug was present before this PR, and not caused by this refactor? Your refactor seems to keep the old path pretty much intact, but it's hard to say without testing. 

In general, if you think this is safe enough to turn on as the new default behavior in 2.2.0, I trust your stance. I just want to make sure we have good tests for both paths if at all possible.

sebastianburckhardt: Yes, the bug was there in the old behavior, before this PR. There were not enough tests for the exception behavior. I have added tests now for both the old and new behavior. But I don't see much point in keeping the old behavior as a default (it is also not documented).

PR #1241, opened 2020-02-25
Analyzer: Bug fixes and Function Analyzer refactor

amdeel:
Hi. Here is a rundown of the changes I made:

- Fixed Orchestrator Analyzer (and other compilation analyzers that pass around a SemanticModel to static methods) bug ("threw an exception of type 'System.ArgumentException' with message 'Syntax node is not within syntax tree").

- Added message when customer passes an input to an activity function when they don't use it in the function.

- FunctionAnalyzer sub-analyzers ArgumentAnalyzer and ReturnTypeAnalyzer now allow any type that serializes to a JSON array to be considered a match and will not flag code with a warning.

- Refactored FunctionAnalyzer to move all logic associated with each sub-analyzer to the sub-analyzers to reduce coupling.


amdeel:
Note: I still need to update/add test cases

src/WebJobs.Extensions.DurableTask.Analyzers/Resources.resx
```diff
@@ -121,7 +121,10 @@
     <value>Activity function call specifies a parameter that doesn't match the function definition parameter.</value>
   </data>
   <data name="ActivityArgumentAnalyzerMessageFormat" xml:space="preserve">
-    <value>Activity function named '{0}' takes a '{1}' but was given a '{2}'.</value>
+    <value>Activity function named '{0}' takes '{1}' but was given '{2}'.</value>
+  </data>
+  <data name="ActivityArgumentAnalyzerMessageFormatNotUsed" xml:space="preserve">
+    <value>Activity function named '{0}' doesn't use an input. Pass null for the input parameter instead.</value>
```
cgillum: Maybe use "its" instead of "an" here?

test/WebJobs.Extensions.DurableTask.Analyzers.Test/ActivityFunction/FunctionReturnTypeAnalyzerTests.cs
```diff
@@ -168,7 +173,7 @@ public static class HelloSequence
             {
                 var outputs = new List<string>();
 
-                outputs.Add(await context.CallActivityAsync<string>(""E1_SayHello"", ""Ben""));
+                outputs.Add(await context.CallActivityAsync<string>(""E1_SayHello"", ""World""));
```
cgillum: :)

test/WebJobs.Extensions.DurableTask.Analyzers.Test/ActivityFunction/FunctionReturnTypeAnalyzerTests.cs
```diff
@@ -69,6 +67,13 @@ public static string SayHello([ActivityTrigger] IDurableActivityContext context)
             return $""Hello {name}!"";
         }
 
+        [FunctionName(""E1_SayHello_Array"")]
+        public static string[] SayHello([ActivityTrigger] IDurableActivityContext context)
+        {
+            string name = context.GetInput<string>();
+            return $""Hello {name}!"";
```
cgillum: This example is a little strange because it's not valid C#. To avoid confusion, could we return `new [] { $@""Hello {name}!"" }` instead?

test/WebJobs.Extensions.DurableTask.Analyzers.Test/ActivityFunction/FunctionReturnTypeAnalyzerTests.cs
```diff
@@ -41,9 +38,9 @@ public static class HelloSequence
             {
                 var outputs = new List<string>();
 
-                outputs.Add(await context.CallActivityAsync<int>(""E1_SayHello"", ""Tokyo""));
-                outputs.Add(await context.CallActivityAsync<int>(""E1_SayHey"", ""Tokyo""));
-                outputs.Add(await context.CallActivityAsync<int>(""E1_SayHello_Tuple"", (""Seattle"", 4)));
+                await context.CallActivityAsync<int>(""E1_SayHello"", ""Tokyo"");
+                await context.CallActivityAsync<int>(""E1_SayHey"", ""Tokyo"");
+                await context.CallActivityAsync<int>(""E1_SayHello_Tuple"", (""Seattle"", 4));
```
cgillum: Why this change? The C# is no longer valid...

test/WebJobs.Extensions.DurableTask.Analyzers.Test/ActivityFunction/FunctionReturnTypeAnalyzerTests.cs
```diff
@@ -54,10 +51,11 @@ public static class HelloSequence
...
 
-                outputs.Add(await context.CallActivityAsync<string>(""E1_SayHello"", ""Tokyo""));
-                outputs.Add(await context.CallActivityAsync<Object>(""E1_SayHey"", ""Tokyo""));
-                outputs.Add(await context.CallActivityAsync<Tuple<string, int>>(""E1_SayHello_Tuple"", (""Seattle"", 4)));
-                outputs.Add(await context.CallActivityAsync(""E1_SayHello_ReturnsString"", ""London""));
+                await context.CallActivityAsync<string>(""E1_SayHello"", ""Tokyo"");
+                await context.CallActivityAsync<string[]>(""E1_SayHello_Array"", ""Tokyo"");
+                await context.CallActivityAsync<Object>(""E1_SayHey"", ""Tokyo"");
+                await context.CallActivityAsync<Tuple<string, int>>(""E1_SayHello_Tuple"", (""Seattle"", 4));
+                await context.CallActivityAsync(""E1_SayHello_ReturnsString"", ""London"");
```
cgillum: Same question here.

test/WebJobs.Extensions.DurableTask.Analyzers.Test/ActivityFunction/FunctionReturnTypeAnalyzerTests.cs
amdeel: Technically it wasn't valid before. I was adding int to a string array. Now I'm just not doing anything with the result. So now it bothers me slightly less.

test/WebJobs.Extensions.DurableTask.Analyzers.Test/ActivityFunction/FunctionReturnTypeAnalyzerTests.cs
amdeel: This is valid C#, we use it in our sample "E1_SayHello"
`return $"Hello {name}!";`

I took most of this from our samples and just changed what I needed to to check the things I'm testing while making it resemble durable functions code.

cgillum: By not valid C#, I meant that it won't compile. This method signature defines the return value as `string[]` but you're returning a single `string`. The "E1_SayHello" sample has a `string` return type.

test/WebJobs.Extensions.DurableTask.Analyzers.Test/ActivityFunction/FunctionReturnTypeAnalyzerTests.cs
cgillum: But why change this at all?

test/WebJobs.Extensions.DurableTask.Analyzers.Test/ActivityFunction/FunctionReturnTypeAnalyzerTests.cs
amdeel: These tests are all probably confusing because they are just a code chunk and you can't tell what you expect to be analyzed and diagnosed just by looking at it. Unlike usual unit tests which are usually easy to tell what is being tested by reading the code. I don't know how to get around that. But I'll fix these details as we notice them.

cgillum: One simple strategy might be to just add comments in the code chunks explaining what the expectation is.

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/FunctionAnalyzer.cs
```diff
@@ -36,86 +37,76 @@ public override void Initialize(AnalysisContext context)
...
-                    ParameterType = inputTypeName,
-                    ExpectedReturnType = returnTypeName,
+                    ReturnTypeNode = returnTypeNode,
                     InvocationExpression = invocationExpression
                 });
             }
         }
 
-        private bool IsCallActivityInvocation(InvocationExpressionSyntax invocationExpression)
+        private void StoreSemanticModel(SyntaxNodeAnalysisContext context)
```
cgillum: nit: consider `SaveSemanticModel` or `SetSemanticModel` as the method name. `Store` usually implies some kind of persistent storage.

src/WebJobs.Extensions.DurableTask.Analyzers/SyntaxNodeUtils.cs
```diff
@@ -236,5 +255,89 @@ private static bool IsSpecifiedAttribute(AttributeSyntax attribute, string attri
...
+        }
+
+        private static bool AreCompatibleIEnumerableTypes(ITypeSymbol invocationType, ITypeSymbol functionType)
+        {
+            if ((invocationType.Kind.Equals(SymbolKind.ArrayType) || invocationType.Kind.Equals(SymbolKind.NamedType))
+                && (functionType.Kind.Equals(SymbolKind.ArrayType) || functionType.Kind.Equals(SymbolKind.NamedType)))
+                if (UnderlyingTypesMatch(invocationType, functionType))
+                {
+                    return ((invocationType.AllInterfaces.Any(i => i.Name.Equals("IEnumerable")))
+                    && (functionType.AllInterfaces.Any(i => i.Name.Equals("IEnumerable"))));
```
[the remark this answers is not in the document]
amdeel: UnderlyingTypesMatch checks if I'm looking at an ArrayType or NamedType (like List or Enumerable; by the way, other classes like string implement IEnumerable), then gets and compares the T from both of them. Then I check that they both implement IEnumerable, because any type that implements IEnumerable is serialized to a JSON Array.

src/WebJobs.Extensions.DurableTask.Analyzers/SyntaxNodeUtils.cs
```diff
@@ -236,5 +255,89 @@ private static bool IsSpecifiedAttribute(AttributeSyntax attribute, string attri
...
+                }
+
+            return false;
+        }
+
+        private static bool UnderlyingTypesMatch(ITypeSymbol invocationType, ITypeSymbol functionType)
+        {
+            if (TryGetUnderlyingType(invocationType, out ITypeSymbol invocationUnderlyingType) && TryGetUnderlyingType(functionType, out ITypeSymbol functionUnderlyingType))
```
ConnorMcMahon: nit: Could simplify to 

```csharp
return TryGetUnderlyingType(invocationType, out ITypeSymbol invocationUnderlyingType)
    && TryGetUnderlyingType(functionType, out ITypeSymbol functionUnderlyingType)
    && invocationUnderlyingType.Name.Equals(functionUnderlyingType.Name);
```

test/WebJobs.Extensions.DurableTask.Analyzers.Test/ActivityFunction/FunctionReturnTypeAnalyzerTests.cs
amdeel: I changed it to make the C# valid. Also, adding my result to an outputs list isn't part of my test, so I wanted to simplify the test code as much as possible to make my test slightly more straightforward. My analyzer is looking for CallActivityAsync calls.

test/WebJobs.Extensions.DurableTask.Analyzers.Test/ActivityFunction/FunctionReturnTypeAnalyzerTests.cs
amdeel: I'm not a big fan of adding a bunch of comments in my test files to explain exactly what my analyzer is doing because if my analyzer logic changes then the comments will be out of date. I still think the test classes are hard to read, but knowing the ArgumentAnalyzer test file contains test cases for the ArgumentAnalyzer is probably good enough.

amdeel: However, if you think that I need to change the names of my methods to clarify what my test cases are exactly, let me know.

test/WebJobs.Extensions.DurableTask.Analyzers.Test/ActivityFunction/FunctionReturnTypeAnalyzerTests.cs
amdeel: Also consider that these tests are taking this string and running my analyzer on them to see which new warnings/errors appear. I just want to keep them as simple as possible, while having the structure of durable functions customer code.

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/ArgumentAnalyzer.cs
```diff
@@ -13,28 +14,106 @@ public class ArgumentAnalyzer
...
+                    {
+                        var diagnostic = Diagnostic.Create(Rule, activityInvocation.ParameterNode.GetLocation(), activityInvocation.Name, definitionTypeName, invocationTypeName);
 
                         context.ReportDiagnostic(diagnostic);
                     }
                 }
             }
         }
+
+        private static void TryGetInvocationInputType(SemanticModel semanticModel, ActivityFunctionCall activityInvocation, out ITypeSymbol invocationInputType)
```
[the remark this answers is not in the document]
amdeel: Should I change this to return ITypeSymbol or make this match the TryGet pattern with bool? I don't care, in this case, if it returns a null value. But I thought I should still make the method a TryGet just to let the caller know that it could return a null value, if I end up caring that it returns null in the future.

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/ArgumentAnalyzer.cs
```diff
@@ -13,28 +14,106 @@ public class ArgumentAnalyzer
...
+            if (SyntaxNodeUtils.TryGetMethodDeclaration(definitionInput, out SyntaxNode methodDeclaration))
+            {
+                var memberAccessExpressionList = methodDeclaration.DescendantNodes().Where(x => x.IsKind(SyntaxKind.SimpleMemberAccessExpression));
+                foreach (var memberAccessExpression in memberAccessExpressionList)
+                {
+                    var identifierName = memberAccessExpression.ChildNodes().Where(x => x.IsKind(SyntaxKind.IdentifierName)).FirstOrDefault();
+                    if (identifierName != null)
+                    {
+                        var identifierNameType = SyntaxNodeUtils.GetSyntaxTreeSemanticModel(semanticModel, identifierName).GetTypeInfo(identifierName).Type.Name;
+                        if (identifierNameType.Equals("IDurableActivityContext") || identifierNameType.Equals("DurableActivityContext"))
```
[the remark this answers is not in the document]
amdeel: Good point. I should definitely put these checks in a single place.

src/WebJobs.Extensions.DurableTask.Analyzers/SyntaxNodeUtils.cs
```diff
@@ -236,5 +255,89 @@ private static bool IsSpecifiedAttribute(AttributeSyntax attribute, string attri
...
+                if (!string.IsNullOrEmpty(typeInfo.Name))
+                {
+                    return typeInfo.ContainingNamespace?.ToString() + "." + typeInfo.Name.ToString() + arrayString;
+                }
+            }
+
+            return "Unknown Type";
+        }
+
+        internal static bool InputMatchesOrCompatibleType(ITypeSymbol invocationType, string invocationTypeName, ITypeSymbol functionType, string functionTypeName)
```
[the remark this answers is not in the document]
amdeel: Good point, cant defend that decision anymore.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -75,7 +75,6 @@
     <PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="2.2.1" />
     <PackageReference Include="Microsoft.Azure.DurableTask.AzureStorage" Version="1.7.1" />
     <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.0.3" />
-    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers" Version="0.2.0"/>
```
cgillum: Any reason to not include the analyzer reference in this PR?

amdeel:
Going to change some test names and maybe add some test comments still.

ConnorMcMahon:
Just make sure the build is passing.

PR #1242, opened 2020-02-25
Abort Durable Task session on Functions runtime failures

### LINKED ISSUE #1081: Orchestrations hang caused by unhandled exception in dependency injection
### Description
This issue was created as part of investigating ICM #159447500 (which contains all the details). To summarize, there appears to be a failure case where an unhandled exception in dependency injection code (for example, an `ObjectDisposedException` in an injected `LoggerFactory`) can result in a hung orchestration.

From what I can tell, it seems that an exception in the customer's .NET class constructor can cause the function execution to be aborted without surfacing any exception information to the Durable Functions extension. As a result, the extension is not aware of any failures and assumes the orchestration scheduled any subsequent actions, when in fact it didn't. Depending on the orchestrator function code, this can result in an orchestration that is permanently hung.

### Repro
The following code demonstrates how to minimally reproduce this issue.
```csharp
public class HelloSequence
{
    // Constructor called by Azure Functions dependency injection infrastructure
    public HelloSequence()
    {
        // Any exception thrown in here stalls the orchestration
        throw new System.Exception("Kah-BOOOOOM!!!");
    }

    public async Task<List<string>> Run(
    {

        outputs.Add(await context.CallActivityAsync<string>("E1_SayHello", "London"));

        return outputs;
    }

    [FunctionName("E1_SayHello")]
    public static string SayHello([ActivityTrigger] IDurableActivityContext context)
    {
        string name = context.GetInput<string>();
        return $"Hello {name}!";
    }
}
```

### Expected behavior
An orchestration should never hang, even if there is an exception in dependency injection logic. At a minimum, the orchestration instance should be terminated.

### Actual behavior
The orchestration hangs because the runtime doesn't execute the replay.

### Known workarounds
Making sure custom DI code doesn't throw unhandled exceptions.

### App Details
This is expected to impact any .NET orchestrator functions that use dependency injection.

### If deployed to Azure
Here is a query that shows an instance of this problem. The region is West EU:
```
DurableFunctionsEvents
| where PreciseTimeStamp >= datetime(2019-11-16 16:21:00) and PreciseTimeStamp <= datetime(2019-11-16 16:22:10)
| where ExecutionId == '9f9bdd1f9e0645669491992380ade65d'
| where ProviderName == "DurableTask-AzureStorage"
| extend ActId = substring(ActivityId, 0, 8), RelActId = substring(RelatedActivityId, 0, 8), ExId = substring(ExecutionId, 0, 4), MsgId = substring(MessageId, 0, 8) 
| project TIMESTAMP, EventStampName, RoleInstance, Pid, Tid, ActId, RelActId, Level, MsgId, ExId, TaskName, TaskEventId, PartitionId, EventType, Age, NewEvents, DequeueCount, Details, LatencyMs, TaskHub, Account, ExtensionVersion, ExecutionId
| order by TIMESTAMP asc
| take 5000
```

```
FunctionsLogs
| where PreciseTimeStamp >= datetime(2019-11-16 16:21:00) and PreciseTimeStamp <= datetime(2019-11-16 16:22:10)
| where FunctionInvocationId == '8d410462-3810-4813-88fb-9fc25e28f093'
| take 5000
```

ConnorMcMahon:
When the Functions JobHost is shutting down, there are often exceptions
occuring in the function execution pipeline unrelated to customer code.
This can lead to cases where the orchestration fails unecessarily, or
in some cases where the exception is largely swallowed, can lead to
orchestrations stuck in an eternally pending state.

We now utilize common logic to try to differentiate between actual user
code failures, and when the failures come from the function pipeline. In
the case where it is a failure in the functions pipeline, we abort the
session, so that Durable Task will try again later.

Resolves #1081

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -14,9 +14,11 @@ using System.Threading;
 using System.Threading.Tasks;
 using DurableTask.AzureStorage;
 using DurableTask.Core;
+using DurableTask.Core.Exceptions;
 using DurableTask.Core.History;
 using DurableTask.Core.Middleware;
 using Microsoft.Azure.WebJobs.Description;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask.Listener;
 using Microsoft.Azure.WebJobs.Host;
 using Microsoft.Azure.WebJobs.Host.Config;
 using Microsoft.Azure.WebJobs.Host.Executors;
@@ -62,6 +64,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         private readonly bool isOptionsConfigured;
         private IDurabilityProviderFactory durabilityProviderFactory;
         private INameResolver nameResolver;
+        private IApplicationLifetimeWrapper hostLifetimeService;
         private DurabilityProvider defaultDurabilityProvider;
         private TaskHubWorker taskHubWorker;
         private bool isTaskHubWorkerStarted;
@@ -146,9 +149,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.HttpApiHandler = new HttpApiHandler(this, logger);
 
 #if !FUNCTIONS_V1
+            this.hostLifetimeService = hostLifetimeService;
+
             // The RPC server is started when the extension is initialized.
             // The RPC server is stopped when the host has finished shutting down.
             hostLifetimeService.OnStopped.Register(this.StopLocalRcpServer);
+#else
+            this.hostLifetimeService = hostLifetimeService ?? new HostLifecycleService();
 #endif
         }
 
@@ -415,7 +422,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 throw new InvalidOperationException(message);
             }
 
-            return new TaskActivityShim(this, info.Executor, name);
+            return new TaskActivityShim(this, info.Executor, this.hostLifetimeService, name);
         }
 
         private async Task OrchestrationMiddleware(DispatchMiddlewareContext dispatchContext, Func<Task> next)
@@ -456,7 +463,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             }
 
             // 1. Start the functions invocation pipeline (billing, logging, bindings, and timeout tracking).
-            FunctionResult result = await info.Executor.TryExecuteAsync(
+            WrappedFunctionResult result = await FunctionExecutionHelper.ExecuteFunctionInOrchestrationMiddleware(
+                info.Executor,
+                this.hostLifetimeService,
                 new TriggeredFunctionData
                 {
                     TriggerValue = context,
@@ -479,6 +488,20 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 #pragma warning restore CS0618
                 },
                 CancellationToken.None);
+
+            if (result.ExecutionStatus == WrappedFunctionResult.FunctionResultStatus.FunctionsRuntimeError)
+            {
+                this.TraceHelper.FunctionAborted(
+                    this.Options.HubName,
+                    context.FunctionName,
+                    context.InstanceId,
+                    $"An internal error occurred while attempting to execute this function. The execution will be aborted and retried. Details: {result.Exception}",
+                    functionType: FunctionType.Orchestrator);
+
+                // This will abort the execution and cause the message to go back onto the queue for re-processing
+                throw new SessionAbortedException(
+                    $"An internal error occurred while attempting to execute '{context.FunctionName}'.", result.Exception);
+            }
 
             if (!context.IsCompleted)
             {
@@ -618,11 +641,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             }
 
             // 3. Start the functions invocation pipeline (billing, logging, bindings, and timeout tracking).
-            FunctionResult result = await entityShim.GetFunctionInfo().Executor.TryExecuteAsync(
+            WrappedFunctionResult result = await FunctionExecutionHelper.ExecuteFunctionInOrchestrationMiddleware(
+                entityShim.GetFunctionInfo().Executor,
+                this.hostLifetimeService,
                 new TriggeredFunctionData
                 {
                     TriggerValue = entityShim.Context,
-
 #pragma warning disable CS0618 // Approved for use by this extension
                     InvokeHandler = async userCodeInvoker =>
                     {
@@ -652,6 +676,20 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 #pragma warning restore CS0618
                 },
                 CancellationToken.None);
+
+            if (result.ExecutionStatus == WrappedFunctionResult.FunctionResultStatus.FunctionsRuntimeError)
+            {
+                this.TraceHelper.FunctionAborted(
+                    this.Options.HubName,
+                    entityContext.FunctionName,
+                    entityContext.InstanceId,
+                    $"An internal error occurred while attempting to execute this function. The execution will be aborted and retried. Details: {result.Exception}",
+                    functionType: FunctionType.Orchestrator);
+
+                // This will abort the execution and cause the message to go back onto the queue for re-processing
+                throw new SessionAbortedException(
+                    $"An internal error occurred while attempting to execute '{entityContext.FunctionName}'.", result.Exception);
+            }
 
             await entityContext.RunDeferredTasks();
 
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -144,9 +144,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
             // The RPC server is started when the extension is initialized.
             // The RPC server is stopped when the host has finished shutting down.
-            hostLifetimeService.OnStopped.Register(this.StopLocalRcpServer);
+            this.hostLifetimeService.OnStopped.Register(this.StopLocalRcpServer);
 #else
-            this.hostLifetimeService = hostLifetimeService ?? new HostLifecycleService();
+            this.hostLifetimeService = HostLifecycleService.NoOp;
 #endif
         }
 
@@ -504,9 +504,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             }
 
             // 1. Start the functions invocation pipeline (billing, logging, bindings, and timeout tracking).
-            WrappedFunctionResult result = await FunctionExecutionHelper.ExecuteFunction(
+            WrappedFunctionResult result = await FunctionExecutionHelper.ExecuteFunctionInOrchestrationMiddleware(
                 info.Executor,
-                this.hostLifetimeService,
                 new TriggeredFunctionData
                 {
                     TriggerValue = context,
@@ -528,7 +527,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     },
 #pragma warning restore CS0618
                 },
-                CancellationToken.None);
+                this.hostLifetimeService.OnStopping);
 
             if (result.ExecutionStatus == WrappedFunctionResult.FunctionResultStatus.FunctionsRuntimeError)
             {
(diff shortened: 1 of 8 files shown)


ConnorMcMahon:
@cgillum 

The example you gave in #1081 may be problematic with my current solution, as we will just continuously retry to execute the function, even though it will never successfully execute. I will have to try it myself to see what happens, but I wanted to highlight that. 

I don't know if we have any way from the extension layer to stop trying if we abort enough times, but maybe we could consider adding something in DurableTask.AzureStorage (or even DurableTask.Core).

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -649,10 +671,24 @@ private async Task EntityMiddleware(DispatchMiddlewareContext dispatchContext, F
                         entityContext.ThrowInternalExceptionIfAny();
                         entityContext.ThrowApplicationExceptionsIfAny();
                     },
-#pragma warning restore CS0618
+                #pragma warning restore CS0618
                 },
                 CancellationToken.None);
 
+            if (result.ExecutionStatus == WrappedFunctionResult.FunctionResultStatus.FunctionsRuntimeError)
```
ConnorMcMahon: I'm not sure I fully understand the implications of calling `SessionAbortedException` from inside of an Entity execution.

cgillum:
I'm concerned about the amount of complexity that this change adds to our shim layer (which is already pretty tricky to comprehend). I feel like what's actually needed to address #1081 is actually quite a bit smaller and doesn't require new public surface area or dependency injection. The fact that it might create new problems is also concerning.

Can we consider a different, more direct approach for this PR? I think you are right that aborting a constructor dependency injection error could put an orchestration in an infinite failure loop. In that case, I would suggest we simply fail the orchestration rather than aborting it. It's the user's code which caused the issue anyways, and they can rely on retry policies if these types of problem are transient.

As far as how to move forward with this PR, I suggest removing the `IFunctionExecutorWrapper` abstraction. I feel like it doesn't add sufficient value to the problem we're trying to solve and creates new (public facing) concepts that folks have to digest. I would prefer instead we either directly code in the behavior we want for orchestrations, entities, and activities individually. Each of these shims are different enough that I don't think there is a lot of value in trying to create a common abstraction. Either that, or try to hide common logic behind helper methods.

Feel free to push back if you disagree.

ConnorMcMahon:
I disagree that the complexity of `FunctionExecutorWrapper.ExecuteFunction` makes it harder to reason about the middleware layers. If you look at what it does, it simply gives more information about why the function executed failed. It allows the middleware/shim layer for activities/orchestrators/entities each make their own decision of how to handle Functions pipelines errors vs user code errors. The wrapper doesn't change the actual flow of the pipelines at all, it just gives extra information at the end.

As for making it an object with an injectable interface, the only reason I did this was to improve testability. It very easily could have been just a set of helper methods (if not just one helper method), but then it would have been impossible to write tests to account for Functions pipelines failures. If you think that exposing these internal types publicly is not worth the testability improvements, I could be convinced, but I really like the ability to test this behavior.

As far as the infinite loop problem, unfortunately I'm not sure we have a clean way to distinguish between user-caused dependency injection issues (i.e. like your example), and transient issues due to host shutdowns. I can investigate this a bit more, but if we can't distinguish between the two, then it may be a matter of failing the orchestration on host shutdowns (with the data loss that entails), vs retrying an orchestration that will always fail an infinite number of times. Neither of which is ideal.

ConnorMcMahon:
@cgillum 

I have reworked this to ensure we are not exposing any new types or using DI. I also made it a bit more robust to ensure we are not going to infinitely retry DI failures.

I'm going to run this through some advanced debugging in various shutdown scenarios to ensure that I have actually fixed the issues this aims to address.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
ConnorMcMahon: Talked with Sebastian and it should be safe to abort Entity executions in the case of an exception.

ConnorMcMahon:
Punting this to 2.2.1.

src/WebJobs.Extensions.DurableTask/Listener/FunctionExecutionHelper.cs
```diff
@@ -0,0 +1,144 @@
...
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Executors;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Listener
+{
+    internal static class FunctionExecutionHelper
+    {
+        public static async Task<WrappedFunctionResult> ExecuteFunctionInOrchestrationMiddleware(
```
ConnorMcMahon, line 13: This method is currently not used in my approach. I figured if we know the app is shutting down, and the exception was thrown in the WebJobs namespace, that is sufficient to determine to abort the session. There may be false positives, but those will just be retried and fail when the host is not shutting down.

There is still theoretically some value in checking whether we got to the `executedUserCode` chunk, but at this point I don't think it is worth the code complexity.

src/WebJobs.Extensions.DurableTask/Listener/FunctionExecutionHelper.cs
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        public static async Task<WrappedFunctionResult> ExecuteFunction(
+            ITriggeredFunctionExecutor executor,
+            IApplicationLifetimeWrapper hostServiceLifetime,
+            TriggeredFunctionData triggerInput,
+            CancellationToken cancellationToken)
+        {
+#pragma warning disable CS0618 // InvokeHandler approved for use by this extension
+            //if (triggerInput.InvokeHandler != null)
```
ConnorMcMahon, line 83: This commented out code only needs to be in place if this method is only used by activities. In the current iteration, it is used for all function executions.

src/WebJobs.Extensions.DurableTask/Listener/FunctionExecutionHelper.cs
```diff
@@ -0,0 +1,144 @@
...
+                        return new WrappedFunctionResult(
+                            WrappedFunctionResult.FunctionResultStatus.UserCodeError,
+                            result.Exception);
+                    }
+                }
+                else
+                {
+                    return new WrappedFunctionResult(
+                        WrappedFunctionResult.FunctionResultStatus.Success,
+                        null);
```
cgillum, line 65: I have a few suggestions for how to simplify this code. I'll batch them all into this one comment:

* Consider checking for the success case first instead of checking for the failure case first.
* Consider removing the `else` to reduce nesting.
* Consider creating static methods that represent each result type to further reduce code.
* Consider removing the `hostServiceLifetime` and using `cancellationToken` instead.
* Consider having a separate check for `!executeduserCode` that is not coupled to the host cancellation.

Here is an example that I think would be easier to read and understand:

```csharp
FunctionResult result = await executor.TryExecuteAsync(triggerInput, cancellationToken);
if (result.Succeeded)
{
    return WrappedFunctionResult.Success();
}

if (cancellationToken.IsCancellationRequested)
{
    return WrappedFunctionResult.FunctionsRuntimeError(result.Exception);
}

if (!executedUserCode)
{
   WrappedFunctionResult.UserCodeError(new InvalidOperationException(
        "The function failed to start executing. " +
        "For .NET functions, this can happen if an unhandled exception is thrown in the function's class constructor."));
}

return WrappedFunctionResult.UserCodeError(result.Exception);
```

src/WebJobs.Extensions.DurableTask/Listener/FunctionExecutionHelper.cs
cgillum: I would like it if we could keep the `executedUserCode` logic somehow. I'm not sure that we should tie it to the host shutdown, however, since there are other potential causes of user code not running (like DI failures).

src/WebJobs.Extensions.DurableTask/Listener/FunctionExecutionHelper.cs
```diff
@@ -0,0 +1,144 @@
...
+                        return new WrappedFunctionResult(
+                            WrappedFunctionResult.FunctionResultStatus.UserCodeError,
+                            result.Exception);
+                    }
+                }
+                else
+                {
+                    return new WrappedFunctionResult(WrappedFunctionResult.FunctionResultStatus.Success, null);
+                }
+            }
```
cgillum, line 120: See my previous comment for a suggestions about how we could make this code easier to read.

src/WebJobs.Extensions.DurableTask/Listener/FunctionExecutionHelper.cs
```diff
@@ -0,0 +1,144 @@
...
+            {
+                bool executedUserCode = false;
+                var triggeredFunctionData = new TriggeredFunctionData()
+                {
+                    TriggerValue = triggerInput.TriggerValue,
+                    ParentId = triggerInput.ParentId,
+                    InvokeHandler = async userCodeHandler =>
+                    {
+                        executedUserCode = true;
+                        await triggerInput.InvokeHandler(userCodeHandler);
```
cgillum, line 36: nit: I suggest removing `async` and `await` here to reduce the number of exceptions that get bubbled up in the failure code path.

src/WebJobs.Extensions.DurableTask/Listener/FunctionExecutionHelper.cs
ConnorMcMahon: I will uncomment it out considering we are going to keep the executedUserCode logic for orchestrations/entities, so this method should only be used for activities.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -146,9 +149,13 @@ public DurableTaskExtension()
             this.HttpApiHandler = new HttpApiHandler(this, logger);
 
 #if !FUNCTIONS_V1
+            this.hostLifetimeService = hostLifetimeService;
+
             // The RPC server is started when the extension is initialized.
             // The RPC server is stopped when the host has finished shutting down.
             hostLifetimeService.OnStopped.Register(this.StopLocalRcpServer);
+#else
+            this.hostLifetimeService = hostLifetimeService ?? new HostLifecycleService();
```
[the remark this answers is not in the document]
ConnorMcMahon: I'm pretty sure we can't pass in the default `HostLifecycleService` as a default parameter, as it is not a value type.

I also like using the conditional compilation here, as it doesn't make sense to register `this.StopLocalRpcServer` against a no-op `hostLifetimeService`. The conditional compilation here highlights that V1 and V2 handle this problem differently.

I think the crux of the lack of clarity here is in how I used conditional compilation to change the behavior of `HostLifecycleService`. Instead, Functions V1 should have a conditionally compiled different implementation of `IApplicationLifetimeWrapper`, that is explicitly called `NoOpLifecycleService`. That way, this line changes to 

```csharp
this.hostLifetimeService = new NoOpLifecylceService();
```

which is more clear.

src/WebJobs.Extensions.DurableTask/Listener/TaskActivityShim.cs
```diff
@@ -54,58 +58,56 @@ public override async Task<string> RunAsync(TaskContext context, string rawInput
...
-
-                // This will abort the execution and cause the message to go back onto the queue for re-processing
-                throw new SessionAbortedException(
-                    $"An internal error occurred while attempting to execute '{this.activityName}'.", e);
-            }
+            WrappedFunctionResult result = await FunctionExecutionHelper.ExecuteFunction(
+                this.executor,
+                this.hostServiceLifetime,
+                triggerInput,
+                CancellationToken.None);
```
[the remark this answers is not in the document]
ConnorMcMahon: It probably would be the right call. I just figured using the `hostServiceLifetime.OnStopping` to make ad-hoc decisions after the functions runtime threw an exception was more similar to what we have currently as opposed to passing in a more appropriate cancellation token that might change the exceptions being thrown by `this.executor.TryExecuteAsync()`.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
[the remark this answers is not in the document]
cgillum: Rather than creating a new class, I think it would be sufficiently clear if you just created a static method or property that represented the creation of a no-op instance. For example, using `HostLifecycleService.NoOp` instead of `new HostLifecycleService()`.

Also, I'm not suggesting a default parameter. I'm suggesting an actual parameter value. But I'm not sure I understand why this constructor would even get called in V1. Presumably it should only get called in V2 since only V2 supports DI. I wonder if the right thing to do is just set the default assignment to the no-op lifecycle service - i.e. 

```csharp
private readonly HostLifecycleService hostLifetimeService = HostLifecycleService.NoOp;
```

And then just overwrite it in the ctor for V2 only.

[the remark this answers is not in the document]
ConnorMcMahon: So I always get confused by the constructor on line 152, which is V1 only which calls this constructor. I think it is dead code, but it always makes me think that this code path is called in V1, which it isn't. 

I think I addressed most of your feedback, though internally I did have to make a new class for `HostLifeCycleService.NoOp`, as I had no good way of making a noop version of `IApplicationLifecycleService` otherwise.

ConnorMcMahon:
Tested on Functions V1, so I feel confident this is safe to merge.

PR #1256, opened 2020-03-03
Initialized DataConverter in V1 constructor

### LINKED ISSUE #1234: Durable Functions above v2.0.0 do not work locally in function v1 runtime, dataConverter null?
### Description
I am creating a simple timed function which acts as an OrchestratorClient, but the client doesn't even trigger because the exception below occurs before hitting any line of code. I use Function v1 runtime (.net 4.7.1) and Durable Function 2.1.1. Everything works as expected if I use a function v2 runtime with .net core, but when using function v1 runtime with .net 4.7.1 I can't catch the first line of code with a breakpoint.

After investigating for a complete day, I saw issue #1215 which seems to have a similar error with the **dataConverter** being null, so after seeing it was introduced after v2.0.0, I decided to try the 2.0.0 package and everything started working. However I do need the JSON serializer DI update so I would like to use 2.1.1 👎 

To replicate the issue, simply create a new v1 timed function, add the durable task 2.1.1 package, change the function's signature to the orchestrator client signature and put a breakpoint at the first line of code. You should not be able to hit the breakpoint as it should fail before.

### Expected behavior
I should be able to use Durable Functions 2.1.1 with Function v1 runtime without issues

### Actual behavior
Code fails before entering the function with the following stack trace :

[2/21/2020 2:47:36 PM] A ScriptHost error has occurred
[2/21/2020 2:47:36 PM] Exception while executing function: [MyFunction]. Microsoft.Azure.WebJobs.Host: **Exception binding parameter 'client'. DurableTask.Core: Value cannot be null.**
[2/21/2020 2:47:36 PM] **Parameter name: dataConverter.**
[2/21/2020 2:47:36 PM] Exception while executing function: [MyFunction]
[2/21/2020 2:47:36 PM] Exception while executing function: [MyFunction]. Microsoft.Azure.WebJobs.Host: Exception binding parameter 'client'. DurableTask.Core: Value cannot be null.
[2/21/2020 2:47:36 PM] Parameter name: dataConverter.
[2/21/2020 2:47:36 PM] Function completed (Failure, Id=eae74910-37ae-457a-9dd3-4bc3099eff36, Duration=416ms)
[2/21/2020 2:47:36 PM]
[2/21/2020 2:47:36 PM] Executed '[MyFunction]' (Failed, Id=eae74910-37ae-457a-9dd3-4bc3099eff36)
[2/21/2020 2:47:36 PM] mscorlib: Exception while executing function: [MyFunction]. Microsoft.Azure.WebJobs.Host: Exception binding parameter 'client'. DurableTask.Core: Value cannot be null.[2/21/2020 2:47:36 PM] Loading the application settings...

[2/21/2020 2:47:36 PM] Parameter name: dataConverter.
[2/21/2020 2:47:36 PM]   Function had errors. See Azure WebJobs SDK dashboard for details. Instance ID is 'eae74910-37ae-457a-9dd3-4bc3099eff36'
[2/21/2020 2:47:36 PM] mscorlib: Exception while executing function: [MyFunction]. Microsoft.Azure.WebJobs.Host: Exception binding parameter 'client'. DurableTask.Core: Value cannot be null.
[2/21/2020 2:47:36 PM] Parameter name: dataConverter.

Basic code provided by Visual Studio :

```csharp
[FunctionName("MyFunction")]
public static Task Run([DurableClient] IDurableClient client, [TimerTrigger("0 0 */4 * * *", RunOnStartup = true)] TimerInfo myTimer)
{
    // This never gets triggered, the exception occurs before
    return client.PurgeInstanceHistoryAsync(
        DateTime.MinValue,
        DateTime.UtcNow.AddDays(-30),
        new List<OrchestrationStatus>
        {
            OrchestrationStatus.Completed
        });
}
```

### Known workarounds
If I use the 2.0.0 nuget my orchestrator client triggers just fine.

### App Details

- **Durable Functions extension version (e.g. v1.8.3)**: 2.1.1 (or anything above 2.0.0)
- **Azure Functions runtime version (1.0 or 2.0)**: 1.0
- **Programming language used**: C# (.NET 4.7.1)
- **IDE**: Visual Studio 2019 fully updated (v16.4.5)
- **Storage Emulator** : Fully updated (5.10.0.0)

amdeel:
resolves #1234

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -164,6 +164,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             : this(options, loggerFactory, nameResolver, orchestrationServiceFactory, shutdownNotification, durableHttpMessageHandlerFactory)
         {
             this.connectionStringResolver = connectionStringResolver;
+
+            var messageSerializerSettingsFactory = new MessageSerializerSettingsFactory();
+            var errorSerializerSettingsFactory = new ErrorSerializerSettingsFactory();
+            this.DataConverter = new MessagePayloadDataConverter(messageSerializerSettingsFactory, errorSerializerSettingsFactory);
         }
 
         /// <summary>
--- a/src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask-net471.xml
+++ b/src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask-net471.xml
@@ -955,6 +955,12 @@
             </summary>
             <remarks>
             <para>
+            A terminated instance will eventually transition into the <see cref="F:Microsoft.Azure.WebJobs.Extensions.DurableTask.OrchestrationRuntimeStatus.Terminated"/> state.
+            However, this transition will not happen immediately. Rather, the terminate operation will be queued in the task hub
+            along with other operations for that instance. You can use the <see cref="M:Microsoft.Azure.WebJobs.Extensions.DurableTask.IDurableOrchestrationClient.GetStatusAsync(System.String,System.Boolean,System.Boolean,System.Boolean)"/>
+            method to know when a terminated instance has actually reached the Terminated state.
+            </para>
+            <para>
             Terminating an orchestration instance has no effect on any in-flight activity function executions
             or sub-orchestrations that were started by the current orchestration instance.
             </para>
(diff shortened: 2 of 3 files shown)


src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -164,6 +164,10 @@ public DurableTaskExtension()
             : this(options, loggerFactory, nameResolver, orchestrationServiceFactory, shutdownNotification, durableHttpMessageHandlerFactory)
         {
             this.connectionStringResolver = connectionStringResolver;
+
+            var messageSerializerSettingsFactory = new MessageSerializerSettingsFactory();
```
ConnorMcMahon: I'm not convinced that this is the place that will actually fix the issue. It looks like this constructor calls into the other constructor that already initializes these settings. In fact, I think this is just a legacy constructor that may no longer be needed for Functions V1 to function.

I would test this manually to see if this works, and if it doesn't consider moving this into `InitializeForFunctionsV1()`

ConnorMcMahon:
This is now addressed in #1258, as I had to fix a bunch of other Functions V1 bugs to test this fix.

PR #1257, opened 2020-03-04
Downgrade Functions V1 .NET version to net461

ConnorMcMahon:
At one point, we upgraded the full framework target framework from
net461 to net471 to reduce package size for our .NET Standard packages,
as well as to utilize some new language features. This has caused two
problems.

1. When Functions V1 customers target net462, they pull in the
netstandard2.0 version of the dll, which is meant for Functions V2/V3
customers.

2. Using any 4.7 or greater release for Functions V1 breaks when using
Microsoft.NET.Sdk.Functions 1.0.30 or greater, due to some conditional
logic relying on #if NET46. Unfortunately, this version of that package
is required to use Microsoft.Azure.WebJobs 2.3.0 instead of 2.2.0.

This PR reverts to net461 to avoid these issues, as well as some
required code changes to make it compatible with this older version.


src/WebJobs.Extensions.DurableTask/EntityScheduler/EntityId.cs
```diff
@@ -11,6 +11,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
     /// </summary>
     public struct EntityId : IEquatable<EntityId>, IComparable
     {
+        private string schedulerId;
```
cgillum: Should this be `readonly`?

src/WebJobs.Extensions.DurableTask/EntityScheduler/EntityId.cs
```diff
@@ -60,7 +68,7 @@ internal static EntityId GetEntityIdFromSchedulerId(string schedulerId)
         /// <inheritdoc/>
         public override string ToString()
         {
-            return GetSchedulerIdFromEntityId(this);
+            return this.schedulerId;
```
cgillum: I'm not 100% sure because I don't always understand how JSON.NET works, but if this object is deserialized from JSON, is it possible that `schedulerId` will be null (because the constructor doesn't necessarily get called)?  I ask because I see comments in the `EntityName` and `EntityKey` properties saying we still need the private setters for deserialization.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -1,16 +1,17 @@
...
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net471</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
     <AssemblyName>Microsoft.Azure.WebJobs.Extensions.DurableTask</AssemblyName>
     <RootNamespace>Microsoft.Azure.WebJobs.Extensions.DurableTask</RootNamespace>
     <MajorVersion>2</MajorVersion>
     <MinorVersion>2</MinorVersion>
     <PatchVersion>0</PatchVersion>
-    <Version>$(MajorVersion).$(MinorVersion).$(PatchVersion)</Version>
+    <Version>$(MajorVersion).$(MinorVersion).$(PatchVersion)-test2</Version>
```
cgillum: I assume we want to remove this before checking in?

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -1,16 +1,17 @@
...
     <FileVersion>$(MajorVersion).$(MinorVersion).$(PatchVersion)</FileVersion>
     <AssemblyVersion>$(MajorVersion).0.0.0</AssemblyVersion>
     <Company>Microsoft Corporation</Company>
+    <LangVersion>7.1</LangVersion>
```
cgillum: Any reason not to target 7.3? Is it because our AppVeyor images don't support it?

ConnorMcMahon: I don't think so. I will try doing so.

src/WebJobs.Extensions.DurableTask/EntityScheduler/EntityId.cs
ConnorMcMahon: Good point!

There are two approaches I can take. One of which is lazily calculate this field when I call ToString() the first time, the other is to serialize the field. I think I prefer the first, as I don't see a reason we should send this field around since it can already be derived.

src/WebJobs.Extensions.DurableTask/EntityScheduler/EntityId.cs
ConnorMcMahon: Can no longer be readonly now that I realized that the constructor is not the only path in which this field needs to be populated.

src/WebJobs.Extensions.DurableTask/EntityScheduler/EntityId.cs
ConnorMcMahon: I went with the first approach.

PR #1262, opened 2020-03-09
Update SignalEntity to call entity from other function app

### LINKED ISSUE #1248: Not able to Signal Entity Function using IDurableClient from different Function App
### Description
I want to use IDurableClient from an EventHub triggered Function to Signal the Entity Function sitting in different Function App. I pass the Entity Function's Storage COnnection string and Taskhub name in DurableClient Attribute like this -
`[DurableClient(ConnectionName = "EntiyFunctionConnectionStringName", TaskHub = "RaghavTaskHub1")]`

When I try to Signal EF, it says it can't find the EF. Actual exception in below section. I have tried these cases -
1) Both the functions running locally at different ports. Both using Storage Emulator locally with `AzureWebJobsStorage = “UseDevelopmentStorage=true”.`
2) EF running in cloud with its own Storage and EH triggered Function running locally. 

Both of above ends up in same error.
However, note that I am successfully able to read Entity State using IDurableClient from different App. So the issue looks to be that for Signal call, Durable Extension is trying to search the EF in the same Function App. 
Also, I am able to signal across Function Apps using Http endpoint.

### Expected behavior
Entity Function should be found and Signal call to successfully return success code.

### Actual behavior
The Entity is "Circuit" in my code and here is the Exception on console -
image

```csharp
[FunctionName(“IngestFunction”)]
public async Task Run(
     [EventHubTrigger("health_change", ConsumerGroup = "testcg", Connection = "EventHubConnectionString")] EventData[] events,
     [DurableClient(ConnectionName = "EntiyFunctionConnectionStringName", TaskHub = "RaghavTaskHub1")] IDurableEntityClient entityClient)
{
     await client.SignalEntityAsync(
                    new EntityId("Circuit", "IngestFunction"),
                    nameof(ICircuit.ReportOperationsStatus),
                    new CircuitBreakerRequest
                    {
                        RequestId = Guid.NewGuid().ToString(),
                        RequestTime = DateTime.UtcNow,
                        SuccessCount = 10,
                        FailureCounts = new Dictionary<string, int>
                        {
                            { "Different App DurableClient", 50 },
                        },
                    },
                    "RaghavTaskHub1",
                    "EntiyFunctionConnectionStringName");

}
```

### Known workarounds
Using Http endpoint to Signal instead of IDurableClient.

### App Details

- **Durable Functions extension version (e.g. v1.8.3)**:
Microsoft.Azure.Functions.Extensions = 1.0.0
Microsoft.Azure.WebJobs.Extensions.DurableTask = 2.1.1
Microsoft.Azure.WebJobs.Extensions.EventHubs = 4.1.1
Microsoft.NET.Sdk.Functions = 3.0.3

- **Azure Functions runtime version (1.0 or 2.0)**:
v3

- **Programming language used**:
C# (.Net Core 3.1)

- **Timeframe issue observed**:
- **Function App name**:
- **Function name(s)**:
- **Azure region**:
- **Orchestration instance ID(s)**:
- **Azure storage account name**:

> If you don't want to share your Function App or storage account name GitHub, please at least share the orchestration instance ID. Otherwise it's extremely difficult to look up information.

amdeel:
resolves #1248

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -262,7 +262,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 throw new ArgumentNullException(nameof(operationName));
             }
 
-            this.config.ThrowIfFunctionDoesNotExist(entityId.EntityName, FunctionType.Entity);
+            if (client == this.client)
+            {
+                this.config.ThrowIfFunctionDoesNotExist(entityId.EntityName, FunctionType.Entity);
+            }
 
             var guid = Guid.NewGuid(); // unique id for this request
             var instanceId = EntityId.GetSchedulerIdFromEntityId(entityId);
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -262,7 +262,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 throw new ArgumentNullException(nameof(operationName));
             }
 
-            if (client == this.client)
+            if (this.client.Equals(client))
             {
                 this.config.ThrowIfFunctionDoesNotExist(entityId.EntityName, FunctionType.Entity);
             }


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -262,7 +262,10 @@ private async Task SignalEntityAsyncInternal(TaskHubClient client, string hubNam
                 throw new ArgumentNullException(nameof(operationName));
             }
 
-            this.config.ThrowIfFunctionDoesNotExist(entityId.EntityName, FunctionType.Entity);
+            if (client == this.client)
```
ConnorMcMahon: I'm not convinced this block is ever true. That being said, it matches our logic for in other places. Do you mind just testing by using a TaskHub that matches the application's taskhub?

amdeel: I can change this to .Equals() to cover the case where the user specifies a TaskHub that matches the applications TaskHub.

ConnorMcMahon: `==` and `Equals()` are equivalent assuming we haven't added our own implementation of `Equals()`, but I am fine with merging the current state.

PR #1263, opened 2020-03-09
Roslyn Analyzer: Comparing function names which use `nameof()`

marcduiker:
I got DF0109 warnings when I used `nameof(ActivityFunctionX)` in my `CallActivityAsync` calls. The analyzer tries to compare `"nameof(ActivityFunctionX)"` with `"ActivityFunctionX"`, which fails.

I therefore made a small change to the NameAnalyzer.cs so it compares function names regardless of `nameof()` usage in either the `FunctionNameAttribute` or the `CallActivityAsync` method. I've extended the `Name_NonIssueCalls` unittest with 3 test cases.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/NameAnalyzer.cs
+++ b/src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/NameAnalyzer.cs
@@ -34,7 +34,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers
         {
             foreach (var activityInvocation in calledFunctions)
             {
-                if (!availableFunctions.Select(x => GetCleanedFunctionName(x.FunctionName)).Contains(GetCleanedFunctionName(activityInvocation.Name)))
+                if (!availableFunctions.Select(x => x.FunctionName).Contains(activityInvocation.Name))
                 {
                     if (SyntaxNodeUtils.TryGetClosestString(activityInvocation.Name, availableFunctions.Select(x => x.FunctionName), out string closestName))
                     {
@@ -50,18 +50,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers
                     }
                 }
             }
-        }
-
-        private static string GetCleanedFunctionName(string functionName)
-        {
-            const string nameofStart = "nameof(";
-            const string nameofEnd = ")";
-            if (functionName.StartsWith(nameofStart) && functionName.EndsWith(nameofEnd))
-            {
-                functionName = functionName.Substring(nameofStart.Length, functionName.Length - nameofStart.Length - nameofEnd.Length);
-            }
-
-            return functionName;
         }
     }
 }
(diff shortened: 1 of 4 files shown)


amdeel:
Thank you so much for this contribution @marcduiker!

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/NameAnalyzer.cs
```diff
@@ -51,5 +51,17 @@ private static string GetClosestString(string name, IEnumerable<string> availabl
                 }
             }
         }
+
+        private static string GetCleanedFunctionName(string functionName)
```
amdeel: I'd like to see this in FunctionAnalyzer, so we can store the function name from CallActivity* correctly. This will allow our ArgumentAnalyzer and FunctionReturnTypeAnalyzers to work as expected without adding these extra method calls. 

Line 73 in FunctionAnalyzer stores that function name. Currently it trims the parenthesis off of the string. You could modify this method to either 1. trim nameof() off, like you currently do or 2. trim the parenthesis if the string doesn't start with nameof(.

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/NameAnalyzer.cs
```diff
@@ -34,7 +34,7 @@ private static string GetClosestString(string name, IEnumerable<string> availabl
         {
             foreach (var activityInvocation in calledFunctions)
             {
-                if (!availableFunctions.Select(x => x.FunctionName).Contains(activityInvocation.Name))
+                if (!availableFunctions.Select(x => GetCleanedFunctionName(x.FunctionName)).Contains(GetCleanedFunctionName(activityInvocation.Name)))
```
amdeel: We already store FunctionName accounting for the nameof() case. This method would only benefit our activityInvocation name.

test/WebJobs.Extensions.DurableTask.Analyzers.Test/ActivityFunction/NameAnalyzerTests.cs
```diff
@@ -102,6 +105,23 @@ public static string SayHelloTupleOnContext([ActivityTrigger] IDurableActivityCo
...
+        public static string SayHelloByMethodName([ActivityTrigger] IDurableActivityContext context)
+        {
+            string name = context.GetInput<string>();
+            return $""Hello {name}!"";
+        }
+    }
+
+    public class SayHelloByClassName
+    {
+        [FunctionName(nameof(SayHelloByClassName))]
```
amdeel: Lets make one of these attributes just a string, so we can test that it works in mismatch scenarios.

marcduiker:
Thanks for the suggestions @amdeel, I'll take care of it shortly!

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/NameAnalyzer.cs
marcduiker: `GetCleanedFunctionName` is now an extension method which is called in the FunctionAnalyzer

marcduiker:
@amdeel  I think I've covered the changes you suggested, let me know if anything else needs to be done 😊.

PR #1268, opened 2020-03-11
Updated sample csproj files

amdeel:


cgillum:
Can we also update the [F# sample project](https://github.com/Azure/azure-functions-durable-extension/blob/dev/samples/fsharp/DurableFSharp.fsproj)?

Issue #939, opened 2019-09-18
End-to-end distributed tracing support


cgillum:
### Problem
Azure Functions has integration with Application Insights and supports end-to-end tracing for various binding types. However, there is no support for end-to-end tracing in Durable Functions.

### Proposed Solution
Durable Functions should support end-to-end tracing and associated visualization in Application Insights.

* A user should be able to see the order of activity function calls from an orchestration, including any sub-orchestrations
* User logs should be able to correlate with these traces
* Ideally this works with external events as well
* Durable HTTP APIs should correctly flow end-to-end trace IDs
* Calls to external services supported by App Insights should flow the correct correlation info
* Should be compatible with "App Map" visualizations

The first step to supporting this is to flow correlation information through the DTFx libraries. Once that is done, it should be surfaced to the Azure Functions / Durable Functions layers. The DTFx work is currently being tracked here: https://github.com/Azure/durabletask/pull/261

### Workarounds
We do have tracing to Application Insights which includes the orchestration instance ID across orchestrator and activity function traces. However, it is not correlated with user logs. It's also not possible to directly see the call chain between orchestrations and specific activity function invocations. It also doesn't work with App Insights visualization features.

idg10:
One aspect of this that you've not mentioned in the bullet list in your proposed solutions is how Eternal Orchestrations https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-eternal-orchestrations should fit into this.

My worry would be that once Durable Functions integrate with Application Insights in the same way that other functions triggers do today, all work started from an Eternal Orchestration would appear to be part of the same massive operation.

If possible, what I'd like to see is each iteration of an Eternal Orchestration effectively appearing as a distinct Operation in Application Insights. So each time you `ContinueAsNew` you start a new Operation for monitoring purposes.

TsuyoshiUshio:
Hi @idg10 , 
Thank you for the comment. 
Durable Functions Distributed Tracing support won't be the same as the other functions.  It will be a separate tracing mechanism.  Since, some people need correlation in Eternal Orchestration scenario, however, now I know some people want start from new when it is ContinueAsNew. Which feature does it fit your use case? 

* Having a configuration to make ContineAsNew starts new tracing for each time.
* Having a configuration to make a specific Orchestrator ContinueAsNew start new tracing for each tiem. But other's is not. 

Or would you prefer other solution? 
@cgillum  If you have any comment to that spec, please leave a comment.

[...]

Issue #593, opened 2019-01-29
Suppress unnecessary dependency tracking for internal storage calls.


kashimiz:
Prompted by [this StackOverflow post](https://stackoverflow.com/questions/54400108/failing-dependency-call-in-azure-durable-functions): Currently, Application Insights' dependency tracking feature reports 404 responses from Azure Storage such the initial call made by `table.CreateIfNotExists()` when the table doesn't exist, which is part of the Durable Functions extension's program flow. We should suppress these to reduce noise in customer's App Insights logs.

JohnReeson:
Link [one identical issue](https://github.com/Azure/azure-webjobs-sdk/issues/2039).

oising:
This one is bothering my app insights OCD.

ProjectPete:
Bump!  
+1 oising!  
    
It's a distracting red, please make this tweak or advise us on a work-around.   

We need a clearer vision for 2020!  :D  
 
Apart from these tiny issues... I flippin lovin' Function Apps and the monitoring/logs around them...  
A true game changer. Interesting times...

Happy New Year to all durable devs & digital devotees x

hugobarona:
Definitely I am looking forward to this change since the 404 dependency errors are really distracting, especially when i am using Live Metrics Stream in the functions in Production while executing several requests per minute.

cgillum:
@amdeel this might be a good one for you to look into. Here is a StackOverflow post that describes how to do this for SQL (we would need to do one slightly differently - i.e. for Azure Storage calls): https://stackoverflow.com/questions/38320886/app-insights-disable-sql-dependency-telemetry

@brettsam would be a great person to consult with on this as well.

shelbaz:
Bump. Looking forward to this update, too many false positive errors in ApplicationInsights because of it.

ConnorMcMahon:
Don't want to delay 2.2.0 for this, so targeting 2.2.1.

ConnorMcMahon:
punting for 2.3.0

corwestermaniddink:
Update please?

oising:
La la la la la la

ravibha:
We are facing a similar issue as well. Seeing 404 dependency failures when durable functions are being executed. Dependency type is 'Azure Table'. Has this been fixed already? I was not able to tell from the backlog item.

Thanks!

cgillum:
[...]

PR #1270, opened 2020-03-11
Distributed Tracing

TsuyoshiUshio:
# The Distributed Tracing Implementation

I create this PR for share/review/discuss the architecture. 
This branch includes references for non-published NuGet package (DurableTask correlation branch:https://github.com/Azure/durabletask/tree/correlation), so **DO NOT MERGE**
Also, I need to resolve conflict against the correlation branch. 
image

# Overview
* Support HttpCorrelationProtocol and W3C TraceContext
* Use the Distributed Tracing enabled DurableTask (https://github.com/Azure/durabletask/tree/correlation/samples/Correlation.Samples)
* Update Testing Framework to support E2E tracing testing 
* Add Configuration for enabling/disabling distributed Tracing
* Add Configuration for specifying Distributed Tracing Protocol

# Supported
* End to End Tracing of Orchestration, SubOrchestration, Activity, and Exception, Retry, Fan-Out-Fan-In, ContinueAsNew
# Not Supported yet
* Durable Entity Tracing
* Http Activity
* JavaScript

# TODO
* Telemetry Suppression (Currently, Durable Functions and Azure Functions Host send duplicate telemetry with different Id.)

# Related Issue:
https://github.com/Azure/azure-functions-durable-extension/issues/939


TsuyoshiUshio:
@amdeel Thank you for the trigger! I'm busy state until Monday, I'll fix the conflict after that, then ping you.

TsuyoshiUshio:
The test fails, we need to push DurableTaskCore.Telemetry package first.

TsuyoshiUshio:
Ready for review and merge. However, the CI fails because:

* DurableTask.Core.Telemetry and DurableTask.AzureStrorage.Telemetry is reside on the myget. 
* TestV1 is not supported. (Distributed Tracing is V2 only feature)

Is there any good ways to run the appveyor CI under this situation?

samples/correlation-csharp/FunctionAppCorrelation/local.settings.json
```diff
@@ -0,0 +1,8 @@
+{
+    "IsEncrypted": false,
```
reyang, line 2: nit: the indentation seems to be a mix of TABs and SPCs, also considering an empty line before EOF.

src/WebJobs.Extensions.DurableTask/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
```diff
@@ -0,0 +1,324 @@
...
+            }
+            else
+            {
+                telemetry.Context.Operation.ParentId = activity.SpanId.ToHexString();
+            }
+        }
+    }
+}
+
+#endif
```
reyang, line 324: nit: put a empty line before EOF.

src/WebJobs.Extensions.DurableTask/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
```diff
@@ -0,0 +1,324 @@
...
+
+        /// <summary>
+        /// Initializes telemetry item.
+        /// </summary>
+        /// <param name="telemetry">Telemetry item.</param>
+        public void Initialize(ITelemetry telemetry)
+        {
+            if (this.IsSuppressedTelemetry(telemetry))
+            {
+                this.SuppressTelemetry(telemetry);
```
reyang, line 98: This would suppress the telemetry within the durable task.
Additional work is required to communicate back to the host and suppress the telemetry there:
1. Durable task puts the suppression flag in the payload before returning back to the Functions host.
2. Once Functions host received the suppression flag, respect the flag and stop emitting telemetry events (context propagation is still required).

samples/correlation-csharp/FunctionAppCorrelation/local.settings.json
```diff
@@ -0,0 +1,8 @@
...
+  "Values": {
+    "AzureWebJobsStorage": "UseDevelopmentStorage=true",
+    "FUNCTIONS_WORKER_RUNTIME": "dotnet",
+    "APPINSIGHTS_INSTRUMENTATIONKEY": "f152bd88-9b36-43ba-a8a6-ebf921f9fc8f"
+  }
+}
```
reyang, line 8: nit, empty line before EOF, there might be other place that need to updated, too.

src/WebJobs.Extensions.DurableTask/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
```diff
@@ -0,0 +1,324 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using DurableTask.Core.Settings;
```
reyang, line 4: Not related to this PR though, the namespace seems to be against the [naming convention](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-namespaces).

TsuyoshiUshio: Do you mean `<Company>.<Technology>*` rule?

samples/correlation-csharp/FunctionAppCorrelation/local.settings.json
TsuyoshiUshio: I"m creating different branch for the full samples. (I didn't want to have a big pr). 
I'll fix this on the branch! Thank you!

src/WebJobs.Extensions.DurableTask/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
TsuyoshiUshio: Done.

samples/correlation-csharp/FunctionAppCorrelation/local.settings.json
TsuyoshiUshio: Thank you! I'll fix this on the samples branch!

src/WebJobs.Extensions.DurableTask/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
TsuyoshiUshio: Hi @reyang 
Thank you for the notice! Very good point. 

This issue might be related. 
https://github.com/Azure/azure-functions-durable-extension/issues/593

Maybe this TelemetryInitializer might be  the one fix on the functions host side. 
https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs

src/WebJobs.Extensions.DurableTask/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
TsuyoshiUshio: Sure. Then it might be a problem of the DurableTask framework side.

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Net.Http;
 using System.Threading;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask.Correlation;
```
ConnorMcMahon, line 7: I think this namespace is unused if Functions V1, so move it down a line so it is only used in non-Functions V1 compilations.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -103,7 +108,15 @@ public DurableTaskExtension()
             IDurableHttpMessageHandlerFactory durableHttpMessageHandlerFactory = null,
             ILifeCycleNotificationHelper lifeCycleNotificationHelper = null,
             IMessageSerializerSettingsFactory messageSerializerSettingsFactory = null,
-            IErrorSerializerSettingsFactory errorSerializerSettingsFactory = null)
+            IErrorSerializerSettingsFactory errorSerializerSettingsFactory = null
+#if !FUNCTIONS_V1
+#pragma warning disable SA1113, SA1001, SA1115
+            ,ITelemetryActivator telemetryActivator = null
```
ConnorMcMahon, line 114: NIT: I would rather duplicate the line of `IErrorSerializerSettingsFactory errorSerializerSettingsFactory = null` with a `,` and ` )` then to do all of the warning disablement, and have the weird comma and paranthesis placement. It forces a bunch of `#` statements that really clutter up the code.

src/WebJobs.Extensions.DurableTask/Correlation/TraceContextBaseExtensions.cs
```diff
@@ -0,0 +1,58 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using DurableTask.Core;
+#if !FUNCTIONS_V1
```
ConnorMcMahon, line 8: You may want to extend this conditional compilation up, as it appears none of these imports are necessary in Functions V1.

ConnorMcMahon: In fact, for many of these correlation files, it may be eaiser to conditionally exclude them from Functions V1 builds in the csproj file: https://stackoverflow.com/questions/14175853/c-is-it-possible-to-conditionally-exclude-some-files-from-compilation.

src/WebJobs.Extensions.DurableTask/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
```diff
@@ -0,0 +1,324 @@
...
+                if (telemetry.Context.Operation.ParentId == null)
+                {
+                    telemetry.Context.Operation.ParentId = traceParent.SpanId;
+                }
+            }
+        }
+
+        internal void SuppressTelemetry(ITelemetry telemetry)
+        {
+            // TODO change the strategy.
```
ConnorMcMahon, line 235: This TODO is very vague. I would either elaborate on what the new strategy should be in the comment, or open a GitHub issue and reference it in your TODO.

src/WebJobs.Extensions.DurableTask/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
```diff
@@ -0,0 +1,324 @@
...
+    /// <summary>
+    /// Telemetry Initializer that sets correlation ids for W3C.
+    /// This source is based on W3COperationCorrelationTelemetryInitializer.cs
+    /// 1. Modified with CorrelationTraceContext.Current
+    /// 2. Avoid to be overriden when it is RequestTelemetry
+    /// Original Source is here https://github.com/microsoft/ApplicationInsights-dotnet-server/blob/2.8.0/Src/Common/W3C/W3COperationCorrelationTelemetryInitializer.cs.
+    /// </summary>
+    [Obsolete("Not ready for public consumption.")]
+    [EditorBrowsable(EditorBrowsableState.Never)]
+#if DEPENDENCY_COLLECTOR
```
ConnorMcMahon, line 28: What is this new condition?

src/WebJobs.Extensions.DurableTask/Options/TraceOptions.cs
```diff
@@ -37,6 +37,19 @@ public class TraceOptions
         /// </value>
         public bool TraceReplayEvents { get; set; }
 
+        /// <summary>
+        /// Gets or sets a flag indicating whether to disable distributed tracing.
+        /// The default value is false.
+        /// </summary>
+        public bool DistributedTracingDisabled { get; set; } = false;
```
ConnorMcMahon, line 44: Let's conditionally include these.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -52,14 +52,21 @@
...
     <PackageReference Include="Microsoft.AspNetCore.Routing" Version="1.0.5" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.WebApiCompatShim" Version="1.0.5" />
+    <PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="2.2.1" />
+    <PackageReference Include="Microsoft.Azure.DurableTask.AzureStorage" Version="1.7.2" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.14" />
-    <PackageReference Include="Microsoft.AspNetCore.Mvc.WebApiCompatShim" Version="2.2.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.WebApiCompatShim" Version="2.1.0" />
```
ConnorMcMahon, line 61: I vaguely remember some weirdness here, where we had to use a specific version to avoid breaking a build. I'll see if I can dig up the issue.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -52,14 +52,21 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="1.0.2" />
     <PackageReference Include="Microsoft.AspNetCore.Routing" Version="1.0.5" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.WebApiCompatShim" Version="1.0.5" />
+    <PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="2.2.1" />
```
ConnorMcMahon, line 55: Might as well upgrade these to the versions of the Telemetry package you are using.

In fact, I would argue that we shouldn't even package a NET full framework compatible version of this package, as it has no benefits for Functions V1 customers. We would want to leave in all of our conditional compilation work so that it is easy to merge into the main branch once that is ready, but no reason to publish a V1 compatible version while they are still separate packages.

samples/correlation-csharp/FunctionAppCorrelation/local.settings.json
```diff
@@ -0,0 +1,8 @@
+{
+    "IsEncrypted": false,
+  "Values": {
+    "AzureWebJobsStorage": "UseDevelopmentStorage=true",
+    "FUNCTIONS_WORKER_RUNTIME": "dotnet",
+    "APPINSIGHTS_INSTRUMENTATIONKEY": "f152bd88-9b36-43ba-a8a6-ebf921f9fc8f"
```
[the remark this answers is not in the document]
TsuyoshiUshio: Done

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
TsuyoshiUshio: Done

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
TsuyoshiUshio: Done

src/WebJobs.Extensions.DurableTask/Correlation/TraceContextBaseExtensions.cs
TsuyoshiUshio: Done

src/WebJobs.Extensions.DurableTask/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
TsuyoshiUshio: Done

src/WebJobs.Extensions.DurableTask/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
TsuyoshiUshio: Removed the condition

src/WebJobs.Extensions.DurableTask/Options/TraceOptions.cs
TsuyoshiUshio: Done

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
TsuyoshiUshio: We talked about it, It is because of the issue that shibayan point out.

TsuyoshiUshio:
Hi @ConnorMcMahon 

I fix all of the issue that you point out! Could you run the AzDO pipeline? 
I seem don't have a permission. 
Once it passed. Let's get merge this.

TsuyoshiUshio:
According to the request of renaming from `DistributedTracingDisabled` to `DistributedTracingEnabled`, I change the PR. https://github.com/Azure/azure-functions-durable-extension/pull/1298

src/WebJobs.Extensions.DurableTask/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
```diff
@@ -0,0 +1,318 @@
...
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Correlation
+{
+    /// <summary>
+    /// Telemetry Initializer that sets correlation ids for W3C.
+    /// This source is based on W3COperationCorrelationTelemetryInitializer.cs
+    /// 1. Modified with CorrelationTraceContext.Current
+    /// 2. Avoid to be overriden when it is RequestTelemetry
+    /// Original Source is here https://github.com/microsoft/ApplicationInsights-dotnet-server/blob/2.8.0/Src/Common/W3C/W3COperationCorrelationTelemetryInitializer.cs.
+    /// </summary>
+    [Obsolete("Not ready for public consumption.")]
```
ConnorMcMahon, line 25: Is there a reason we are marking this as obsolete? In general, since this is internal, I don't see much benefit to this.

src/WebJobs.Extensions.DurableTask/Correlation/StringExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// <param name="traceparent">string representations of traceParent.</param>
+        /// <returns>TraceParent object.</returns>
+        public static TraceParent ToTraceParent(this string traceparent)
+        {
+            if (!string.IsNullOrEmpty(traceparent))
+            {
+                var substrings = traceparent.Split('-');
+                if (substrings.Length != 4)
+                {
+                    throw new ArgumentException($"Traceparent doesn't respect the spec. {traceparent}");
```
ConnorMcMahon, line 25: Maybe put the schema `{version}-{traceId}-{spanId}-{traceFlags}` in the exception message, so if we see this we don't need to look up the spec.

src/WebJobs.Extensions.DurableTask/Correlation/StringExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+                return new TraceParent
+                {
+                    Version = substrings[0],
+                    TraceId = substrings[1],
+                    SpanId = substrings[2],
+                    TraceFlags = substrings[3],
+                };
+            }
+
+            return null;
```
ConnorMcMahon, line 37: Is returning null valid here?

src/WebJobs.Extensions.DurableTask/Correlation/StringExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// StringExtensions are the extension method parse string into <see cref="TraceParent"></see> object.
+    /// </summary>
+    public static class StringExtensions
+    {
+        /// <summary>
+        /// Convert a traceParent string to TraceParent object.
+        /// </summary>
+        /// <param name="traceparent">string representations of traceParent.</param>
+        /// <returns>TraceParent object.</returns>
+        public static TraceParent ToTraceParent(this string traceparent)
```
ConnorMcMahon, line 18: I don't know if I love putting an extension method on such a primitive type like `string`. I would almost prefer we just call this static method `TraceParent.FromString()`, and move it under `TraceParent` class.

src/WebJobs.Extensions.DurableTask/Correlation/TraceContextBaseExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        /// <summary>
+        /// Create DependencyTelemetry from the Activity.
+        /// </summary>
+        /// <param name="context">TraceContext.</param>
+        /// <returns>DependencyTelemetry.</returns>
+        public static DependencyTelemetry CreateDependencyTelemetry(this TraceContextBase context)
+        {
+            var telemetry = new DependencyTelemetry { Name = context.OperationName };
+            telemetry.Start(); // TODO Check if it is necessary.
```
ConnorMcMahon, line 40: In general, before we merge this all to dev, I would like to make sure we have issues tracking all of the TODOs so that they don't just stay there forever. Doesn't have to be addressed now.

src/WebJobs.Extensions.DurableTask/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
ConnorMcMahon: I think you may want to mark the interface `ITelemetryInitializer` with this attribute, as that _is_ public. I wish it didn't have to be, but I haven't found a good way to have non-public interfaces be injected via dependency injection yet.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -149,7 +166,10 @@ public DurableTaskExtension()
             // The RPC server is started when the extension is initialized.
             // The RPC server is stopped when the host has finished shutting down.
             hostLifetimeService.OnStopped.Register(this.StopLocalRcpServer);
+            this.telemetryActivator = telemetryActivator;
+            this.telemetryActivator?.Initialize();
 #endif
+
```
ConnorMcMahon, line 172: nit: unecessary whitespace.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -102,8 +111,16 @@ public DurableTaskExtension()
             IApplicationLifetimeWrapper hostLifetimeService,
             IDurableHttpMessageHandlerFactory durableHttpMessageHandlerFactory = null,
             ILifeCycleNotificationHelper lifeCycleNotificationHelper = null,
+#if !FUNCTIONS_V1
```
ConnorMcMahon, line 114: small nit: we don't need to include `IMessageSerializerSettingsFactory messageSerializerSettingsFactory = null,` in the conditional compilation. 

Overall, very big improvement over previous iteration!

src/WebJobs.Extensions.DurableTask/Options/TraceOptions.cs
```diff
@@ -37,6 +37,22 @@ public class TraceOptions
         /// </value>
         public bool TraceReplayEvents { get; set; }
 
+#if !FUNCTIONS_V1
+
```
ConnorMcMahon, line 41: nit: extra whitespace.

src/WebJobs.Extensions.DurableTask/Options/TraceOptions.cs
```diff
@@ -37,6 +37,22 @@ public class TraceOptions
...
+        public bool DistributedTracingEnabled { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets a protocol for distributed Tracing.
+        /// Possible values are "HttpCorrelationProtocol" and "W3CTraceContext".
+        /// The default value is "HttpCorrelationProtocol".
+        /// </summary>
+        public string DistributedTracingProtocol { get; set; } = "HttpCorrelationProtocol";
+#endif
+
```
ConnorMcMahon, line 55: nit: I think we want this whitespace before the #endif. 

I recommend building in both netstandard and net461 to see if you get stylecop errors in either. That should help diagnose whitespace issues in conditionally compiled segments.

src/WebJobs.Extensions.DurableTask/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
TsuyoshiUshio: I remove the Obsolute attribute. The reason it was there is, I took this code from https://github.com/microsoft/ApplicationInsights-dotnet-server/blob/2.8.0/Src/Common/W3C/W3COperationCorrelationTelemetryInitializer.cs as I leave the comment on it. After that I modify it. So the first time, I just keep the same structure, however, since I modify it, it means nothing. I removed it and remove warning suppression at the TelemetryActivator.

src/WebJobs.Extensions.DurableTask/Correlation/StringExtensions.cs
TsuyoshiUshio: Fixed. It is very good idea!

src/WebJobs.Extensions.DurableTask/Correlation/StringExtensions.cs
TsuyoshiUshio: It won't happen for the normal case. It will happen if the traceparent is null or empty. Do you prefer Exception instead? @ConnorMcMahon

TsuyoshiUshio: It won't happen. However, I change it to the Exception.

src/WebJobs.Extensions.DurableTask/Correlation/StringExtensions.cs
TsuyoshiUshio: Done

src/WebJobs.Extensions.DurableTask/Correlation/TraceContextBaseExtensions.cs
TsuyoshiUshio: This TODO is not needed. Removed.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
TsuyoshiUshio: Done

src/WebJobs.Extensions.DurableTask/Options/TraceOptions.cs
TsuyoshiUshio: I've done the publishing. Both frameworks works.

src/WebJobs.Extensions.DurableTask/Options/TraceOptions.cs
TsuyoshiUshio: Done

TsuyoshiUshio:
@ConnorMcMahon I've done the all fix.

PR #1271, opened 2020-03-11
Allows out-of-proc Orchestration to SignalEntity

davidmrdavid:
This is the C# component of: https://github.com/Azure/azure-functions-durable-js/issues/134

This PR extends the orchestrator shim to accept SignalEntity messages from an out-of-process language like JavaScript. It's quite a _simple_ change but it also contains some preliminary work for requesting Locks from Entities by out-of-process languages. I'd be happy to remove the latter if you all believe it would be better suited for its own PR.

Also, this is my first PR here so I'm happy to take suggestions on formatting and others.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -370,7 +370,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             {
                 throw new ArgumentNullException(nameof(retryOptions));
             }
-            
             return this.CallDurableTaskFunctionAsync<TResult>(functionName, FunctionType.Activity, false, null, null, retryOptions, input, null);
         }
 
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -370,6 +370,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             {
                 throw new ArgumentNullException(nameof(retryOptions));
             }
+
             return this.CallDurableTaskFunctionAsync<TResult>(functionName, FunctionType.Activity, false, null, null, retryOptions, input, null);
         }
 
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -370,7 +370,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             {
                 throw new ArgumentNullException(nameof(retryOptions));
             }
-
+            
             return this.CallDurableTaskFunctionAsync<TResult>(functionName, FunctionType.Activity, false, null, null, retryOptions, input, null);
         }
 
--- a/src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
@@ -190,9 +189,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
             [JsonProperty("input")]
             internal object Input { get; set; }
-
-            [JsonProperty("instanceIDs")]
-            internal string[] InstanceIDs { get; set; }
 
             [JsonProperty("fireAt")]
             internal DateTime FireAt { get; set; }
--- a/src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
@@ -40,7 +40,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             CallEntity = 7,
             CallHttp = 8,
             SignalEntity = 9,
-            Lock = 10,
         }
 
         internal async Task HandleOutOfProcExecutionAsync(JObject executionJson)
@@ -132,6 +131,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                                 tasks.Add(this.context.CallEntityAsync(entityId, action.EntityOperation, action.Input));
                                 break;
                             }
+
                         case AsyncActionType.SignalEntity:
                             {
                                 // We do not add a task because this is 'fire and forget'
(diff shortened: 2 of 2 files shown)


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -370,7 +370,6 @@ Task<TResult> IDurableOrchestrationContext.CallActivityWithRetryAsync<TResult>(s
             {
                 throw new ArgumentNullException(nameof(retryOptions));
             }
-
```
ConnorMcMahon: nit: I think our style cop configuration will yell at this. This will create a warning, which will fail on release builds.

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -38,6 +39,8 @@ private enum AsyncActionType
             WaitForExternalEvent = 6,
             CallEntity = 7,
             CallHttp = 8,
+            SignalEntity = 9,
+            Lock = 10,
```
ConnorMcMahon: Let's wait on Lock until we can fully support it.

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -124,9 +127,18 @@ private async Task ProcessAsyncActions(AsyncAction[][] actions)
                             tasks.Add(this.context.CallSubOrchestratorWithRetryAsync(action.FunctionName, action.RetryOptions, action.InstanceId, action.Input));
                             break;
                         case AsyncActionType.CallEntity:
-                            var entityId = EntityId.GetEntityIdFromSchedulerId(action.InstanceId);
-                            tasks.Add(this.context.CallEntityAsync(entityId, action.EntityOperation, action.Input));
-                            break;
+                            {
```
ConnorMcMahon: nit: I'm not necessarily opposed to using `{}`s in switch statements, but it currently doesn't match our existing code, so I would remove them.

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
davidmrdavid: Understood, will remove it

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
davidmrdavid: The `{}` are there because, otherwise, the compiler **complains** about having two variables `entityId` declared in the same scope, where the scope is the whole `switch`-statement. In other words, the `{}` help me define a new scope. 

I was certainly unsure how to handle this. I suppose I could declare `entityId` in the beginning of the switch and simply assign it in the `callEntity` and `signalEntity` cases. What do you think?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
davidmrdavid: I see, thanks for the suggestion. Is there a way I can locally run the style cop? So that I can test it in the future.

I'll add the new line :)

davidmrdavid: In fact, I'm adding newlines after every `{}`, since the editor displays a warning / green swiggly line otherwise.

ConnorMcMahon: Yep style-cop integrates very nicely with VS, (and probably VS Code, though I haven't tried that).

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
ConnorMcMahon: I'm fine leaving it. Forgot about how variable context works in non-scoped switch statements.

The other option that would work would be to inline `EntityId.GetEntityIdFromSchedulerId(action.InstanceId)`, but I am not convinced that is any better than just leaving the scoping `{}`.

I wouldn't recommend declaring `entityId` before the switch statement, as it only really makes sense in two of the cases, and may even throw an error if a non-entity instance id is used.

Let's just leave it.

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -178,6 +191,9 @@ private class AsyncAction
             [JsonProperty("input")]
             internal object Input { get; set; }
 
+            [JsonProperty("instanceIDs")]
```
ConnorMcMahon: nit: This field is no longer needed.

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
```
ConnorMcMahon: nit: I don't think we are using Linq anymore.

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -124,9 +126,20 @@ private async Task ProcessAsyncActions(AsyncAction[][] actions)
...
+                                var entityId = EntityId.GetEntityIdFromSchedulerId(action.InstanceId);
+                                tasks.Add(this.context.CallEntityAsync(entityId, action.EntityOperation, action.Input));
+                                break;
+                            }
+
+                        case AsyncActionType.SignalEntity:
+                            {
+                                // We do not add a task because this is 'fire and forget'
+                                var entityId = EntityId.GetEntityIdFromSchedulerId(action.InstanceId);
+                                this.context.SignalEntity(entityId, action.EntityOperation, action.Input);
```
sebastianburckhardt: why not  support the more general version of signal  (scheduled signals) since you are at it? It's just one more argument being passed through. You can perhaps just reuse the FireAt field already there for the time.

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
davidmrdavid: done

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
davidmrdavid: done

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
davidmrdavid: Ah, I didn't know about this variant. I'll take a stab at it

ConnorMcMahon: Good catch Sebastian. I forgot all about that.

davidmrdavid: I added the scheduled signal as a separate action code and tested via JavaScript that it can be reached

Issue #1276, opened 2020-03-18
Allow for Durable Entities Proxy to implement multiple interfaces


alvpickmans:
**Is your feature request related to a problem? Please describe.**
As per [documentation](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-dotnet-entities#restrictions-on-entity-interfaces) regarding interfaces in durable entities, it specifies some constraints they must comply with. Although it doesn't explicitly say DEs classes can only implement a single interface, the code enforces it [here](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/EntityScheduler/TypedInvocationExtensions.cs#L104-L107) and [here](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/EntityScheduler/Proxy/EntityProxyFactory.cs#L91). As a simple example, the a DE class implementing the below `IPrimaryJob` below is not currently permitted:
```c#
       public interface IJob
        {
            void SetStartDate(DateTime date);
            void SetEndDate(DateTime date);
            Task<TimeSpan> GetDuration();
            void Delete();
        }

        public interface IPrimaryJob : IJob
        {
            void SetId(string Id);
            Task<string> GetId();
        }
```

**Describe the solution you'd like**
Being able to use multiple interfaces on DEs classes.

**Describe alternatives you've considered**
Duplicating interfaces to comply with current proxy limitations on DEs.

**Additional context**
I see the value of using global or more generic interfaces that children of these can extend to ensure and enforce a minimum common set of methods.

I have a PR ready that implements this with corresponding test.

sebastianburckhardt:
I believe this is working as intended. The code snippets that throw an exception are executed only when using the "shorter version of the entity ID parameter", which is a feature for simplifying the common case where there is only one implementation per interface by omitting the entity name. This is documented [https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-dotnet-entities#shorter-option-for-specifying-the-target](here). 

If there is more than one implementation, one simply has to use the more verbose form of passing an entity id, which includes the entity name.

alvpickmans:
The issue I encountered is not multiple classes implementing the same interface, but instead a single class implementing more than one interface. Taking the example below, the runtime throws the following exception:
```
[...]

PR #1277, opened 2020-03-18
Durable Entity proxy with multiple interfaces

alvpickmans:
Addresses #1276 

This PR allows for Durable Entities classes to be proxied by more than one implemented interface. This current restriction is not explicitly defined on [documentation](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-dotnet-entities#restrictions-on-entity-interfaces).

All tests are passing, except for [`ReplaySafeLogger_LogsOnlyOnce`](https://github.com/Azure/azure-functions-durable-extension/blob/dev/test/Common/DurableTaskEndToEndTests.cs#L4310-L4336) on tests for V1 (not sure how to/if address it).
image

@cgillum @ConnorMcMahon

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/EntityScheduler/TypedInvocationExtensions.cs
+++ b/src/WebJobs.Extensions.DurableTask/EntityScheduler/TypedInvocationExtensions.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Linq;
 using System.Reflection;
 using System.Threading.Tasks;
 
@@ -101,11 +102,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             var type = typeof(T);
 
             var interfaces = type.GetInterfaces();
-            if (interfaces.Length > 1)
-            {
-                throw new InvalidOperationException("Only a single interface can be implemented on an entity");
-            }
-
             const BindingFlags bindingFlags = BindingFlags.IgnoreCase | BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance;
 
             var method = type.GetMethod(context.OperationName, bindingFlags);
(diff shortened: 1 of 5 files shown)


ConnorMcMahon:
Thanks for the submission @alvpickmans.

Today we are hoping to release 2.2.0, but I will definitely look at this as a merge candidate for 2.2.1!

src/WebJobs.Extensions.DurableTask/EntityScheduler/TypedInvocationExtensions.cs
```diff
@@ -114,8 +110,8 @@ internal static MethodInfo FindMethodForContext<T>(IDurableEntityContext context
                 return method;
             }
 
-            var entityInterface = interfaces[0];
-            return entityInterface.GetMethod(context.OperationName, bindingFlags);
+            var methods = interfaces.SelectMany(i => i.GetMethods(bindingFlags));
+            return methods.FirstOrDefault(m => m.Name.Equals(context.OperationName));
```
sebastianburckhardt, line 114: Previously this was throwing an `AmbiguousMatchException` when there are multiple matches for the name. It may not be important to do the same check across all interfaces but we could at least do it within each interface, e.g. something like
```
interfaces.Select(i => i.GetMethod(context.OperationName, bindingFlags)).FirstOrDefault(m => m != null);

```

test/Common/TestEntityClasses.cs
```diff
@@ -50,6 +50,24 @@ public interface IAsyncCounter
...
+            void SetEndDate(DateTime date);
+
+            Task<TimeSpan> GetDuration();
+
+            void Delete();
+        }
+
+        public interface IPrimaryJob : IJob
+        {
+            void SetId(string Id);
```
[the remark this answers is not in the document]
alvpickmans: detailed eye in there :)

src/WebJobs.Extensions.DurableTask/EntityScheduler/TypedInvocationExtensions.cs
alvpickmans: fair enough, it does make sense. I was a bit restless tbh by specifying a `BindingFlag.IgnoreCase` without getting the method itself by name.

oising:
Thank you for this! It's a small thing with broad scope to improve many aspects of design.

PR #1283, opened 2020-03-20
Updated TokenSource serializer

amdeel:
Makes a copy of the input serializer and sets TypeNameHandling to Objects to assure correct serialization of ITokenSource in case customers use their own custom serialization settings.

Updated the past implementation to now copy the input settings and add TypeNameHandling.Objects to avoid changing the input JsonSerializer.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
@@ -114,6 +114,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private class TokenSourceConverter : JsonConverter
         {
+            private static JsonSerializer tokenSerializer;
+
             private enum TokenSourceType
             {
                 None = 0,
@@ -127,6 +129,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
             public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
             {
+                this.CreateTokenSourceSerializer(serializer);
+
                 JToken json = JToken.ReadFrom(reader);
                 if (json.Type == JTokenType.Null)
                 {
@@ -147,7 +151,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 else if (jsonObject.TryGetValue("$type", StringComparison.Ordinal, out JToken clrTypeValue))
                 {
                     Type runtimeType = Type.GetType((string)clrTypeValue, throwOnError: true);
-                    return jsonObject.ToObject(runtimeType, serializer);
+                    return jsonObject.ToObject(runtimeType, tokenSerializer);
                 }
                 else
                 {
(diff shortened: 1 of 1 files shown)


src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
```diff
@@ -170,10 +174,57 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
...
-                    var currTypeNameSetting = serializer.TypeNameHandling;
-                    serializer.TypeNameHandling = TypeNameHandling.Objects;
-                    serializer.Serialize(writer, value);
-                    serializer.TypeNameHandling = currTypeNameSetting;
+                    this.CreateTokenSourceSerializer(serializer);
+                    tokenSerializer.Serialize(writer, value);
+                }
+            }
+
+            private void CreateTokenSourceSerializer(JsonSerializer serializer)
```
ConnorMcMahon: small nits:

1. The method could be `static`.
2. In general, instead of instantiating a variable if it doesn't exist in a void method, and then referencing the variable that was instantiated, an approach often used is to use `GetTokenSourceSerializer(JsonSerializer serializer)`. This way you never directly call `tokenSerializer`, you just call your `Get` method, that you know will never be null. You have effectively accomplished that here, but the Get method approach is a bit less error prone in general.

Neither of these nits are dealbreakers for merging for me.

src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
```diff
@@ -170,10 +174,57 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
...
+                        MissingMemberHandling = serializer.MissingMemberHandling,
+                        NullValueHandling = serializer.NullValueHandling,
+                        ObjectCreationHandling = serializer.ObjectCreationHandling,
+                        PreserveReferencesHandling = serializer.PreserveReferencesHandling,
+                        ReferenceResolver = serializer.ReferenceResolver,
+                        ReferenceLoopHandling = serializer.ReferenceLoopHandling,
+                        StringEscapeHandling = serializer.StringEscapeHandling,
+                        TraceWriter = serializer.TraceWriter,
+
+                        // Enforcing TypeNameHandling.Objects to make suire ITokenSource gets serialized/deserialized correctly
```
ConnorMcMahon: typo: "sure"

src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
```diff
@@ -170,11 +174,60 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
...
+            }
+
+            private static JsonSerializer GetTokenSourceSerializer(JsonSerializer serializer)
+            {
+                if (tokenSerializer == null)
+                {
+                    if (serializer.TypeNameHandling == TypeNameHandling.Objects
+                    || serializer.TypeNameHandling == TypeNameHandling.All)
+                    {
+                        tokenSerializer = serializer;
```
ConnorMcMahon: make sure to return here, otherwise you will still be overwriting `tokenSerializer`.

Issue #1290, opened 2020-03-24
Allow specifying "orchestrationClient" instead of "durableClient" in function.json (v2.x)


cgillum:
## Problem statement
We made a breaking change from Durable Functions v1 to v2 by renaming the [durable client binding](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-bindings#orchestration-client) from `orchestrationClient` to `durableClient`. This is fine, and we documented the change, but we recently decided that we want to make it easier for customers to upgrade from v1.x to v2.x.

For JavaScript customers specifically, the only change they need to make is to update their orchestration client bindings in their **function.json** files.  Ideally we can remove this requirement to make the transition much smoother. Without this backwards compatibility, it could be hard for them to figure out why things aren't working when they upgrade their extension bundles from v1 to v2.

## Proposed solution
I would like it if we could support both "durableClient" **or** "orchestrationClient" in functions.json. It might be as simple as defining a new public `OrchestrationClientAttribute` class in the .NET extension and having it derive from `DurableClientAttribute`. Perhaps something like this:

```csharp
[Binding]
[AttributeUsage(AttributeTargets.Parameter)]
[Obsolete($"{nameof(OrchestrationClientAttribute)} is obsolete. Use {nameof(DurableClientAttribute)} instead.")]
public class OrchestrationClientAttribute : DurableClientAttribute
{
}
```

A few notes:
* We want to mark this as `[Obsolete]` so that C# developers don't try to use it.
* We will need to remove the `sealed` part of the `DurableClientAttribute` declaration or else the C# compiler won't let us derive from that base attrbute.
* I don't think anyone has every tried this before (deriving from a base class attribute) and I'm not actually sure if this will even work. We'll need to test it. If it doesn't work, we probably will need to define two unrelated attributes that are treated the same everywhere in code.

If successful, we should be able to take a JavaScript sample, replace `"type":"durableClient"` with `"type":"orchestrationClient"` in function.json and have everything still work.

cgillum:
@davidmrdavid this might be a good work item for you to tackle for the next extension release. What do you think? @ConnorMcMahon should be able to provide some guidance as well.

FYI @anthonychu

ConnorMcMahon:
I think we have tried base class attributes before and ran into some issues. Unfortunately, I don't remember those issues, so it's possible they don't apply here.

[...]

PR #1295, opened 2020-03-27
Backwards compatibility for orchestrationClient

davidmrdavid:
This PR addresses: https://github.com/Azure/azure-functions-durable-extension/issues/1290

There changes are few and simple (_except I apparently pushed other people's commits, more on this later_):

1. `DurableClientAttribute` is no longer a `sealed` class, so we can subclass it.

2. The declaration of a new dummy subclass of `DurableClientAttribute` named `OrchestrationClientAttribute`

3. A new binding rule in `DurableTaskExtension.cs` that explicitly accepts `OrchestrationClientAttribute`. This solution is less than ideal, because there's a handful of copy-pasted lines from the rule accepting `DurableClientAttribute`. Ideally, we would have a single line allowing us to select between the two attributes. If anyone knows how to do this, *please let me know*

So, while this works, I'm still looking for a solution with less repetition. I'm open to input!

Finally, in  https://github.com/Azure/azure-functions-durable-extension/issues/1290 we discuss the possibility of raising warnings on the logs of `func host start` for JS users using `orchestrationClient` instead of `durableClient`. This PR assumes that this is change we would implement exclusively on the JS side of things. If this is not the case, please let me know.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -252,6 +252,22 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             rule.BindToInput<IDurableEntityClient>(this.GetClient);
             rule.BindToInput<IDurableClient>(this.GetClient);
 
+            // We add a binding rule to support the now deprecated `orchestrationClient` binding
+            // A cleaner solution would be to have the prior rule have an OR-style selector between
+            // DurableClientAttribute and OrchestrationClientAttribute, but it's unclear if that's
+            // possible (for now).
+#pragma warning disable CS0618 // Type or member is obsolete
+            var backwardsCompRule = context.AddBindingRule<OrchestrationClientAttribute>()
+#pragma warning restore CS0618 // Type or member is obsolete
+                .AddConverter<string, StartOrchestrationArgs>(bindings.StringToStartOrchestrationArgs)
+                .AddConverter<JObject, StartOrchestrationArgs>(bindings.JObjectToStartOrchestrationArgs)
+                .AddConverter<IDurableClient, string>(bindings.DurableOrchestrationClientToString);
+
+            backwardsCompRule.BindToCollector<StartOrchestrationArgs>(bindings.CreateAsyncCollector);
+            backwardsCompRule.BindToInput<IDurableOrchestrationClient>(this.GetClient);
+            backwardsCompRule.BindToInput<IDurableEntityClient>(this.GetClient);
+            backwardsCompRule.BindToInput<IDurableClient>(this.GetClient);
+
             string storageConnectionString = null;
             var providerFactory = this.durabilityProviderFactory as AzureStorageDurabilityProviderFactory;
             if (providerFactory != null)
--- a/test/Common/DurableTaskHostExtensions.cs
+++ b/test/Common/DurableTaskHostExtensions.cs
@@ -105,7 +105,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
             ITestOutputHelper output)
         {
             var startFunction = typeof(ClientFunctions)
-                .GetMethod(nameof(ClientFunctions.GetEntityClientBindingTest));
+                .GetMethod(nameof(ClientFunctions.GetEntityClientBindingBackCompTest));
 
             var clientRef = new IDurableEntityClient[1];
             var args = new Dictionary<string, object>
@@ -126,7 +126,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
             ITestOutputHelper output)
         {
             var startFunction = typeof(ClientFunctions)
-                .GetMethod(nameof(ClientFunctions.GetOrchestrationClientBindingTest));
+                .GetMethod(nameof(ClientFunctions.GetOrchestrationClientBindingBackCompTest));
 
             var clientRef = new IDurableOrchestrationClient[1];
             var args = new Dictionary<string, object>
--- a/test/Common/DurableTaskHostExtensions.cs
+++ b/test/Common/DurableTaskHostExtensions.cs
@@ -96,47 +96,5 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
             IDurableOrchestrationClient client = clientRef[0];
             return client;
         }
-
-        /// <summary>
-        /// Helper function for the IDurableEntityClientBindingBackComp test. Gets an IDurableEntityClient.
-        /// </summary>
-        public static async Task<IDurableEntityClient> GetEntityClientBindingBackCompTest(
-            this ITestHost host,
-            ITestOutputHelper output)
-        {
-            var startFunction = typeof(ClientFunctions)
-                .GetMethod(nameof(ClientFunctions.GetEntityClientBindingBackCompTest));
-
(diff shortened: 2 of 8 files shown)


davidmrdavid:
Hmm, I see a few too many commits being pushed. Perhaps I pulled from the wrong source? Sorry about that, I'll look into fixing that next!

cgillum:
Yeah, that's definitely weird. I typically see it if by branches aren't fully up-to-date. I don't yet really understand what causes it though.

One thing I would suggest is to exclude your changes under the `docs` directory. This is actually all the auto-generated documentation that existed before we migrated to docs.microsoft.com. We haven't gotten around to deleting this content from our repo. 😅

cgillum:
Also, it looks like the AppVeyor build is failing due to the usage of an obsolete API. 

>   DurableTaskExtension.cs(259,60): error CS0618: 'OrchestrationClientAttribute' is obsolete: '{OrchestrationClientAttribute is obsolete. Use DurableClientAttribute instead.' [C:\projects\azure-functions-durable-extension\src\WebJobs.Extensions.DurableTask\WebJobs.Extensions.DurableTask.csproj]

We should suppress that warning in the source code since we're doing this intentionally. I think Visual Studio should have some tool tips to help suppress warnings in source code.

davidmrdavid:
After *alot* of work trying to fix my git state, the diff appears right now. Sorry for the clutter folks, I'll be more careful next time. My mistake was to merge accidentally master and to repeatedly fail to perform a `git diff` from local to remote.

cgillum:
Glad to see everything is working! :)

However, I'm still a little concerned about the Git history represented in this PR. If you're certain that your local branch is good, would you mind locally squashing your changes and doing a force push to make sure that GitHub also thinks things are looking clean?

I imaging the following commands could work for you assuming your branch is in a sane state. They will do the following:

1. Rebase your commits on top of the latest from `dev` to ensure maximum cleanliness
2.  Squash your changes into a single commit
3. Overwrite the history of this PR to match the *new* history of your local branch

```bash
git checkout dev
git pull
git checkout djusto/backwards_compatibility
git rebase dev

git reset --soft 4b81619ddd31043f63a1e351e33367c8de65f368
git commit -m "Backwards compatibility for orchestrationClient"

git push --force
```

davidmrdavid:
@cgillum , thanks for the helpful commands and their explanation. I spent a little bit of time today also reading up on how to get out of these bad git states. Anyhow, it should be looking better now :)

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -252,6 +252,22 @@ void IExtensionConfigProvider.Initialize(ExtensionConfigContext context)
             rule.BindToInput<IDurableEntityClient>(this.GetClient);
             rule.BindToInput<IDurableClient>(this.GetClient);
 
+            // We add a binding rule to support the now deprecated `orchestrationClient` binding
+            // A cleaner solution would be to have the prior rule have an OR-style selector between
+            // DurableClientAttribute and OrchestrationClientAttribute, but it's unclear if that's
+            // possible (for now).
+#pragma warning disable CS0618 // Type or member is obsolete
+            var backwardsCompRule = context.AddBindingRule<OrchestrationClientAttribute>()
```
amdeel: Why is this needed? Shouldn't this work without these changes with inheritance?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -252,6 +252,22 @@ void IExtensionConfigProvider.Initialize(ExtensionConfigContext context)
...
+#pragma warning disable CS0618 // Type or member is obsolete
+            var backwardsCompRule = context.AddBindingRule<OrchestrationClientAttribute>()
+#pragma warning restore CS0618 // Type or member is obsolete
+                .AddConverter<string, StartOrchestrationArgs>(bindings.StringToStartOrchestrationArgs)
+                .AddConverter<JObject, StartOrchestrationArgs>(bindings.JObjectToStartOrchestrationArgs)
+                .AddConverter<IDurableClient, string>(bindings.DurableOrchestrationClientToString);
+
+            backwardsCompRule.BindToCollector<StartOrchestrationArgs>(bindings.CreateAsyncCollector);
+            backwardsCompRule.BindToInput<IDurableOrchestrationClient>(this.GetClient);
+            backwardsCompRule.BindToInput<IDurableEntityClient>(this.GetClient);
```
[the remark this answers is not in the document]
ConnorMcMahon: Note that if you go the route where you get rid of `backwardsCompRule`, you can disregard this feedback.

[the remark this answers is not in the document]
cgillum: We already discourage users from using `[OrchestrationClient]` by putting the `[Obsolete]` attribute on it.  But the whole reason for `backCompRule` is that inheritance by itself is not sufficient, per David's investigation last week.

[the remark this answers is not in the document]
ConnorMcMahon: Good to know inheritance didn't work. I must either be misremembering the behavior of binding rules, or they have changed since I last looked at the code.

In that case, while I still don't see a good reason to include support for the backwards compatible binding with a V2 only feature, it probably doesn't hurt to include the support, so this nit can be disregarded.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
ConnorMcMahon: As mentioned on another issue, it sounds like you already tried removing this, but it didn't work.

davidmrdavid: @amdeel @ConnorMcMahon 
Yup! This code is only here because inheritance didn't work by itself. 
It appears that inheritance is enough to _register_ the binding, so a `func host start` runs successfully. However, invoking the function will trigger a runtime exception, so inheritance-alone only seems to take us half of the way there.

I don't think this is the expected behaviour, since it surprises all of us. I would love to follow-up with the WebJobs folks to tell them about our usecase

davidmrdavid:
My latest commit(s) adds _quite simple_ unit tests as a sanity check that my changes work. They are slight variants on already existing tests for the `DurableClientAttribute`.

I also manually tested the usage of `orchestrationClient` as a binding in JS and C#, and a handful of samples seem to continue working. All in all, I think this is ready.

I'll wait for another 'lgtm' and them I'll squash-and-merge. Thanks folks!

test/Common/DurableTaskHostExtensions.cs
```diff
@@ -96,5 +96,47 @@ internal static class DurableTaskHostExtensions
             IDurableOrchestrationClient client = clientRef[0];
             return client;
         }
+
+        /// <summary>
+        /// Helper function for the IDurableEntityClientBindingBackComp test. Gets an IDurableEntityClient.
+        /// </summary>
+        public static async Task<IDurableEntityClient> GetEntityClientBindingBackCompTest(
```
ConnorMcMahon, line 103: nit: I realize you just copied the approach from previous tests, but since this logic is only used for one test, I think I prefer having it actually appear in the test instead of in a helper method. Otherwise just looking at the test doesn't give us a clear idea of what is happening.

davidmrdavid: Sounds good, will fix and merge!

Issue #1172, opened 2020-01-19
Allow using App Insights to query current state of orchestrations and entities


anthonychu:
Add custom dimensions to App Insights requests so that we can use the requests table to query recently run orchestrations and their current status.

Orchestrations: add instanceId and status.
Entities: add entityId and operationName.

Need to verify this is possible from the extension and that we can write the query based on those custom dimensions.

The goal is to create specialized monitor views in the portal for orchestrations and entities.

anthonychu:
Ran some tests. Looks like this is going to work.

```csharp
using System.Diagnostics;
// ...
var activity = Activity.Current;
activity.AddTag("DurableFunctionsType", "Orchestration");
activity.AddTag("DurableFunctionsInstanceId", instanceId);
activity.AddTag("DurableFunctionsStatus", status);
```

App Insights query to pull the 30 most recently run orchestrators and their current status:

```
requests
| extend DurableFunctionsInstanceId = tostring(customDimensions["DurableFunctionsInstanceId"]),
         DurableFunctionsStatus = tostring(customDimensions["DurableFunctionsStatus"]),
         DurableFunctionsType = tostring(customDimensions["DurableFunctionsType"])
| where DurableFunctionsType == "Orchestration"
| summarize arg_max(timestamp, *) by DurableFunctionsInstanceId 
| order by timestamp desc
| limit 30
```

Thanks to @brettsam for helping with this.

anthonychu:
@cgillum Thoughts on how this might look like for entities? Does a "status" make sense or would something else be more interesting (e.g. has state, no state)?

cgillum:
I'm not entirely sure what would make sense for the current "status" of an entity. Maybe the size of its current state in bytes? Adding @sebastianburckhardt in case he has thoughts on this.

sebastianburckhardt:
We currently already return selected status info in `TaskEntityShim.GetStatus`, as a JSON object. Those same fields probably would make sense to use:
```csharp
return this.dataConverter.MessageConverter.Serialize(new EntityStatus()
                {
                    EntityExists = this.context.State.EntityExists,
                    QueueSize = this.context.State.Queue?.Count ?? 0,
                    LockedBy = this.context.State.LockedBy,
                    CurrentOperation = opStatus,
                });
```
Most of the time these fields are uninteresting (because requests are typically processed instantly). But if operations have high latency or entities are locked one could potentially see that here, so it can help with troubleshooting.

[...]

PR #1301, opened 2020-04-02
App Insights Tags for Orchestrators and Entities

davidmrdavid:
This addresses: https://github.com/Azure/azure-functions-durable-extension/issues/1172

This PR aims to add custom App Insights dimensions to enable querying recent orchestrators and entity calls.

For Orchestrators, we add the following fields:
- `DurableFunctionsType` as "Orchestrator"
- `DurableFunctionsInstanceID`
- `DurableFunctionsStatus`

For Entities, we add the following fields:
- `DurableFunctionsType` as "Entity".  I believe this tag name is a confusing, I'm open to improvements. Overall, it was unclear to me if these tags should also be prefixed with `DurableFunctions-` or `DurableEntity`
- `DurableEntityInstanceID`
- `DurableEntityStatus`

I added these tags at the beginning of the "execute" methods, but if mid-execution their state is changed, then maybe this is not a good idea. Please let me know what you think. Overall, the current approach is rather naive so I expect to iterate on this.

Also, I realize that `DurableEntityStatus` returns a string-version of what are really JSON-codified fields. I'd be happy to expand those into proper tags as well.

**Update 1:**
Below is a screenshot of how the changes look for Entities. I'm keep the screenshot small as to avoid leaking any sensitive information. This is the counter entity from the precompiled samples.
[
appInsights1
](url)

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
@@ -225,8 +225,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             }
 #if !FUNCTIONS_V1
             var activity = Activity.Current;
-            // activity may be null when running this repo's tests but does not appear
-            // empty when actually running the samples
+            /* activity may be null when running this repo's tests but does not appear
+             * empty when actually running the samples
+             */
             if (activity != null)
             {
                 string state = this.context.State.EntityExists ? $"has state ({this.context.State.EntityState.Length} characters)" : "no state";
--- a/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
@@ -148,7 +148,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         public override async Task<string> Execute(OrchestrationContext innerContext, string serializedInput)
         {
-            string status;
+#if !FUNCTIONS_V1
+            // for reporting the status of the entity on App Insights
+            OrchestrationRuntimeStatus statusAppInsights;
+#endif
             if (this.operationBatch.Count == 0 && this.lockRequest == null)
             {
                 // we are idle after a ContinueAsNew - the batch is empty.
@@ -168,8 +171,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 this.Config.GetIntputOutputTrace(serializedInput),
                 FunctionType.Entity,
                 isReplay: false);
-            status = "Running";
-
+#if !FUNCTIONS_V1
+            statusAppInsights = OrchestrationRuntimeStatus.Running;
+#endif
             if (this.NumberEventsToReceive > 0)
             {
                 await this.doneProcessingMessages.Task;
@@ -213,8 +217,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     description,
                     functionType: FunctionType.Entity,
                     isReplay: false);
-                status = "Failed";
-
+#if !FUNCTIONS_V1
+                statusAppInsights = OrchestrationRuntimeStatus.Failed;
+#endif
             }
             else
             {
@@ -226,11 +231,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     continuedAsNew: true,
                     functionType: FunctionType.Entity,
                     isReplay: false);
-                status = "Completed";
+#if !FUNCTIONS_V1
+                statusAppInsights = OrchestrationRuntimeStatus.Completed;
+#endif
             }
 
 #if !FUNCTIONS_V1
-            // Adding "Tags" to activity allows using App Insights to query current state of entities 
+            // Adding "Tags" to activity allows using App Insights to query current state of entities
             var activity = Activity.Current;
 
             // The activity may be null when running unit tests, but should be non-null otherwise
@@ -238,7 +245,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             {
                 activity.AddTag("DurableFunctionsType", "Entity");
                 activity.AddTag("DurableFunctionsInstanceId", this.context.InstanceId);
-                activity.AddTag("DurableFunctionsRuntimeStatus", status);
+                activity.AddTag("DurableFunctionsRuntimeStatus", Enum.GetName(statusAppInsights.GetType(), statusAppInsights));
             }
 #endif
 
--- a/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
@@ -228,13 +228,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             /* activity may be null when running this repo's tests but does not appear
              * empty when actually running the samples
              */
-            if (activity != null)
+            /*if (activity != null)
             {
                 string state = this.context.State.EntityExists ? $"has state ({this.context.State.EntityState.Length} characters)" : "no state";
                 activity.AddTag("DurableFunctionsType", "Entity");
                 activity.AddTag("DurableFunctionsInstanceId", this.context.InstanceId);
                 activity.AddTag("DurableFunctionsState", state);
-            }
+            }*/
 #endif
 
             // The return value is not used.
--- a/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
@@ -223,19 +223,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     functionType: FunctionType.Entity,
                     isReplay: false);
             }
-            //#if !FUNCTIONS_V1
-            //var activity = Activity.Current;
-            /* activity may be null when running this repo's tests but does not appear
-             * empty when actually running the samples
-             */
-            /*if (activity != null)
-            {
-                string state = this.context.State.EntityExists ? $"has state ({this.context.State.EntityState.Length} characters)" : "no state";
-                activity.AddTag("DurableFunctionsType", "Entity");
-                activity.AddTag("DurableFunctionsInstanceId", this.context.InstanceId);
-                activity.AddTag("DurableFunctionsRuntimeStatus", state);
-            }*/
-            //#endif
 
             // The return value is not used.
             return string.Empty;
--- a/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
@@ -223,8 +223,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     functionType: FunctionType.Entity,
                     isReplay: false);
             }
-#if !FUNCTIONS_V1
-            var activity = Activity.Current;
+            //#if !FUNCTIONS_V1
+            //var activity = Activity.Current;
             /* activity may be null when running this repo's tests but does not appear
              * empty when actually running the samples
              */
@@ -233,9 +233,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 string state = this.context.State.EntityExists ? $"has state ({this.context.State.EntityState.Length} characters)" : "no state";
                 activity.AddTag("DurableFunctionsType", "Entity");
                 activity.AddTag("DurableFunctionsInstanceId", this.context.InstanceId);
-                activity.AddTag("DurableFunctionsState", state);
+                activity.AddTag("DurableFunctionsRuntimeStatus", state);
             }*/
-#endif
+            //#endif
 
             // The return value is not used.
             return string.Empty;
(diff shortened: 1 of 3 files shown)


cgillum:
I had the same question about adding this information before or after the actual function execution. I think the right way to go is to do this at the _end_ of the execute method so that we have the right status value.

BTW, were you able to verify that this data gets correctly populated in App Insights? It would be great if you could provide some screenshots in your PR description showing what this looks like.

sebastianburckhardt:
Now that I see the context where the information shows up, I think including the entire `GetStatus()` result for the entity status is probably overkill. Perhaps going back to something simple that just shows the state size as suggested by @cgillum is better. This also avoids the overhead of allocating multiple objects and calling JSON serialization on every entity operation.

For example
```csharp
activity.AddTag("DurableEntityStatus", this.context.State.EntityExists ? $"has state ({this.context.State.EntityState.Length} characters)" : "no state");
```

As noted by others this should be done at the end of the operation to make sense (so it displays the final state, not the state before the operation).

anthonychu:
I'm open to suggestions about the prefix. However it feels like things that are the same between orchestrations and entities (like instance id) should share the same key.

As for what we output as statuses for entities, outputting nothing is an option as well. We can always add later.

The goal of this is to be able to list a summary of the last x orchestrations or entities that ran. Someone can click on a single instance id and the portal will render more details from the `traces` table using a query similar to [this](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-diagnostics#single-instance-query).

Should `...ID` be `...Id` to match other places in code?

anthonychu:
Maybe we should also match the naming convention to those we use already here: https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-diagnostics#single-instance-query

e.g., `DurableFunctionsStatus` should be `DurableFunctionsState`.

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -178,6 +179,18 @@ public override async Task<string> Execute(OrchestrationContext innerContext, st
...
+#if !FUNCTIONS_V1
+            var activity = Activity.Current;
+            /* activity may be null when running this repo's tests but does not appear
+             * empty when actually running the samples
+             */
+            if (activity != null)
+            {
+                activity.AddTag("DurableFunctionsType", "Orchestration");
+                activity.AddTag("DurableOrchestrationInstanceId", this.context.InstanceId);
+                activity.AddTag("DurableFunctionState", this.context.GetSerializedCustomStatus());
```
anthonychu, line 191: I think the "state" should be one of [these](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/OrchestrationRuntimeStatus.cs).

The custom status could be useful, but it can get quite large. I think we can add it later based on feedback.

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -148,6 +148,7 @@ internal void Rehydrate(string serializedInput)
 
         public override async Task<string> Execute(OrchestrationContext innerContext, string serializedInput)
         {
+            string status;
```
cgillum, line 151: Consider naming this variable `statusForTracking`, or something similar to help communicate the purpose of this data.

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -167,6 +168,7 @@ public override async Task<string> Execute(OrchestrationContext innerContext, st
                 this.Config.GetIntputOutputTrace(serializedInput),
                 FunctionType.Entity,
                 isReplay: false);
+            status = "Running";
```
cgillum, line 171: Rather than hardcoded strings, consider using `OrchestrationRuntimeStatus.Running.ToString()`. You could also consider using `OrchestrationRuntimeStatus` as the type of variable (instead of `string`) and then call `.ToString()` at the very end. The result is exactly the same, but it will help future maintainers understand what the acceptable range of status values are.

cgillum:
It looks like the App Veyor build failed because the C# compiler issued a warning when compiling against the .NET Framework (i.e. Functions v1). Basically, to the compiler it looks like you created a variable and assigned a value but never used it (we only use it for .NET Standard/Functions v2). A couple possible solutions:
1. Use `#if` for all usage of this variable
2. Suppress the warning for this specific instance.

I'm personally fine with either of these options. I'm open to other ideas as well.

davidmrdavid:
@cgillum , Yup! I finally figured out how to run the AppVeyor style cop on locally on Visual Studio, so those surprises should not be happening anymore. Fixed it using your first suggestion

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
davidmrdavid: That's a good point, hard-coded strings are flimsy. Incorporated your feedback in my latest changes!

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
davidmrdavid: Did so and left an explanatory comment as well :)

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
davidmrdavid: Good point, I incorporated these changes, to the extend possible, more details coming up in a separate comment

davidmrdavid:
Summary of the latest changes:

- To facilitate querying, I made Entities and Orchestrators share the same keys. A user should be able to filter between them by checking on the value of `DurableFunctionsType`
- Unfortunately, I cannot use probe the `OrchestrationRuntimeState ` returned by `          dispatchContext.GetProperty<OrchestrationRuntimeState>();` as a reliable way to report the `DurableFunctionsRuntimeStatus`. In my experience, that value was always set to "Running", which is unhelpful. 
    - As a result, you'll see that my current approach simply tracks the `TraceHelper` calls into a variable and uses that to report a value instead.
    - In general, it would be worthwhile to explore _why_ that value is always set to "Running" and whether or not we're in need of some refactoring as a result. Additionally, I also found the `string` retuned by `OrchestrationRuntimeState.Status` to be always `null`, adding to my concerns.

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -166,7 +171,9 @@ public override async Task<string> Execute(OrchestrationContext innerContext, st
                 this.Config.GetIntputOutputTrace(serializedInput),
                 FunctionType.Entity,
                 isReplay: false);
-
+#if !FUNCTIONS_V1
```
ConnorMcMahon, line 174: I think we can get rid of this chunk, because it will always be overwritten as `Completed` or `Failed`.

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -147,6 +148,10 @@ internal void Rehydrate(string serializedInput)
 
         public override async Task<string> Execute(OrchestrationContext innerContext, string serializedInput)
         {
+#if !FUNCTIONS_V1
```
ConnorMcMahon, line 151: In general, it is best to keep the conditional compilation to a minimum, so defining the variable up here and then having to conditional compilation every time we want to set it makes the code harder to read. I would almost prefer we just remove the conditional compilation around defining/setting this variable, and just suppress any warnings we get in Functions V1 for the variable never being referenced.

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -48,6 +49,10 @@ public override void RaiseEvent(OrchestrationContext unused, string eventName, s
 
         public override async Task<string> Execute(OrchestrationContext innerContext, string serializedInput)
         {
+#if !FUNCTIONS_V1
```
ConnorMcMahon, line 52: Similar comment here.

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
ConnorMcMahon: I just realized Chris said either option works for him. There is technically nothing wrong with this over suppressing the warning, but as mentioned, I prefer the suppressing the warning, as it has less impact on readability imo.

davidmrdavid: I think you're right, after trying the alternative, I think this is much cleaner. Fixing it in the next commit!

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
davidmrdavid: Fixed in upcoming commit!

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
davidmrdavid: I think you're right. Although I wonder if it would make sense for an Entity to ever be just "Running". What if its Locked?

davidmrdavid: I removed it for the time being, but I think the discussion is worthwhile

davidmrdavid: On that note, I'm now also unsure if my current approach event captures a "Running" Orchestrator

davidmrdavid:
This is how the changes reflect on App Insights:
AppInsightsPic

Query:
```
requests
| extend DurableFunctionsInstanceId = tostring(customDimensions["DurableFunctionsInstanceId"]),
         DurableFunctionsStatus = tostring(customDimensions["DurableFunctionsRuntimeStatus"]),
         DurableFunctionsType = tostring(customDimensions["DurableFunctionsType"])
| summarize arg_max(timestamp, *) by DurableFunctionsInstanceId
| order by timestamp desc
| limit 30
```

I don't know where that first empty row is coming from, but it shouldn't be coming from my changes as I have "Orchestrator" or "Entity" hard-coded for`DurableFunctionsType`

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
davidmrdavid: In addition, I'm unsure how to test this for failure states. It would be good if we had that discussion.

sebastianburckhardt: This appears to report a different status to AppInsights than what is the DTFx status of the orchestration (as visible in the status table)? 

Entities are "Running" while they wait for messages to arrive, which is most of the time. They should never be "Completed" or "Failed" which would indicate a serious bug. 

If we want to provide a different information, we should perhaps not use the same enumeration because I think that may be confusing?

davidmrdavid: Thanks for pointing this out! Indeed, that would be confusing.
For cross-referencing, what and where is the status table you're referencing?

ConnorMcMahon: Did we ever make a decision on this?

davidmrdavid: Not yet. We should discuss it early next week.

ConnorMcMahon:
Pushing this to our next release.

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -46,8 +47,29 @@ public override void RaiseEvent(OrchestrationContext unused, string eventName, s
             this.context.RaiseEvent(eventName, serializedEventData);
         }
 
+#if !FUNCTIONS_V1
+        public void ReportToAppInsights(OrchestrationRuntimeStatus status, string instanceId)
```
reyang, line 51: Minor feedback on the naming:
* [AppInsights](https://www.appdirect.com/products/appinsights) is a different product not owned by Microsoft. Suggest to use either `AzureMonitor` or `ApplicationInsights`.
* This function itself doesn't do any specific thing regarding ApplicationInsights, it simply tags the current activity. Is is possible to have a name that reflects to what the function is actually doing? (e.g. TagCurrentActivity)?

davidmrdavid: This is good feedback, thank you. I'll change it to `TagCurrentActivity` in my next commit

reyang: As we've discussed last Friday, this code would only update the current activity, instead of emit extra event to ApplicationInsights. The actual code that emits data is probably somewhere in the Azure Functions host. If the intention is to be able to emit multiple events while the process is running, the host code has to be changed.

davidmrdavid: Incorporated these changes in my newest commit

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
davidmrdavid: Ended up dropping updating the status for Entities, so they're always "running". I could also just completely avoid showing a status for entities, we should discuss.

davidmrdavid: @sebastianburckhardt , does that sound appropriate to you?

Issue #1281, opened 2020-03-19
Start orchestration cross function app


mwikstrom:
**Is your feature request related to a problem? Please describe.**
We want to separate orchestration logic and web api logic into different function apps.

In particular, we want http triggers in the api app to query, start and terminate orchestrations.

We assumed that this would be a possible setup if we ensured that both apps were configured with the same task hub name.

However, when attempting to start an orchestration from the api app an exception is thrown with the following message:

>The function 'BuildOrchestration' doesn't exist, is disabled, or is not an orchestrator function. Additional info: No orchestrator functions are currently registered!

There is however an orchestrator function named 'BuildOrchestration' in the other function app, and we believe that if this check would be ignored then things would work as we want since interaction between the two apps goes via the common task hub.

**Describe the solution you'd like**
I would like to have an option that skips the check whether a started orchestration is mapped to a function in the currrent app, or even better to be able to register "external orchestrator functions".

**Describe alternatives you've considered**
We're currently using the following hack:

```c#
/// <summary>
///     Starts an orchestration that is hosted by another function app, using the same task hub name.
/// </summary>
/// <remarks>
///     This is a reflection hack that register the specified orchestrator function name in the current app
///     just before calling `StartNewAsync(...)`.
///     
///     IMPORTANT! Do NOT use this method for functions that DO exist in the current app, doing so will
///     overwrite their registration causing undefined behavior.
/// </remarks>
public static Task StartNewExternalAsync<TInput>(
    this IDurableOrchestrationClient client,
    string orchestratorFunctionName,
    string instanceId,
    TInput input)
{
    Requires.ArgNotNull(client, nameof(client));
    Requires.ArgNotNull(orchestratorFunctionName, nameof(orchestratorFunctionName));
    Requires.ArgNotNull(instanceId, nameof(instanceId));

    var clientType = client.GetType();
    var configField = clientType.GetField("config", BindingFlags.NonPublic | BindingFlags.Instance);

    if (configField is null || configField.FieldType != typeof(DurableTaskExtension))
    {
        throw new NotSupportedException($"Cannot extract '{nameof(DurableTaskExtension)}' from '{clientType}'");
    }

    var config = (DurableTaskExtension)configField.GetValue(client)!;
            
    var registerMethod = typeof(DurableTaskExtension).GetMethod(
        "RegisterOrchestrator",
        BindingFlags.NonPublic | BindingFlags.Instance);

    if (registerMethod is null)
    {
[...]

PR #1306, opened 2020-04-07
Updated DurableClient ThrowIfFunctionDoesNotExist calls to work with correctly when calling a function in another app

### LINKED ISSUE #1248: Not able to Signal Entity Function using IDurableClient from different Function App

amdeel:
resolves #1248 
related to #1281 

Updated ThrowIfDoesNotExist check to only be called if the connectionName matches the connectionName in DurableTaskOptions.StorageProvider. This prevents an exception from being thrown when a user is signaling an entity or starting an orchestrator in another app.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -210,7 +210,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             if (string.IsNullOrEmpty(taskHubName))
             {
-                return this.SignalEntityAsyncInternal(this.client, this.TaskHubName, entityId, null, operationName, operationInput);
+                return this.SignalEntityAsyncInternal(this, this.TaskHubName, entityId, null, operationName, operationInput);
             }
             else
             {
@@ -225,7 +225,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     ConnectionName = connectionName,
                 };
 
-                TaskHubClient taskHubClient = ((DurableClient)this.config.GetClient(attribute)).client;
+                var taskHubClient = (DurableClient)this.config.GetClient(attribute);
                 return this.SignalEntityAsyncInternal(taskHubClient, taskHubName, entityId, null, operationName, operationInput);
             }
         }
@@ -235,7 +235,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             if (string.IsNullOrEmpty(taskHubName))
             {
-                return this.SignalEntityAsyncInternal(this.client, this.TaskHubName, entityId, scheduledTimeUtc, operationName, operationInput);
+                return this.SignalEntityAsyncInternal(this, this.TaskHubName, entityId, scheduledTimeUtc, operationName, operationInput);
             }
             else
             {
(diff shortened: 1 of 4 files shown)


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -297,6 +297,18 @@ private async Task SignalEntityAsyncInternal(TaskHubClient client, string hubNam
                 isReplay: false);
         }
 
+        private bool ConnectionNameMatchesCurrentApp(DurableClient client)
```
ConnorMcMahon, line 300: We can technically be even pickier, by verifying that this DurableClient is also targeting the same task hub that the current application is targeting.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -225,7 +225,7 @@ Task IDurableEntityClient.SignalEntityAsync(EntityId entityId, string operationN
                     ConnectionName = connectionName,
                 };
 
-                TaskHubClient taskHubClient = ((DurableClient)this.config.GetClient(attribute)).client;
+                var taskHubClient = (DurableClient)this.config.GetClient(attribute);
```
ConnorMcMahon, line 228: nit: this variable is now incorrectly named, as it is no longer a `TaskHubClient`, but a `DurableClient`

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -250,19 +250,19 @@ Task IDurableEntityClient.SignalEntityAsync(EntityId entityId, DateTime schedule
...
 
-                TaskHubClient taskHubClient = ((DurableClient)this.config.GetClient(attribute)).client;
-                return this.SignalEntityAsyncInternal(taskHubClient, taskHubName, entityId, scheduledTimeUtc, operationName, operationInput);
+                var durableClient = (DurableClient)this.config.GetClient(attribute);
+                return this.SignalEntityAsyncInternal(durableClient, taskHubName, entityId, scheduledTimeUtc, operationName, operationInput);
             }
         }
 
-        private async Task SignalEntityAsyncInternal(TaskHubClient client, string hubName, EntityId entityId, DateTime? scheduledTimeUtc, string operationName, object operationInput)
+        private async Task SignalEntityAsyncInternal(DurableClient clientAttribute, string hubName, EntityId entityId, DateTime? scheduledTimeUtc, string operationName, object operationInput)
```
ConnorMcMahon, line 258: nit: I would leave the name as `client` or rename it to `durableClient`.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
amdeel: Good point!

test/Common/DurableClientBaseTests.cs
```diff
@@ -57,7 +57,6 @@ public async Task StartNewAsync_is_calling_overload_method_with_specific_instanc
 
         [Theory]
         [Trait("Category", PlatformSpecificHelpers.TestCategory)]
-        [InlineData("")]
```
ConnorMcMahon, line 60: Is this no longer an invalid value?

amdeel: It is not, we assign a guid to instanceId when the string is null or empty. (DurableClient line 111)

Issue #1273, opened 2020-03-16
Local HTTP listener causing startup issues


ConnorMcMahon:
Tracking this information as an issue here from @olitomlinson. Quotes below from him:

I've got an open support case (120031225000535) because my DF app failed to start up for approx 30 minutes and then eventually corrected itself without intervention.

The error message during startup was :

`Failed to bind to address http://127.0.0.1:17071: address already in use.
Only one usage of each socket address (protocol/network address/port) is normally permitted`

My non-scientific googling has brought me to this GitHub issue. I don't really know if this work item impacts my particular issue, it probably doesn't, but are you in a position to shed any light? No worries if not, ill just keep following up with support to get to the root cause. Thanks in advance!

ConnorMcMahon:
@olitomlinson 

This is the port we are opening. We should be properly shutting it down, so it is concerning you are seeing this cause stalling on restarting your application.

That being said, I find it interesting that you would even have this code starting up. We only try to turn it on for non-.NET apps, as it is only used in our out-of-proc language SDKs (i.e. JavaScript and soon Python).

To mitigate in the meantime, can you make sure your application has `FUNCTIONS_WORKER_RUNTIME` set to `dotnet`? That should prevent us from starting up any listener on this port. The other option that should help mitigate is to `localRpcEndpointEnabled` to `false` in your host.json.

olitomlinson:
@ConnorMcMahon thanks for the swift reply!

Yes it appears that I don't have `FUNCTIONS_WORKER_RUNTIME` set in my configuration at all. I will apply it. 

Btw is this configuration mandatory? I guess I'm confused because it doesn't _feel_ mandatory, but it kinda actually is?

ConnorMcMahon:
I know it is something we flag as a warning in our internal tooling for investigating customer issues. That being said, I'm not sure if it is explicitly required, because we are often able to infer based on the code we end up seeing.

In general, I think our tooling now sets this automatically when you select your language, but I may be incorrect about that. Assuming your function application has been around for a long time, it could be from before we started setting that in the default app creation in our tools.

cgillum:
It just occurred to me that this listener behavior could be very problematic for apps that use slots because of how multiple instances may be running simultaneously on the same VM. We probably need to prioritize making the listener port selection dynamic like @anthonychu  suggested to avoid these kinds of problems.

olitomlinson:
[...]

PR #1307, opened 2020-04-07
Allow for dynamic port usage for local rpc endpoint if 17071 in use

ConnorMcMahon:
Our initial implementation of Local HTTP for out-of-proc languages always uses the port 17071. This can cause issues with slots, as well as caused some noisy telemetry when hosts would shut down occasionally.

We now check to see if the default port is available before attempting to bind to it. There is still some degree of race condition if the extensions are instantiated before the port is being listened to, but in general the OS should provide some protections against this.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
@@ -83,7 +83,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private int GetAvailablePort()
         {
-            // If we are able to successfully start a listener looking on the default port without 
+            // If we are able to successfully start a listener looking on the default port without
             // an exception, we can use the default port. Otherwise, let the TcpListener class decide for us.
             try
             {
--- a/src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
@@ -5,7 +5,9 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Net;
 using System.Net.Http;
+using System.Net.Sockets;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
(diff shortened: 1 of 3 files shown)


ConnorMcMahon:
I still want to do some end-to-end testing for this PR in Functions V2 and V3, and make sure it doesn't break anything in V1.

src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
```diff
@@ -18,20 +20,36 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
...
+                .Build();
+#else
+            // Just use default port for internal Uri. No need to check for port availability since
+            // we won't be listening to this endpoint.
+            this.InternalRpcUri = new Uri($"http://127.0.0.1:{DefaultPort}/durabletask/");
+            this.localWebHost = new NoOpWebHost();
+#endif
         }
 
+        public Uri InternalRpcUri { get; private set; }
```
cgillum, line 51: Looks like we can remove the setter completely since we only assign it in the constructor.

src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
```diff
@@ -63,27 +81,25 @@ public async Task StopAsync()
...
+                listener.Start();
+                listener.Stop();
+                return DefaultPort;
             }
-
-            if (listenUri.AbsolutePath.Length > 1)
+            catch (SocketException)
             {
-                throw new ArgumentException($"The listen URL must not contain a path.", nameof(listenUri));
+                var listener = new TcpListener(IPAddress.Loopback, 0);
```
cgillum, line 97: It might be worth adding a comment with this URL: https://stackoverflow.com/a/150974/2069 :)

ConnorMcMahon:
Tested in FunctionsV1 and FunctionsV3 to be sure neither broke (Functions V1 due to conditional compilation, FunctionsV3 to verify no synchronous calls in HTTP pipeline). 

Also tested end-to-end in Durable JS, and verified using RPC URI with a different port.

CastleArg:
is there any way of getting my hands on this fix?  I've had a durable function go down without warning and cannot start even after deleting and recreating the app because of this port in use error

cgillum:
Related: https://github.com/Azure/azure-functions-durable-extension/issues/1273

PR #1332, opened 2020-04-21
Analyzer: Fixed several FunctionAnalyzer bugs

### LINKED ISSUE #1320: AD0001 build warnings due to System.ArgumentException in FunctionAnalyzer
### Description
After a build of my .NETCore 3.1 Function App using DurableTask v2.2.1 with the DurableTask.Analyzers v0.2.1, I see warnings in the build output: `CSC : warning AD0001: Analyzer 'Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers.FunctionAnalyzer' threw an exception of type 'System.ArgumentException' with message 'Syntax node is not within syntax tree'.`

### Expected behavior
No build warnings, or specific (DFxxx) warnings related to the DurableTask Analyzer.

### Actual behavior
Copied from build output:

```
1>------ Build started: Project: DurableFunctions.Demo.DotNetCore, Configuration: Debug Any CPU ------
1>CSC : warning AD0001: Analyzer 'Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers.FunctionAnalyzer' threw an exception of type 'System.ArgumentException' with message 'Syntax node is not within syntax tree'.
1>CSC : warning AD0001: Analyzer 'Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers.FunctionAnalyzer' threw an exception of type 'System.ArgumentException' with message 'Syntax node is not within syntax tree'.
1>CSC : warning AD0001: Analyzer 'Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers.FunctionAnalyzer' threw an exception of type 'System.ArgumentException' with message 'Syntax node is not within syntax tree'.
1>DurableFunctions.Demo.DotNetCore -> C:\dev\git\personal\demos-azure-durable-functions\src\DurableFunctions.Demo.DotNetCore\bin\Debug\netcoreapp3.1\bin\DurableFunctions.Demo.DotNetCore.dll
1>Done building project "DurableFunctions.Demo.DotNetCore.csproj".
2>------ Build started: Project: DurableFunctions.Demo.DotNetCore.Test, Configuration: Debug Any CPU ------
2>DurableFunctions.Demo.DotNetCore.Test -> C:\dev\git\personal\demos-azure-durable-functions\src\DurableFunctions.Demo.DotNetCore.Test\bin\Debug\netcoreapp3.1\DurableFunctions.Demo.DotNetCore.Test.dll
========== Build: 2 succeeded, 0 failed, 0 up-to-date, 0 skipped ==========
```

### Known workarounds
I temporarily removed the dependency to the Analyzer by modifying the DurableTask extension nuspec:

- Go to the local folder where the DurableTask package is stored. For me this is: C:\Users\%User%\.nuget\packages\microsoft.azure.webjobs.extensions.durabletask\2.2.1
- Open the nuspec file and comment out (or remove) the two dependencies for Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers, one for `.NETFramework4.6.1`, and one for ` .NETStandard2.0`.
- Save the nuspec
- In the IDE, remove the package reference to the DurableTask extension.
- Add the DurableTask extension again. Since the package it's still in the local NuGet folder it should not download it from nuget.org. It will load the extension but not with the Analyzer now. 

### App Details

- **Durable Functions extension version (e.g. v2.1.1)**:
- **Durable Functions extension version (e.g. v0.2.1)**:

### LINKED ISSUE #1322: Analyzer: DF0110 JsonArray Compatible types bug
When comparing mismatch JsonArray compatible types and the method declaration returns Task<T>, the ReturnTypeAnalyzer will flag code incorrectly.

Ex:
var rows = await context.CallActivityAsync<IEnumerable<string>>("GetRowsFromBatch", (jobId, batchNumber));

[FunctionName("GetRowsFromBatch")]
        public static async Task<List<string>> GetRowsFromBatchActivity(
            [ActivityTrigger] IDurableActivityContext inputs,
            ILogger log)

### LINKED ISSUE #1324: Analyzer: DF0108 ValueTuple bug
When using ValueTuples in CallActivityAsync, the ArgumentAnalyzer compares the ValueTuple to Tuple, so the comparison triggers a warning.

Ex:
var (jobId, batchNumber, outputLogDirectory) = context.GetInput<(string, int, string)>();

await context.CallActivityAsync("ReportBatch", (jobId, batchNumber, outputLogDirectory));

[FunctionName("ReportBatch")]
        public static async Task ReportBatchActivity(
            [ActivityTrigger] IDurableActivityContext inputs,
            ILogger log)
        {
            var (jobId, batchNumber, outputLogDirectory) = inputs.GetInput<(string, int, string)>();
        }

amdeel:
Made several FunctionAnalyzer changes and bug fixes:
resolves #1320 -- 'System.ArgumentException' with message 'Syntax node is not within syntax tree'. due to SemanticModel being passed around without getting new SemanticModel for the node being analyzed

resolves #1322 -- Bug where ReturnTypeAnalyzer didn't properly handle function return types of Task<T> when T was a mismatch JsonArray serialized type (types that implement IEnumerable)

resolves #1324 -- Bug where ValueTuples would be incorrectly labeled as Tuples causing incorrect warnings when using ValueTuples

Adds logic to ArgumentAnalyzer to allow passed input to be a subclass or implementation of the defined input.
Adds logic to ReturnTypeAnalzyer to allow expected return type to be a base/interface for the defined return type.

Updated all FunctionAnalyzer tests to remove clutter and to clearly define test cases. This change will make adding to tests easier and, especially, reading and reviewing test code to be a much more enjoyable process.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/FunctionReturnTypeAnalyzer.cs
+++ b/src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/FunctionReturnTypeAnalyzer.cs
@@ -55,7 +55,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers
                 definitionReturnType = taskTypeArgument;
             }
 
-            return SyntaxNodeUtils.InputMatchesOrCompatibleType(invocationReturnType, definitionReturnType);
+            return SyntaxNodeUtils.InputMatchesOrCompatibleType(invocationReturnType, definitionReturnType)
+                || SyntaxNodeUtils.TypeNodeImplementsOrExtendsType(definitionReturnType, invocationReturnType.ToString());
         }
 
         private static bool TryGetTaskTypeArgument(ITypeSymbol returnType, out ITypeSymbol taskTypeArgument)
--- a/src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/FunctionReturnTypeAnalyzer.cs
+++ b/src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/FunctionReturnTypeAnalyzer.cs
@@ -3,6 +3,7 @@
 
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 
@@ -49,20 +50,23 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers
 
         private static bool InputMatchesOrTaskOrCompatibleType(ITypeSymbol invocationReturnType, ITypeSymbol definitionReturnType)
         {
-            return SyntaxNodeUtils.InputMatchesOrCompatibleType(invocationReturnType, definitionReturnType) || DefinitionReturnsTask(invocationReturnType, definitionReturnType);
-        }
-
-        private static bool DefinitionReturnsTask(ITypeSymbol invocationReturnType, ITypeSymbol definitionReturnType)
-        {
-            if (definitionReturnType.Name.Equals("Task") && definitionReturnType is INamedTypeSymbol namedType)
+            if (TryGetTaskTypeArgument(definitionReturnType, out ITypeSymbol taskTypeArgument))
             {
-                var taskTypeArgument = namedType.TypeArguments.FirstOrDefault();
-                if (taskTypeArgument != null)
-                {
-                    return invocationReturnType.Equals(taskTypeArgument);
-                }
+                definitionReturnType = taskTypeArgument;
             }
 
+            return SyntaxNodeUtils.InputMatchesOrCompatibleType(invocationReturnType, definitionReturnType);
+        }
+
+        private static bool TryGetTaskTypeArgument(ITypeSymbol returnType, out ITypeSymbol taskTypeArgument)
+        {
+            if (returnType.Name.Equals("Task") && returnType is INamedTypeSymbol namedType)
+            {
+                taskTypeArgument = namedType.TypeArguments.FirstOrDefault();
+                return taskTypeArgument != null;
+            }
+
+            taskTypeArgument = null;
             return false;
         }
 
--- a/src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/FunctionReturnTypeAnalyzer.cs
+++ b/src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/FunctionReturnTypeAnalyzer.cs
@@ -35,7 +35,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers
                     TryGetInvocationReturnType(semanticModel, activityInvocation, out ITypeSymbol invocationReturnType);
                     TryGetDefinitionReturnType(semanticModel, functionDefinition, out ITypeSymbol definitionReturnType);
 
-                    if (!InputMatchesOrTaskOrCompatibleType(invocationReturnType, definitionReturnType))
+                    if (!IsValidReturnTypeForDefinition(invocationReturnType, definitionReturnType))
                     {
                         var invocationTypeName = SyntaxNodeUtils.GetQualifiedTypeName(invocationReturnType);
                         var functionTypeName = SyntaxNodeUtils.GetQualifiedTypeName(definitionReturnType);
@@ -48,7 +48,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers
             }
         }
 
-        private static bool InputMatchesOrTaskOrCompatibleType(ITypeSymbol invocationReturnType, ITypeSymbol definitionReturnType)
+        private static bool IsValidReturnTypeForDefinition(ITypeSymbol invocationReturnType, ITypeSymbol definitionReturnType)
         {
             if (TryGetTaskTypeArgument(definitionReturnType, out ITypeSymbol taskTypeArgument))
             {
--- a/src/WebJobs.Extensions.DurableTask.Analyzers/SyntaxNodeUtils.cs
+++ b/src/WebJobs.Extensions.DurableTask.Analyzers/SyntaxNodeUtils.cs
@@ -343,8 +343,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers
             }
 
             return invocationType.Equals(definitionType)
-                || AreEqualTupleTypes(invocationType, definitionType)
                 || AreCompatibleIEnumerableTypes(invocationType, definitionType);
+                //|| AreEqualTupleTypes(invocationType, definitionType);
         }
 
         private static bool AreEqualTupleTypes(ITypeSymbol invocationType, ITypeSymbol definitionType)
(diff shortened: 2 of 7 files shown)


src/WebJobs.Extensions.DurableTask.Analyzers/SyntaxNodeUtils.cs
```diff
@@ -312,7 +313,7 @@ internal static string GetQualifiedTypeName(ITypeSymbol typeInfo)
                     var tupleUnderlyingType = namedTypeInfo.TupleUnderlyingType;
                     if (tupleUnderlyingType != null)
                     {
-                        return $"System.Tuple<{string.Join(", ", tupleUnderlyingType.TypeArguments.Select(x => x.ToString()))}>";
+                        return $"({string.Join(", ", tupleUnderlyingType.TypeArguments.Select(x => x.ToString()))})";
```
ConnorMcMahon, line 316: Is the goal of this to return a type name tuples that is valid for both `Tuple` and `ValueTuple`?

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/FunctionReturnTypeAnalyzer.cs
```diff
@@ -49,20 +50,23 @@ public class FunctionReturnTypeAnalyzer
 
         private static bool InputMatchesOrTaskOrCompatibleType(ITypeSymbol invocationReturnType, ITypeSymbol definitionReturnType)
```
ConnorMcMahon, line 51: This name is getting a bit hard to parse.

I would consider something like `IsReturnTypeValidForDefinition()`, and have a brief comment covering all of the cases (i.e. direct type match, compatible type match, or Task containing the matching type).

test/WebJobs.Extensions.DurableTask.Analyzers.Test/ActivityFunction/FunctionReturnTypeAnalyzerTests.cs
```diff
@@ -100,14 +104,20 @@ public static Object SayHello([ActivityTrigger] IDurableActivityContext context)
...
+            return tuple;
+        }
+
         [FunctionName(""E1_SayHello_ReturnsString"")]
         public static string SayHelloDirectInput([ActivityTrigger] string name)
         {
             return $""Hello {name}!"";
         }
 
         [FunctionName(""E1_SayHello_ArrayAndNamedType"")]
```
ConnorMcMahon, line 119: Not sure I understand what `ArrayAndNamedType` means exactly. I was able to look up that `INamedTypeSymbol` is any type other than "an array, a pointer, a type parameter", but that isn't immediately obvious what this function is trying to accomplish. Is the idea that ValueTuple is a NamedType, and List<string>() is an array type? If that is the case, then I would make the name more descriptive, like "E1_SayHello_InputNamedType_ReturnsArray" or something along those lines.

test/WebJobs.Extensions.DurableTask.Analyzers.Test/ActivityFunction/FunctionReturnTypeAnalyzerTests.cs
```diff
@@ -100,14 +104,20 @@ public static Object SayHello([ActivityTrigger] IDurableActivityContext context)
...
-        public static List<string> SayHelloTuple([ActivityTrigger] Tuple<string, int> tuple)
+        public static List<string> SayHelloTuple([ActivityTrigger] ValueTuple<string, int> tuple)
```
ConnorMcMahon, line 120: small nit: It may be worthwhile naming the method name of our functions what their function names currently are. Then we could use `nameof()` to reference it in more locations and make name changes easier to manage.

src/WebJobs.Extensions.DurableTask.Analyzers/SyntaxNodeUtils.cs
amdeel: Actually no. This fixes a bug with Tuples/ValueTuples and also covers a new test case that came up. Tuples (and usually ValueTuples) will actually bypass this check and go to the line
`return typeInfo.ToString();`

Previously I was treating Tuples and ValueTuples as the same thing. I realized my mistake and this change fixes a case where the ITypeSymbol for a ValueTuple will be something like "(string jobId, int batchNumber)" instead of "(string, int)"

This happens in this case (this example omits the function definition, you can assume it takes a (string, int) ValueTuple):
```
 var (jobId, batchNumber) = context.GetInput<(string, int)>();

var rows = await context.CallActivityAsync<IEnumerable<(string, string)>>("GetRowsFromBatch", (jobId, batchNumber));
```

GetQualifiedTypeName is basically a ToStringInAStandardWay method. It is usually not used in the step to compare types and is usually used only when creating the message to display to the user. But in this case, I actually need to get a "different version" (TupleUnderlyingType) of the INamedTypeSymbol to be able to compare them. This particular ValueTuple situation is the only one that I ran into where I needed to use this.

Previously I didn't understand Tuples and thought that Tuples and (string, int) (ValueTuples) were the same thing. They are not and I was incorrectly comparing them.

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/FunctionReturnTypeAnalyzer.cs
amdeel: Very good point

test/WebJobs.Extensions.DurableTask.Analyzers.Test/ActivityFunction/FunctionReturnTypeAnalyzerTests.cs
amdeel: For the types that are serialized to JsonArrays (implement IEnumerable) there are IArrayTypeSymbols and INamedTypeSymbols. This test is comparing the two here. I agree, it would be a good idea to rename it.

test/WebJobs.Extensions.DurableTask.Analyzers.Test/ActivityFunction/FunctionReturnTypeAnalyzerTests.cs
amdeel: Unfortunately, I can't use nameof() here to help manage test code changes because this is just one big string representing a document to be scanned by my analyzer. I can go through my test code and make sure the method names match the function names, though, just to clean it up a bit.

ConnorMcMahon: I didn't realize from the GitHub page that this was just a raw string. This seems difficult to maintain.

I'm curious if we could make this code as a SampleFunctions.cs, and then have the test read the contents of that .cs file as a string. This would allow you to make use of intellisense and other important dev features for maintaining these samples, while still working within your test framework.

amdeel: Good suggestion. However, for every test case, there would have to be a cs file. I don't think it would improve much. However, I do think that these tests are very hard to read/understand and hard to add to and review, so I decided to make some changes to them. It will be much easier to identify the test cases now. I made these changes only for the function analyzer tests, for now, keeping the PR on theme with function analyzers. In the future, I will make changes to the rest of the tests to match.

PR #1336, opened 2020-04-24
Handle constructor failures for non-static functions

sebastianburckhardt:
Fixes a bug where constructor failures for non-static orchestrators or entities lead to incorrect behavior, and added testcases.

The desired behavior is to treat constructor exceptions as if they were thrown by the function call.

For entities this means that general entity behaviors (e.g. locking) keep working  even if the function cannot be constructed.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Listener/FunctionExecutionHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/FunctionExecutionHelper.cs
@@ -13,6 +13,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Listener
         public static async Task<WrappedFunctionResult> ExecuteFunctionInOrchestrationMiddleware(
             ITriggeredFunctionExecutor executor,
             TriggeredFunctionData triggerInput,
+            DurableCommonContext context,
             CancellationToken cancellationToken)
         {
 #pragma warning disable CS0618 // InvokeHandler approved for use by this extension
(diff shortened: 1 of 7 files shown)


src/WebJobs.Extensions.DurableTask/Listener/FunctionExecutionHelper.cs
```diff
@@ -24,42 +25,43 @@ internal static class FunctionExecutionHelper
...
-                    {
-                        executedUserCode = true;
-                        return triggerInput.InvokeHandler(userCodeHandler);
-                    },
-#if !FUNCTIONS_V1
-                    TriggerDetails = triggerInput.TriggerDetails,
-#endif
-                };
-#pragma warning restore CS0618
+                context.ExecutorCalledBack = false;
```
ConnorMcMahon, line 28: I'm confused as to what `context.ExecutorCalledBack` is used for, as it appears to be the exact same valuethat the local variable `executedUserCode` was measuring before. They both are false until they are set to true immediately before the rest of the invoke handler is called.

From what I can tell, embedding this boolean on the DurableCommonContext just requires more work on the caller of this method.

src/WebJobs.Extensions.DurableTask/Listener/FunctionExecutionHelper.cs
```diff
@@ -24,42 +25,43 @@ internal static class FunctionExecutionHelper
...
+                    // We want to treat this case exactly as if the function itself is throwing the exception.
+                    // So we execute the middleware directly, instead of via the executor.
+                    try
+                    {
+                        var exception = result.Exception ?? new InvalidOperationException("The function failed to start executing.");
 
-                return WrappedFunctionResult.UserCodeFailure(result.Exception);
+                        await triggerInput.InvokeHandler(() => Task.FromException<object>(exception));
+
+                        return WrappedFunctionResult.Success();
```
ConnorMcMahon, line 57: So the idea is that some entity functionality will not actually throw an exception even when we manually invoke the handled with a thrown exception. In those cases, we want to treat the operation as a success?

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -2014,6 +2014,59 @@ public async Task ThrowExceptionOnLongTimer(bool extendedSessions, string storag
...
+        }
+
+        [Theory]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        [MemberData(nameof(TestDataGenerator.GetBooleanAndFullFeaturedStorageProviderOptions), MemberType = typeof(TestDataGenerator))]
+        public async Task HandleUncallableFunctions(bool extendedSessions, string storageProvider)
+        {
+            using (ITestHost host = TestHelpers.GetJobHost(
+                this.loggerProvider,
+                nameof(this.HandleUncallableOrchestrator),
```
ConnorMcMahon, line 2052: This should be HandleUncallableFunctions, otherwise you will end up with taskhub collision between the two tests.

test/Common/TestOrchestrations.cs
```diff
@@ -953,6 +953,50 @@ public static async Task<string> RollbackSignalsOnExceptions([OrchestrationTrigg
...
+            {
+            }
+
+            // try an uncallable activity
+            try
+            {
+                var result = await ctx.CallActivityAsync<int>(nameof(UnconstructibleClass.UncallableActivity), null);
+            }
+#if FUNCTIONS_V1
+                catch (FunctionFailedException e) when (e.Message == "The activity function 'UncallableActivity' failed: \"Exception has been thrown by the target of an invocation.\". See the function execution logs for additional details.")
```
ConnorMcMahon, line 990: be careful with indentation in your conditional compilation. Make sure that you aren't getting stylecop warnings in either compilation mode.

src/WebJobs.Extensions.DurableTask/Listener/FunctionExecutionHelper.cs
sebastianburckhardt: Right. For entities there is not a one-to-one correspondence between calling the middleware executor (which corresponds to delivering a message to the entity scheduler) and the function execution (which corresponds to executing an operation on the entity). Some messages do not cause an entity operation execution, and some messages can cause more than one entity operation execution. And yes, the entity operation exceptions are caught by the entity scheduler.

test/Common/DurableTaskEndToEndTests.cs
sebastianburckhardt: Makes sense. I did not actually pay attention to the meaning of this parameter.

src/WebJobs.Extensions.DurableTask/Listener/FunctionExecutionHelper.cs
sebastianburckhardt: I think my intention was to save an object allocation... instead of creating an extra indirection through a `TriggereredFunctionData` wrapper around the userCodeHandler, I just use a boolean flag on the already existing DurableCommonContext object to check if the executor called back.

The previous version had a bug also, I think, since it created this `TriggeredFunctionData` wrapper but then doesn't pass it on as a parameter to the `executor.TryExecuteAsync`.

src/WebJobs.Extensions.DurableTask/Listener/FunctionExecutionHelper.cs
sebastianburckhardt: An interesting effect of this is that you can use an undeclared entity class as a lock.

src/WebJobs.Extensions.DurableTask/Listener/FunctionExecutionHelper.cs
ConnorMcMahon: Good catch on that bug. I wonder what doofus wrote the original code here...

I definitely understand your point about removing an object allocation. My concern is that it then puts a burden on the `InvokeHandler` implementations in the respective middleware to set `context.ExecutorCalledBack` to `true`, which would be easy to forget.

That being said, I doubt we will be touching the existing InvokeHandlers too much, and I don't anticipate us adding any new orchestration middleware in the near future, so I could go either way on the issue.

PR #1339, opened 2020-04-26
Check for interface methods after we've found them

withinboredom:
It's possible that an interface might only implement another interface, consider for example:

```csharp
public interface IPhoneToUser : ISignalEntityIndex<UserId> {}
```

where `ISignalEntityIndex<UserId>` contains methods to signal. There may be no need for `IPhoneToUser` to expose methods itself.

This PR checks for methods after we've discovered all the methods on the interface and will throw if there are no methods.


ConnorMcMahon:
Thanks for the contribution @withinboredom!

PR #1349, opened 2020-05-12
DurableClient changes to allow for use of APIs across function apps

### LINKED ISSUE #1281: Start orchestration cross function app

amdeel:
Allows all orchestrator and entity operations to work across function apps.

These changes include significant IDurableOrchestrationClient and IDurableEntityClient changes, adding new APIs that include a DurableConnectionDetails optional parameter and marking the old APIs as Obsolete. The old APIs have been retained to avoid breaking changes.

resolves #1281

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityClient.cs
@@ -26,9 +26,31 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// <param name="entityId">The target entity.</param>
         /// <param name="operationName">The name of the operation.</param>
         /// <param name="operationInput">The input for the operation.</param>
+        /// <param name="connectionDetails">The storage account connection details.</param>
+        /// <returns>A task that completes when the message has been reliably enqueued.</returns>
+        Task SignalEntityAsync(EntityId entityId, string operationName, object operationInput = null, DurableClientConnectionDetails connectionDetails = null);
+
+        /// <summary>
+        /// Signals an entity to perform an operation, at a specified time.
+        /// </summary>
+        /// <param name="entityId">The target entity.</param>
+        /// <param name="scheduledTimeUtc">The time at which to start the operation.</param>
+        /// <param name="operationName">The name of the operation.</param>
+        /// <param name="operationInput">The input for the operation.</param>
+        /// <param name="connectionDetails">The storage account connection details.</param>
+        /// <returns>A task that completes when the message has been reliably enqueued.</returns>
+        Task SignalEntityAsync(EntityId entityId, DateTime scheduledTimeUtc, string operationName, object operationInput = null, DurableClientConnectionDetails connectionDetails = null);
+
+        /// <summary>
+        /// Signals an entity to perform an operation.
+        /// </summary>
+        /// <param name="entityId">The target entity.</param>
+        /// <param name="operationName">The name of the operation.</param>
+        /// <param name="operationInput">The input for the operation.</param>
         /// <param name="taskHubName">The TaskHubName of the target entity.</param>
         /// <param name="connectionName">The name of the connection string associated with <paramref name="taskHubName"/>.</param>
         /// <returns>A task that completes when the message has been reliably enqueued.</returns>
+        [Obsolete]
         Task SignalEntityAsync(EntityId entityId, string operationName, object operationInput = null, string taskHubName = null, string connectionName = null);
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityClient.cs
```diff
@@ -59,7 +59,17 @@ public interface IDurableEntityClient
         /// </summary>
         /// <param name="query">Return entity instances that match the specified query conditions.</param>
         /// <param name="cancellationToken">Cancellation token that can be used to cancel the query operation.</param>
+        /// <param name="connectionDetails">The storage account connection details.</param>
         /// <returns>Returns a page of entity instances and a continuation token for fetching the next page.</returns>
+        Task<EntityQueryResult> ListEntitiesAsync(EntityQuery query, CancellationToken cancellationToken, DurableConnectionDetails connectionDetails = null);
```
amdeel, line 64: I avoided/removed my changes to IDurableEntityClient that added new methods with DurableConnectionDetails as an optional parameter for the other methods in this interface to avoid errors that a user's method call was ambiguous. Adding these new methods would be a breaking change that we can put off for now.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
```diff
@@ -116,12 +116,123 @@ public interface IDurableOrchestrationClient
...
+        /// <param name="orchestratorFunctionName">The name of the orchestrator function to start.</param>
+        /// <param name="instanceId">The ID to use for the new orchestration instance.</param>
+        /// <param name="input">JSON-serializable input value for the orchestrator function.</param>
+        /// <typeparam name="T">The type of the input value for the orchestrator function.</typeparam>
+        /// <returns>A task that completes when the orchestration is started. The task contains the instance id of the started
+        /// orchestratation instance.</returns>
+        /// <exception cref="ArgumentException">
+        /// The specified function does not exist, is disabled, or is not an orchestrator function.
+        /// </exception>
+        [Obsolete]
```
amdeel, line 235: I kept all of these old methods around to avoid breaking mocks

src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
```diff
@@ -398,5 +398,15 @@ public virtual bool ValidateDelayTime(TimeSpan timespan, out string errorMessage
...
         }
+
+        /// <summary>
+        ///  Returns true if the stored connection string, ConnectionName, matches the input DurabilityProvider ConnectionName.
+        /// </summary>
+        /// <param name="durabilityProvider">The DurabilityProvider used to check for matching connection string names.</param>
+        /// <returns>A boolean indicating whether the connection names match.</returns>
+        public virtual bool ConnectionNameMatches(DurabilityProvider durabilityProvider)
+        {
+            return this.ConnectionName.Equals(durabilityProvider.ConnectionName);
```
amdeel, line 409: This change makes the logic for comparing connection strings up to the DurabilityProvider. I believe this change should be sufficient for testing locally.

cgillum:
I'm a little uncomfortable with this design. Adding a new overload to all our client APIs is problematic for several reasons:

- It's a breaking change from an interface perspective (it will break users' test code).
- All users who upgrade to this version will suddenly get warnings in their client functions because of the `[Obsolete]` attributes on the existing methods.
- This model would need to be replicated to all of our language SDKs, and feature drift across languages is already becoming hard to manage.
- It's more public API surface area that we have to preserve and document.

Why not just use the existing attribute properties like we've been doing already? That seems to be sufficient for #1281 and would work automatically for all languages without breaking changes. That's also more consistent with how existing bindings work across Functions. Let me know if I'm missing any context.

PR #1359, opened 2020-05-20
Handle out-of-process exceptions properly

ConnorMcMahon:
In order to prevent non-determinism errors when out-of-proc
orchestrations throw an exception, we need to read state from the
RpcException. We can then properly replay the actions performed by the
out-of-proc orchestration before throwing the exception.

See https://github.com/Azure/azure-functions-durable-extension/pull/1171 for the equivalent PR for Durable 1.x.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
@@ -3,18 +3,21 @@
 
 using System;
 using System.Collections.Generic;
+using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
+using DurableTask.Core.Exceptions;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Converters;
 using Newtonsoft.Json.Linq;
+using static Microsoft.Azure.WebJobs.Extensions.DurableTask.TaskOrchestrationShim;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 {
     /// <summary>
     /// Not intended for public consumption.
     /// </summary>
-    public class OutOfProcOrchestrationShim
+    internal class OutOfProcOrchestrationShim
     {
         private readonly IDurableOrchestrationContext context;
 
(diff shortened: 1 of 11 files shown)


src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -6,7 +6,7 @@
     <RootNamespace>Microsoft.Azure.WebJobs.Extensions.DurableTask</RootNamespace>
     <MajorVersion>2</MajorVersion>
     <MinorVersion>2</MinorVersion>
-    <PatchVersion>1</PatchVersion>
+    <PatchVersion>2</PatchVersion>
```
davidmrdavid, line 9: was this pushed as an accident or are we increasing the version w/ this PR too?

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -42,9 +44,9 @@ private enum AsyncActionType
             ScheduledSignalEntity = 10,
         }
 
-        internal async Task HandleOutOfProcExecutionAsync(JObject executionJson)
+        public async Task HandleDurableTaskReplay(OrchestrationInvocationResult executionJson)
```
davidmrdavid, line 47: a small summary comment here would be greatly appreciated

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -101,15 +101,65 @@ public override async Task<string> Execute(OrchestrationContext innerContext, st
                     this.context.IsReplaying));
             }
 
-            object returnValue;
+            await this.InvokeFunctionAsync(orchestratorInfo, innerContext);
```
davidmrdavid, line 104: is this achieving the replay or what is this call for?

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -101,15 +101,65 @@ public override async Task<string> Execute(OrchestrationContext innerContext, st
...
+                    this.context.ContinuedAsNew,
+                    this.context.IsReplaying));
+            }
+#if !FUNCTIONS_V1
+            DurableTaskExtension.TagActivityWithOrchestrationStatus(status, this.context.InstanceId);
+#endif
+            return serializedOutput;
+        }
+
+        private async Task InvokeFunctionAsync(
```
davidmrdavid, line 137: a one-line summary comment here would be appreciated as well

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
ConnorMcMahon: accidental.

PR #1368, opened 2020-05-28
Remove more duplicate task hub names in tests

ConnorMcMahon:
Another round of removing duplicate task hub names in our tests to make
our CI more reliable.


ConnorMcMahon:
I think the issue is more on us. They probably just ramped up parallelism, and some of our tests just don't play friendly in parallel. 

I am thinking we may need to take a slightly different approach to our task hub generation. Maybe appending some random 3-4 chars at the end to prevent conflicts?

PR #1385, opened 2020-06-12
Make E2E tests more robust in CI pipeline

ConnorMcMahon:
Our end-to-end tests currently rely on having unique taskhub names so
that tests can be run safely in parallel. This has been proven to be
hard to catch/maintain this restriction, so we add a random 4
alphanumeric suffix to each task hub to make sure that there are never
duplicate taskhub names.

### DIFF
--- a/test/Common/TestHelpers.cs
+++ b/test/Common/TestHelpers.cs
@@ -113,27 +113,27 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
             // Delete blobs
             CloudBlobClient blobClient = storageAccount.CreateCloudBlobClient();
             CloudBlobContainer leases = blobClient.GetContainerReference($"{taskHub}-leases");
-            deletionTasks.Add(leases.DeleteAsync());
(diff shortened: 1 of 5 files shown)


ConnorMcMahon:
This does have a couple of potential downsides.

1. It makes it a bit harder to find the taskhub name in the storage account when a test fails.
2. We will no longer reuse taskhubs across test runs. This is a double edged sword, because it means  we can have multiple CI runs happen in parallel, but it will also cause the number of taskhubs on our storage account to increase every time we have a CI run. The downside could be mitigated by periodically cleaning out the storage account.

test/Common/TestHelpers.cs
```diff
@@ -89,9 +89,15 @@ internal static class TestHelpers
                 testNameResolver);
         }
 
+        // Create a valid task hub from the test name, and add a random g
```
davidmrdavid, line 92: I think this comment might be incomplete / cut-off?

ConnorMcMahon: Haha probably was debating between leaving the comment or not and my brain just gave up :)

PR #1386, opened 2020-06-16
Add TaskEventId to function traces

### LINKED ISSUE #1382: Include Task Event ID in ETW and App Insights traces
When looking at orchestration traces, it's not possible to distinguish one activity function call from another if they have the same name. This creates various challenges when doing end-to-end tracing.

In the Durable Task Framework, each activity and sub-orchestration has an auto-incremented "event ID" that can be used to differentiate these tasks. These IDs should be included in the logs. Note that we already include this information in our internal DurableTask-AzureStorage ETW events, but those are not visible in the high-level Durable Functions traces.

We could expose the DurableTask-AzureStorage ETW traces (which have this information), but those are very low level and high-volume, making them harder for ordinary users to understand.

cgillum:
Resolves #1382.

With this PR, both our ETW traces and the Application Insight traces will have a new `TaskEventId` column for activity function executions. The `TaskEventId` integer value can be used to distinguish two different activity function executions that have the same name. For example, in the `HelloSequence` example, you'd see that each `SayHello` execution has a different `TaskEventId` value (0, 1, and 2).

For orchestrator and entity functions, this value will always be -1 since there isn't a way to associate a single event to a single orchestration function execution (there could be multiple events corresponding to a single execution).

We already have this data in the **DurableTask-AzureStorage** traces. This PR adds the same data to the **WebJobs-Extensions-DurableTask** traces (and to AppInsights).

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -291,6 +291,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 .BindToTrigger(new EntityTriggerAttributeBindingProvider(this, context, storageConnectionString, this.TraceHelper));
 
             this.taskHubWorker = new TaskHubWorker(this.defaultDurabilityProvider, this, this);
+            this.taskHubWorker.AddActivityDispatcherMiddleware(this.ActivityMiddleware);
             this.taskHubWorker.AddOrchestrationDispatcherMiddleware(this.EntityMiddleware);
             this.taskHubWorker.AddOrchestrationDispatcherMiddleware(this.OrchestrationMiddleware);
 
@@ -463,6 +464,17 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             }
 
             return new TaskActivityShim(this, info.Executor, this.hostLifetimeService, name);
+        }
+
+        private Task ActivityMiddleware(DispatchMiddlewareContext dispatchContext, Func<Task> next)
+        {
+            if (dispatchContext.GetProperty<TaskActivity>() is TaskActivityShim shim)
+            {
+                TaskScheduledEvent @event = dispatchContext.GetProperty<TaskScheduledEvent>();
+                shim.SetTaskEventId(@event?.EventId ?? -1);
+            }
+
+            return next();
         }
 
         private async Task OrchestrationMiddleware(DispatchMiddlewareContext dispatchContext, Func<Task> next)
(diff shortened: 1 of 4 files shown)


src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -465,6 +466,17 @@ TaskActivity INameVersionObjectManager<TaskActivity>.GetObject(string name, stri
             return new TaskActivityShim(this, info.Executor, this.hostLifetimeService, name);
         }
 
+        private Task ActivityMiddleware(DispatchMiddlewareContext dispatchContext, Func<Task> next)
```
davidmrdavid, line 469: Nitpick:
I find the `Middleware`-suffixed method names to be a bit mysterious, in part because they're not given much documentation. I would really appreciate a header comment here to that can clarify
(1) how to think of the dispatchContext 
(2) how the function works, especially around its `return next()`  line (line 477)

While I realize this is not strictly related to the feature, the header comment would help illuminate other features that re-use the "AddMiddleware" pattern.

src/WebJobs.Extensions.DurableTask/Listener/TaskActivityShim.cs
```diff
@@ -122,6 +127,11 @@ public override string Run(TaskContext context, string input)
             throw new NotImplementedException();
         }
 
+        internal void SetTaskEventId(int taskEventId)
```
davidmrdavid, line 130: When I first encountered `taskEventId`, I thought it referred to ETW / logger Event IDs N ow it's clear that it serves as an ID to differentiate between activity executions.

A short IntelliSense-friendly comment on the `taskEventId` instance variable could help me, and future newcomers, avoid this confusion :)

davidmrdavid:
It mostly looks good to me, but I think there's an opportunity here to document some of the functionality that this PR interacts with. Please see my nitpicky comments above, and let me know if we can include those clarifications 😄

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
cgillum: Good idea. I'm probably the only one who understands this stuff so more documentation would be helpful. :)

cgillum:
Thanks @davidmrdavid for the feedback! I've added more comments in my PR which will hopefully make things a bit less mysterious. This is a good reminder for me to get back into that habit Let me know if the updated comments are helpful.

I didn't go into too much detail on the mechanics of the middleware, though, because that would be pretty involved. However, it is modeled after the [ASP.NET Core middleware design](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/?view=aspnetcore-3.1) so I'm hoping that folks that are already familiar with ASP.NET Core will recognize the pattern I'm using with the context parameter and the `next` callback.

PR #1390, opened 2020-06-23
Long running durable timers

### LINKED ISSUE #14: Allow timers to last longer than 7 days

bachuv:
These changes allow durable timers to run longer than 7 days and resolves #14 . 

Durable timer messages use queue messages which don't last longer than 7 days. To get around this, if a timer is scheduled for longer than 7 days, then the timer will go off every 3 days until the difference between the current time and the scheduled time is less than 3 days.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
@@ -54,6 +54,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         public override JObject ConfigurationJson => this.storageOptionsJson;
 
+        public override TimeSpan MaximumDelayTime => MaxTimerDuration;
+
         /// <inheritdoc/>
         public async override Task<IList<OrchestrationState>> GetAllOrchestrationStates(CancellationToken cancellationToken)
         {
@@ -139,7 +141,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             if (timespan > MaxTimerDuration)
             {
-                errorMessage = null;
+                errorMessage = $"The Azure Storage provider supports a maximum of {MaxTimerDuration.TotalDays} days for time-based delays";
                 return false;
             }
 
--- a/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
@@ -22,8 +22,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
     /// </summary>
     internal class AzureStorageDurabilityProvider : DurabilityProvider
     {
-        private static readonly TimeSpan MaxTimerDuration = TimeSpan.FromDays(6);
-
         private readonly AzureStorageOrchestrationService serviceClient;
         private readonly string connectionName;
         private readonly JObject storageOptionsJson;
@@ -54,7 +52,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         public override JObject ConfigurationJson => this.storageOptionsJson;
 
-        public override TimeSpan MaximumDelayTime => MaxTimerDuration;
+        public override TimeSpan MaximumDelayTime { get; set; } = TimeSpan.FromDays(6);
+
+        public override TimeSpan LongRunningTimerIntervalLength { get; set; } = TimeSpan.FromDays(3);
 
         /// <inheritdoc/>
         public async override Task<IList<OrchestrationState>> GetAllOrchestrationStates(CancellationToken cancellationToken)
@@ -139,9 +139,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         public override bool ValidateDelayTime(TimeSpan timespan, out string errorMessage)
         {
-            if (timespan > MaxTimerDuration)
+            if (timespan > this.MaximumDelayTime)
             {
-                errorMessage = $"The Azure Storage provider supports a maximum of {MaxTimerDuration.TotalDays} days for time-based delays";
+                errorMessage = $"The Azure Storage provider supports a maximum of {this.MaximumDelayTime.TotalDays} days for time-based delays";
                 return false;
             }
 
--- a/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
@@ -139,7 +139,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             if (timespan > MaxTimerDuration)
             {
-                errorMessage = $"The Azure Storage provider supports a maximum of {MaxTimerDuration.TotalDays} days for time-based delays";
+                errorMessage = null;
                 return false;
             }
 
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -362,16 +362,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             return result;
         }
 
-        public string ValidOutOfProcTimer(DateTime fireAt)
+        public bool ValidOutOfProcTimer(DateTime fireAt, out string errorMessage)
         {
             this.ThrowIfInvalidAccess();
 
-            if (!this.durabilityProvider.ValidateDelayTime(fireAt.Subtract(this.InnerContext.CurrentUtcDateTime), out string errorMessage))
+            if (!this.durabilityProvider.ValidateDelayTime(fireAt.Subtract(this.InnerContext.CurrentUtcDateTime), out errorMessage))
             {
-                return errorMessage;
+                return false;
             }
 
-            return string.Empty;
+            return true;
         }
 
         /// <inheritdoc />
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -85,7 +85,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         internal bool IsCompleted { get; set; }
 
-        internal bool IsLongRunningTimer { get; set; }
+        internal bool IsLongRunningTimer { get; private set; }
 
         internal ExceptionDispatchInfo OrchestrationException { get; set; }
 
@@ -320,11 +320,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             }
 
             this.Config.TraceHelper.FunctionListening(
-                    this.Config.Options.HubName,
-                    this.OrchestrationName,
-                    this.InstanceId,
-                    reason: $"CreateTimer:{fireAt:o}",
-                    isReplay: this.InnerContext.IsReplaying);
+                this.Config.Options.HubName,
+                this.OrchestrationName,
+                this.InstanceId,
+                reason: $"CreateTimer:{fireAt:o}",
+                isReplay: this.InnerContext.IsReplaying);
 
             T result = default;
 
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -85,6 +85,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         internal bool IsCompleted { get; set; }
 
+        internal bool IsLongRunningTimer { get; set; }
+
         internal ExceptionDispatchInfo OrchestrationException { get; set; }
 
         internal bool IsOutputSet => this.serializedOutput != null;
@@ -311,10 +313,18 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             DateTime intervalFireAt = fireAt;
             TimeSpan longRunningTimerIntervalLength = TimeSpan.FromDays(3);
 
-            if (!this.durabilityProvider.ValidateDelayTime(fireAt.Subtract(this.InnerContext.CurrentUtcDateTime), out string errorMessage))
+            if (fireAt.Subtract(this.InnerContext.CurrentUtcDateTime) > this.durabilityProvider.MaximumDelayTime)
             {
-                intervalFireAt = this.InnerContext.CurrentUtcDateTime.AddDays(longRunningTimerIntervalLength.TotalDays);
+                this.IsLongRunningTimer = true;
+                intervalFireAt = this.InnerContext.CurrentUtcDateTime.Add(longRunningTimerIntervalLength);
             }
+
+            this.Config.TraceHelper.FunctionListening(
+                    this.Config.Options.HubName,
+                    this.OrchestrationName,
+                    this.InstanceId,
+                    reason: $"CreateTimer:{fireAt:o}",
+                    isReplay: this.InnerContext.IsReplaying);
 
             T result = default;
 
@@ -322,13 +332,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             {
                 this.IncrementActionsOrThrowException();
                 Task<T> timerTask = this.InnerContext.CreateTimer(intervalFireAt, state, cancelToken);
-
-                this.Config.TraceHelper.FunctionListening(
-                    this.Config.Options.HubName,
-                    this.OrchestrationName,
-                    this.InstanceId,
-                    reason: $"CreateTimer:{fireAt:o}",
-                    isReplay: this.InnerContext.IsReplaying);
 
                 result = await timerTask;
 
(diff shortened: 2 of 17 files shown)


src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
```diff
@@ -139,7 +139,7 @@ public override bool ValidateDelayTime(TimeSpan timespan, out string errorMessag
         {
             if (timespan > MaxTimerDuration)
             {
-                errorMessage = $"The Azure Storage provider supports a maximum of {MaxTimerDuration.TotalDays} days for time-based delays";
+                errorMessage = null;
```
cgillum, line 142: I see why you did this, but putting myself in the shoes of someone who is not an expert in this code, it's hard to understand the logic behind returning a null error message. A comment would help, but even then it's still weird to have a validate method that never results in an exception.

I recommend a slightly different design, which might be cleaner. Instead of changing `ValidateDelayTime`, add a new abstract property called `MaximumDelayTime` in the `DurabilityProvider` base class. Here in `AzureStorageDurabilityProvider` it could simply return the `MaxTimerDuration` value (6 days). Then, in `DurableOrchestrationContext`, revert your change to the `ValidateDelayTime` logic and instead add a check before that which compares the `fireAt` value with the new `MaximumDelayTime` property. It could look something like this:

```csharp
if (fireAt.Subtract(this.InnerContext.CurrentUtcDateTime) > this.MaximumDelayTime)
{
    internvalFireAt = this.InnerContext.CurrentUtcDateTime.AddDays(longRunningTimerIntervalLength.TotalDays);
}
```

If it turns out that we actually don't need `ValidateDelayTime` anymore, then let's make it obsolete by adding the `[Obsolete]` attribute to it, since our design no longer care about maximum delays anymore (I can't think of any other provider which would care about this either).

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -308,29 +308,53 @@ async Task<T> IDurableOrchestrationContext.CreateTimer<T>(DateTime fireAt, T sta
         {
             this.ThrowIfInvalidAccess();
 
+            DateTime intervalFireAt = fireAt;
+            TimeSpan longRunningTimerIntervalLength = TimeSpan.FromDays(3);
+
             if (!this.durabilityProvider.ValidateDelayTime(fireAt.Subtract(this.InnerContext.CurrentUtcDateTime), out string errorMessage))
             {
-                throw new ArgumentException(errorMessage, nameof(fireAt));
+                intervalFireAt = this.InnerContext.CurrentUtcDateTime.AddDays(longRunningTimerIntervalLength.TotalDays);
```
cgillum, line 316: Instead of `.AddDays(longRunningTimerIntervalLength.TotalDays)`, just do `.Add(longRunningTimerIntervalLength)`. Same comment in the places below where you are also using `AddDays`.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -308,29 +308,53 @@ async Task<T> IDurableOrchestrationContext.CreateTimer<T>(DateTime fireAt, T sta
...
-                this.InstanceId,
-                reason: $"CreateTimer:{fireAt:o}",
-                isReplay: this.InnerContext.IsReplaying);
+            while (this.InnerContext.CurrentUtcDateTime < fireAt && !cancelToken.IsCancellationRequested)
+            {
+                this.IncrementActionsOrThrowException();
+                Task<T> timerTask = this.InnerContext.CreateTimer(intervalFireAt, state, cancelToken);
 
-            T result = await timerTask;
+                this.Config.TraceHelper.FunctionListening(
```
cgillum, line 326: I suggest we move this trace outside of the `while` loop so that we only trace once. The fact that we use multiple messages should probably not be exposed at this layer of logs.

bachuv:
The tests, `AzureStorage_TimerLimitExceeded_ThrowsException()` and `ThrowExceptionOnLongTimer()`, fail now because of the changes that make durable timers last longer than 7 days. Would it be best to remove them and see if they can be replaced with a test that has a shorter interval time and max time?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -545,7 +545,7 @@ private async Task OrchestrationMiddleware(DispatchMiddlewareContext dispatchCon
                     $"An internal error occurred while attempting to execute '{context.FunctionName}'.", result.Exception);
             }
 
-            if (!context.IsCompleted)
+            if (!context.IsCompleted && !context.IsLongRunningTimer)
```
cgillum, line 548: Good idea!

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -311,25 +313,26 @@ async Task<T> IDurableOrchestrationContext.CreateTimer<T>(DateTime fireAt, T sta
...
+                this.IsLongRunningTimer = true;
+                intervalFireAt = this.InnerContext.CurrentUtcDateTime.Add(longRunningTimerIntervalLength);
             }
 
+            this.Config.TraceHelper.FunctionListening(
+                    this.Config.Options.HubName,
+                    this.OrchestrationName,
+                    this.InstanceId,
+                    reason: $"CreateTimer:{fireAt:o}",
+                    isReplay: this.InnerContext.IsReplaying);
```
cgillum, line 327: nit: only use 4 spaces for indentation. It looks like these parameters are double-indented with 8 spaces.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -85,6 +85,8 @@ internal bool IsReplaying
 
         internal bool IsCompleted { get; set; }
 
+        internal bool IsLongRunningTimer { get; set; }
```
cgillum, line 88: Can we make this `private set` instead of `set`? We should always try to minimize the visibility of fields where reasonably possible.

MaximRouiller:
This is exactly what I needed. I have things to schedule that may be every 7 days, 14 days, 4 weeks, or 6 weeks. 

I'd love for this to be merged yesterday 😛

jedjohan:
Oh yes, was just implementing a complicated (to me) while loop to crank more than 7 days out of durable timers. This makes me really happy. Will there be a new maximum delay implemented ? Keep up the good work, thanks !

bachuv:
Glad to hear that @jedjohan! No, there isn't a new maximum delay. If a timer is scheduled for longer than the maximum delay, then a new timer is scheduled every three days until the difference is less than three days. This ensures that users won't have to worry about staying within a maximum time.

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
```diff
@@ -54,6 +52,10 @@ internal class AzureStorageDurabilityProvider : DurabilityProvider
 
         public override JObject ConfigurationJson => this.storageOptionsJson;
 
+        public override TimeSpan MaximumDelayTime { get; set; } = TimeSpan.FromDays(6);
+
+        public override TimeSpan LongRunningTimerIntervalLength { get; set; } = TimeSpan.FromDays(3);
```
ConnorMcMahon, line 57: Setting this to 3 days means that in-flight orchestrations that have timers of lengths between 3 and 6 days (the current maximum) will end up with non-determinism errors. 

In order to mitigate this, we have a few options:

1. Make long-running timers opt-in. This means adding a flag on DurableTaskOptions, and will probably greatly reduce the number of customers who make use of the feature.
2. Make a note as a "Breaking Change" in the release notes. We generally try to avoid breaking changes for in-flight orchestrations, as many customers don't read the release notes, and might shoot themselves in the foot this way.
3. Set the timer interval to 6 days, as no customers have timers currently longer than 6 days due to previous limits in place.

@cgillum thoughts?

cgillum: @ConnorMcMahon I don't yet understand why this change would cause non-determinism failures for in-flight orchestrations. Are you saying that this PR will cause us to change the fire time for already scheduled timers? I didn't get that impression when reading the code. My understanding was that this 3 day interval value only applies to timers that are scheduled for longer than the 6 day maximum. Otherwise we don't behave differently. If I'm misunderstanding the logic and you are certain about this problem, then I'm totally fine with option (3).

ConnorMcMahon: Ah I misread this code in this approach. I think you are right, as the internalFireAt should equal fireAt if we don't exceed the maximum timer. That makes sense, and completely nullifies this concern.

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -127,6 +127,14 @@ private async Task ProcessAsyncActions(AsyncAction[][] actions)
                         case AsyncActionType.CreateTimer:
                             using (var cts = new CancellationTokenSource())
                             {
+                                DurableOrchestrationContext ctx = this.context as DurableOrchestrationContext;
```
ConnorMcMahon, line 130: We should handle the case where ctx is null (i.e. where this.context is not of type `DurableOrchestrationContext`). We can do this by throwing an exception.

This should only happen in test code where we pass in a mock IDurableOrchestrationContext, but getting a nice exception as opposed to a null ref exception is always preferred.

test/FunctionsV2/PlatformSpecificHelpers.FunctionsV2.cs
```diff
@@ -78,6 +78,9 @@ private static IWebJobsBuilder AddDurableTask(this IWebJobsBuilder builder, IOpt
                 case TestHelpers.AzureStorageProviderType:
                     // This provider is built into the default AddDurableTask() call below.
                     break;
+                case TestHelpers.AzureStorageProviderModifiedType:
```
ConnorMcMahon, line 81: I think you could make this a bit more generalized. Instead of adding a new string type, have the `GetJobHost` method take a type of `IDurabilityProviderFactory`, and if that is provided, use that instead of the type specified in the string `storageProvider`. 

That way, you could remove some of the special logic around `"azure_storage_modified"`, and your tests would look like:

```csharp
            using (ITestHost host = TestHelpers.GetJobHost(
                this.loggerProvider,
                nameof(this.LongRunningTimer),
                extendedSessions,
                storageProviderType: "azure_storage",
                storageProviderFactory = typeof(AzureStorageShortenedMaxTimerDurabilityProviderFactory))
```

test/FunctionsV2/AzureStorageModifiedDurabilityProviderFactory.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
+{
+    internal class AzureStorageModifiedDurabilityProviderFactory : AzureStorageDurabilityProviderFactory
```
ConnorMcMahon, line 9: This name is very vague.

I would go for something like `AzureStorageShortenedTimerDurabilityProviderFactory`.

test/FunctionsV2/TimerTests.cs
```diff
@@ -0,0 +1,69 @@
...
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests;
+using Microsoft.Azure.WebJobs.Host.TestCommon;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace WebJobs.Extensions.DurableTask.Tests.V2
```
ConnorMcMahon, line 12: I realized we could theoretically support this test for Functions V1, we would just have to write some reflection based code in PlatformSpecificHelpers.FunctionsV1. Not sure if that is worth the effort though.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -332,9 +357,23 @@ async Task<T> IDurableOrchestrationContext.CreateTimer<T>(DateTime fireAt, T sta
                 expirationTime: fireAt,
                 isReplay: this.InnerContext.IsReplaying);
 
+            this.IsLongRunningTimer = false;
+
             return result;
         }
 
+        public string ValidOutOfProcTimer(DateTime fireAt)
```
ConnorMcMahon, line 365: General approach for these types of problems is to do something like 

`bool ValidateOutOfProcTimer(DateTIme fireAt, out string errorMessage)`.

That way the caller doesn't have to know that an empty string means that it is valid, and a non-empty string means it isn't.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -308,22 +310,45 @@ async Task<T> IDurableOrchestrationContext.CreateTimer<T>(DateTime fireAt, T sta
...
+
+                TimeSpan remainingTime = fireAt.Subtract(this.InnerContext.CurrentUtcDateTime);
+                if (remainingTime <= TimeSpan.Zero)
+                {
+                    break;
+                }
+                else if (remainingTime < this.durabilityProvider.LongRunningTimerIntervalLength)
+                {
+                    TimeSpan timerStartsIn = fireAt.Subtract(this.InnerContext.CurrentUtcDateTime);
+                    intervalFireAt = this.InnerContext.CurrentUtcDateTime.Add(timerStartsIn);
```
cgillum, line 345: Are we doing redundant math here? I would have expected the following:
```suggestion
                    intervalFireAt = this.InnerContext.CurrentUtcDateTime.Add(remainingTime);
```

...and removing line 344.

test/FunctionsV2/TimerTests.cs
cgillum: Probably not worth backporting to V1. I'm thinking we should try to limit new features to V2 and just backport important bug fixes to V1.

test/FunctionsV2/TimerTests.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        public TimerTests(ITestOutputHelper output)
+        {
+            this.output = output;
+            this.loggerProvider = new TestLoggerProvider(output);
+        }
+
+        [Theory]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        [MemberData(nameof(TestDataGenerator.GetBooleanOptions), MemberType = typeof(TestDataGenerator))]
```
cgillum, line 27: Using a test data generator for a single boolean value is probably overkill. I suggest you instead use `[InlineData(true)]` and `[InlineData(false)]`.

MaximRouiller:
Can't wait for this to be merged. 🤩

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -127,6 +127,13 @@ private async Task ProcessAsyncActions(AsyncAction[][] actions)
...
+
+                                if (ctx != null)
+                                {
+                                    tasks.Add(ctx.OutOfProcCreateTimer(ctx, action.FireAt, cts.Token));
+                                }
```
ConnorMcMahon, line 135: Shouldn't line 137 be in an else case of this?

bachuv: Yeah it should, just fixed it.

PR #1391, opened 2020-06-23
WIP: Enabling Linux Telemetry

davidmrdavid:
**Do not merge, this PR is here for discussion**

### Summary and Intent
This PR contains a naive first-attempt at enabling telemetry to be forwarded to Kusto in Azure's Linux offerings in App Service. This PR is by no means the _right_ way to implement this and so my hope is to use it as a way to communicate what it takes to write telemetry in our Linux servers and to use that to design a cleaner, and more robust, alternative.

#### Linux Dedicated
In Linux Dedicated, we write our logs to the file `/var/log/functionsLogs/durableevents.log`. A background process monitors changes to this file and, informed by the configuration we made in that codebase, those logs get redirected to Kusto. So really the job on the Durable Functions - side of things is fairly easy: _write a structured log to the specified file every time we want to emit a telemetry event_. Every line in  `/var/log/functionsLogs/durableevents.log` needs be parseable by a regular expression, which is available in the azure-internal changes we've made, and our code needs to produce new lines where the fields are ordered as per the regular expression's expectations. 

### Linux Consumption
The story is rather similar for Linux Consumption. The **only** difference is that, instead of writing to a file, we write our logs to `STDOUT` and provide a unique prefix to our logged lines which tells Azure what kind of log this is and, therefore, to which Kusto table to send it to.

### Problem Areas with the code

Our events can contain many kinds of fields: over 20 if I'm not mistaken. This means that it is not easy to elegantly generate the regex-parseable string that Azure expects. Currently, my "solution" is to dynamically populate a `DurableFunctionsLinuxLog` object, which exports a public dictionary that the object-user can populate by mapping a field (e.g. `FunctionName`) to its value (e.g. `DavidsFunction`), and then to produce the regex-parseable string via a `AsString` method.

This method implementation also bothers me because, in order to preserve the ordering of fields, I need to create this large array of strings (the fields) and use their position in the array to enforce the order in which values are written to the log-string.

Finally, this approach greatly expands the amount of work that the `ETWEventSource` class needs to do, as now we need to manually populate that dictionary.

### Performance
As it's been mentioned before, Durable Functions Telemetry is "quite chatty" and so I/O operations will probably need to be batched for performance reasons. The `azure-functions-host` codebase implemented a batching strategy for Linux Dedicated in this [file](https://github.com/Azure/azure-functions-host/blob/35cf323fa3464a08b410a518bcab006e801301fe/src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLogger.cs). There seem to be many hand-tuned parameters in it and so, for the sake of making progress, I just borrowed it directly to our codebase and called it `LinuxDedicatedLogger`. I've intentionally left it mostly un-modified because we may want to understand it better, and "make it our own". 

We currently do not have a "batching" strategy for Linux Consumption. There may already be one in `azure-functions-host` as well, but I have not seen it yet.

As an aside, the comments in this PR currently claim that I'm writing asynchronously to the console or to the log-file, but this is not the case yet. Again, this is just a naive _first_ attempt.

### On propagating this to DTFx
To enable Telemetry at the DTFx level, we'll need similar changes. In my opinion, once we design a clean solution, we should **move** these classes and their initialization to DTFx and simply export them to the extension (this repo) so that they can be used.

### A potential clean solution
At its core, the problem here is that `EventSource` is not natively supported at the Azure-level in its Linux offerings, so we have to do all this manual work that needs to be replicated at every `EventSource` subclass. 

So, in my mind, an ideal solution would be to create a drop-in replacement for `EventSource`. This replacement would be a  **subclass** of `EventSource` named, say, `MultiPlatformEventSource` which, on a windows machine, would simply act as `EventSource` (its parent) would. Otherwise, it would _do the right thing_ and print either to the console or write to the log file. Afterwards, all we'd need to enable Linux Telemetry for us, or for any other project, would be to change all `EventSource` inheritors to inherit from `MultiPlatformEventSource`. It'd be _like magic_.

To figure out what to print in `MultiPlatformEventSource`, we could leverage **reflection** on the parameter names of the `EventSource` method that's getting called and use that to, in the worst case scenario, generate the log-line via our current `DurableFunctionsLinuxLog`-based strategy. 

This is what I originally tried to do, but reflection is hard, and I'm under the impression that it is not very efficient in C#, which would defeat the purpose.

Please let me know your thoughts! Finally, another friendly reminder that this PR is deliberately naive because the time-pressing part was to make the Azure-internal changes. Thanks!

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableFunctionsLinuxLog.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableFunctionsLinuxLog.cs
@@ -1,0 +1,69 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    /// <summary>
+    /// A string-serializable collection of data representing a durable event
+    /// to be logged in Linux.
+    /// </summary>
+    public class DurableFunctionsLinuxLog
+    {
+        /// <summary>
+        /// Turns Log into a string.
+        /// </summary>
+        /// <returns>String-representation of the log.</returns>
+        public static string AsString(Dictionary<string, string> record)
+        {
+            string log = "";
+            List<string> orderedColumns = new List<string>
+            {
+                "Account", "ActiveActivities", "ActiveOrchestrators",
+                "Age", "AppName", "ContinuedAsNew", "CreatedTimeFrom",
+                "CreatedTimeTo", "DequeueCount", "Details", "Duration",
+                "ETag", "Episode", "EventCount", "EventName", "EventType",
+                "Exception", "ExceptionMessage", "ExecutionId", "ExtensionVersion",
+                "FromWorkerName", "FunctionName", "FunctionState", "FunctionType",
+                "Input", "InstanceId", "IsCheckpointComplete", "IsExtendedSession",
+                "IsReplay", "LastCheckpointTime", "LatencyMs", "MessageId", "MessagesRead",
+                "MessagesSent", "MessagesUpdated", "NewEventCount", "NewEvents", "NextVisibleTime",
+                "OperationId", "OperationName", "Output", "PartitionId", "PendingOrchestratorMessages",
+                "PendingOrchestrators", "Reason", "RelatedActivityId", "RequestCount", "RequestId",
+                "RequestingExecutionId", "RequestingInstance", "RequestingInstanceId", "Result",
+                "RuntimeStatus", "SequenceNumber", "SizeInBytes", "SlotName", "StatusCode",
+                "StorageRequests", "Success", "TableEntitiesRead", "TableEntitiesWritten",
+                "TargetExecutionId", "TargetInstanceId", "TaskEventId", "TaskHub", "Token",
+                "TotalEventCount", "Version", "VisibilityTimeoutSeconds", "WorkerName",
+            };
+
+            // to represent the value of a column
+            string value;
+
+            // we write each column in order
+            foreach (string column in orderedColumns)
+            {
+                value = "";
+
+                // Details and ExceptionMessage are double-quote-wrapped
+                // messages, so their default value is '""'
+                if (column == "Details" || column == "ExceptionMessage")
+                {
+                    value = "\"\"";
+                }
+
+                if (record.ContainsKey(column))
+                {
+                    value = record[column];
+                }
+
+                log += value + ",";
+            }
+
+            int logLength = log.Length;
+            log = log.Remove(logLength - 1);
+
+            return log;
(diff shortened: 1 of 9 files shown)


cgillum:
Thanks David! This writeup (and the other one you sent me) was very helpful for me to understand what's going on. I have two proposals about what we could consider as next steps:

### Option 1 - Event Source on Linux
According to [this documentation](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/logging-tracing#logging-events), EventSource _can_ be used on the Linux OS, but it doesn't use ETW. Rather, it requires the use of [LTTng](https://lttng.org/) to capture these traces. If the Linux infrastructure could use LTTng and push that to Kusto, then we could potentially have a more generic scenario that does not require new code for logging to a file. However, it's not clear to me if/how LTTng and FluentD can be used together. Adding support for LTTng could potentially be very costly since it requires updating the Linux core infrastructure.

### Option 2 - Modernize the Durable logging code
Think of this as a variation of your current PR that is general purpose and flexible enough to support a variety of environments. The basic idea is to standardize on the [.NET Core ILogger abstractions](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/logging-tracing#ilogger-and-logging-frameworks) and implement providers for each of the environments we support.

* Windows - an EventSource (ETW) provider
* Linux Dedicated - a file system provider
* Linux Consumption - a console provider

The advantage of standardizing on `ILogger` instead of EventSource is that it allows us to decouple our core logging logic from the provider that actually serializes the logs. If done right, it also has the benefit of allowing customers to inject their own providers. Since the Functions runtime already supports `ILogger` natively, it also allows us to write our logs to Application Insights without requiring any code changes.

This option will not necessarily address the performance concerns, but it shouldn't make anything worse. I actually started investigating this exact approach in the `DurableTask.SqlServer` codebase and I have a working prototype showing how I can emit Durable logs into ETW, Application Insights, a file, the console, or any combination of these. I'd love to show it to you if you're interested.

davidmrdavid:
Thanks, @cgillum . I'm glad the documents were helpful.

Abstraction is the name of the game, so I'd be happy to proceed with the `ILogger` approach. Please loop me in on any previews of the SQL backend once those become available. 

Three more comments, in order of urgency:
1) A friendly reminder that we should reach a decision soon as to whether or not to merge this, or a slightly improved version of this, in order to actually produce some logs for Durable Python. Don't we? In doing this, we should be careful to make sure the performance appears reasonable and logging is non-blocking.

2) While we still have Python Durable in beta, we should have a conversation about extracting the implicit fields that are being logged in our ETW events. We have to find a good solution for that before Python Durable comes out of beta.

3) Relatedly, just a reminder that we don't have to opt for writing all the durable logs to the same file. Just pointing out that we still have to think about that...but that one **can wait.**

davidmrdavid:
Closing this because we merged this feature in another branch

PR #1392, opened 2020-06-23
Updated DurableActivityContext GetInputAsJson to handle TypeNameHandling All

### LINKED ISSUE #1389: Serialization Issue in 2.2.2
### Description
After upgrading to version 2.2.2, we receive the following exception when trying to run an Activity from an Orchestrator: "Exception binding parameter [parametername]", where the innerException is "Error reading JArray from JsonReader. Current JsonReader item is not an array: StartObject. Path '', line 1, position 1."

It appears that the framework serializes Activity parameters with " TypeNameHandling.Object", and tries to deserialize with whatever the system default it (in this case TypeNameHandling.All).  This is leading to a mismatch that newtonsoft cannot resolve.

Our app requires the following code in `Startup.cs` in order to deserialize out internal domain objects to/from Storage Queues correctly.

```
 public class MessageSerializerSettingsFactory : IMessageSerializerSettingsFactory
    {
        public JsonSerializerSettings CreateJsonSerializerSettings() => new JsonSerializerSettings
        {
            TypeNameHandling = TypeNameHandling.All
        };
    }
```
This is registered via:
```
 builder.Services.AddSingleton<IMessageSerializerSettingsFactory, MessageSerializerSettingsFactory>();
```

Sample project here: https://github.com/tolzon/Function222Serilization

### App Details

- **Durable Functions extension version 2.2.2**:
- **Azure Functions runtime version 3.1**:
- **Programming language used C#**:

amdeel:
resolves #1389 

When the custom setting TypeNameHandling.All is used, GetInputAsJson would throw an exception when attempting to parse the Json to a JArray. This fix adds a check to handle that case.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActivityContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActivityContext.cs
@@ -59,7 +59,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
                 if (objectArray?.Length != 1)
                     {
-                        throw new ArgumentException("The serialized input is expected to be a JSON array with one element.");
+                        throw new ArgumentException("The serialized input is expected to be an object array with one element.");
                     }
 
                 var token = JToken.Parse(this.messageDataConverter.Serialize(objectArray[0]));
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActivityContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActivityContext.cs
@@ -55,30 +55,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             if (this.serializedInput != null && this.parsedJsonInput == null)
             {
-                if (this.messageDataConverter.JsonSettings.TypeNameHandling == TypeNameHandling.All)
-                {
-                    var objectArray = this.messageDataConverter.Deserialize<object[]>(this.serializedInput);
+                var objectArray = this.messageDataConverter.Deserialize<object[]>(this.serializedInput);
 
-                    if (objectArray?.Length != 1)
+                if (objectArray?.Length != 1)
                     {
                         throw new ArgumentException("The serialized input is expected to be a JSON array with one element.");
                     }
 
-                    var token = JToken.Parse(this.messageDataConverter.Serialize(objectArray[0]));
+                var token = JToken.Parse(this.messageDataConverter.Serialize(objectArray[0]));
 
-                    this.parsedJsonInput = token;
-                }
-                else
-                {
-                    JArray array = JArray.Parse(this.serializedInput);
-
-                    if (array?.Count != 1)
-                    {
-                        throw new ArgumentException("The serialized input is expected to be a JSON array with one element.");
-                    }
-
-                    this.parsedJsonInput = array[0];
-                }
+                this.parsedJsonInput = token;
             }
 
             return this.parsedJsonInput;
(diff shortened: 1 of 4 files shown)


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActivityContext.cs
```diff
@@ -55,13 +55,30 @@ internal JToken GetInputAsJson()
...
                 {
-                    throw new ArgumentException("The serialized input is expected to be a JSON array with one element.");
+                    var objectArray = this.messageDataConverter.Deserialize<object[]>(this.serializedInput);
+
+                    if (objectArray?.Length != 1)
+                    {
+                        throw new ArgumentException("The serialized input is expected to be a JSON array with one element.");
+                    }
+
+                    var token = JToken.Parse(this.messageDataConverter.Serialize(objectArray[0]));
```
amdeel, line 67: I couldn't find a method that allowed me to do this easier, but I'd like to know if there is an easier way.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActivityContext.cs
```diff
@@ -55,13 +55,30 @@ internal JToken GetInputAsJson()
         {
             if (this.serializedInput != null && this.parsedJsonInput == null)
             {
-                JArray array = JArray.Parse(this.serializedInput);
-                if (array?.Count != 1)
+                if (this.messageDataConverter.JsonSettings.TypeNameHandling == TypeNameHandling.All)
```
ConnorMcMahon, line 58: I think the fundamental problem here is not just with TypeNameHandling.All, but that we don't use the JsonSettings on the message data converter to deserialize here.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActivityContext.cs
ConnorMcMahon: It makes me wonder if we should even use this method inside of `GetInput(Type destinationType)`. I realize we have to for the case where the destination type is string so that we support the out-of-proc case, but maybe we just do something like below for the other case: 

`return this.messageDataConvert.Deserialize(this.serializedInput, destinationType.GetArrayType())[0]`;

On the other hand, we probably want to fix this method too, as it is used in other places.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActivityContext.cs
amdeel: You're right, I could make this the only path instead. For some reason initially I just figured that because in other cases it deserializes to JArray I should keep it like that because its simpler. But that doesn't make sense, I can just choose to deserialize to an object[] every time.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -4461,6 +4461,34 @@ public async Task CustomIMessageSerializerSettingsFactory()
             }
         }
 
+        [Fact]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        public async Task CustomSerializerSettings_TypeNameHandlingAll()
+        {
+            string[] orchestratorFunctionNames =
+            {
+                nameof(TestOrchestrations.SayHelloInline),
```
ConnorMcMahon, line 4470: This test passes even with the customer serializer settings before your source code changes. In order to make sure this code tests the bug fix, we need to run an orchestrator that uses a more complicated type.

amdeel: Oh, oops, I forgot to use an orchestrator that uses an activity function. That is where the input gets turned into an object[] in DTFx and where/why this serialization issue occurs.

amdeel: Good catch!

PR #1394, opened 2020-06-24
Analyzer: Added AwaitAnalyzer, fixed bugs, refactored code

### LINKED ISSUE #1288: Analyzer - Does not pick up activities found in dependencies
We currently have a number of function activities which we share code and reference using the `<FunctionsInDependencies>true</FunctionsInDependencies>` project setting. These activities seem to still show the DF109 warning from the analyzer

Checks are performed for activities found in dependencies and no warning shown

Warning is currently shown for activities found in dependencies

```csharp
Project A

        var jobEntity = await context.CallActivityWithRetryAsync<JobEntity>("GetJobEntity", _retryOptions,
            new InvestigationJobMessage { InvestigationKey = message.InvestigationKey, JobNumber = message.JobNumber });

Project B, dependency of A
        [FunctionName(JobConstants.GetJobEntityActivityName)]
        public async Task<JobEntity> GetJob([ActivityTrigger] InvestigationJobMessage message, ILogger log)
        {
```

None

### App Details

- **Durable Functions extension version (e.g. v1.8.3)**: 2.2.0
- **Azure Functions runtime version (1.0 or 2.0)**: 3.0
- **Programming language used**: .NET

image

### LINKED ISSUE #1311: Analyzer: Flag custom awaits in an orchestrator
Make an OrchestratorAnalyzer that flags awaits on customer defined code or non-durable code.

### LINKED ISSUE #1373: NullReferenceException in Analyzer v0.2.3 (included in Durable Functions v2.2.2)
### Description
I'm updating my project from v2.2.0 to v2.2.2 of Durable Functions, and am seeing the FunctionAnalyzer throw an NRE.  We're not blocked, but it would be nice to know a good way to disable the analyzer so that I can pick up the latest version of the package without stopping treating warnings as errors (couldn't figure out how to not let AD0001 be promoted to an error).

```
3>CSC : warning AD0001: Analyzer 'Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers.FunctionAnalyzer' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'.
```

### Expected behavior
No NREs in analyzers.

### Actual behavior
NRE in analyzer!

The code is not public, and is not in a position to be made public yet.  I also don't know how to provide a reduction, since I can't even get a stack trace out of the analyzer to know what part of my code it's choking on.  I did reach out to @amdeel a few minutes ago to share a ms-internal Azure DevOps link, so hopefully that's enough to go on.

### Known workarounds
None.

### App Details

- **Durable Functions extension version (e.g. v1.8.3)**: v2.2.2
- **Azure Functions runtime version (1.0 or 2.0)**: v3.0

### LINKED ISSUE #1378: Analyzer 0.2.3 does not recognize type inheritance
### Description
I have an `ActivityFunction` defined that takes an `IEnumerable<MyClass>`, but when I call it from my Orchestrator, the object I'm passing in is a `List<MyClass>`. The Analyzer throws a DF0108 warning for this even though it shouldn't. 

### Expected behavior
The analyzer should recognize that a `List<MyClass>` is also an `IEnumerable<MyClass>` and not throw a warning.

### Actual behavior
DF0108 warning:

```csharp
Activity function named 'Foo' takes '(System.Guid, System.Collections.Generic.IEnumerable<MyClass>)' but was given '(System.Guid, System.Collections.Generic.List<MyClass>)'.
```

### App Details

- **Durable Functions extension version (e.g. v1.8.3)**: DurableTask v2.2.2, DurableTask.Analyzer v0.2.3

amdeel:
**Added Await Analyzer**
resolves #1311 

This analyzer is to cover all cases that are not currently covered with the MethodAnalyzer. The AwaitAnalyzer flags customer-written methods that use the await keyword in an orchestrator (or any method called from an orchestrator/code required to be deterministic).

_AwaitAnalyzer.cs, AwaitAnalyzerTests.cs_

**Type comparison refactor** (For FunctionAnalyzers ArgumentAnalyzer and ReturnTypeAnalyzer)
resolves #1378 

This refactor checks for equal types, compatible json types (types serialized to json arrays/implement IEnumerable), and subclass or implementation types (that are only checked in one direction, ex: ArgumentAnalyzer checks that the input to a function is either a matching type or subclass, but not the other way around). This works for generic types, Tuples, ValueTuples, and any nested types by checking all type arguments for each requirement (see: NightmareTest in ArgumentAnalyzerTests.cs).

_SyntaxNodeUtils.cs lines 360-550_

**Test Updates**

In the last analyzer PR, I started updating the tests to a common format, starting with FunctionAnalyzer tests. This PR continues adding those changes to the other test files.

_All *Tests.cs_

**Refactor of various files**
resolves #1373 

Specifically the refactor in the FunctionAnalyzers, ArgumentAanlyzer, ReturnTypeAnalyzer, and NameAnalyzer, solved an issue with a NullReferenceException. This also inspired a bit of a refactor with various files in areas to make the code a little bit more readable.

**Changed NameAnalyzer for FunctionsInDependencies**
resolves #1288 

When a customer is using `<FunctionsInDependencies>true</FunctionsInDependencies>` the NameAnalyzer will not be able to find the function in the project and it will flag the code as if the function does not exist. Now, if a customer is using a constant variable but the function is not found in the project, we will trust the customer that this function does exist and will no longer flag their code with a warning.

_NameAnalyzer.cs, NameAnalyzerTests.cs_


src/WebJobs.Extensions.DurableTask.Analyzers/SyntaxNodeUtils.cs
```diff
@@ -308,133 +345,208 @@ private static bool IsSpecifiedAttribute(AttributeSyntax attribute, string attri
...
+                || type.ToString().Equals("Microsoft.Azure.WebJobs.DurableActivityContextBase"));
+        }
 
-                var arrayString = "";
-                if (typeInfo.Kind.Equals(SymbolKind.ArrayType))
-                {
-                    arrayString = "[]";
-                    typeInfo = ((IArrayTypeSymbol)typeInfo).ElementType;
-                }
+        public static bool IsMatchingDerivedOrCompatibleType(ITypeSymbol subclassOrMatching, ITypeSymbol superOrMatching)
```
amdeel, line 360: As you will see in the code below, I have subclassOrMatching and superOrMatching listed as parameters throughout most of these methods. The order of these parameters matters. An input to a function can be a subclass, but not the other way around. The expected return type can be a super, but not the other way around.

I dislike that this method would be so easy to misuse. Of course, this method is unlikely to get misused in this project but I wanted input on this approach.

What is a better way to handle cases where parameters have order that matter? I thought about making this an extension method, but ultimately kept it matching the other methods that follow for consistency. On second thought, I could always make the public method an extension method and the private methods I could keep with parameters that require the specific order. Thoughts?

AtOMiCNebula:
For the AwaitAnalyzer, will there be a way to mark customer-written async functions as acceptable to invoke from orchestrators?  I have a few extension methods for `IDurableOrchestrationContext` that help keep my code organized, and I'd prefer to not suppress these warnings away.  They themselves don't invoke any non-`IDurableOrchestrationContext` APIs, but I'm not sure how easy that is to auto-detect.

amdeel:
@AtOMiCNebula Good question. One solution for this would be to change the AwaitAnalyzer to not flag code when awaits are used on those custom methods contained in a class called "DurableContextExtensions" (or some standard name or set of names).

I discussed this with @cgillum but don't know if we decided this was the solution we wanted to go with. Chris, what do you think of this solution?

cgillum:
I think looking for a class named `DurableContextExtensions` is a nice simple way to handle this. Another option might be to bring back the `[Deterministic]` attribute as a way of marking specific extension methods (or maybe any method?) as safe for use with `async`. This would be more work though.

Thoughts?

test/WebJobs.Extensions.DurableTask.Analyzers.Test/ActivityFunction/NameAnalyzerTests.cs
```diff
@@ -55,31 +56,35 @@ public static class HelloSequence
         {
             // Matching names (strings)
 
-            await context.CallActivityAsync<string>(""Test_MatchingStrings"", ""Tokyo"");
+            await context.CallActivityAsync<string>(""Test_MatchingStrings"", ""Sandra Bland"");
```
cgillum, line 59: I'm not comfortable with these activity name changes. Let's stick with the city name convention.

test/WebJobs.Extensions.DurableTask.Analyzers.Test/Orchestrator/AwaitAnalyzerTests.cs
```diff
@@ -0,0 +1,208 @@
...
+                Message = Resources.AwaitAnalyzerMessageFormat,
+                Severity = Severity,
+                Locations =
+                    new[] {
+                            new DiagnosticResultLocation("Test0.cs", 20, 17)
+                        }
+            };
+
+            VerifyCSharpDiagnostic(test, expectedDiagnostics);
+        }
```
cgillum, line 201: What happens if I have code like this?
```csharp
[FunctionName("AwaitAnalyzerTestCases")]
public static async Task Run([OrchestrationTrigger] IDurableOrchestrationContext context)
{
    await Method(context);
}

private static async Task Method(IDurableOrchestrationContext context)
{
    await context.CallActivityAsync("Hello", "World");
}
```

Will this pass without a warning? It should, since it doesn't violate any determinism code constraints.

amdeel:
I changed my mind about the AwaitAnalyzer. It makes a lot of things more complicated compared to situations it could actually be helpful. I want to get rid of it entirely.

AtOMiCNebula:
Ok!  I see value in having the await-analyzer, we just need a way to hint to it that some things are acceptable.  Hopefully it can be revisited later. :smile:

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/FunctionReturnTypeAnalyzer.cs
```diff
@@ -22,27 +21,25 @@ public class FunctionReturnTypeAnalyzer
...
+            SemanticModel semanticModel,
+            IEnumerable<ActivityFunctionDefinition> functionDefinitions,
+            IEnumerable<ActivityFunctionCall> functionInvocations)
         {
-            foreach (var activityInvocation in calledFunctions)
+            foreach (var invocation in functionInvocations)
             {
-                var functionDefinition = availableFunctions.Where(x => x.FunctionName == activityInvocation.Name).FirstOrDefault();
-                if (functionDefinition != null && activityInvocation.ReturnTypeNode != null)
+                var definition = functionDefinitions.Where(x => x.FunctionName == invocation.FunctionName).FirstOrDefault();
```
ConnorMcMahon, line 31: similar nit about `Find()` here.

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/NameAnalyzer.cs
```diff
@@ -28,28 +29,35 @@ private static string GetClosestString(string name, IEnumerable<string> availabl
...
+            CompilationAnalysisContext context,
+            SemanticModel semanticModel,
+            IEnumerable<ActivityFunctionDefinition> functionDefinitions,
+            IEnumerable<ActivityFunctionCall> functionInvocations)
         {
-            foreach (var activityInvocation in calledFunctions)
+            foreach (var invocation in functionInvocations)
             {
-                if (!availableFunctions.Select(x => x.FunctionName).Contains(activityInvocation.Name))
+                // If invocation uses constant and there is no matching function name in function definition, trust the customer for correctness in case they are using <FunctionsInDependencies>true</FunctionsInDependencies>
```
ConnorMcMahon, line 39: Do we have any knowledge into whether the customer has set `FunctionsInDependencies` in their csproj? If we do, I think it would be preferable to only filter out constants in that case.

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/Orchestrator/AwaitAnalyzer.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public static bool RegisterDiagnostic(SyntaxNode method, CompilationAnalysisContext context, SemanticModel semanticModel)
+        {
+            var diagnosedIssue = false;
+
+            foreach (SyntaxNode descendant in method.DescendantNodes())
+            {
+                if (descendant.IsKind(SyntaxKind.AwaitExpression))
+                {
+                    var invocationExpression = descendant.ChildNodes().Where(x => x.IsKind(SyntaxKind.InvocationExpression)).FirstOrDefault();
```
ConnorMcMahon, line 32: Same nit about `.Find()`.

amdeel:
I could demote my AwaitAnalyzer to just Info (grey squiggle and a message in the error list), but even that can add a lot of clutter and I still think it's for the best to just get rid of it.

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/ArgumentAnalyzer.cs
```diff
@@ -3,16 +3,16 @@
...
 using Microsoft.CodeAnalysis.Diagnostics;
-using System;
 using System.Collections.Generic;
 using System.Linq;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers
 {
     public class ArgumentAnalyzer
     {
-        public const string DiagnosticId = "DF0108";
```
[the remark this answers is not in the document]
amdeel: That's a very good point. And, actually, I'm going to take out the AwaitAnalyzer, so I can revert the changed Ids. In the future I will have to be cautious when I want to change them.

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/NameAnalyzer.cs
amdeel: We don't currently have that information. I could see if doing this would be easy to add that check, but if I have to parse their xml I think I'll just leave it. I figure that if they are using a constant it's unlikely to be an incorrect name here anyway. This does only affect the NameAnalyzer.

test/WebJobs.Extensions.DurableTask.Analyzers.Test/Binding/OrchestratorContextAnalyzerTests.cs
```diff
@@ -306,7 +383,85 @@ public static class HireEmployee
...
+            SyntaxNodeUtils.version = DurableVersion.V2;
+
+            VerifyCSharpDiagnostic(test, expectedDiagnostics);
+
+            VerifyCSharpFix(test, V2ExpectedFix);
+        }
+
+        // Tests SyntaxKind.GenericName
+        [TestMethod]
+        public void OrchestrationContext_V2_UsingList()
```
ConnorMcMahon, line 399: I notice there is a lot of duplicate tests for v1 vs v2. Consider using parameterized tests for cases that apply to both.

test/WebJobs.Extensions.DurableTask.Analyzers.Test/Entity/DispatchEntityNameAnalyzerTests.cs
```diff
@@ -111,25 +152,58 @@ public class MyEmptyEntity : IMyEmptyEntity
...
+        {
+            var test = @"
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+
+ public class MyEmptyEntity : IMyEmptyEntity
+    {
+        [FunctionName(""MyEmptyEntity"")]
+        public static Task Run([EntityTrigger] IDurableEntityContext ctx) => ctx.DispatchAsync<ILogger>();
```
ConnorMcMahon, line 197: This scenario might be another one that is tricky to define if customers have interfaces defined in a dependency. I think that is less likely than having activity functions defined in a dependency, so I am fine leaving it unless someone complains.

test/WebJobs.Extensions.DurableTask.Analyzers.Test/EntityInterface/EntityInterfaceReturnTypeAnalyzerTests.cs
```diff
@@ -128,21 +130,104 @@ public interface IEntityExample
...
+
+    namespace VSSample
+    {
+        public static class HelloSequence
+        {
+            [FunctionName(""ReturnTypeAnalyzerTestCases"")]
+            public static async Task Run(
+            [OrchestrationTrigger] IDurableOrchestrationContext context)
+            {
+                context.SignalEntityAsync<IEntityExample>();
```
ConnorMcMahon, line 151: I'm a bit confused about what this diagnostic is trying to detect. First, it looks like `SignalEntityAsync()` is [missing some required parameters](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.webjobs.extensions.durabletask.durableentityproxyextensions.signalentityasync?view=azure-dotnet#Microsoft_Azure_WebJobs_Extensions_DurableTask_DurableEntityProxyExtensions_SignalEntityAsync__1_Microsoft_Azure_WebJobs_Extensions_DurableTask_IDurableEntityClient_Microsoft_Azure_WebJobs_Extensions_DurableTask_EntityId_System_Action___0__). Second, is the concern that the customer is calling an entity method with the signal functionality that has a return value? I don't know if that is necessarily an anti-pattern.

test/WebJobs.Extensions.DurableTask.Analyzers.Test/Orchestrator/DateTimeAnalyzerTests.cs
```diff
@@ -31,57 +31,60 @@ public void dateTimeNow()
...
-            " + allTests;
+            [FunctionName(""DateTimeAnalyzerTestCases"")]
+            " + allNonDiagnosticTestCases;
 
             VerifyCSharpDiagnostic(test);
         }
         
         [TestMethod]
-        public void DateTimeInOrchestrator_Now_Namespace()
+        public void DateTime_Now_WithNamespace()
```
ConnorMcMahon, line 71: Since all of the orchestration pieces of code are almost identical except for which value goes on the other side of `var dateTime=`, it may be possible to parameterize these tests.

src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/ActivityFunction/ArgumentAnalyzer.cs
```diff
@@ -24,126 +24,117 @@ public class ArgumentAnalyzer
...
+            SemanticModel semanticModel,
+            IEnumerable<ActivityFunctionDefinition> functionDefinitions,
+            IEnumerable<ActivityFunctionCall> functionInvocations)
         {
-            foreach (var activityInvocation in calledFunctions)
+            foreach (var invocation in functionInvocations)
             {
-                var functionDefinition = availableFunctions.Where(x => x.FunctionName == activityInvocation.Name).FirstOrDefault();
-                if (functionDefinition != null)
+                var definition = functionDefinitions.Where(x => x.FunctionName == invocation.FunctionName).FirstOrDefault();
```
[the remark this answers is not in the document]
amdeel: Find is only implemented for List, but I could change these to Lists.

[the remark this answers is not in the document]
ConnorMcMahon: I didn't actually realize that Find() wasn't a linq method. In fact, after reading [this SO answer](https://stackoverflow.com/a/9335036/9035640), I think I prefer the Linq() approach. I would stick with what you have, unless you want to use the overload of .FirstOrDefault() that takes a predicate instead of chaining a .Where().FirstOrDefault()

test/WebJobs.Extensions.DurableTask.Analyzers.Test/EntityInterface/EntityInterfaceReturnTypeAnalyzerTests.cs
amdeel: These EntityInterface tests are actually analyzing the entity interfaces and producing diagnostics on the interface code (aside from InterfaceAnalyzer itself, which is the initial analyzer that runs to store the entity interfaces and trigger the other analyzers to run afterward. This analyzer can also produce diagnostics on this code if the type argument on SignalEntityAsync is not an interface that is defined in user code). What I mean is that these are primarily analyzers for the entity interface itself:

- content (only methods)

- return type (methods must only have the return types void, Task, or Task<T>)

- parameters (methods must have one or 0 parameters).

`public interface IEntityExample
        {
            public static void methodTestVoid();

            public static Task methodTestTask(string test);

            public static Task<T> methodTestTaskT(string test);
        }`

My analyzers determine that an interface is an "entity interface" by saving the user defined interface that is used in SignalEntityAsync. I know that SignalEntityAsync does use parameters, but these tests are for the type argument on SiganlEntityAsync specifically. Also, the test is only comparing the diagnostics before and after my analyzer runs (also, to be clear, it checks that no other analyzer that I wrote produced a diagnostic that I did not intend in the test). I thought it would be extra clutter to add the arguments and would take away from the understanding of the test. I can add the parameters, but I didn't think it was necessary.

I can also write comments on the top of these tests to explain what they are testing if that would help.

amdeel: I'm actually not sure if adding comments to my tests would be better for easier understanding or would be worse because the comments would potentially be out of date when logic in the analyzers changes. Do you have an opinion on this?

test/WebJobs.Extensions.DurableTask.Analyzers.Test/Orchestrator/DateTimeAnalyzerTests.cs
amdeel: That makes sense, considering these are all just strings.

amdeel: I'm going to save this suggestion for a future test refactor and not update the tests again in this PR.

test/WebJobs.Extensions.DurableTask.Analyzers.Test/Entity/DispatchEntityNameAnalyzerTests.cs
amdeel: This analyzer is checking that the identifier used in this type argument matches the function name, actually. The tests are just checking that it will flag the type argument (and not throw any exceptions) when the type argument is different SyntaxKinds.

src/WebJobs.Extensions.DurableTask.Analyzers/SyntaxNodeUtils.cs
ConnorMcMahon: In general, I do think that an extension method would probably be the best approach for this comparison of two things of the same type, as then it is clear which is the "object" and which is the "subject".

Issue #1183, opened 2020-01-29
Support long-running operations without forcing them to violate the HTTP specification


idg10:
The 202-based polling pattern that Durable Functions supports (both for consuming external HTTP endpoints, and also for presenting the status of Durable Functions in progress) is at odds with the HTTP specification, and also does not conform Microsoft's guidelines at https://github.com/Microsoft/api-guidelines/blob/vNext/Guidelines.md

We would like Durable Functions to be able to consume long-running operations that work in the way described in those guidelines. Currently, Durable Functions require an operation status endpoint to return a 202 status for as long as the operation is running. If the operation status endpoint returns a 200 status, Durable Functions interprets this as meaning that the operation is complete.

That is directly incompatible with section 13.2.5 of Microsoft's API guidelines at https://github.com/Microsoft/api-guidelines/blob/vNext/Guidelines.md#1325-operation-resource which say:

> The GET operation against an operation MUST return:
> 1. The operation resource, it's state, and any extended state relevant to the particular API.
> 2. 200 OK as the response code.

The reason we'd prefer to have long-running endpoints work the way that these guidelines recommend is that they are consistent with the HTTP specification in a way that the approach taken by Durable Functions is not.

Durable Functions expect a GET against the URL representing an operation to return a 202 if the operation is incomplete, but the problem with this, from an HTTP Specification point of view, is that the meaning of a 202 status at this point is *not* "the operation hasn't finished", it is "we haven't finished retrieving the operation's status". That's not how Durable Functions interprets it of course, but it's what a 202 status at this point implies based on what the HTTP Spec says.

Where a 202 *does* come into this is as the return code from the endpoints that *starts* the operation. In that context it makes sense, because the HTTP operation it describes is typically a POST asking to do a thing, and the 202 correctly indicates that the work to do the thing is underway but not complete, and the response can include a header saying where to find out more. (Opinion is divided on exactly what that header should be. Azure uses either of 2 different headers for this as described at https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-async-operations#monitor-status-of-operation while the Microsoft guidelines show a non-standard `Operation-Location` header in https://github.com/Microsoft/api-guidelines/blob/vNext/Guidelines.md#1327-the-typical-flow-polling and there's a proposal at https://github.com/Microsoft/api-guidelines/pull/101 to use the standard `Content-Location` instead; Durable Functions expects `Location`).

[...]

PR #1398, opened 2020-06-25
DurableHttpRequest Content exception during location polling

### LINKED ISSUE #1344: DurableHttpRequest Throws Exception When Content Is Supplied
### Description
After instantiating a DurableHttpRequest with content and calling CallHttpAsync an exception is thrown specifying a BadRequest was made.

### Expected behavior
The Content (JSON in this case) supplied to the initial DurableHttpRequest should not be passed to the new DurableHttpRequest created in the CreateLocationPollRequest method.

### Actual behavior
The Content from the original request is copied into this request that polls the Location URI returned 
in the header of the initial request. This causes an exception to the thrown and the Azure Function to terminate.

```csharp
private DurableHttpRequest CreateLocationPollRequest(DurableHttpRequest durableHttpRequest, string locationUri)
{
	DurableHttpRequest newDurableHttpRequest = new DurableHttpRequest(
		method: HttpMethod.Get,
		uri: new Uri(locationUri),
		headers: durableHttpRequest.Headers,
		content: durableHttpRequest.Content, // This is the line causing the issue
		tokenSource: durableHttpRequest.TokenSource);

	// Do not copy over the x-functions-key header, as in many cases, the
	// functions key used for the initial request will be a Function-level key
	// and the status endpoint requires a master key.
	newDurableHttpRequest.Headers.Remove("x-functions-key");

	return newDurableHttpRequest;
}
```

### Known workarounds
After cloning this repo and referencing it in my project instead of the NuGet package I commented out the line using the initial Content and published my Azure function. It no longer throws an exception and correctly waits until the task (a CopyDatabase operation in my case) is complete.

bachuv:
This change resolves #1344 . The location polling request is a GET request, so removing the content should be fine.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -292,7 +292,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 method: HttpMethod.Get,
                 uri: new Uri(locationUri),
                 headers: durableHttpRequest.Headers,
-                content: durableHttpRequest.Content,
                 tokenSource: durableHttpRequest.TokenSource);
 
             // Do not copy over the x-functions-key header, as in many cases, the


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -292,7 +292,6 @@ private DurableHttpRequest CreateLocationPollRequest(DurableHttpRequest durableH
                 method: HttpMethod.Get,
                 uri: new Uri(locationUri),
                 headers: durableHttpRequest.Headers,
```
cgillum, line 294: Maybe this can be a separate PR, but I wonder if we need to reduce the set of headers we send to the polling endpoint. I recall there were some issues with this that Connor had to fix at one point.

ConnorMcMahon: Yeah. I think the assumption we made when designing this feature was that the request would be identical to the initial request, just with a different URI.

I think it has been made clear that in practice this can lead to some problems. I'm not sure what the proper default polling request looks like honestly. I'm a bit afraid to make to many drastic changes do to it technically being a breaking change, but I think we can spin this as more of a bug fix. Once we tackle #1183, we will have given everyone a wide enough escape valve that if they don't like our defaults, they can use their own thankfully.

PR #1399, opened 2020-06-25
Swapping Microsoft.Azure.Services.AppAuthentication nuget dependency for Azure.Identity

### LINKED ISSUE #1232: Investigate swapping Microsoft.Azure.Services.AppAuthentication nuget dependency for Azure.Identity
Right now there is a lot of duplication in terms of Azure SDKs and the Azure team is trying to consolidate on a common set of SDKs. Right now we have a dependency on [Microsoft.Azure.Services.AppAuthentication](https://www.nuget.org/packages/Microsoft.Azure.Services.AppAuthentication/) for our Durable HTTP feature. However, the guidance we're hearing is that we should instead use [Azure.Identity](https://www.nuget.org/packages/Azure.Identity/).

Hopefully we can simply replace our usage of **Microsoft.Azure.Services.AppAuthentication** with **Azure.Identity**. We would want to make sure that it continues to work well with managed identities in Azure and the local user identity when running locally.

We could stick with **Microsoft.Azure.Services.AppAuthentication**, but there is some risk that this library could be deprecated.

**Additional context**
This is just an internal implementation detail and should not impact end-user apps.

bachuv:
Resolves #1232 by using DefaultAzureCredential from Azure.Identity instead of AzureServiceTokenProvider from Microsoft.Azure.Services.AppAuthentication.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
+++ b/src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
@@ -75,8 +75,8 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.DurableTask.AzureStorage" Version="1.7.6" />
     <PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="2.2.5" />
-    <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.0.3" />
     <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers" Version="0.2.3" />
+    <PackageReference Include="Azure.Identity" Version="1.1.1" />
   </ItemGroup>
 
   <!-- NuGet Publishing Metadata -->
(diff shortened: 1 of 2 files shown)


cgillum:
The code change looks good to me. I'll sign off once you are able to verify:
1. It works when running locally (using the current user's Azure identity)
2. It works in Azure when using managed identities.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -76,7 +76,7 @@
     <PackageReference Include="Microsoft.Azure.DurableTask.AzureStorage" Version="1.7.6" />
     <PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="2.2.5" />
     <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers" Version="0.2.3" />
-    <PackageReference Include="Azure.Identity" Version="1.1.1" />
+    <PackageReference Include="Azure.Identity" Version="1.2.0-preview.4" />
```
cgillum, line 79: We can't ship with a preview version of a dependency. Is this change to 1.2.0 required?

bachuv: No, it's not required. I changed it back to 1.1.1.

cgillum:
@ConnorMcMahon do you have any more feedback on this PR?

PR #1407, opened 2020-07-08
Adding DefaultAzureCredentialOptions to ManagedIdentityTokenSource()

### LINKED ISSUE #1279: Configure ManagedIdentityTokenSource to work with government clouds
This issue is pulled from a comment made in the ["Allow custom polling behavior in CallHttpAsync() for long-running endpoints"](https://github.com/Azure/azure-functions-durable-extension/issues/1183) issue.

The comments say:
"Finally, the `ManagedIdentityTokenSource` that you can provide doesn't let you override the `azureAdInstance` constructor for the underlying `AzureServiceTokenProvider`, which prevents you from using it against Azure Government."

bachuv:
These changes resolve #1279 by adding an optional `ManagedIdentityOptions` parameter to `ManagedIdentityTokenSource()`. `ManagedIdentityOptions` is a new type that is introduced with these changes.

The issue mentioned wanted the ability to configure `azureAdInstance` and `tenantId`. `ManagedIdentityOptions.AuthorityHost` is equivalent to `azureAdInstance` and `ManagedIdentityOptions.TenantId` is equivalent to `tenantId`.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
@@ -150,7 +150,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
                         if (!string.IsNullOrEmpty(optionsString))
                         {
-                            DefaultAzureCredentialOptions options = JsonConvert.DeserializeObject<DefaultAzureCredentialOptions>(optionsString);
+                            ManagedIdentityOptions options = JsonConvert.DeserializeObject<ManagedIdentityOptions>(optionsString);
                             return new ManagedIdentityTokenSource(resourceString, options);
                         }
 
--- a/src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
@@ -147,28 +147,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     {
                         string resourceString = (string)jsonObject.GetValue("resource", StringComparison.Ordinal);
 
-                        if (jsonObject.TryGetValue("options", out JToken optionsValue))
+                        if (jsonObject.TryGetValue("options", out JToken optionsToken))
                         {
-                            JObject optionsJObject = (JObject)optionsValue;
-                            string authorityHostValue = (string)optionsJObject.GetValue("authorityhost", StringComparison.Ordinal);
-                            string tenantIdValue = (string)optionsJObject.GetValue("tenantid", StringComparison.Ordinal);
-
-                            if (!string.IsNullOrEmpty(authorityHostValue) || !string.IsNullOrEmpty(tenantIdValue))
-                            {
-                                ManagedIdentityOptions managedIdentityOptions = new ManagedIdentityOptions();
-
-                                if (!string.IsNullOrEmpty(authorityHostValue))
-                                {
-                                    managedIdentityOptions.AuthorityHost = new Uri(authorityHostValue);
-                                }
-
-                                if (!string.IsNullOrEmpty(tenantIdValue))
-                                {
-                                    managedIdentityOptions.TenantId = tenantIdValue;
-                                }
-
-                                return new ManagedIdentityTokenSource(resourceString, managedIdentityOptions);
-                            }
+                            ManagedIdentityOptions managedIdentityOptions = optionsToken.ToObject<JObject>().ToObject<ManagedIdentityOptions>();
+                            return new ManagedIdentityTokenSource(resourceString, managedIdentityOptions);
                         }
 
                         return new ManagedIdentityTokenSource(resourceString);
@@ -198,24 +180,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     writer.WritePropertyName("resource");
                     writer.WriteValue(tokenSource.Resource);
 
-                    if (tokenSource.Options != null && ((tokenSource.Options.AuthorityHost != null) || (!string.IsNullOrEmpty(tokenSource.Options.TenantId))))
+                    if (tokenSource.Options != null)
                     {
                         writer.WritePropertyName("options");
-                        writer.WriteStartObject();
-
-                        if (tokenSource.Options.AuthorityHost != null)
-                        {
-                            writer.WritePropertyName("authorityhost");
-                            writer.WriteValue(tokenSource.Options.AuthorityHost);
-                        }
-
-                        if (!string.IsNullOrEmpty(tokenSource.Options.TenantId))
-                        {
-                            writer.WritePropertyName("tenantid");
-                            writer.WriteValue(tokenSource.Options.TenantId);
-                        }
-
-                        writer.WriteEndObject();
+                        writer.WriteRawValue(JsonConvert.SerializeObject(tokenSource.Options));
                     }
 
                     writer.WriteEndObject();
--- a/src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
@@ -146,7 +146,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                         tokenSourceKind == TokenSourceType.AzureManagedIdentity)
                     {
                         string resourceString = (string)jsonObject.GetValue("resource", StringComparison.Ordinal);
-                        string optionsString = (string)jsonObject.GetValue("defaultazurecredentialoptions", StringComparison.Ordinal);
+                        string optionsString = (string)jsonObject.GetValue("options", StringComparison.Ordinal);
 
                         if (!string.IsNullOrEmpty(optionsString))
                         {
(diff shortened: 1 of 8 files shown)


src/WebJobs.Extensions.DurableTask/ManagedIdentityTokenSource.cs
```diff
@@ -34,13 +36,19 @@ public ManagedIdentityTokenSource(string resource)
...
+        public DefaultAzureCredentialOptions Options { get; }
+
         /// <inheritdoc/>
         public async Task<string> GetTokenAsync()
         {
             var scopes = new string[] { this.Resource };
             TokenRequestContext context = new TokenRequestContext(scopes);
 
-            DefaultAzureCredential defaultCredential = new DefaultAzureCredential();
+            DefaultAzureCredential defaultCredential = new DefaultAzureCredential(this.Options);
```
cgillum, line 51: Is a `null` value allowed for this constructor? It might be safer to do something like this:
```suggestion
            DefaultAzureCredential defaultCredential = this.Options != null ?
                new DefaultAzureCredential(this.Options) :
                new DefaultAzureCredential();
```

src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
```diff
@@ -143,7 +145,16 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
                     if (Enum.TryParse((string)kindValue, out TokenSourceType tokenSourceKind) &&
                         tokenSourceKind == TokenSourceType.AzureManagedIdentity)
                     {
-                        return new ManagedIdentityTokenSource((string)jsonObject.GetValue("resource", StringComparison.Ordinal));
+                        string resourceString = (string)jsonObject.GetValue("resource", StringComparison.Ordinal);
+                        string optionsString = (string)jsonObject.GetValue("defaultazurecredentialoptions", StringComparison.Ordinal);
```
cgillum, line 149: Could we just use "options" as the name instead?

cgillum:
Sorry, one more thing: can you update the the [CallHttpActionOrchestrationWithManagedIdentity](https://github.com/Azure/azure-functions-durable-extension/blob/aa6676b577c749bdbfe101adfc9c0b802416eb1d/test/FunctionsV2/OutOfProcTests.cs#L102) unit test to cover the "options" setting?

bachuv:
`DurableAzureCredentialOptions.Transport` is the type of `HttpPipelineTransport` and defaults to the instance of `HttpClientTransport`. Users could configure `DefaultAzureCredentialOptions` and introduce a possible `HttpClient` like in the code below which sets up `DefaultAzureCredentialOptions`:

```
HttpClientHandler handler = new HttpClientHandler();
handler.Credentials = new NetworkCredential(username, password);

DefaultAzureCredentialOptions options = new DefaultAzureCredentialOptions();
options.Transport = new HttpClientTransport(new HttpClient(handler));
```
Supporting `DefaultAzureCredentialOptions.Transport` introduces nested serialization and if it's created in an orchestrator method then it introduces this warning: `System.Net.Http.HttpClientHandler violates the orchestrator deterministic code constraint`.

Is it possible to not support `DefaultAzureCredentialOptions.Transport`? Another option could be to create a wrapper class that doesn’t support `DefaultAzureCredentialOptions.Transport`. This would be like `DurableHttpRequest` and `DurableHttpResponse` for the Durable HTTP feature.

@cgillum @ConnorMcMahon Any thoughts on how to handle `DefaultAzureCredentialOptions.Transport`?

cgillum:
Hmm...this is problematic. There is no way we can safely serialize all these fields (especially not `Transport`). I considered using a custom serializer to serialize only the things that are known to be serializable, but then that could lead to customer confusion because we're not preserving all the properties they set.

The safest approach might be to create our own type instead of relying on `DefaultAzureCredentialOptions`. We could call it `ManagedIdentityOptions` or something similar, and expose our own subset of `DefaultAzureCredentialOption`'s properties. I think we would start by supporting in the properties required to solve #1279, like `AuthorityHost` and `ManagedIdentityClientId` (I would prefer to shorten it to just `ClientId`) to support user-assigned identities. If we get customers asking for more fields, we can add them later. On the HTTP activity side, we could use this data and initialize own own private `DefaultAzureCredentialsOptions` object from the values we get from our deserialized `ManagedIdentityOptions` field.

Having our own type could also simplify documentation and cross-language support.

@bachuv @ConnorMcMahon thoughts?

bachuv:
@cgillum I agree with that approach. It would definitely make documentation and serialization easier. I can create the new type and add the 2 fields that were mentioned in #1279.

src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
```diff
@@ -143,7 +145,33 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
                     if (Enum.TryParse((string)kindValue, out TokenSourceType tokenSourceKind) &&
                         tokenSourceKind == TokenSourceType.AzureManagedIdentity)
                     {
-                        return new ManagedIdentityTokenSource((string)jsonObject.GetValue("resource", StringComparison.Ordinal));
+                        string resourceString = (string)jsonObject.GetValue("resource", StringComparison.Ordinal);
```
ConnorMcMahon, line 148: Are these changes really necessary? It seems like your custom `ManagedIdentityOptions` should serialize/deserialize correctly already?

src/WebJobs.Extensions.DurableTask/ManagedIdentityTokenSource.cs
```diff
@@ -34,13 +36,40 @@ public ManagedIdentityTokenSource(string resource)
...
             var scopes = new string[] { this.Resource };
             TokenRequestContext context = new TokenRequestContext(scopes);
 
-            DefaultAzureCredential defaultCredential = new DefaultAzureCredential();
+            DefaultAzureCredential defaultCredential;
+            DefaultAzureCredentialOptions defaultAzureCredentialOptions = new DefaultAzureCredentialOptions();
+
+            if (this.Options != null && ((this.Options.AuthorityHost != null) || (!string.IsNullOrEmpty(this.Options.TenantId))))
+            {
+                if (this.Options.AuthorityHost != null)
```
ConnorMcMahon, line 56: It seems like the body of this if statement could just be the code used, and we don't need the surrounding if-else statement? Essentially we use the default credential options and only edit them if the fields on this.Options are populated?

The only difference would be needing to stick `?.` operators after `this.Options`.

src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs
bachuv: These changes are there to correctly serialize/deserialize the json given from out of process requests. I followed the json format from the tests in OutOfProcTests.cs (shown below). Let me know if there's a better way to format the `options` in the out of process request to avoid these changes.

```
var executionJson = @"
{
    ""isDone"": false,
    ""actions"": [
        [{
            ""actionType"": ""CallHttp"",
            ""httpRequest"": {
                ""method"": ""GET"",
                ""uri"": ""https://example.com"",
                ""tokenSource"": {
                    ""kind"": ""AzureManagedIdentity"",
                    ""resource"": ""https://management.core.windows.net"",
                    ""options"": {
                        ""authorityhost"": ""https://login.microsoftonline.com/"",
                        ""tenantid"": ""example_tenant_id""
                    }
                }
            }
        }]
    ]
}";
```

src/WebJobs.Extensions.DurableTask/ManagedIdentityTokenSource.cs
```diff
@@ -51,25 +51,18 @@ public async Task<string> GetTokenAsync()
...
-
-                defaultCredential = new DefaultAzureCredential(defaultAzureCredentialOptions);
+                defaultAzureCredentialOptions.AuthorityHost = this.Options.AuthorityHost;
             }
-            else
+
+            if (!string.IsNullOrEmpty(this.Options?.TenantId))
             {
-                defaultCredential = new DefaultAzureCredential();
+                defaultAzureCredentialOptions.InteractiveBrowserTenantId = this.Options.TenantId;
```
cgillum, line 61: I'm only noticing this for the first time now, but what affect does setting `InteractiveBrowserTenantId` have in a Durable HTTP context? The property name seems to imply the interactive use of a browser, which wouldn't apply here. I noticed there is also a [SharedTokenCacheTenantId](https://docs.microsoft.com/en-us/dotnet/api/azure.identity.defaultazurecredentialoptions.sharedtokencachetenantid?view=azure-dotnet#Azure_Identity_DefaultAzureCredentialOptions_SharedTokenCacheTenantId) property. How is that one different from our perspective?

bachuv: In the linked issue #1279, a customer asked for access to set the `tenantId` through `AzureServiceTokenProvider` (the current method to get tokens). In that context, `tenantId` refers to the customer's AAD tenantId.

In the docs for `DefaultAzureCredentialOptions`, it says that `InteractiveBrowserTenantId` can be set through the `AZURE_TENANT_ID` environment variable. Since `AZURE_TENANT_ID` is equivalent to an AAD tenantId, it seems like the better option compared to `SharedTokenCacheTenantId`. I think the name `InteractiveBrowserTenantId` just isn't very clear.

test/Common/DurableHttpTests.cs
```diff
@@ -120,6 +121,106 @@ public void DeserializeCallActivity()
...
+            MockTokenSource mockTokenSource = new MockTokenSource("dummy token", options);
+
+            TestDurableHttpRequest request = new TestDurableHttpRequest(
+                httpMethod: HttpMethod.Get,
+                headers: headers,
+                tokenSource: mockTokenSource);
+
+            string serializedTestDurableHttpRequest = JsonConvert.SerializeObject(request);
+
+            Assert.Contains("\"authorityhost\":\"https://dummy.login.microsoftonline.com/", serializedTestDurableHttpRequest);
```
ConnorMcMahon, line 145: For slightly more rigorous testing, we could build up the whole expected JSON object as a string, and do something like `Assert.True(JToken.DeepEquals(JObject.Parse(expectedJson), JObject.Parse(serializedTestDurableHttpRequest))`

Issue #1000, opened 2019-10-29
Provide unit testing instructions for interfaces in Durable 2.x


hunyhabib:
### Description
while trying to setup a mock using MOQ with IDurableOrchestrationContext I get the below exception: 

```
Unsupported expression: it => it.CallActivityAsync(It.IsAny<string>(), It.IsAny<object>())
    Extension methods (here: DurableContextExtensions.CallActivityAsync) may not be used in setup / verification expressions.
  Stack Trace: 
    Guard.IsOverridable(MethodInfo method, Expression expression)
    InvocationShape.ctor(LambdaExpression expression, MethodInfo method, IReadOnlyList`1 arguments, Boolean exactGenericTypeArguments)
    ExpressionExtensions.<Split>g__Split|4_1(Expression e, Expression& r, InvocationShape& p)
    ExpressionExtensions.Split(LambdaExpression expression)
    Mock.GetMatchingInvocationCount(Mock mock, LambdaExpression expression, List`1& invocationsToBeMarkedAsVerified)
    Mock.Verify(Mock mock, LambdaExpression expression, Times times, String failMessage)
    Mock`1.Verify[TResult](Expression`1 expression, Func`1 times)
```

### Expected behavior
no exception to be thrown and mock get init correctly during setup

### Actual behavior
Mentioned exception has been thrown

### Relevant source code snippets
```csharp
  durableOrchestrationContextMock = new Mock<IDurableOrchestrationContext>();
            durableOrchestrationContextMock
                .Setup(x => x.CallActivityAsync(It.IsAny<string>(), It.IsAny<BaseJobInfo>()))
                .Returns(Task.CompletedTask);
```

### App Details

- **Durable Functions extension version 1.0.0**:
- **Microsoft.Azure.WebJobs.Extensions.DurableTask version 2.0.0-beta3**:
- **Programming language used**: c#

ConnorMcMahon:
@hunyhabib 

This is an unfortunate side affect of our new design in Durable Functions 2.0. We moved from abstract classes to interfaces, and in order to add support for many of our overloads that just call other core methods, we had to use extension methods instead of virtual methods with a default implementation.

This works well until Moq comes into the scene. With our old approach, since the methods were virtual, Moq was able to override the implementation. However, with extension methods, they are simply syntactic sugar for static methods, so there is no way for Moq to override them.

[...]

PR #1422, opened 2020-08-04
Improve unit testability of interfaces

ConnorMcMahon:
Our current approach to default paramter passing for method overloads on
our interfaces used extension methods. Extension methods have the
downside of not being mockable, especially by tooling such as Moq. This
means that customers had to find out which actual interface method was
being called by their method overload in order to use these tools, which
is unecessarily complicated and this burden should not be placed on our
customers.

Since we have only one implementation of each of these interfaces, we
gain little to no benefit to having these method overloads as extension
methods for "default implementations". By making all of these method
overloads explicit interface methods, and moving their implementations
to the implementing classes, they all become mockable and customers can
write unit tests as they expect to be able to.

If we encourage all of our customers to use Moq for unit testing after
this change, we can then add new interface methods, as they effectively
will no longer be breaking changes.

Address #1000


src/WebJobs.Extensions.DurableTask/EntityScheduler/Proxy/DurableEntityProxyExtensions.cs
```diff
@@ -15,155 +15,7 @@ public static class DurableEntityProxyExtensions
...
-        /// <param name="scheduledTimeUtc">The time at which to start the operation.</param>
-        /// <param name="operation">A delegate that performs the desired operation on the entity.</param>
-        /// <typeparam name="TEntityInterface">Entity interface.</typeparam>
-        public static void SignalEntity<TEntityInterface>(this IDurableEntityContext context, EntityId entityId, DateTime scheduledTimeUtc, Action<TEntityInterface> operation)
-        {
-            operation(EntityProxyFactory.Create<TEntityInterface>(new EntityContextProxy(context, scheduledTimeUtc), entityId));
-        }
-
-        private static string ResolveEntityName<TEntityInterface>()
+        public static string ResolveEntityName<TEntityInterface>()
```
ConnorMcMahon, line 18: This class should probably be renamed now.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/DurableContextExtensions.cs
```diff
@@ -18,544 +10,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
...
-        /// <param name="showHistory">Boolean marker for including execution history in the response.</param>
-        /// <returns>Returns a task which completes when the status has been fetched.</returns>
-        public static Task<DurableOrchestrationStatus> GetStatusAsync(this IDurableOrchestrationClient client, string instanceId, bool showHistory)
-        {
-            return client.GetStatusAsync(instanceId, showHistory, showHistoryOutput: false, showInput: true);
-        }
-
         /// <summary>
         /// Returns an instance of ILogger that is replay safe, ensuring the logger logs only when the orchestrator
         /// is not replaying that line of code.
```
ConnorMcMahon, line 15: I left this method as an extension, as I saw no reason for customers to moq it.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/DurableContextExtensions.cs
```diff
@@ -18,544 +10,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
...
-        /// </summary>
-        /// <param name="client">The client object.</param>
-        /// <param name="instanceId">The ID of the orchestration instance to query.</param>
-        /// <param name="showHistory">Boolean marker for including execution history in the response.</param>
-        /// <returns>Returns a task which completes when the status has been fetched.</returns>
-        public static Task<DurableOrchestrationStatus> GetStatusAsync(this IDurableOrchestrationClient client, string instanceId, bool showHistory)
-        {
-            return client.GetStatusAsync(instanceId, showHistory, showHistoryOutput: false, showInput: true);
-        }
-
```
cgillum, line 558: Seeing all this deleted extension method code is making me very happy. 😊

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityClient.cs
```diff
@@ -43,6 +43,44 @@ public interface IDurableEntityClient
...
+
+        /// <summary>
+        /// Signals an entity to perform an operation, at a specified time.
+        /// </summary>
+        /// <typeparam name="TEntityInterface">Entity interface.</typeparam>
+        /// <param name="entityId">The target entity.</param>
+        /// <param name="scheduledTimeUtc">The time at which to start the operation.</param>
+        /// <param name="operation">A delegate that performs the desired operation on the entity.</param>
+        /// <returns>A task that completes when the message has been reliably enqueued.</returns>
+        Task SignalEntityAsync<TEntityInterface>(EntityId entityId, DateTime scheduledTimeUtc, Action<TEntityInterface> operation);
```
cgillum, line 82: Have you considered using optional parameters and creating a smaller number of method overloads for things like this? I feel like it would be much easier for a developer to mock a single method rather than four. It also reduces our documentation burden. Is this an option we can safely consider given we're already planning on making a breaking change to these interfaces?

This question applies to all the interfaces you're updating (entity client, orchestration client, orchestration context, etc.)

src/WebJobs.Extensions.DurableTask/EntityScheduler/Proxy/DurableEntityProxyExtensions.cs
cgillum: Would it make sense to combine it with our other extension method class so that we have just one?

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityClient.cs
ConnorMcMahon: So I considered handling default parameters on this interfaces, but I remember running into some oddness around default parameters with interfaces in the past. [This article](https://michael-mckenna.com/optional-arguments-in-c-sharp-interfaces/) explains why they can be dangerous. Though I did write all of these as explicit interface implementations, so it may not be an issue...

Even if it is an issue, since we only have one implementation, I think we could get away with it as long as we wrote a test that would catch these kind of errors using reflection.

ConnorMcMahon: So initial investigations makes it look like this should not be a problem for our production code at least.

That being said, [another issue](https://stackoverflow.com/questions/12413583/an-expression-tree-may-not-contain-a-call-or-invocation-that-uses-optional-argum) popped up when I tested out using default parameters with Moq. Essentially, since optional parameters are compile time attributes, Moq can't handle them. The workaround is for customers to use `It.IsAny<T>()` in their `.Setup()` expression trees for all of the optional parameters, even if they don't use them in their production code. 

It is a bit more annoying as a test writer, but it is still a vast improvement to the previous problem, as intellisense gives you fairly clear instructions on how to fix the `.Setup()`.

So essentially the tradeoff is between:

1. Explicitly have all overloads as interface methods, and maintain the documentation + explicit interface implementations for them, in order to maximize test writeability.
2. Simplify our burden by using parameters, and make test writers explicitly handle the optional parameters in their Setup operations with help from intellisense.

I don't have strong opinions about which set of trade offs is better in this case.

cgillum: If you're open to either option, I would cast my vote for (2) to help us reduce our maintenance burdens.

ConnorMcMahon: Also worth noting is that this makes adding a new optional parameter break customer tests, where has having them as an their own interface method with an implementation means customers using Moq won't be broken by adding that new method.

Though, it's fair to argue that adding optional parameters to these interfaces causes bloat to our API surface that would better be handled by some data object that contains all of the optional parameters, and that we should be hesitant about adding them regardless...

cgillum: *sigh*

If I understand correctly, you're saying this will basically break _everyone's_ test code since their existing Moq code would be incompatible? In that case, I think we have to go with your current approach.

But honestly, I'm already convinced we need to just move back to abstract classes to avoid all these problems with interfaces. I think we can do this without introducing any breaking changes - just have them implement the interfaces but have all future updates only target the abstract classes (and update all documentation to use the abstract classes instead of interfaces). That's how I would prefer we evolve our APIs moving forward. This interface stuff is becoming too difficult to deal with IMO.

ConnorMcMahon: I don't think it would break most people's test code because the current interface method that customer's can mock is already the version with all the optional parameters provided, which is what they would have to mock with the default parameter approach. I would have to verify that is the case for most/all methods we would use default parameters.

Essentially what we are doing by making some of the parameters optional on the interface is we are giving them intellisense guidance when writing their Moq setup expressions; right now, they get runtime failures, or if they use a Moq analyzer, get compile time errors but they don't know which of the method overloads is the correct one to Moq.

I am not opposed to going back to abstract class implementations moving forward, but I think that abstract classes would have similar issues with Moq and optional parameters moving forward.

Essentially if we did something like adding an optional cancellation token to an API (an example I know we have done in the past), it would break Moq consumers even if it was a virtual method on an abstract class.

src/WebJobs.Extensions.DurableTask/EntityScheduler/Proxy/DurableEntityProxyExtensions.cs
ConnorMcMahon: I renamed the class, as it didn't make sense to put it in the extension method class since this isn't an extension method.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityClient.cs
cgillum: Looking at this a second time, I'm realizing that my original interpretation of this change and the current state of the world was completely backwards! As you mentioned, the interface _already_ defined the most generic overload and used optional parameters. Therefore this is non-breaking for existing Moq-based test code. I think I now understand that the point of this PR is to add all the extension methods as explicit interface methods so the users can actually mock the overloads that they're using.

PR #1431, opened 2020-08-12
Added AppLease settings for GeoRedundancy feature

### LINKED ISSUE #1128: Support for Geo-redundant Application
We want to deploy our application in multiple regions to support failover incase of region failure. 

**Approach**
Deploy application to multiple regions and use same TaskHub across all regions and make sure only one application is running at any given point in time. Turn off always on flag in the application so that the application wont pick up orchestrations as soon an the region recovers. Once a regions fails, start the application in backup region so that all the orchestrations in-flight will be picked up by backup application.

**Problem**
The application is picking up orchestrations from TaskHub as soon as it starts even before any http requests are received with always on flag off. 

**Repro-steps** 
(All the below steps are done using Azure Portal)
1. Turn of always on flag and start the application and create an orchestration
2. when the orchestration is running, stop the application, wait for few minutes and start the application
3. Ideally, the application is not supposed to pickup in-flight orchestrations from taskhub until it receives a http request, but I see that the application is picking up all orchestrations from taskhub

**Proposed Solution** 
Feature to make sure that an orchestration can only be picked up by one instance of application even when multiple instances are running and using the same taskhub.

amdeel:
resolves #1128 

Added the DurableTaskOptions for the AppLease feature that allows for Geo-Redundant applications for backup in case of failure.

DTFx changes: https://github.com/Azure/durabletask/pull/432


src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -195,6 +197,45 @@ public string HubName
...
+        /// If true, takes a lease on the task hub container, allowing for only one app to process messages in a task hub at a time.
+        /// </summary>
+        public bool UseAppLease { get; set; } = true;
+
+        /// <summary>
+        /// If UseAppLease is true, gets or sets the AppLeaaseOptions used for acquiring the lease to start the application.
+        /// </summary>
+        public AppLeaseOptions AppLeaseOptions { get; set; } = AppLeaseOptions.DefaultOptions;
+
+        internal string AppName
```
cgillum, line 210: We actually already have an AppName property defined [here](https://github.com/Azure/azure-functions-durable-extension/blob/ffa4e192b25984b9381bf32ce52fe01ff25b21e7/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs#L29). Should we just reuse that so that we don't have redundant settings?

amdeel: I thought at the time that the AppName shouldnt be an empty string, but I can just use this instead.

cgillum: If the current app name is an empty string and if that isn't valid, maybe we can use `Environment.MachineName` instead when assigning a value for `AzureStorageOrchestrationServiceSettings.AppName`?

ConnorMcMahon: I can't think of any Antares scenario where the environment variable wouldn't be present, but machine name seems reasonable default for those running outside of Antares.

amdeel: That's what I'm currently doing in this PR. But I did run a test to make sure an empty string was fine. It turns out it is.

ConnorMcMahon: Hmm,since empty string works, My preference would be to use that AppName property.

PR #1436, opened 2020-08-17
Updated DurableClient ConnectionName matching function and EntityKey validity checks

### LINKED ISSUE #1425: EntityKey is not being checked for illegal characters
When trying to create an `EntityKey` with an illegal character (which is not a valid character to use in the `PartitionKey` of Table Storage) the entity fails to persist. No exceptions are thrown which is incredibly frustrating as it took me hours to determine the root cause.

looks like the fix was introduced for `Orchestrations`, but never for `Entities` #601 

**Repro**
I was trying to create an `EntityKey` which contained a DateTime respresentation, i.e. 2020/08/07 <- the forward slashes are illegal.

amdeel:
Moved the logic to determine if the connectionName matches to the DurabilityProvider.

Added a check to SignalEntity to check the EntityKey for validity.
resolves #1425

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
@@ -408,5 +408,15 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             errorMessage = null;
             return true;
         }
+
+        /// <summary>
+        ///  Returns true if the stored connection string, ConnectionName, matches the input DurabilityProvider ConnectionName.
+        /// </summary>
+        /// <param name="durabilityProvider">The DurabilityProvider used to check for matching connection string names.</param>
+        /// <returns>A boolean indicating whether the connection names match.</returns>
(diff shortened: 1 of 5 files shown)


src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
```diff
@@ -408,5 +408,15 @@ public virtual bool ValidateDelayTime(TimeSpan timespan, out string errorMessage
             errorMessage = null;
             return true;
         }
+
+        /// <summary>
+        ///  Returns true if the stored connection string, ConnectionName, matches the input DurabilityProvider ConnectionName.
+        /// </summary>
+        /// <param name="durabilityProvider">The DurabilityProvider used to check for matching connection string names.</param>
+        /// <returns>A boolean indicating whether the connection names match.</returns>
+        public virtual bool ConnectionNameMatches(DurabilityProvider durabilityProvider)
```
cgillum, line 417: Any particular reason for making this public instead of internal? I'd prefer to make it internal, if possible, since it doesn't look like any outside code needs direct access to this method.

PR #1440, opened 2020-08-18
Updated default JsonSerializerSetting to include DateParseHandling.None

amdeel:
This change prevents strings that look like dates from being serialized into DateTime objects.

related to Azure/azure-functions-durable-js#208

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActivityContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActivityContext.cs
@@ -62,7 +62,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     throw new ArgumentException("The serialized input is expected to be an object array with one element.");
                 }
 
-                this.parsedJsonInput = this.messageDataConverter.ConvertToJToken(this.messageDataConverter.Serialize(objectArray[0]));
+                this.parsedJsonInput = MessagePayloadDataConverter.ConvertToJToken(this.messageDataConverter.Serialize(objectArray[0]));
             }
 
             return this.parsedJsonInput;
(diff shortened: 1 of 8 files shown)


afrancoc2000:
Hi, I know this is closed and solved but I'm still getting the issue, my orchestator sends an object like this to the activity: 

```
[[{
    "fechaMedida":"2020-07-23T00:00:00-05:00",
    "tipoMedicion":"medida"
}]]
```

And I get an object like this in the activity:

```
[[{
    "fechaMedida":"2020-07-23T05:00:00+00:00"
    "tipoMedicion":"medida"
}]]

```
I have "version": "[2.*, 3.0.0)" in the hosts.json for the extension bundle that in theory uses this version but I'm still getting the error, do I need to do something special to get this fix?

Thanks

ConnorMcMahon:
Last I checked, some apps using extension bundles still had 2.2.2 as their extension version, which is several versions out of date. We are working to get that time to extension bundle deployment far quicker.

If you can share the following information I can verify if that is the case:
App Name **or** Orchestration Instance ID:
Region:
Timestamp (UTC):

afrancoc2000:
Hi, you know I also tried it downloading the nugets directly using "func extensions install" and checked the version 2.4.1 was there and the problem was the same in my localhost, my app name is fapp-epm-np-agpe-bch-dllo, region East US and it runned yesterday all day long so 2021-03-12T23:26:00.471 must do it.

davidmrdavid:
Hi @afrancoc2000,
Thank you for reaching out! Do you mind posting your question in our _issues_ board here: https://github.com/Azure/azure-functions-durable-extension/issues

That would help us focus on your question without the clutter of the previous comments in this thread. Feel free to tag me in there. Thank you! ⚡ ⚡

PR #1448, opened 2020-08-24
Update samples and XML doc comments

cgillum:
This is for the v2.3.0 release. I looked through all our sample projects and made sure they were up-to-date in terms of the versions of our nuget packages they use.

Related to https://github.com/Azure/azure-functions-durable-extension/issues/1444

### DIFF
--- a/samples/VSSample.Tests/VSSample.Tests.csproj
+++ b/samples/VSSample.Tests/VSSample.Tests.csproj
@@ -1,12 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.2</TargetFramework>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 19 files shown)


samples/VSSample.Tests/VSSample.Tests.csproj
```diff
@@ -1,12 +1,12 @@
...
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.2</TargetFramework>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="2.2.0" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="2.3.*" />
```
ConnorMcMahon, line 9: Do we want to encourage people to use lose package reference versioning like this?

cgillum: It makes it easier to stay up-to-date with the latest bug fixes, so yes, I think we should encourage folks to do this.

PR #1468, opened 2020-09-09
Doc update contribution from @donald93

scbedd:
[PR for context](https://github.com/Azure/azure-docs-sdk-dotnet/pull/1286)

A user contributed the following. Unfortunately the reference content is 1-way, so I need to make the fix here.


ConnorMcMahon (approved):
LGTM!

Issue #1463, opened 2020-09-04
Enable durable entity scheduled signals > 7 days


jeffhollan:
With the 2.3 release, Durable Timers can now be set for longer than 7 days.  However, it seems that durable entities cannot schedule a signal > 7 days.  I assumed they would be the same, but this appears not to be the case.

The behavior I see now is that if I try to schedule > 7 days, I don't get any errors but the scheduling doesn't go through, and appears to put the entity in a bit of a bad state where future signals are all buffered.  If I schedule for < 7 days I see it is scheduled as expected.

Hoping it's a trivial fix to make it so the > 7 day feature can be used by entities as well.

ConnorMcMahon:
Adding @bachuv. My understanding is that the approach we use for orchestrations will not work for entities. We essentially use replay behavior to break 7+ day timers in 2.3.0 into multiple timers behind the scenes, and we just handle them intelligently as a part of the replay.

For entities, I don't believe we have any replays, as each entity execution is a fresh "orchestration execution", so I don't know if this same approach would work. 

At the very least, we should be able to properly throw errors for > 7 day executions in entities so that customers don't silently get into a bad state.

sebastianburckhardt:
I can take a look and see if I can modify the entity message processing logic so it breaks > 7 day schedules into shorter steps.

sebastianburckhardt:
This has been implemented and merged.

PR #1475, opened 2020-09-14
Support scheduling entity signals without limit on duration

sebastianburckhardt:
Addresses issue #1463.

If scheduling signals beyond what the durability provider can take responsibility for, the signal is scheduled for a shorter duration. When receiving a signal ahead of the time it is scheduled to execute, it is then rescheduled (instead of being processed). This keeps going until the scheduled time has been reached. 

This implementation uses the same parameters (DurabilityProvider.MaximumDelayTime and DurabilityProvider.LongRunningTimerIntervalLength) as the corresponding functionality for Timers.


ConnorMcMahon (approved):
LGTM. Let's wait to merge until we get a green CI run.

PR #1482, opened 2020-09-15
Structured Logging for Linux

davidmrdavid:
**WIP**: _Do not merge, here just to get feedback and pointers on next steps_

This PR implements a simple, perhaps naive, `ILogger` and `ILoggerProvider` to write JSON-like logs in Linux. Another class, `SystemEnviroment` helps to read enviroment variables to determine if we're in Linux Consumption or Linux Dedicated.

Unfortunately, as I'm unfamiliar with .NET patterns, I'm unsure how to combine these pieces together to utilize the [new structured logging backend](https://github.com/Azure/durabletask/pull/425). So this _draft_ PR is here to help clarify that in my mind.

Below are the next steps, in my mind. _Please let me know if those are in the right track._
**(1)** Attach the new logging  provider (`LinuxLoggerProvider`) to the LoggerFactory here: https://github.com/Azure/azure-functions-durable-extension/blob/ffa4e192b25984b9381bf32ce52fe01ff25b21e7/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs#L102-L120
**(1.a)** I suppose we only want to attach the new logging provider if we know to be in a linux enviroment, is that right?

**(2)** In my mind, we'd somehow be able to _listen_ all these ETW write events (preview below):
https://github.com/Azure/azure-functions-durable-extension/blob/c01a22628685b4d427243552ab3f9d4e03076266/src/WebJobs.Extensions.DurableTask/EtwEventSource.cs#L20-L50

...and then get a JSON object representing those write events, which we can turn into a string and call `LinuxLogger.Log` on that string. I'm just unsure as to how to do those two steps cleanly: listening and getting the JSON from there. _Any thoughts?_. I remember the ILogger abstraction supported EventSource events by default, but I couldn't find much information on its usage so I can't think of a clean, ILogger-friendly, solution. My expectation is that we'll refactor the `EtwEventSource` class to use the ILogger abstraction and that the pieces should fall from there.

I think that's all for now. Looking forward to hearing from y'all!


src/WebJobs.Extensions.DurableTask/LinuxLogger.cs
```diff
@@ -0,0 +1,54 @@
+﻿using Microsoft.Extensions.Logging;
```
cgillum, line 1: Let's add the copyright header to these files.

src/WebJobs.Extensions.DurableTask/LinuxLogger.cs
```diff
@@ -0,0 +1,54 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    public class LinuxLogger : ILogger
```
cgillum, line 9: Since this is a platform-specific thing, I would suggest naming it `AppServiceLinuxLogger` or something similar.

src/WebJobs.Extensions.DurableTask/LinuxLogger.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        private readonly bool inConsumption;
+        private readonly string consolePrefix = "MS_DURABLE_FUNCTION_EVENTS_LOGS";
+        private readonly string loggingPath = "/var/log/functionsLogs/durableevents.log";
```
cgillum, line 13: These bottom two can be made into `const` variables, i.e. 
```csharp
private const string ConsolePrefix = "...";
private const string FileLoggingPath = "...";
```

src/WebJobs.Extensions.DurableTask/LinuxLogger.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public LinuxLogger(bool inConsumption)
+        {
+            this.inConsumption = inConsumption;
+        }
+
+        public IDisposable BeginScope<TState>(TState state)
+        {
+            // TODO: I this we can leave this unimplemented
+            throw new NotImplementedException();
```
cgillum, line 23: It might not be a good idea to throw here because some code we don't control could theoretically invoke this and cause issues. We might want to do some exploratory testing and see whether it functions correctly if you just return `null`.

src/WebJobs.Extensions.DurableTask/LinuxLogger.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        public bool IsEnabled(LogLevel logLevel)
+        {
+            // TODO: assuming these logs should never be disabled
+            return true;
```
cgillum, line 29: I would say we only want it enabled if we're on an App Service Linux platform. But then again, the other way you could do it (and maybe you are, I haven't gone through the whole PR yet) is to just not add this logger at all if we're not on App Service Linux.

src/WebJobs.Extensions.DurableTask/LinuxLogger.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+        {
+            string jsonString = ""; // assume that, somehow, this becomes a string-representation of a JSON-representation of an ETW event....
+            // @Chris, any ideas on how to set the right value here? I suppose we'll need to be listening to ETW events as well and transform them into a JSON-string here.
```
cgillum, line 35: I think we might be able to do something like this:
```csharp
JObject json = JObject.FromObject(state); // the state object is basically a dictionary of key-value log pairs
// TODO: maybe add some fields to the JSON object?
string jsonString = json.ToString(Formatting.None);
```

src/WebJobs.Extensions.DurableTask/LinuxLogger.cs
```diff
@@ -0,0 +1,54 @@
...
+                // In Linux Consumption, we write to console
+                string consoleLine = this.consolePrefix + " " + jsonString;
+                Console.WriteLine(consoleLine);
+            }
+            else
+            {
+                // In Linux Dedicated, we write to file
+                using (var streamWriter = new StreamWriter(this.loggingPath, true))
+                {
+                    streamWriter.WriteLine(jsonString);
```
cgillum, line 47: Synchronous writes to files worry me a little. We might need to use `WriteLineAsync` instead to avoid blocking I/O. However, the crazy side of me wants to do it in a way that is _completely_ asynchronous - i.e. don't even await. Something like:
```csharp
Task unused = streamWriter.WriteLineAsync(jsonString);
```

Though I have some questions about whether this will cause problems if there is an exception writing to a file. I would need to consult with the latest .NET literature to understand how the CLR handles exceptions for unobserved tasks (they used to crash the process, but I think that policy was changed in more recent versions of .NET).

src/WebJobs.Extensions.DurableTask/LinuxLoggerProvider.cs
```diff
@@ -0,0 +1,24 @@
+﻿using Microsoft.Extensions.Logging;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    internal class LinuxLoggerProvider : ILoggerProvider
```
cgillum, line 8: Same comments about prefixing with "AppService" and adding copyright headers at the top of each new file.

davidmrdavid:
looping in @wenhzha, who I think we need to add to these repos still.

src/WebJobs.Extensions.DurableTask/SystemEnvironment.cs
```diff
@@ -0,0 +1,80 @@
...
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    /// <summary>
+    /// Provides utilities for inspecting the process enviroment and platform, currently mostly
+    /// for checking if we are running on a Linux enviroment or not.
+    /// </summary>
+    /// Mostly duplicated functionality from:
+    /// https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Environment/EnvironmentExtensions.cs
+    /// We should consider abstracting away this functionality from both repos, to avoid code duplication.
+    internal class SystemEnvironment
```
cgillum, line 15: This seems like a lot of abstraction just to read some environment variables. I'm curious to understand why the Functions team decided to do it this way.

src/WebJobs.Extensions.DurableTask/LinuxLoggerProvider.cs
```diff
@@ -0,0 +1,24 @@
...
+            // TODO: I do not know if it's ok to throw away the categoryName
+            bool inLinuxConsumption = SystemEnvironment.Instance.IsLinuxConsumtpion();
+            LinuxLogger logger = new LinuxLogger(inLinuxConsumption);
+            return logger;
+        }
+
+        public void Dispose()
+        {
+            // TODO: Not exactly sure how to implement this
+            throw new NotImplementedException();
```
cgillum, line 21: I would just make this a no-op since we have nothing to dispose.

cgillum (commented):
Happy to see the progress on this! I did a very quick scan and have some feedback that should hopefully help you continue to make progress on this PR.

cgillum:
@wenhzha I asked Paul to add you to **azure-functions-team** team on GitHub. You should now have access to all the Functions GitHub repos, including the Durable ones.

src/WebJobs.Extensions.DurableTask/LinuxLogger.cs
davidmrdavid: Hmm, I'm not the biggest fan of adding `null`s but I trust your judgement here. Let's leave this issue as unresolved so we can do some testing for it

src/WebJobs.Extensions.DurableTask/LinuxLogger.cs
davidmrdavid: Good point!

Let's say we were to go for the more conservative approach, so `await writer.WriteLineAsync(jsonString);`.
If so, we'd have to add the `async` keyword to the method declaration, right? Not sure how this works in .NET, but I'd assume this would then deviate from the ILogger required method names, so I'm not sure if it would get called correctly.

cgillum: That's true. I think logging in general is supposed to have a minimal impact on runtime performance, which is why they did not design this API to by async-compatible, so my original proposal is probably the way to go.

src/WebJobs.Extensions.DurableTask/LinuxLogger.cs
davidmrdavid: Great, that makes sense. I have yet to see how we'd call `logger.Log(state)` with a `state` that contains the ETW data, but we can worry about that when we get to it :)

src/WebJobs.Extensions.DurableTask/SystemEnvironment.cs
davidmrdavid: I'll reach out to see what they say

davidmrdavid: Although, from my understanding, it appears that their file is doing a lot more than just enviroment variable reading for them. I also think this design somehow facilitates testing for them, if I recall correctly

**The Functions version of this file:** https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Environment/EnvironmentExtensions.cs

src/WebJobs.Extensions.DurableTask/LinuxLogger.cs
cgillum: `null` is just the lazy way to go (and I tend to be lazy sometimes :D). If you want to make this 100% robust, you can instead implement a no-op `IDisposable` class and return a static instance of that.

cgillum: Something like [this](https://github.com/StephenCleary/Disposables/blob/master/src/Nito.Disposables/NoopDisposable.cs) would work.

src/WebJobs.Extensions.DurableTask/SystemEnvironment.cs
ConnorMcMahon: I know the functions team has done a lot of abstractions around the environment variables and file system to put them behind interfaces that they code against so they can easily mock them for unit tests. Without also introducing their infrastructure for things like IEnvironment, it is unlikely that we will see the same testing benefits here.

src/WebJobs.Extensions.DurableTask/LinuxLogger.cs
cgillum: So that's a good point, actually. The DTFx log statements all pass in a `state` object that implements `IReadOnlyDictionary<string, object>` (see [here](https://github.com/Azure/durabletask/blob/ce153b3f9fe7ad053d7b6cc02300707079521a23/src/DurableTask.Core/Logging/StructuredLogEvent.cs#L29)). However, WebJobs.Extensions.DurableTask does *not* write logs in this format. It's possible that the implementation of `ILogger.LogInformation(...)` etc. creates a usable `state` object that we can serialize as-is, but we'd need to test it out to be sure.

However, we have to also be aware that our customer `ILogger` might also capture logs written by components we don't care about, like Functions runtime logs or customer logs. We don't want any of those logs to show up in our Durable-specific Kusto tables so we'll need to do some kind of filtering here. For example, I can imagine we might do something like this:

```csharp
if (state?.GetType().FullName.Contains("DurableTask") != true)
{
   // ignore log statements that do not come from Durable sources
   return;
}
```

src/WebJobs.Extensions.DurableTask/NoOpDisposable.cs
```diff
@@ -0,0 +1,39 @@
...
+        private NoOpDisposable()
+        {
+        }
+
+#if NETSTANDARD2_1
+        /// <summary>
+        /// Does nothing.
+        /// </summary>
+        public ValueTask DisposeAsync() => new ValueTask();
+#endif
```
cgillum, line 25: Since we don't target .NET Standard 2.1 (and don't plan to), let's remove all the `IAsyncDisposable` stuff.

src/WebJobs.Extensions.DurableTask/LinuxLogger.cs
davidmrdavid: Sounds good. I went with the unobserved Task approach, assuming that was your original proposal.

RE: Will Unobserved exceptions crash the process?
The **remarks** section here would suggest not: https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskscheduler.unobservedtaskexception?view=netcore-3.1

It also seems to apply to .NET Standard 2.0, which I think is the one we're on, as  our `.csproj` file

src/WebJobs.Extensions.DurableTask/SystemEnvironment.cs
davidmrdavid: That sounds about right to me.
That being said, and looking at the code in **our** `SystemEnviroment.cs`, I'm not immediately convinced it's abusing abstractions that much. It's certainly verbose, but the alternative would be to just inline a bunch of these calls into two functions instead of four small ones? Not sure if we're gaining that much by it, and I could imagine this utility file growing in the future.

I don't feel strongly though!

src/WebJobs.Extensions.DurableTask/AppServiceLinuxLoggerProvider.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Text;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    internal class AppServiceLinuxLoggerProvider : ILoggerProvider
+    {
+        public ILogger CreateLogger(string categoryName)
+        {
+            // TODO: I do not know if it's ok to throw away the categoryName
```
wenhzha, line 15: I think what you are trying to with the passed category is to have different type of ILogger, based on our sync, you mentioned there would be one for storage, one for regular, and the other? Are they all going to the same kusto table or seperate? If latter, I think that would mean you need set the ConsolePrefix tio corresponding prefix that matches with category in your Logger class in order to have traces pumped into different kusto table.

davidmrdavid: @wenhzha , I see what you mean! However, my thinking was that there would be only **two** types of loggers: one for Linux Consumption and another for Linux Dedicated. And that these loggers would figure out how to write the data from the three **sources** of logs: storage, durableTask, durableExtension.

In this case, they all go to the same table :) : DurableFunctionsEventLogs

davidmrdavid: Although perhaps the right answer is that there would be three types of loggers/ ILoggers: one for Linux Consumption, one for Linux Dedicated, and another for Windows ETW events.

@cgillum , is that what you were thinking as well?

davidmrdavid:
@cgillum , just looking to revive this thread as we're close to merging the platform-level changes. 
Please consider the EventSource instance here:

https://github.com/Azure/azure-functions-durable-extension/blob/c01a22628685b4d427243552ab3f9d4e03076266/src/WebJobs.Extensions.DurableTask/EtwEventSource.cs#L6-L50

In EasyAuth, @ConnorMcMahon 's solution was to conditionally emit logs inside each **event** if they were in a linux enviroment. In our case, are we following that approach too or do we want to refactor our EventSource instances, like EtwEventSoruce above, to also respect by the ILogger interface? I ask this because I know that ILogger has some kind of premier support for ETWEvents so I'm unsure if you were planning on somehow using that

davidmrdavid:
Following-up on my previous comment, this appears to be the key utility to have an ETW implementation of ILogger: https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.eventsource.loggingeventsource?view=dotnet-plat-ext-3.1

In it we have this sample:

```C#
class MyEventListener : EventListener 
{
    protected override void OnEventSourceCreated(EventSource eventSource) 
    {
        if (eventSource.Name == "Microsoft-Extensions-Logging") 
        {
            // initialize a string, string dictionary of arguments to pass to the EventSource.
            // Turn on loggers matching App* to Information, everything else (*) is the default level (which is EventLevel.Error)
            var args = new Dictionary&lt;string, string&gt;() { { "FilterSpecs", "App*:Information;*" } };
            // Set the default level (verbosity) to Error, and only ask for the formatted messages in this case.
            EnableEvents(eventSource, EventLevel.Error, LoggingEventSource.Keywords.FormattedMessage, args);
        }
    }

    protected override void OnEventWritten(EventWrittenEventArgs eventData) 
    {
        // Look for the formatted message event, which has the following argument layout (as defined in the LoggingEventSource.
        // FormattedMessage(LogLevel Level, int FactoryID, string LoggerName, string EventId, string FormattedMessage);
        if (eventData.EventName == "FormattedMessage")
            Console.WriteLine("Logger {0}: {1}", eventData.Payload[2], eventData.Payload[4]);
    }
}
```

I'm just unsure as to what this is doing. I'm under the impression this is some kind of EventSource **listener** and that the first method somehow registers an EventSource to be listenable and that the second gets invoked at every listened write event as a starting point for us to call some other logger in response.

ConnorMcMahon:
@davidmrdavid, this link looks useful: https://developers.redhat.com/blog/2019/12/23/tracing-net-core-applications/. It looks like EventListener allows you to listen to when ETW events are fired, and then you can do custom code in there. You could imagine that when you receive various ETW events, you emit them with the ILogger that emits logs how the various Linux platforms expect them.

davidmrdavid:
Thanks Connor! My latest PR includes a naive EventListener now, thanks to your link. Next up, I need to refactor some things around, but I get the feeling that we're close

src/WebJobs.Extensions.DurableTask/AppServiceLinuxLoggerProvider.cs
cgillum: @davidmrdavid The EtwEventSource class already takes care of getting events into the _DurableFunctionsEvents_ Kusto table. We don't need any `ILogger` help for that, so at most we'd want just two. I don't have a strong preference about whether consumption and dedicated are combined into a single ILogger or multiple ILoggers.

src/WebJobs.Extensions.DurableTask/SystemEnvironment.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// The singleton instance of SystemEnvironment, no explicit constructor is exported.
+        /// </summary>
+        public static SystemEnvironment Instance => PrivInstance.Value;
+
+        /// <summary>
+        /// Gets a value indicating whether the application is running in a Linux Consumption (dynamic)
+        /// App Service environment.
+        /// </summary>
+        /// <returns>true if running in a Linux Consumption App Service app; otherwise, false.</returns>
+        public bool IsLinuxConsumtpion()
```
cgillum, line 47:
```suggestion
        public bool IsLinuxConsumption()
```

src/WebJobs.Extensions.DurableTask/ETWEventListener.cs
```diff
@@ -0,0 +1,27 @@
+﻿using Microsoft.Extensions.Logging;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.Tracing;
+using System.Text;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    internal class ETWEventListener : EventListener
```
cgillum, line 9: Two things:
1. The name of this class doesn't look right. There shouldn't be any "ETW" involved at all in Linux scenarios.
2. I'm not familiar with the `EventListener` class. Does it work in Linux, and how is its `OnEventWritten` method (below) getting invoked?

ConnorMcMahon: @cgillum, the way that I understand the way EventListener and how it works, is that it allows you to listen to "ETW" events emitted by EventSource implementations, even if actual ETW events aren't being emitted to the OS like they are on Windows. 

This allows you to set up a listener for these  "ETW" events, and we can instead write our ETW-equivalent log.

This is at least my understanding of the class, but @davidmrdavid, it may be worth spinning up a Linux VM and testing this code path there to ensure EventListener is behaving as expected in non-Windows environments.

ConnorMcMahon: From what I could grasp reading around (see [here](https://medium.com/criteo-labs/c-in-process-clr-event-listeners-with-net-core-2-2-ef4075c14e87) and [here](https://github.com/aspnet/Logging/issues/572)), EventSource implementations on Linux write to LTTng, and EventListener on Linux listens to LTTng events and calls `OnEventWritten` based on those.

Because of this, I would second Chris's first critique, and we should just call this `EventSourceListener`.

davidmrdavid: @cgillum 

I'll have to answer these two backwards for clarity.

(2) 

My understanding is that `EventListener`'s `OnEventWritten` method gets invoked every time an `EventSource` write event is emitted. So, for us, that would be a callback resulting from executing any event in `EtwEventSource`. For example: `OnEventWritten` would be called right after `FunctionScheduled` or `FunctionAwaited`.

You can read about it [here](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.eventsource.loggingeventsource?view=dotnet-plat-ext-3.1) and [here](https://developers.redhat.com/blog/2019/12/23/tracing-net-core-applications/)

_Why does this work?_

**_Currently_**,  our internal logic always calls upon `EtwEventSource` events but, on Linux, they have no effect. Therefore, with the Listener, they would have the effect of invoking the `OnEventWritten` listener which, in turn, would write a log for the linux platform to consume. So we've turned ETW events having no effects, to now having a logging effect..

_Why am I using this approach?_

Well, when we switched to using the `ILogger` abstraction, I thought the vision was to use it as an interface to control **all** logging in the Durable stack. So my understanding was that maybe there was an **EventSource-*implementation* of the ILogger interface that we were going to refactor `EtwEventSource` into. That was what had me confused for so long, because I couldn't find it. Now I realize it probably doesn't exist and that maybe I misunderstood this vision.

I had also read [here](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.eventsource.loggingeventsource?view=dotnet-plat-ext-3.1) that `ILogger` has an `EventSource` bridge: `EventListener`. So, after realizing there was no EventSource-implementation of the ILogger interface, I put those two together and _assumed_ that the plan was to use something like this instead :) . Now, I'm not so sure that was the plan either but that's what this thread is for!

(1)

From my answer to (2), hopefully its clear that I was thinking of leaving the **effect-less** ETW events in Linux and to use the Listener as a way to give them a callback that triggers a logging action. If you were to agree with that plan, then the naming would be correct.

src/WebJobs.Extensions.DurableTask/AppServiceLinuxLoggerProvider.cs
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    internal class AppServiceLinuxLoggerProvider : ILoggerProvider
```
ConnorMcMahon, line 11: I'm not sure another ILoggerProvider is the right way to go about this, as I think the functions host already gives us an ILoggerProvider and I don't want to conflict with anything there.

We could probably put this in a new interface or just stick the logic in the extension's `Initialize()` method if we don't expect dependency injection to be required in the future. This new interface could take an `INameResolver` so it can read environment variables directly, and won't need to take a category name since we only care about environmentvariables when building our Logger.

src/WebJobs.Extensions.DurableTask/SystemEnvironment.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            return !string.IsNullOrEmpty(this.GetEnvironmentVariable(AzureWebsiteInstanceId));
+        }
+
+        /// <summary>
+        /// Retrieves the value from an environment variable from the current process.
+        /// </summary>
+        public string GetEnvironmentVariable(string name)
+        {
+            return Environment.GetEnvironmentVariable(name);
```
ConnorMcMahon, line 77: This should use `INameResolver` to make it easier to test.

src/WebJobs.Extensions.DurableTask/SystemEnvironment.cs
ConnorMcMahon: I think this would be cleaner as a set of helper static methods for `INameResolver`. That way we can make this super unit testable by mocking `INameResolver`.

ConnorMcMahon (changes requested):
It's coming along.

Some general feedback that should improve testability.

src/WebJobs.Extensions.DurableTask/ETWEventListener.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            this.logger = logger;
+        }
+
+        protected override void OnEventWritten(EventWrittenEventArgs eventData)
+        {
+            if (DurableTaskExtension.InLinux)
+            {
+                // TODO: Not sure that LogLevel here is right
+                this.logger.Log(LogLevel.Information, eventData.EventId, eventData, null, null);
```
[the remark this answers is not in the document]
davidmrdavid: `EventWrittenEventArgs`  does have an EventLevel but it needs to be manually converted to a `LogLevel` type, apparently there's no automatic translation. That's ok though, I can take care of that :)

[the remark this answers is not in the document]
davidmrdavid: Yup! You're right! I haven't updated my ILogger implementation to take care of that yet. That's the place where I'm thinking it would make most sense to include that logic

src/WebJobs.Extensions.DurableTask/AppServiceLinuxLogger.cs
```diff
@@ -0,0 +1,64 @@
...
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    public class AppServiceLinuxLogger : ILogger
+    {
+        private const string ConsolePrefix = "MS_DURABLE_FUNCTION_EVENTS_LOGS";
+        private const string LoggingPath = "/var/log/functionsLogs/durableevents.log";
+        private readonly bool inConsumption;
```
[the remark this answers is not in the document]
davidmrdavid: I'll look into separating them into two separate loggers then.

src/WebJobs.Extensions.DurableTask/SystemEnvironment.cs
davidmrdavid: @ConnorMcMahon , you mentioned in a later comment (which for some reason I can't comment on) that you recommended that I refactor this to use `INameResolver`. I've been very unlucky at finding information about that pattern, for instance [this](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.webjobs.inameresolver?view=azure-dotnet) page has no such description. 

Two questions:
(1) What should I know about it?
(2) As a .NET/C# newcomer, where can I find this info? I swear I've been googling for it and results haven't been very helpful in giving me a straightforward answer. Oh well, maybe this particular query was unlucky.....

davidmrdavid: Nevermind these questions, I asked in a C# chat and ended up getting answers to these two questions :)

davidmrdavid:
Alright, my latest changes include most of the feedback here, so I'll be requesting a re-review for y'all.

The one thing that I haven't incorporated yet is the resolution of INameResolver-formatted strings in SystemEnviroment.
I was hoping those would automagically get resolved but the failing tests indicate otherwise.

@ConnorMcMahon @cgillum , would it make sense to re-use the nameResolver here: 

https://github.com/Azure/azure-functions-durable-extension/blob/ffa4e192b25984b9381bf32ce52fe01ff25b21e7/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs#L115

And to give SystemEnviroment (below) a reference to it? I'm juts unsure if that pre-existing `nameResolver` will always do the right thing as just read the enviroment variables. Alternatively, I can create a new one and use it only in `SystemEnviroment`.

https://github.com/Azure/azure-functions-durable-extension/blob/274a90d82a976beb428760fad27d5b788822c070/src/WebJobs.Extensions.DurableTask/SystemEnvironment.cs#L12-L20

Also, it appears that I have to call `INameResolver`'s `Resolve` method explicitly, is that right? I thought this happened behind the curtain, but maybe that's just for WebJobs users.

src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
```diff
@@ -0,0 +1,33 @@
...
+            this.logger = logger;
+        }
+
+        protected override void OnEventSourceCreated(EventSource eventSource)
+        {
+            if (eventSource.Name == "DurableTask-Core" ||
+                eventSource.Name == "DurableTask-AzureStorage" ||
+                eventSource.Name == "WebJobs-Extensions-DurableTask")
+            {
+                this.EnableEvents(eventSource, EventLevel.LogAlways, EventKeywords.All);
```
ConnorMcMahon, line 24: @cgillum, I know you cleaned up the noise in the DurableTask-Core logs, but are we also looking for verbose logs here too?

src/WebJobs.Extensions.DurableTask/LinuxConsumptionLogger.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    internal class LinuxConsumptionLogger : LinuxAppServiceILogger
```
ConnorMcMahon, line 9: You may need some extra metadata for LinuxConsumption, as it does not handle things like Tenant and RoleInstance for you like windows and linux dedicated do.

If you make `GenerateJsonStringToLog` virtual (or some subset of the method), you can override it here to add this extra metadata.

src/WebJobs.Extensions.DurableTask/LinuxDedicatedLogger.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    internal class LinuxDedicatedLogger : LinuxAppServiceILogger
+    {
+        private const string LoggingPath = "/var/log/functionsLogs/durableevents.log";
+
+        public override void Log(string jsonString)
+        {
+            using (var writer = new StreamWriter(LoggingPath, true))
+            {
+                // TODO: ensure this won't crash the process if an exception
```
ConnorMcMahon, line 19: I'm not sure what best practices are if for some reason we lose the ability to write telemetry. On the one hand, customer crashes are bad, but on the other hand, having silent loss of telemetry is also bad. It has less impact on customers, but if the customer does have impact, we don't even know why we lost telemetry.

If we do want to avoid crashes, you can just wrap in a try-catch, and just have a comment in the catch block saying we don't want to crash the process or something along those lines.

ConnorMcMahon (changes requested):
We are getting really close here. 

I have a few suggestions (some of which are probably going to pop up in your linux consumption testing), and I would love to see some tests. It may be hard to test linux consumption logger with a hardcoded path like that, but let's at least write some tests verifying we are using the correct logger in the right circumstances.

src/WebJobs.Extensions.DurableTask/ETWEventListener.cs
cgillum: Thanks @ConnorMcMahon and @davidmrdavid for the explanations. I now understand what `EventListener` does and I think it's actually a better solution than the `ILogger` vision I had originally.

However, regarding (2) I think we still need to remove ETW from the class name. This class doesn't have anything to do with ETW as far as I can tell, so including ETW in the name is misleading. I'm not understanding the point about effect-less ETW events in Linux. I'm not sure how there can be a concept of "effect-less ETW events in Linux if there are no ETW events in Linux. Apologies in advance if I'm misunderstanding something.

src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
cgillum: Heh, so there are actually two event sources named `DurableTask-Core` now 😅. We definitely want to only write events emitted from the newer one, which is the "cleaned up" one. We can filter out the "old" one by checking `eventSource.Guid != new Guid("7DA4779A-152E-44A2-A6F2-F80D991A5BEE")`.

src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
```diff
@@ -0,0 +1,33 @@
...
+            }
+        }
+
+        protected override void OnEventWritten(EventWrittenEventArgs eventData)
+        {
+            this.logger.Log(LogLevel.Information, eventData.EventId, eventData, null, null);
```
cgillum, line 30: @davidmrdavid you're gonna hate me for this (if you don't already by now 😅), but now that we're using `EventListener` instead of following my original `ILogger` vision (which was flawed - I think you're going in a much more sustainable direction), I suggest we not use `ILogger` at all and instead just inline the logic that you wrote in the `LinuxAppServiceILogger`/`LinuxDedicatedLogger` code (it's up to you if you want to maintain an abstraction that differentiates between Linux dedicated and Linux Consumption, but I would suggest an abstraction that is *not* ILogger). Hopefully you'll agree with me that this will result in a much simpler solution with far less boilerplate.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceILogger.cs
```diff
@@ -0,0 +1,61 @@
...
+        public string GenerateJsonStringToLog<TState>(TState maybeEventData)
+        {
+            string jsonString = "";
+            if (maybeEventData is EventWrittenEventArgs)
+            {
+                var eventData = maybeEventData as EventWrittenEventArgs;
+
+                var values = eventData.Payload;
+                var keys = eventData.PayloadNames;
+                var keyValuePairs = keys.Zip(values, Tuple.Create);
```
cgillum, line 36: This is a very clean way of generating the JSON. However, I'm worried about performance since this is high-volume logging code. `Tuple.Create` will result in the creation of a new object in the managed heap which will need to be garbage collected, and since we will call this method tons of times with lots of data, I expect it to generate tons of garbage and that the CLR garbage collector will eat up a lot of CPU trying to keep up.

As an alternative, I suggest we do some simple looping, like...

```csharp
var values = eventData.Payload;
var keys = eventData.PayloadNames;

var json = new JObject();
for (int i = 0; i < values.Count; i++)
{
   json.Add(keys[i], values[i]);
}
```

src/WebJobs.Extensions.DurableTask/LinuxDedicatedLogger.cs
davidmrdavid: I'll wrap this in a try-catch and maybe we can come up with something to log on a failure as well? It would be good if we could set up a monitor that detects these failures as well

src/WebJobs.Extensions.DurableTask/LinuxAppServiceILogger.cs
```diff
@@ -0,0 +1,61 @@
...
+
+                JObject json = new JObject();
+                foreach (Tuple<string, object> keyAndValue in keyValuePairs)
+                {
+                    json.Add(keyAndValue.Item1, keyAndValue.Item2.ToString());
```
cgillum, line 41: I'm a little worried about the use of `ToString()` here. It should work fine for most types but it might cause problems for others, like `DateTime`.  Have you considered just assigning the value directly instead of using `ToString()`. That might actually even be more efficient, especially in cases where the value is a primitive type like `int` (calling ToString() would result in "boxing" which creates more garbage that the GC has to deal with). 

If we _have_ to use `ToString()`, then I suggest having some type-specific overrides for things like `DateTime` so that we can call `ToString("o")` which generates a culture agnostic timestamp instead of the vanilla `ToString()` which generates something very different and less computer friendly.

src/WebJobs.Extensions.DurableTask/LinuxConsumptionLogger.cs
davidmrdavid: Right. I was looking at your EasyAuth PR and was not sure how you were populating those _extra_ fields. Do those correspond to some enviroment variables that I can parse?

src/WebJobs.Extensions.DurableTask/LinuxDedicatedLogger.cs
cgillum: Most guidance I've read suggests that it's better to err on the side of not crashing if logging fails. That said, we're certainly free to make our own decisions based on what we think is necessary for supporting our own product. :)

However, I'm not sure how much value adding a try/catch will add if we're going to swallow the exception anyways. It would be a bit of boilerplate to even set it up so that try/catch would work since this is a non-async method. Alternatively, we could simply add a comment here saying that we're intentionally ignoring exceptions and achieve the same outcome of clear communication of intent.

src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
```diff
@@ -0,0 +1,33 @@
...
+        public EventSourceListener(LinuxAppServiceILogger logger)
+        {
+            this.logger = logger;
+        }
+
+        protected override void OnEventSourceCreated(EventSource eventSource)
+        {
+            if (eventSource.Name == "DurableTask-Core" ||
+                eventSource.Name == "DurableTask-AzureStorage" ||
+                eventSource.Name == "WebJobs-Extensions-DurableTask")
```
cgillum, line 22: Should we consider accepting anything that has `DurableTask` in the name?  It would be convenient if this could continue working even for `DurableTask-SqlServer` and whatever the name of Sebastian's E1 backend is.

cgillum (commented):
Sorry for the long delay from me. I really like where this is going but I'm now thinking we should actually ditch the `ILogger` abstraction completely and just embrace `EventListener`. See my comments below.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceILogger.cs
davidmrdavid: The functional programmer in me dies a little whenever I see a zip replaced with a for-loop 😄 
But yeah, I was also thinking these would end up being regretful abstractions....thanks for letting me know! Will change it

src/WebJobs.Extensions.DurableTask/LinuxAppServiceILogger.cs
davidmrdavid: Right, boxing can be expensive. I was thinking we'd have to call `toString` at some point since we're writing to the console/file, so I'm not sure what you mean by "just assigning the value directly" because I believe that would just postpone the `toString` call, no? Or is there some secret C#/.NET trick I can use here that I'm unaware of?

As for your second suggestion:
I can definitely look into type-aware transformations!

src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
davidmrdavid: Maybe! Could this end up catching customer-managed eventsources if they also implement an EventSource with "DurableTask" in its name? That's mostly what I was trying to avoid.

davidmrdavid: Although I suppose that the linux-platform-level code is already filtering EventSources that it doesn't recognize, so maybe this is a non-concern. I was simply trying to be robust by being explicit about what we capture

src/WebJobs.Extensions.DurableTask/LinuxAppServiceILogger.cs
cgillum: That’s interesting that you consider yourself a functional programmer. I’ve been wanting to offer better Durable support for F# developers for a while. We should talk about this after Python GA. 😊

src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
cgillum: In that case, can you also add DurableTask-SqlServer?

src/WebJobs.Extensions.DurableTask/ETWEventListener.cs
ConnorMcMahon: I agree with the rename to avoid ETW in the name. My initial explanation of  this was poor, as I was still thinking of EventSource in ETW terms. Since .NET Core 2.x, it is more fair to say that EventSource writes system events. On Windows, these are ETW events, on Linux (and macOS?), these are LTTng, but really, the implementation doesn't matter. EventListener is OS-aware enough to listen to these system events regardless, so keeping this abstraction at the layer of "Events" rather than "ETW" is correct.

src/WebJobs.Extensions.DurableTask/LinuxDedicatedLogger.cs
ConnorMcMahon: Does an exception thrown in this layer not bubble up and cause issues higher up in the system. I guess I am unfamiliar with how uncaught exceptions (especially in non-awaited async code) actually affects the process health.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceILogger.cs
ConnorMcMahon: I also love the idea of Functional programming and tried to take inspiration from it where I can, but in practice I am too addicted to state to practice it.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceILogger.cs
ConnorMcMahon: I am also concerned with the current approach, more from a correctness stance than the performance stamp.

The current code would convert `[("number", 5)]` into `{"number": "5"}`, where just letting Newtonsoft handle it would `{"number": 5}`. These are semantically different objects.

And I can't speak to the performance gains as much, but Newtonsoft (at least until .NET Core 3.0), was essentially treated as a standard .NET library, and it is probably doing some fairly performant stuff under the covers.

Essentially, this just boils down into "let the library do its magic thing".

cgillum: If there isn't a specific reason to call `ToString()` then I would avoid doing so. It's not clear to me that you'd actually be delaying anything, but even if you were I'd say it's best to leave the data as-is.

The act of serializing to a string will certainly result in at least a single string allocation, but that won't necessarily require any of the JObject property values to be boxed. Most likely the serializer will just copy the necessary bytes into the output string without additional allocations. Writing to the console/file also shouldn't require any additional per-field allocations since you're just copying a single, already-serialized JSON string into the I/O stream. For those reasons I feel that it's not necessarily the case that we're delaying inevitable boxing, which means this proposed change could actually have a noticeable performance impact under load.

src/WebJobs.Extensions.DurableTask/LinuxDedicatedLogger.cs
cgillum: @davidmrdavid 
> I'll wrap this in a try-catch and maybe we can come up with something to log on a failure as well? It would be good if we could set up a monitor that detects these failures as well

But where do we log the failure to log if we can't log anything? :)  I suppose it depends on whether the failure is related to writing to disk or if it's just a bug that is specific to certain logging payloads. I guess if we want to protect ourselves from those, then adding exception handling logic would be worthwhile. But just remember that we can't use `await` here, so you'd need to either add a task continuation for error handling or wrap the call in another async/await method that uses try/catch internally (you'd still need to ignore the task returned by that method, though).

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -295,6 +296,30 @@ void IExtensionConfigProvider.Initialize(ExtensionConfigContext context)
...
+            }
+            else if (inLinuxConsumption)
+            {
+                linuxLogger = new LinuxAppServiceLogger(writeToConsole: true);
+            }
+
+            if (linuxLogger != null)
+            {
+                this.eventSourceListener = new EventSourceListener(linuxLogger);
+            }
```
cgillum, line 321: Consider wrapping these up in a separate, private method since this method is already fairly long.

src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public EventSourceListener(LinuxAppServiceLogger logger)
+        {
+            this.logger = logger;
+        }
+
+        protected override void OnEventSourceCreated(EventSource eventSource)
+        {
+            if ((eventSource.Name == "DurableTask-Core"
+                  && eventSource.Guid != new Guid("7DA4779A-152E-44A2-A6F2-F80D991A5BEE")) ||
```
cgillum, line 22: Let's add a comment explaining that 7DA4... is the old, legacy DurableTask-Core provider and we want to explicitly ignore that.

src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
cgillum: Regarding Connor's point about verbosity - that's a good question. We log everything on Windows but that's much lower overhead compared to Linux. It makes me wonder if we should add some environment variable switches that allow us to dial down some of this logging. I'm not sure what the right way to approach this is, though, so perhaps we should consider it for a separate PR.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -0,0 +1,72 @@
...
+        private const string LoggingPath = "/var/log/functionsLogs/durableevents.log";
+        private readonly bool writeToConsole;
+
+        public LinuxAppServiceLogger(bool writeToConsole)
+        {
+            // If writeToConsole is False, we write to a file
+            this.writeToConsole = writeToConsole;
+        }
+
+        public string GenerateJson<TState>(TState maybeEventData)
```
cgillum, line 26: Now that we've removed the unnecessary abstractions, let's remove the `TState` type parameter and just use `EventWrittenEventArgs` directly.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -0,0 +1,72 @@
...
+                    json.Add(keys[i], (JToken) values[i]);
+                }
+
+                jsonString = json.ToString(Newtonsoft.Json.Formatting.None);
+            }
+
+            return jsonString;
+        }
+
+        public void Log<TState>(TState state)
```
cgillum, line 48: Similarly, let's do the simple thing here:
```suggestion
        public void Log(EventWrittenEventArgs eventData)
```

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            string jsonString = this.GenerateJson(state);
+            if (jsonString == "")
```
cgillum, line 51:
```suggestion
            if (string.IsNullOrEmpty(jsonString))
```

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -0,0 +1,72 @@
...
+            }
+
+            if (this.writeToConsole)
+            {
+                string consoleLine = ConsolePrefix + " " + jsonString;
+                Console.WriteLine(consoleLine);
+            }
+            else
+            {
+                using (var writer = new StreamWriter(LoggingPath, true))
```
cgillum, line 63: Let's use named parameters so that readers know what `true` is referring to:
```suggestion
                using (var writer = new StreamWriter(LoggingPath, append: true))
```

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -0,0 +1,72 @@
...
+            string jsonString = this.GenerateJson(state);
+            if (jsonString == "")
+            {
+                return;
+            }
+
+            if (this.writeToConsole)
+            {
+                string consoleLine = ConsolePrefix + " " + jsonString;
+                Console.WriteLine(consoleLine);
```
cgillum, line 59: Do you happen to know other other components log to the console? A part of me wonders if it would be better to do this in a non-blocking way, like this:
```suggestion
                Task unused = Console.Out.WriteLineAsync(consoleLine);
```
I'm not sure whether this could cause logs to be written out of order though...

cgillum (approved):
I think we're in a good place with this now. A few suggestions, but nothing blocking.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
davidmrdavid: We actually no longer need this, the check for is the string was empty was forced on us by the ILogger abstraction. So I've just removed the conditional altogether.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
davidmrdavid: Yeah I should have caught on to this. thanks

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -0,0 +1,88 @@
...
+                Task unused = Console.Out.WriteLineAsync(consoleLine);
+            }
+            else
+            {
+                // We write to a file in Linux Dedicated
+                using (var writer = new StreamWriter(LoggingPath, append: true))
+                {
+                    // We're ignoring exceptions in the unobserved Task
+                    Task unused = writer.WriteLineAsync(jsonString);
+                }
```
cgillum, line 84: So I just now noticed two problems with this block of code:
1. Because we're not awaiting the task, there could be a race condition where we dispose the `StreamWriter` before we finish writing to the file. This could cause us to lose logs due to exceptions.
2. I'm a little worried about the performance impact of opening and closing file handles for every log statement by creating and disposing the `StreamWriter`. Does other Linux dedicated logging do this? I'm wondering if we need to consider keeping the file open indefinitely in append mode to reduce disk I/O. The risk there, however, is that I don't know what impact this might have on the infrastructure that reads the log files while the app is running.

So (1) above should be consider blocking. (2) would be non-blocking but something that makes me a little nervous. We should probably plan on doing some performance testing comparing Windows and Linux to see if there is a noticeable difference before and after this change.

cgillum (changes requested):

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
davidmrdavid: Ahh, I think you're right.

@cgillum the Functions Host approach appears to block via `await` but also batches its messages such that it's only writing in 30 second intervals. So it's a compromise. You can find their file logger implementation here: https://github.com/Azure/azure-functions-host/blob/35cf323fa3464a08b410a518bcab006e801301fe/src/WebJobs.Script.WebHost/Diagnostics/LinuxAppServiceFileLogger.cs

I can certainly use this approach, in fact, I had a port of it in my _old_ telemetry PR here: https://github.com/Azure/azure-functions-durable-extension/pull/1391/files


It also reminds me that, as part of the logging setup, I need to create a directory in the linux host when running on linux dedicated, I'll add that change shortly.

but ended up forgetting about it.

davidmrdavid: Please let me know if you'd prefer I go straight away for the linux host approach to file logging

davidmrdavid: I ended up going for the continuation approach, as we discussed. I'm likely to add in a few more changes before merging, but I believe this is mostly done now.

cgillum (approved):

davidmrdavid:
Ugh, I think I'm gonna have to introduce another race condition.

**Context:**
In linux dedicated, we write our to a file. Obviously, we don't want that file to grow indefinitely. That could cause all sorts of problems.

I was thinking the linux platform was making sure this wouldn't occur, but after reviewing the Functions Host code I see that they're manually handling the process of:
(1) checking for the size (in bytes) of the log file
(2) if size exceeds a magic number of bytes, renaming that file to a slightly different name, turning it into an _file to delete._
(3) if there are more than a certain number of _files to delete_, they delete them.

This likely means we'll also have to manually monitor and manage the size and lifetime of our log file.

**Problem:**

Given our non-blocking approach, this process of managing the log files and their lifetimes may interfere with a log-writing task, yielding an exception. You could imagine an unobserved / un-awaited log task taking place while another thread is deleting or moving a now old log file. The outcome there would be missing logs, most likely, which isn't that big of a deal, but it is a concern nonetheless. 

**Next steps:**
I don't personally see a clear way to manage the lifetime and size of our log file while also maintaining a non-blocking approach to writing our logs. If y'all do, please let me know. 

In the meantime, I'll just add the logfile management logic on top of what we already have.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -62,6 +65,7 @@ public LinuxAppServiceLogger(bool writeToConsole, string containerName, string t
             this.tenant = JToken.FromObject(tenant);
             this.sourceMoniker = JToken.FromObject(
                 string.IsNullOrEmpty(stampName) ? string.Empty : "L" + stampName.Replace("-", "").ToUpperInvariant());
+            this.procID = Process.GetCurrentProcess().Id;
```
cgillum, line 68: Does the return value of `Process.GetCurrentProcess()` need to be disposed after use?

davidmrdavid: Yup! Thanks for the heads up

ConnorMcMahon: Turns out that `Process.GetCurrentProcess()` should not be used at all for ids. https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1837.

I need to fix this in the EasyAuth codebase before releasing 1.4.0 on Linux dedicated.

cgillum: It looks like the `Environment.ProcessId` API is only available in .NET 5?

ConnorMcMahon: Ah that makes sense. I thought this warning was a part of the new code warnings being added in .NET 5, but I don't see this API in .NET Standard 2.0, so the API may be new as well. In that case, just dispose of it.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -0,0 +1,161 @@
...
+            int countArchives = 0,
+            int maxLogfileSizeInMb = 10)
+        {
+            this.countArchives = countArchives;
+            this.maxLogfileSizeInMb = maxLogfileSizeInMb;
+
+            // If writeToConsole is False, we write to a file
+            for (int count = 1; count <= MaxArchives; count++)
+            {
+                string archivedPath = LoggingPath + count;
```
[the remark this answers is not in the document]
ConnorMcMahon: In fact, I don't see us actually use `this.archivedPaths` after we set it. Is it still necessary?

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -219,6 +220,92 @@ public async Task ParentInstanceId_Not_Assigned_In_Orchestrator(bool extendedSes
...
+                    await host.StartAsync();
+                    var client = await host.StartOrchestratorAsync(orchestratorName, input: "World", this.output);
+                    var status = await client.WaitForCompletionAsync(this.output);
+                    await host.StopAsync();
+                }
+
+                this.output.WriteLine(sw.ToString());
+
+                // Ensure the console included some basic data
+                Assert.Contains(expectedOutput, sw.ToString());
```
ConnorMcMahon: A bit more validation would be nice. I.e. splitting into newlines and validating that the JSON contains the correct values for the new columns.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -2951,7 +3038,7 @@ public async Task DurableEntity_ScheduledSignal(bool extendedSessions)
 
                 var now = DateTime.UtcNow;
 
-                await client.SignalEntity(this.output, now + TimeSpan.FromSeconds(1), "delayed", null);
+                await client.SignalEntity(this.output, now + TimeSpan.FromSeconds(2), "delayed", null);
```
ConnorMcMahon, line 3044: I'll open a PR to make this 5 seconds again, as 1 is a bit too aggressive I have found.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -219,6 +220,92 @@ public async Task ParentInstanceId_Not_Assigned_In_Orchestrator(bool extendedSes
...
+                // Ensure the console included some basic data
+                Assert.Contains(expectedOutput, sw.ToString());
+            }
+        }
+
+        [Fact]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        public async Task WritesToFile()
+        {
+            LinuxAppServiceLogger.LoggingPath = Path.Combine(Directory.GetCurrentDirectory(), "logfile.log");
```
ConnorMcMahon, line 282: We may want to delete the path if it already exists, or add some suffix to our path, to ensure that we aren't just validating a previous run succeeded.

ConnorMcMahon (commented):
I think using Serilog is a great choice. 

I have some general feedback, but great progress!

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -0,0 +1,161 @@
...
+            // We write to console in Linux Consumption
+            if (this.writeToConsole)
+            {
+                // We're ignoring exceptions in the unobserved Task
+                string consoleLine = ConsolePrefix + " " + jsonString;
+                Task unused = Console.Out.WriteLineAsync(consoleLine);
+            }
+            else
+            {
+                // We write to a file in Linux Dedicated
```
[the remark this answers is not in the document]
davidmrdavid: No commented out code here, as far as I can tell.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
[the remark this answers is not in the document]
davidmrdavid: yeah archived paths was from an older approach where I was managing files manually. This was deleted in a recent clean-up PR

test/Common/DurableTaskEndToEndTests.cs
davidmrdavid: Thanks! Yeah this was rather flaky. 1 is to few, 10 is too many.
I did have a flaky test with 5 seconds, so I suspect that somewhere around 2 to 4 might be a sweet spot. Aka, I'd try 3 :)

test/Common/DurableTaskEndToEndTests.cs
davidmrdavid: I'll remove it before the test, if it exists. That's fair

test/Common/DurableTaskEndToEndTests.cs
ConnorMcMahon: It's flaky at 5 due to the fact the whole test stops at 10 seconds, and we already wait 5 seconds before we call signal.

Increasing that timeout accordingly should resolve flakiness.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -0,0 +1,161 @@
...
+                { "Tid", Thread.CurrentThread.ManagedThreadId },
+            };
+
+            for (int i = 0; i < values.Count; i++)
+            {
+                json.Add(keys[i], JToken.FromObject(values[i]));
+            }
+
+            // Generate string-representation of JSON
+            string jsonString = json.ToString(Newtonsoft.Json.Formatting.None);
```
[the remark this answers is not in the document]
davidmrdavid: Added both!

test/Common/DurableTaskEndToEndTests.cs
davidmrdavid: I agree that this could use a bit more validation.

So, I can (and will) validate that the output is a valid JSON-parseable string. I can also validate some minimal columns that are always present (EventId, TimeStamp, Tid, Pid). _Does that sound sufficient_? Beyond that, I worry that this may be overly specific, especially considering that our logs are so noisy and irregular in shape.

I'm **always** in favor of more testing, just want to also be mindful of our deadline here.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -219,6 +221,145 @@ public async Task ParentInstanceId_Not_Assigned_In_Orchestrator(bool extendedSes
...
+                // This orchestrator should error out on null inputs
+                var client = await host.StartOrchestratorAsync(orchestratorName, input: null, this.output);
+                var status = await client.WaitForCompletionAsync(this.output);
+                await host.StopAsync();
+            }
+
+            // Ensure the logging file was at least generated
+            Assert.True(File.Exists(LinuxAppServiceLogger.LoggingPath));
+
+            // Ensure newlines are removed by checking the the number of lines is equal to the number of "TimeStamp" columns,
```
davidmrdavid, line 352: I'll removed the duplicate "The"

test/Common/DurableTaskEndToEndTests.cs
ConnorMcMahon: Sounds good to me!

ConnorMcMahon (approved):
LGTM!

PR #1485, opened 2020-09-17
Resource URL check for ManagedIdentityTokenSource

### LINKED ISSUE #1481: Add resource url check to ManagedIdentityTokenSource
Need to check if a customer passes in `https://management.core.windows.net/` as the resource url to `ManagedIdentityTokenSource` and change it to be `https://management.core.windows.net/.default`. This should be added to support old docs using `https://management.core.windows.net/` and help align with the new authentication method using Azure.Identity libraries to get Azure AD tokens.

bachuv:
Resolves #1481. This change supports docs that use `https://management.core.windows.net`, `https://management.core.windows.net/`, `https://graph.microsoft.com`, or `https://graph.microsoft.com/` as the `ManagedIdentityTokenSource` resource URL. It replaces those URLs with `https://management.core.windows.net/.default` and `https://graph.microsoft.com/.default` which are the resource URLs used in the new Azure.Identity libraries.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ManagedIdentityTokenSource.cs
+++ b/src/WebJobs.Extensions.DurableTask/ManagedIdentityTokenSource.cs
@@ -32,6 +32,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             {
                 this.Resource = "https://management.core.windows.net/.default";
             }
+
+            if (this.Resource.Equals("https://graph.microsoft.com") || this.Resource.Equals("https://graph.microsoft.com/"))
+            {
+                this.Resource = "https://graph.microsoft.com/.default";
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


src/WebJobs.Extensions.DurableTask/ManagedIdentityTokenSource.cs
```diff
@@ -32,6 +32,11 @@ public ManagedIdentityTokenSource(string resource, ManagedIdentityOptions option
             {
                 this.Resource = "https://management.core.windows.net/.default";
             }
+
+            if (this.Resource.Equals("https://graph.microsoft.com") || this.Resource.Equals("https://graph.microsoft.com/"))
```
cgillum, line 36: nit: maybe make this an `else if` instead of an `else` to avoid unnecessary string comparisons.

cgillum (approved):

src/WebJobs.Extensions.DurableTask/ManagedIdentityTokenSource.cs
bachuv: Do you mean make the `if` that checks the graph URL an `else if`? Just want to confirm before I make any changes.

cgillum: Yep!

PR #1487, opened 2020-09-18
fix flaky tests for CleanEntityStorage

sebastianburckhardt:
Fixes the tests I saw failing occasionally in other PRs.

The problem is that these tests check the number of empty entities removed, but this number can be off if other tests that run before these also create empty entities.

The solution is to clean the storage prior to the tests, so it starts with a known count of zero.

### DIFF
--- a/test/Common/DurableTaskEndToEndTests.cs
+++ b/test/Common/DurableTaskEndToEndTests.cs
@@ -3986,6 +3986,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
             {
                 await host.StartAsync();
 
+                // before the test, clean storage so we can properly validate the count later
+                var irrelevantEntityId = new EntityId("Counter", "unused"); // we don't access this entity, just use it to construct a client
+                var entityClient = await host.GetEntityClientAsync(irrelevantEntityId, this.output);
+                await entityClient.InnerClient.CleanEntityStorageAsync(true, true, CancellationToken.None);
+
                 // construct unique names for this test
                 string prefix = Guid.NewGuid().ToString("N").Substring(0, 6);
                 var emptyEntityId = new EntityId("Counter", $"{prefix}-empty");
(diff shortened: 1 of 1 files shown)


test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -3986,6 +3986,11 @@ public async Task DurableEntity_CleanEntityStorage(string storageProvider)
             {
                 await host.StartAsync();
 
+                // before the test, clean storage so we can properly validate the count later
+                var irrelevantEntityId = new EntityId("Counter", "unused"); // we don't access this entity, just use it to construct a client
```
ConnorMcMahon, line 3990: I'm afraid this will lead to some other race condition in these tests if they are both ran concurrently. The general approach I would take to fixing the flakiness of these tests would be to make `DurableEntity_CleanEntityStorage_Many` have a taskhub name of `nameof(DurableEntity_CleanEntityStorage_Many)`, instead of `nameof(DurableEntity_CleanEntityStorage)`. This way the tests won't interfere with each other as they are in separate taskhubs.

ConnorMcMahon (changes requested):

test/Common/DurableTaskEndToEndTests.cs
sebastianburckhardt: I see! I suppose I did not actually understand the reasons why this fails sometimes - I did not know the tests run concurrently. With a different test name the problem should go away.

ConnorMcMahon (approved):
LGTM!

test/Common/DurableTaskEndToEndTests.cs
ConnorMcMahon: I actually added a PR to the v1 branch that auto adds a randomized suffix of 4 alphanumeric characters to ensure that we get unique task hubs, as this happens a lot by accident, and we don't have any linting to catch it. 

I may look into porting that at some point, so we don't have to be quite so concerned about creating a unique task hub for each test.

PR #1501, opened 2020-09-30
Fix unintentional breaking change with IDurableClient.RaiseEventAsync

### LINKED ISSUE #1472: IDurableOrchestrationClient.RaiseEventAsync optional parameter/overload breaking change
### Description

Code changes made in the following PR:

https://github.com/Azure/azure-functions-durable-extension/pull/1422/files#diff-cb0ed55b71b2814605559f0131ed6f70R174
https://github.com/Azure/azure-functions-durable-extension/pull/1422/files#diff-cb0ed55b71b2814605559f0131ed6f70R197

Results in the compiler selecting the incorrect overload of RaiseEventAsync in user code.

### Expected behavior

Calling RaiseEventAsync with the following parameters:

```csharp
var instanceId = "my-instance-id";
var eventName = "myEvent";
var eventData = "myData";

await client.RaiseEventAsync(instanceId, eventName, eventData);
```
Should use the following method:

```csharp
Task RaiseEventAsync(string instanceId, string eventName, object eventData = null);
```

### Actual behavior

With the same user code above, the following overload is selected by the compiler instead:

```csharp
Task RaiseEventAsync(string taskHubName, string instanceId, string eventName, object eventData = null, string connectionName = null);
```

Additionally, due to the differing order of the parameters - the value supplied for instanceId is set for taskHubName causing an exception from the underlying WindowAzure.Storage provider with "Invalid container name length".

### Known workarounds

I was able to workaround this by specifying the optional parameter eventData in my user code:

```csharp
await client.RaiseEventAsync(instanceId, eventName, eventData: eventData);
```

### App Details

- **Durable Functions extension version**: 2.3.0
- **Programming language used**: netcore 3.1

Discovered while upgrading existing function app code from DurableTask v2.1.1 to v2.3.0

ConnorMcMahon:
While refactoring IDurableOrchestrationClient in #1422, the eventData parameter for an overload of `IDurableOrchestrationClient.RaiseEventAsync()` was accidentally made optional. This made C# bind `RaiseEventAsync(string, string, string)` to `RaiseEventAsync(string taskHubName, string instanceId, string eventName, object eventData, string connectionName = null)` instead of `RaiseEventAsync(string instanceId, string eventName, object eventData = null)` as intended.

This PR reverts that unintentional change, and adds a test to ensure
this won't happen again in the future.

Fixes #1472

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
@@ -194,7 +194,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// <param name="connectionName">The name of the connection string associated with <paramref name="taskHubName"/>.</param>
         /// <returns>A task that completes when the event notification message has been enqueued.</returns>
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1030:UseEventsWhereAppropriate", Justification = "This method does not work with the .NET Framework event model.")]
-        Task RaiseEventAsync(string taskHubName, string instanceId, string eventName, object eventData = null, string connectionName = null);
+        Task RaiseEventAsync(string taskHubName, string instanceId, string eventName, object eventData, string connectionName = null);
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


amdeel (approved):
Unfortunate that we have two of these with differently ordered parameters.

PR #1502, opened 2020-09-30
Gracefully handle extended sessions for out-of-proc languages

### LINKED ISSUE #1452: Handle extended sessions "gracefully" with out-of-proc languages
Currently, if using extended sessions with any non-.NET language, the customer's orchestrations will silently not execute as expected. This is very undesirable behavior, so it is preferable to do one of the following instead:

1. Log a warning that extended sessions is true, but that we will use extended sessions = false instead.
2. Throw an exception on functions host startup, so that we fail fast and it is apparent to the customer they are in an unsupported scenario.

ConnorMcMahon:
Currently, when extended sessions is set to true for out-of-proc
languages, orchestrations will silently fail. Now, if we detect
out-of-proc languages when extended sessions are true, we emit a warning
and set extended sessions to false.

Fixes #1452.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
+++ b/src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
@@ -273,6 +273,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
             string runtimeLanguage = environmentVariableResolver.Resolve("FUNCTIONS_WORKER_RUNTIME");
             if (this.ExtendedSessionsEnabled &&
+                runtimeLanguage != null && // If we don't know from the environment variable, don't assume customer isn't .NET
                 !string.Equals(runtimeLanguage, "dotnet", StringComparison.OrdinalIgnoreCase))
             {
                 traceHelper.ExtensionWarningEvent(
--- a/src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
+++ b/src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
@@ -258,7 +258,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             traceHelper.TraceConfiguration(this.HubName, configurationJson.ToString(Formatting.None));
         }
 
-        internal void Validate()
+        internal void Validate(INameResolver environmentVariableResolver, EndToEndTraceHelper traceHelper)
         {
             if (string.IsNullOrEmpty(this.HubName))
             {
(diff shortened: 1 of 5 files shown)


src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -271,6 +271,17 @@ internal void Validate()
                     "See documentation on Task Hubs for information on how to set this: https://docs.microsoft.com/azure/azure-functions/durable/durable-functions-task-hubs");
             }
 
+            string runtimeLanguage = environmentVariableResolver.Resolve("FUNCTIONS_WORKER_RUNTIME");
+            if (!string.Equals(runtimeLanguage, "dotnet", StringComparison.OrdinalIgnoreCase))
```
cgillum, line 275: Shouldn't we also be checking if `this.ExtendedSessionsEnabled == true`? Otherwise, won't we always log this for every non-.NET function app even if they set it to false?

cgillum (changes requested):

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
ConnorMcMahon: Ah, good catch!

cgillum: Also, should we consider the case where this environment variable is not set at all? There are many .NET apps which don't set this environment variable, and I'm worried that we'd be inadvertently disabling extended sessions for those apps that upgrade to this extension version.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -271,6 +271,18 @@ internal void Validate()
                     "See documentation on Task Hubs for information on how to set this: https://docs.microsoft.com/azure/azure-functions/durable/durable-functions-task-hubs");
             }
 
+            string runtimeLanguage = environmentVariableResolver.Resolve("FUNCTIONS_WORKER_RUNTIME");
+            if (this.ExtendedSessionsEnabled &&
+                !string.Equals(runtimeLanguage, "dotnet", StringComparison.OrdinalIgnoreCase))
```
cgillum, line 276: GitHub doesn't always show review comment threads the way I expect, so I'll post this comment again for this latest iteration:

> Also, should we consider the case where this environment variable is not set at all? There are many .NET apps which don't set this environment variable, and I'm worried that we'd be inadvertently disabling extended sessions for those apps that upgrade to this extension version.

cgillum (changes requested):

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
ConnorMcMahon: I'm going to add a test for this case too.

davidmrdavid (approved):
Looks good to me!
My only request is for the `Validate` method to be updated with a header comment saying just what it is validating.

cgillum (approved):
LGTM - :shipit:

PR #1507, opened 2020-10-02
Use new partition management by default

ConnorMcMahon:


cgillum (approved):

ConnorMcMahon:
For some reason, this has caused some entity scheduling tests to fail. For the normal scheduled signal, it sort of makes sense, as it can take up to an extra 5 seconds for a worker to start processing messages as opposed to before. It doesn't make sense for the self-scheduling entity, as once we have processed a single message, the partition should be up and running for the rest of them.

PR #1515, opened 2020-10-06
Adding Missing Fields to Linux Logger: ProviderName, TaskName, ActivityId, RelatedActivityId, StampName, etc

davidmrdavid:
_This was originally a small test PR, but now it's grown in scope_

This PR adds missing fields to our LinuxLogger class, providing now a fuller scope of logs and, hopefully, enough to make our telemetry and existing diagnostics tools more effective in debugging linux apps.

The logging validation test has been updated accordingly

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
+++ b/src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
@@ -35,7 +35,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         // logging metadata
         private readonly JToken roleInstance;
         private readonly JToken tenant;
-        private readonly JToken sourceMoniker;
         private readonly JToken procID;
 
         // if true, we write to console (linux consumption), else to a file (linux dedicated).
@@ -59,8 +58,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.roleInstance = JToken.FromObject("App-" + containerName);
             this.tenant = JToken.FromObject(tenant);
 
-            this.sourceMoniker = JToken.FromObject(
-                string.IsNullOrEmpty(stampName) ? string.Empty : "L" + stampName.Replace("-", "").ToUpperInvariant());
             using (var process = Process.GetCurrentProcess())
             {
                 this.procID = process.Id;
--- a/src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
+++ b/src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
@@ -36,6 +36,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         private readonly JToken roleInstance;
         private readonly JToken tenant;
         private readonly JToken procID;
+        private readonly string stamp;
+        private readonly string primaryStamp;
 
         // if true, we write to console (linux consumption), else to a file (linux dedicated).
         private readonly bool writeToConsole;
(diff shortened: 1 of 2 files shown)


src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -59,8 +58,6 @@ internal class LinuxAppServiceLogger
             this.roleInstance = JToken.FromObject("App-" + containerName);
             this.tenant = JToken.FromObject(tenant);
 
-            this.sourceMoniker = JToken.FromObject(
-                string.IsNullOrEmpty(stampName) ? string.Empty : "L" + stampName.Replace("-", "").ToUpperInvariant());
```
cgillum, line 63: Is it correct that you were never actually using this value? I don't see it referenced anywhere else. If so, I guess my next question is whether there is an opportunity to make the test more robust.

davidmrdavid: @cgillum, sorry this is just a test PR that I was running because, for some reason, the Azure Pipelines CI stopped running our test. I should have made this a draft, will take a mental note. Didn't mean to be reviewed

davidmrdavid: But, to answer your questions:
I removed the usage of `this.sourceMoniker` on a previous PR, that's why you did not see that here.
That PR is this one:https://github.com/Azure/azure-functions-durable-extension/pull/1482

My test should have caught on to the fact that `this.sourceMoniker` was not getting logged (and in fact they do) but Azure Pipelines is, somehow, no longer automatically running the tests for new PRs. The tests you're currently seeing that are running for this PR were manually triggered.

davidmrdavid: In other words, the Azure Pipelines CI is acting up.
The result is that we're currently getting "green checkmarks" in al our PRs but those are deceiving because the tests are not getting run. For instance:
here: https://github.com/Azure/azure-functions-durable-extension/pull/1513
here: https://github.com/Azure/azure-functions-durable-extension/pull/1514

So, I'm manually running this test and, if it passes, I'm going to propagate this change to dev and master because, otherwise, we have a failing build. After that, I'll investigate what's going on with the CI.

I found out that the CI is not running our tests as I was inspecting the artifacts page of Azure  Pipelines, so that's why we did not catch it when promoting dev to master. I'm looking to fix that now

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -451,6 +458,18 @@ public async Task OutputsValidJSONLogs()
...
                 }
 
+                // recording if ActivityId and RelatedActivityId are seen
+                // we expect to see them, at some point, in a trivial orchestrator
+                if (keys.Contains("ActivityId"))
+                {
+                    foundActivityId = true;
+                }
+
+                if (keys.Contains("ActivityId"))
```
davidmrdavid, line 468: Should have been RelatedActivityId. Will fix

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -428,11 +432,14 @@ public async Task OutputsValidJSONLogs()
...
                 List<string> keys = json.Properties().Select(p => p.Name).ToList();
+                Assert.Contains("EventStampName", keys);
+                Assert.Contains("EventPrimaryStampName", keys);
+                Assert.Contains("ProviderName", keys);
+                Assert.Contains("TaskName", keys);
                 Assert.Contains("EventId", keys);
                 Assert.Contains("TimeStamp", keys);
                 Assert.Contains("RoleInstance", keys);
                 Assert.Contains("Tenant", keys);
-                Assert.Contains("SourceMoniker", keys);
```
davidmrdavid, line 435: this is what was breaking the test in master

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -428,11 +432,14 @@ public async Task OutputsValidJSONLogs()
 
                 // (2) Contains minimal expected fields
                 List<string> keys = json.Properties().Select(p => p.Name).ToList();
+                Assert.Contains("EventStampName", keys);
+                Assert.Contains("EventPrimaryStampName", keys);
+                Assert.Contains("ProviderName", keys);
+                Assert.Contains("TaskName", keys);
```
davidmrdavid, line 438: the rest of this PR should enables us to log these 4 properties in linux. That's it!

bachuv (approved):

PR #1519, opened 2020-10-06
Promoting dev to master for 2.3.1 release, now with fixed CI test and extra logging fields

davidmrdavid:
As in title.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
+++ b/src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
@@ -35,8 +35,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         // logging metadata
         private readonly JToken roleInstance;
         private readonly JToken tenant;
-        private readonly JToken sourceMoniker;
         private readonly JToken procID;
+        private readonly string stamp;
+        private readonly string primaryStamp;
 
         // if true, we write to console (linux consumption), else to a file (linux dedicated).
         private readonly bool writeToConsole;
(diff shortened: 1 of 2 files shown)


bachuv (approved):

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -58,9 +59,11 @@ internal class LinuxAppServiceLogger
             this.writeToConsole = writeToConsole;
             this.roleInstance = JToken.FromObject("App-" + containerName);
             this.tenant = JToken.FromObject(tenant);
+            this.stamp = stampName;
+
+            var finalCharIndex = stampName.Length - 1;
+            this.primaryStamp = char.IsLetter(stampName[finalCharIndex]) ? stampName.Remove(finalCharIndex) : stampName;
```
cgillum, line 65: Hey @davidmrdavid, so having though about this a bit more, this logic will be problematic for ASEs. Take a look at the results of this Kusto query to see what I mean:
```kusto
DurableFunctionsEvents
| where TIMESTAMP > ago(3d)
| summarize by EventPrimaryStampName, EventStampName
| order by EventPrimaryStampName desc, EventStampName asc
| take 500
```
I think it's fine to ship as-is for now since I don't expect a lot of Linux ASE usage (at least not right away), but let's open a new issue to track fixing this for a subsequent release.

davidmrdavid: Will do!

PR #1522, opened 2020-10-06
Linux Logger now supports the legacy regex logging path

davidmrdavid:
This PR has grown in scope multiple times over...
Currently, this PR implements a hybrid logging strategy supporting regex in linux dedicated and JSON for linux consumption. It also contains tests for both cases.


src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -100,18 +110,30 @@ private string GenerateJsonStr(EventWrittenEventArgs eventData)
...
                 { "TaskName", eventData.EventName },
                 { "EventId", eventData.EventId },
                 { "TimeStamp", DateTime.UtcNow },
-                { "RoleInstance", this.roleInstance },
-                { "Tenant", this.tenant },
                 { "Pid", this.procID },
                 { "Tid", Thread.CurrentThread.ManagedThreadId },
             };
 
+            if (!string.IsNullOrEmpty(this.stamp) && !string.IsNullOrEmpty(this.primaryStamp))
```
ConnorMcMahon, line 121: Is there a reason these environment variables were null?

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -57,12 +53,26 @@ internal class LinuxAppServiceLogger
...
-            this.stamp = stampName;
 
-            var finalCharIndex = stampName.Length - 1;
-            this.primaryStamp = char.IsLetter(stampName[finalCharIndex]) ? stampName.Remove(finalCharIndex) : stampName;
+            // Since the values below are obtained via a NameResolver, they might be null.
+            // Attempting to serialize a null value results in exceptions, or even worse, wrong logs,
+            // so we need to be careful.
+            if (!string.IsNullOrEmpty(containerName))
+            {
+                this.roleInstance = JToken.FromObject("App-" + containerName);
```
ConnorMcMahon, line 62: Is there a reason we are using JToken.FromObject here instead of just explicitly adding the string?

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
davidmrdavid: Well, in my test, the `tenant` (`"WEBSITE_STAMP_DEPLOYMENT_ID`) variable was empty.

As a result, and for safety, I made extra checks for all variables that I can't guarantee are set. That includes the stamp name (`"WEBSITE_HOME_STAMPNAME`") and, **when running in dedicated**, the container name.

davidmrdavid: I suppose I could re-run the smoke test and see if we **actually** need to be careful with the stamp name, but I'm under the impression that these environment variables may change in the future and so I wanted to be extra careful.

ConnorMcMahon (commented):
A couple of questions, but otherwise looks good.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
davidmrdavid: It's just an optimization, since internally JObjects only store JTokens, I'm just doing that work in advance. Might be a little overkill, happy to revert to strings only

cgillum: I would recommend against this. Vanilla strings are much easier to debug and parsing to JToken early like this is likely to confuse the future readers of this code because it doesn't map to any obvious intent. Generally speaking, if you're doing a performance optimization that might not be obvious to the reader, then you should include a comment explaining why you're doing what you're doing.

davidmrdavid: Yeah you're right. I'll remove it, was a silly thing to do

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -491,6 +491,133 @@ public async Task OutputsValidJSONLogs()
...
+                }
+
+                // recording if ActivityId and RelatedActivityId are seen
+                // we expect to see them, at some point, in a trivial orchestrator
+                if (keys.Contains("ActivityId"))
+                {
+                    foundActivityId = true;
+                }
+
+                if (keys.Contains("ActivityId"))
```
davidmrdavid, line 595: This is still wrong, should be RelatedActivityId. Not sure why that change keeps on getting removed

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
davidmrdavid: done!

test/Common/DurableTaskEndToEndTests.cs
davidmrdavid: done!

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -57,12 +130,26 @@ internal class LinuxAppServiceLogger
...
-            this.primaryStamp = char.IsLetter(stampName[finalCharIndex]) ? stampName.Remove(finalCharIndex) : stampName;
+            // Since the values below are obtained via a NameResolver, they might be null.
+            // Attempting to serialize a null value results in exceptions, or even worse, wrong logs,
+            // so we need to be careful.
+            if (!string.IsNullOrEmpty(containerName))
+            {
+                this.roleInstance = "App-" + containerName;
+            }
+
+            if (!string.IsNullOrEmpty(tenant))
```
cgillum, line 142: What's the purpose of this `if` check? I'm not noticing any particular behavior different vs. just always assigning `this.tenant = tenant;`.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -129,9 +229,46 @@ private string GenerateJsonStr(EventWrittenEventArgs eventData)
...
+
+                foreach (string column in this.orderedRegexCol)
+                {
+                    string val = "";
+                    if (jsonKeys.Contains(column))
+                    {
+                        try
+                        {
+                            val = (string)json[column];
+                            val = val.Replace("\n", "\\n").Replace("\r", "\\r").Replace("\"", "'").Replace(",", ";;");
```
cgillum, line 248: What's the reason for replacing double quotes `"` with single quotes `'`?

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -129,9 +229,46 @@ private string GenerateJsonStr(EventWrittenEventArgs eventData)
...
+                        }
+                        catch
```
cgillum, line 250: What are we expecting to catch here? InvalidCastException? Anything else?

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -129,9 +229,46 @@ private string GenerateJsonStr(EventWrittenEventArgs eventData)
...
+                        {
+                        }
+                    }
+
+                    if (column == "Details" || column == "ExceptionMessage")
+                    {
+                        val = '"' + val + '"';
```
cgillum, line 257: What's the reason for surrounding these values with quotes? It seems there are some undocumented assumptions here about the expected format of these logs. :)

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -129,9 +229,46 @@ private string GenerateJsonStr(EventWrittenEventArgs eventData)
...
+            {
+                // This path supports the legacy regex-based parser in Linux Dedicated.
+                // In the future, we'll be able to remove this and use JSON logging only
+                IList<string> jsonKeys = json.Properties().Select(p => p.Name).ToList();
+                List<string> regexValues = new List<string>();
+
+                foreach (string column in this.orderedRegexCol)
+                {
+                    string val = "";
+                    if (jsonKeys.Contains(column))
```
cgillum, line 243: Why not simply use `json.TryGetValue(column, out JToken val)`? That would in theory be an O(1) operation. What you're doing now appears to be O(N) where N is the number of properties in the JSON object (constructing jsonKeys is N, and the Contains operation is another N on top of that). It would also allow you to remove at least two lines of code and a redundant dictionary lookup on line 247.

cgillum (changes requested):

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
davidmrdavid: Yeah sorry, this commit you're reviewing is incomplete. I'm going to turn this into a draft PR to avoid getting reviews for in-progress work :)

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
davidmrdavid: I didn't know that was an option :) . I'll add it to my bag of tricks and use it here

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
davidmrdavid: Just an InvalidCast exception. I'm documenting this in my next commit

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
davidmrdavid: Documenting this in my next commit

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
davidmrdavid: Right, I did notice this but had not made the change yet as I was hoping this push would go un-noticed. That's a bad assumption on my part, that's why I made this back into a draft. In any case, I'm pushing a new commit with explanations for all of this and responding to your feedback

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -129,9 +227,60 @@ private string GenerateJsonStr(EventWrittenEventArgs eventData)
...
+                            // (1) Escaping newline-like characters such as \n and \r to keep logs being 1 line
+                            // (2) Escaping double-quotes (") to be single-quotes (') because some fields in our regex string are
+                            //     deliniated by double-quotes. Note, this was a convention copied from the Functions Host regex
+                            //     which also uses double-quotes to capture columns that may contain commas inside them.
+                            // (3) Escaping commas (",") for ";;" because commas separate our columns and so they have the potential to
+                            //     disrupt parsing.
+                            // Note: In retrospective, perhaps the regex-string could have been designed to avoid these awkward
+                            // parsing problems, but it wasn't because (1) it followed conventions from other regex-strings in our system
+                            // and because we asssumed we'd have a JSON-based logger that would have avoided these problems.
+                            val = (string)val;
```
davidmrdavid, line 256: ah, these two variables have the same name. That's not great. I'll give them different ones. Also, I suppose JTokens can always be casted to string, so this try-catch and probably be removed

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -57,12 +130,24 @@ internal class LinuxAppServiceLogger
...
+            }
+
+            this.tenant = tenant;
+
+
+            if (!string.IsNullOrEmpty(stampName))
+            {
+                this.stamp = stampName;
+                var finalCharIndex = stampName.Length - 1;
+                this.primaryStamp = char.IsLetter(stampName[finalCharIndex]) ? stampName.Remove(finalCharIndex) : stampName;
```
cgillum, line 149: I recommend we add a TODO here to update this logic to account for ASEs.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -129,9 +227,54 @@ private string GenerateJsonStr(EventWrittenEventArgs eventData)
...
 
-            // Generate string-representation of JSON. Also remove newlines.
-            string jsonString = json.ToString(Newtonsoft.Json.Formatting.None);
-            jsonString = jsonString.Replace("\n", "\\n").Replace("\r", "\\r");
+            string jsonString = "";
+            if (!this.writeToConsole)
+            {
+                // This path supports the legacy regex-based parser in Linux Dedicated.
+                // In the future, we'll be able to remove this and use JSON logging only
+                IList<string> jsonKeys = json.Properties().Select(p => p.Name).ToList();
```
cgillum, line 235: Looks like `jsonKeys` is no longer used and can be deleted.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -129,9 +227,54 @@ private string GenerateJsonStr(EventWrittenEventArgs eventData)
...
+                    regexValues.Add(val);
+                }
+
+                jsonString = string.Join(",", regexValues);
+            }
+            else
+            {
+                // Generate string-representation of JSON. Also remove newlines.
+                jsonString = json.ToString(Newtonsoft.Json.Formatting.None);
+                jsonString = jsonString.Replace("\n", "\\n").Replace("\r", "\\r");
```
cgillum, line 275: Is this last step necessary? I did a quick test an it seems like `JObject.ToString()` already replaces newline characters with `\\n`, etc. If we don't already, it would be great to have a test that confirms the correct behavior.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -129,9 +227,54 @@ private string GenerateJsonStr(EventWrittenEventArgs eventData)
                 json.Add("RelatedActivityId", eventData.RelatedActivityId);
             }
 
-            // Generate string-representation of JSON. Also remove newlines.
-            string jsonString = json.ToString(Newtonsoft.Json.Formatting.None);
-            jsonString = jsonString.Replace("\n", "\\n").Replace("\r", "\\r");
+            string jsonString = "";
```
cgillum, line 230: Since we might actually be writing a comma-separated string, I recommend not calling this `jsonString`. It makes reading the code confusing.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
davidmrdavid: I tested that too and, indeed, we can remove it

cgillum (approved):
No blockers detected! Thanks for continuing to push on this!

ConnorMcMahon (commented):
I realize I forgot to add this small feedback last week.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -355,22 +380,35 @@ public async Task RemovesNewlinesFromExceptions()
...
+            // which corresponds to the number of JSONs logged
+
             int countTimeStampCols = Regex.Matches(string.Join("", lines), "\"TimeStamp\":").Count;
             Assert.Equal(lineCount, countTimeStampCols);
+            */
+
+            // If every line can be parsed, we know newlines were removed
+            foreach (string line in lines)
+            {
+                System.Text.RegularExpressions.Match match = Regex.Match(line, TestHelpers.RegexPattern);
```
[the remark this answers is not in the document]
davidmrdavid: Kind of. We do check that the consumption output can be parsed as JSON, and that it contains to minimal fields :) , as seen below: 

https://github.com/Azure/azure-functions-durable-extension/blob/18e13e1096251d61a54852d23608209cd597f37b/test/Common/DurableTaskEndToEndTests.cs#L271-L297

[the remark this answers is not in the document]
davidmrdavid: I left some debugging statements in there I should remove (the `this.output.WriteLine` bit). I'll remove it shortly

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -487,6 +524,135 @@ public async Task OutputsValidJSONLogs()
...
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        public async Task NoExceptionsWhenEnvVarsAreMissing()
+        {
+            // Set a different logging path, since the CI is Windows-based instead of linux.
+            LinuxAppServiceLogger.LoggingPath = Path.Combine(Directory.GetCurrentDirectory(), "logfile.log");
+            File.Delete(LinuxAppServiceLogger.LoggingPath); // To ensure the test generates the path
+            string orchestratorName = nameof(TestOrchestrations.ThrowOrchestrator);
+
+            // Simulate linux dedicated via enviroment variables
+            var nameResolver = new SimpleNameResolver(new Dictionary<string, string>()
```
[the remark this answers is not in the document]
davidmrdavid: You're right. The test suite has been evolving with the many iterations of this and, eventually, this one just became redundant. Thanks for the suggestion: removing it in the upcoming commit

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -333,8 +334,166 @@ public async Task WritesToFile()
...
+                        this.output.WriteLine(line);
+                        JObject json = JObject.Parse(jsonStr);
+
+                        List<string> keys = json.Properties().Select(p => p.Name).ToList();
+                        Assert.Contains("EventStampName", keys);
+                        Assert.Contains("EventPrimaryStampName", keys);
+                        Assert.Contains("ProviderName", keys);
+                        Assert.Contains("TaskName", keys);
+                        Assert.Contains("EventId", keys);
+                        Assert.Contains("EventTime", keys);
```
cgillum, line 402: I was going through email just now and noticed a person from CSS talking about how they need to use the `EventTimestamp` field of the FunctionsLogs table for Linux apps to get more accurate timestamps (compared to PreciseTimeStamp). However, here we're using `EventTime`. Can you double check to make sure we shouldn't actually be using `EventTimestamp` as this column name?

davidmrdavid: You're right, that's what it should be. Thanks for the heads up.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -72,18 +161,33 @@ internal class LinuxAppServiceLogger
...
+                    "<staticLogFileName value=\"true\" />" +
+                    "</appender>" +
+                    "<root>" +
+                    "<level value=\"ALL\" />" +
+                    "<appender-ref ref=\"RollingFile\" />" +
+                    "</root>" +
+                    "</log4net>";
+                var xml = new XmlDocument();
+                xml.LoadXml(xmlStr);
+
```
davidmrdavid, line 186: I'd like to know if we're ok with maximumFileSize being set to 300MB. If so, after these currently running tests pass, I'm pretty sure we're ready to release

davidmrdavid: @ConnorMcMahon

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -342,15 +343,20 @@ private void InitializeLinuxLogging()
                 // The logging service for linux works by capturing EventSource messages,
                 // which our linux platform does not recognize, and logging them via a
                 // different strategy such as writing to console or to a file.
-                this.eventSourceListener = new EventSourceListener(linuxLogger);
+
+                // Since our logging payload can be quite large, linux telemetry by default
+                // disables verbose-level telemetry to avoid a performance hit.
+                bool enableVerbose = this.Options.Tracing.AllowVerboseLinuxTelemetry;
+                this.eventSourceListener = new EventSourceListener(linuxLogger, enableVerbose);
```
davidmrdavid, line 350: Added a new tracing option to let customers enable verbose telemetry in linux, disabled by default

src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
```diff
@@ -52,7 +55,13 @@ protected override void OnEventSourceCreated(EventSource eventSource)
...
         protected override void OnEventWritten(EventWrittenEventArgs eventData)
         {
-            this.logger.Log(eventData);
+            // When disabling verbose logs, we skip Verbose DurableTask-Core telemetry
+            if (!(this.disableVerbose
+               && eventData.EventSource.Name == "DurableTask-Core"
+               && eventData.Level == EventLevel.Verbose))
+            {
+                this.logger.Log(eventData);
+            }
```
davidmrdavid, line 64: Logic to filter verbose events from DurableTask-Core

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -309,21 +331,174 @@ public async Task WritesToFile()
...
+                        Assert.Contains("Tenant", keys);
+                        Assert.Contains("Pid", keys);
+                        Assert.Contains("Tid", keys);
+
+                        // Ensuring no DurableTask-Core Verbose logs are found
+                        if ((int)json["Level"] == (int)EventLevel.Verbose)
+                        {
+                            Assert.False(json["ProviderName"].Equals("DurableTask-Core"));
+                        }
+                    }
```
davidmrdavid, line 406: Testing that filtering verbose durable-task-core events is "on" by default

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -309,21 +331,174 @@ public async Task WritesToFile()
...
+                        Assert.Contains("Tenant", keys);
+                        Assert.Contains("Pid", keys);
+                        Assert.Contains("Tid", keys);
+
+                        // Ensuring DurableTask-Core Verbose logs are found
+                        if (((int)json["Level"] == (int)EventLevel.Verbose)
+                            && string.Equals((string)json["ProviderName"], "DurableTask-Core"))
+                        {
+                            foundVerboseLog = true;
+                        }
```
davidmrdavid, line 483: Test to make sure that verbose logs can be enabled via Tracing Options

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
ConnorMcMahon: I don't necessarily feel comfortable setting 300MB. I'm checking what LWAWS uses, but the biggest I have seen so far is 10MB.

davidmrdavid: resolving since we're using another logger now

src/WebJobs.Extensions.DurableTask/BasicLogWriter.cs
```diff
@@ -0,0 +1,134 @@
...
+                        {
+                            File.Delete(this.LogFileBackup);
+                        }
+                        catch (Exception)
+                        {
+                        }
+                    }
+#if !FUNCTIONS_V1
+                    Syscall.rename(this.LogFile, this.LogFileBackup);
+                    this.fileSize = 0;
```
cgillum, line 92: While this works great for our purposes, it's weird to me that we have this `#if` condition here but no `#else` condition. Maybe just for good feelings we could add an `#else` that does a `File.Move(this.LogFile, this.LogFileBackup)`?

src/WebJobs.Extensions.DurableTask/BasicLogWriter.cs
```diff
@@ -0,0 +1,134 @@
...
+                AppendAllBytes(this.LogFile, dataToWrite);
+            }
+            catch (Exception)
+            {
+            }
+        }
+
+        private static void AppendAllBytes(string fileName, byte[] data)
+        {
+            using (FileStream fileStream = new FileStream(fileName, File.Exists(fileName) ? FileMode.Append : FileMode.OpenOrCreate, FileAccess.Write))
```
cgillum, line 115: Non-blocking, but I wonder if we could reduce disk I/O by caching the `File.Exists` result, or checking if `this.fileSize > 0` instead of checking `File.Exists(fileName)`. I suppose this is more robust, though, in that it gracefully handles the case where someone manually deletes the log file while things are running, so maybe we should leave it as-is since our perf numbers are already good enough.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceFileLogger.cs
```diff
@@ -0,0 +1,174 @@
...
+        private readonly string logFileDirectory;
+        private readonly string logFilePath;
+        private readonly string archiveFilePath;
+        private readonly BlockingCollection<string> buffer;
+        private readonly List<string> currentBatch;
+        private readonly CancellationTokenSource cancellationTokenSource;
+        private Task outputTask;
+
+        /// <summary>
+        /// TBD.
```
cgillum, line 33: Do you plan on filling in these TBDs or should be just remove them?

src/WebJobs.Extensions.DurableTask/LinuxAppServiceFileLogger.cs
```diff
@@ -0,0 +1,174 @@
...
+                    // Ignored
+                }
+
+                this.currentBatch.Clear();
+            }
+        }
+
+        private async Task WriteLogs(IEnumerable<string> currentBatch)
+        {
+            Directory.CreateDirectory(this.logFileDirectory);
```
cgillum, line 139: What happens if the directory already exists?

src/WebJobs.Extensions.DurableTask/BasicLogWriter.cs
cgillum: Oh, maybe we're not using this anymore...

src/WebJobs.Extensions.DurableTask/BasicLogWriter.cs
cgillum: Oh, maybe we're not using this anymore...

src/WebJobs.Extensions.DurableTask/LinuxAppServiceFileLogger.cs
davidmrdavid: @cgillum I do plan on filling them. I just pushed that commit / did the port late last night so I didn't want to necessarily comment at that point, especially because  I did not know what the performance results would be. I'll get to it shortly

src/WebJobs.Extensions.DurableTask/LinuxAppServiceFileLogger.cs
davidmrdavid: Nothing! The semantics of CreateDirectory are to do nothing if the path exists, no exceptions either

src/WebJobs.Extensions.DurableTask/BasicLogWriter.cs
davidmrdavid: We're not! I just pushed the port late last night so I didn't want to necessarily remove this at that point. We also still have the log4net dependency somewhere in here :p . I'll get to it shortly

src/WebJobs.Extensions.DurableTask/LinuxAppServiceFileLogger.cs
```diff
@@ -0,0 +1,175 @@
...
+using System.Threading.Tasks;
+
+#if !FUNCTIONS_V1
+using Mono.Unix.Native;
+#endif
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    /// <summary>
+    /// The File logger for linux dedicated. Manages file rolling and is concurrency-safe.
```
ConnorMcMahon, line 19: Can we add a comment where we copied this from, with some notes about how we changed this?

src/WebJobs.Extensions.DurableTask/Options/TraceOptions.cs
```diff
@@ -26,6 +26,19 @@ public class TraceOptions
...
+        /// </summary>
+        /// <remarks>
+        /// The default behaviour is false, which disables verbose logs. When set
+        /// to true, performance may be affected due to the amount of verbose logs.
+        /// We recommend setting this to true primarily for debugging purposes.
+        /// <value>
+        /// <c>true</c> to enable verbose telemetry; <c>false</c> otherwise.
+        /// </value>
+        /// </remarks>
+        public bool AllowVerboseLinuxTelemetry { get; set; }
```
ConnorMcMahon, line 40: nit: This could just be me, but I generally like to explicitly default this to `false`, just so it is a bit more apparent what the default value is by looking. 

Some people would probably disagree with me about that, as default values for `bool` in .NET should be common knowledge.

davidmrdavid: I agree with you. I just didn't want to mess with the format of this too much. I'll get on it

src/WebJobs.Extensions.DurableTask/LinuxAppServiceFileLogger.cs
davidmrdavid: Good point, will do

src/WebJobs.Extensions.DurableTask/Options/TraceOptions.cs
davidmrdavid: Done!

src/WebJobs.Extensions.DurableTask/LinuxAppServiceFileLogger.cs
davidmrdavid: Done!

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -357,8 +357,7 @@ private void InitializeLinuxLogging()
         public void Dispose()
         {
             // Not flushing the linux logger may lead to lost logs
-            int timeoutMilliseconds = 3000;
-            LinuxAppServiceLogger.Logger?.Stop(new TimeSpan(timeoutMilliseconds));
+            LinuxAppServiceLogger.Logger?.Stop(TimeSpan.FromSeconds(10));
```
ConnorMcMahon, line 360: Is there a reason we transitioned from 3 to 10 seconds?

davidmrdavid: I figured it wouldn't hurt to buy us a bit more time, especially during very noisy orchestrations. Ideally, we'd never hit the timeout anyways . I can revert back to 3 seconds if you feel strongly though :)

ConnorMcMahon: I don't have any particularly strong concerns, was just curious.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceFileLogger.cs
```diff
@@ -111,6 +111,8 @@ private async Task ProcessLogQueue(object state)
                 await Task.Delay(TimeSpan.FromSeconds(this.FlushFrequencySeconds), this.cancellationTokenSource.Token).ContinueWith(task => { });
             }
 
+            await this.InternalProcessLogQueue();
```
ConnorMcMahon, line 114: I'm not sure if it is best practices to await after a cancellation token is cancelled. @cgillum, thoughts?

ConnorMcMahon: To clarify, I think it makes sense to, I'm just not sure what the expectations are in .NET when a token is cancelled.

davidmrdavid: Got it. I see how this could be controversial.  Note that this gets called in here:

```C#
        /// <summary>
        /// Flushes the write buffer, stops writing to logfile afterwards.
        /// </summary>
        /// <param name="timeSpan">Timeout in milliseconds for flushing task.</param>
        public void Stop(TimeSpan timeSpan)
        {
            this.cancellationTokenSource.Cancel(); // this is not awaited

            try
            {
                this.outputTask?.Wait(timeSpan);
            }
            catch (Exception)
            {
                // ignored
            }
        }
```

which in turn is called in the dispose method of DurableTaskExtension.cs

```C#
        public void Dispose()
        {
            // Not flushing the linux logger may lead to lost logs
            // 40 seconds timeout because we write normally every 30 seconds, so we're just
            // adding an extra 10 seconds to flush.
            LinuxAppServiceLogger.Logger?.Stop(TimeSpan.FromSeconds(40));
            this.HttpApiHandler?.Dispose();
            this.eventSourceListener?.Dispose();
        }
```

davidmrdavid: Without the highlighted change though, cancelling a token while the buffer still has contents can result is missed logs. I've also already notified the azure-functions-host folks about this potential problem

cgillum: I think of CancellationToken as just a tool for delivering notifications. Any best practices as far as what to do with that notification would be specific to the application. In this case, I agree we should flush out our log buffer. You just may need to be careful about potential race conditions. For example, the listener could be disposed at the same time that we're flushing the logs. We just need to be careful that this won't cause any problems for us.

davidmrdavid: I'm giving the flush operation _at minimum_ 10 seconds to flush, and 40 seconds in the best case scenario. I believe that _should_ be enough if we assume the app is otherwise in a healthy state.

cgillum (commented):
LGTM! - :shipit:

PR #1528, opened 2020-10-09
Stop logging replays to EventSource by default

cgillum:
Part of our effort to shrink our logging footprint in App Service.  The previous behavior can be re-enabled by setting `traceReplayEvents: true` in host.json.

This PR is meant to be merged into https://github.com/Azure/azure-functions-durable-extension/pull/1522.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
@@ -102,19 +102,19 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             FunctionType functionType,
             bool isReplay)
         {
-            EtwEventSource.Instance.FunctionScheduled(
-                hubName,
-                LocalAppName,
-                LocalSlotName,
-                functionName,
-                instanceId,
-                reason,
-                functionType.ToString(),
-                ExtensionVersion,
-                isReplay);
-
             if (this.ShouldLogEvent(isReplay))
             {
+                EtwEventSource.Instance.FunctionScheduled(
+                    hubName,
+                    LocalAppName,
+                    LocalSlotName,
+                    functionName,
+                    instanceId,
+                    reason,
+                    functionType.ToString(),
+                    ExtensionVersion,
+                    isReplay);
+
                 this.logger.LogInformation(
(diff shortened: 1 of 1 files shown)


davidmrdavid:
@cgillum , this seems to be failing because of the new tests I added to validate linux logging. Apparently, there's some kind of weird race condition where the file is being read while still in use by another process. I have personally not seen that in a while in my testing, so I'm a little concerned, but I wouldn't mind commenting those out during development and for the sake of merging this into my other PR

davidmrdavid:
@cgillum Do you mind re-running these tests one more time, just in case they were simply being flaky?

ConnorMcMahon (approved):
LGTM!

cgillum:
@davidmrdavid I got all the same errors in the second run. Should I wait to merge your file logging changes and then try again?

davidmrdavid:
@cgillum , let me first check that the tests are still passing on my PR

cgillum:
Alternatively I can rebase this on `dev`, merge it if it passes, and then you can merge it into your branch.

davidmrdavid:
@cgillum, my PR is now passing the tests. Pulling and merging with that branch, and a re-run of the tests, should make this one pass as well. If possible, please give that a shot and, if it passes, let's merge it directly. Otherwise, I'm happy with just rebasing on dev :) .

davidmrdavid (approved):
LGTM. After this, and adding an extra `EventTime` field for more accurate sorting, I think we'll be pretty much done!

The one thing that's unclear to me is how this is configurable. It isn't yet, right?

cgillum:
> 

@davidmrdavid take a look at the `ShouldLogEvent(...)` method. You'll see that the field it uses to determine whether or not to trace a replay (`traceReplayEvents`) goes back this line in DurableTaskExtension.cs:

```csharp
```

Any property on the `DurableTaskOptions` class (which is referred to by `this.Options` in the expression above) can be configured using host.json. In this case, it corresponds to the `traceReplayEvents` field that I mentioned in the PR description. You can find all of our host.json settings documented either in the code or in [docs.microsoft.com](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-bindings#durable-functions-2-0-host-json).

PR #1541, opened 2020-10-27
Allow Netherite provider to be implemented separately

sebastianburckhardt:
Support for using the EventSourced backend, as in the work-in-progress PR azure/durabletask#461.

Besides adding a durability provider, the changes include:
- turn off message sorter logic for backend that delivers messages in order
- support poll-free version of WaitForOrchestration for backends that support client responses
- change tracing options for unit tests

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -83,6 +83,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         string IDurableEntityClient.TaskHubName => this.TaskHubName;
 
+        /// <inheritdoc/>
+        public override string ToString()
+        {
+            return $"DurableClient[backend={this.config.GetBackendInfo()}]";
+        }
+
         /// <inheritdoc />
         HttpResponseMessage IDurableOrchestrationClient.CreateCheckStatusResponse(HttpRequestMessage request, string instanceId, bool returnInternalServerErrorOnFailure)
         {
@@ -541,10 +547,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                          tasks.Add(CheckForOrphanedLockAndFixIt(state, status.LockedBy));
                     }
 
-                    if (removeEmptyEntities && !status.EntityExists && status.LockedBy == null && status.QueueSize == 0
-                        && now - state.LastUpdatedTime > TimeSpan.FromMinutes(this.config.Options.EntityMessageReorderWindowInMinutes))
+                    if (removeEmptyEntities)
                     {
-                        tasks.Add(DeleteIdleOrchestrationEntity(state));
+                        bool isEmptyEntity = !status.EntityExists && status.LockedBy == null && status.QueueSize == 0;
+                        bool safeToRemoveWithoutBreakingMessageSorterLogic = now - state.LastUpdatedTime > this.config.MessageReorderWindow;
+                        if (isEmptyEntity && safeToRemoveWithoutBreakingMessageSorterLogic)
+                        {
+                            tasks.Add(DeleteIdleOrchestrationEntity(state));
+                        }
                     }
                 }
 
--- a/src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
@@ -58,6 +58,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         public virtual bool SupportsEntities => false;
 
         /// <summary>
+        /// Specifies whether the backend's WaitForOrchestration is implemented without polling.
+        /// </summary>
+        public virtual bool SupportsPollFreeWait => false;
+
+        /// <summary>
+        /// Specifies whether this backend delivers messages in order.
+        /// </summary>
+        public virtual bool GuaranteesOrderedDelivery => false;
+
+        /// <summary>
         /// JSON representation of configuration to emit in telemetry.
         /// </summary>
         public virtual JObject ConfigurationJson => EmptyConfig;
@@ -86,6 +96,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         /// <inheritdoc/>
         public int MaxConcurrentTaskActivityWorkItems => this.GetOrchestrationService().MaxConcurrentTaskActivityWorkItems;
+
+        internal string GetBackendInfo()
+        {
+            return this.GetOrchestrationService().ToString();
+        }
 
         private IOrchestrationService GetOrchestrationService()
         {
@@ -278,7 +293,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// <returns>Returns a task which completes when the state has been fetched.</returns>
         public virtual Task<IList<OrchestrationState>> GetOrchestrationStateWithInputsAsync(string instanceId, bool showInput = true)
         {
-            throw this.GetNotImplementedException(nameof(this.GetOrchestrationStateAsync));
+            throw this.GetNotImplementedException(nameof(this.GetOrchestrationStateWithInputsAsync));
         }
 
         /// <summary>
@@ -417,6 +432,52 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         internal virtual bool ConnectionNameMatches(DurabilityProvider durabilityProvider)
         {
             return this.ConnectionName.Equals(durabilityProvider.ConnectionName);
+        }
+
+        /// <summary>
+        /// Returns the instance id of the entity scheduler for a given entity id.
+        /// </summary>
+        /// <param name="entityId">The entity id.</param>
+        /// <returns>The instance id of the scheduler.</returns>
+        protected string GetSchedulerIdFromEntityId(EntityId entityId)
+        {
+            return EntityId.GetSchedulerIdFromEntityId(entityId);
+        }
+
+        /// <summary>
+        /// Reads the state of an entity from the serialized entity scheduler state.
+        /// </summary>
+        /// <param name="state">The orchestration state of the scheduler.</param>
+        /// <param name="serializerSettings">The serializer settings.</param>
+        /// <param name="result">The serialized state of the entity.</param>
+        /// <returns>true if the entity exists, false otherwise.</returns>
+        protected bool TryGetEntityStateFromSerializedSchedulerState(OrchestrationState state, JsonSerializerSettings serializerSettings, out string result)
+        {
+            if (state != null
+                && state.OrchestrationInstance != null
+                && state.Input != null)
+            {
+                var schedulerState = JsonConvert.DeserializeObject<SchedulerState>(state.Input, serializerSettings);
+
+                if (schedulerState.EntityExists)
+                {
+                    result = schedulerState.EntityState;
+                    return true;
+                }
+            }
+
+            result = null;
+            return false;
+        }
+
+        /// <summary>
+        /// Converts the DTFx representation of the orchestration state into the DF representation.
+        /// </summary>
+        /// <param name="orchestrationState">The orchestration state.</param>
+        /// <returns>The orchestration status.</returns>
+        protected DurableOrchestrationStatus ConvertOrchestrationStateToStatus(OrchestrationState orchestrationState)
+        {
+            return DurableClient.ConvertOrchestrationStateToStatus(orchestrationState);
         }
     }
 }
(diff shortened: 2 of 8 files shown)


davidmrdavid:
@sebastianburckhardt , if this is WIP, should we consider making it a draft PR? :)

sebastianburckhardt:
Not sure what the difference is between draft and WIP?

Since we just decided to restructure the projects I can't be quite sure what will change. Currently figuring this out. My guess is that in its eventual form, this PR will stay similar, as far as modifications to existing files are concerned, but it will not include the new durability provider.

davidmrdavid:
@sebastianburckhardt , a _draft_ is a GitHub UI feature to acknowledge that a PR is work-in-progress. I've used it in the past to make a clean separation between PRs that need immediate reviews for immediate merge consideration and the PRs that do not. You can read about it here: 
https://github.blog/2019-02-14-introducing-draft-pull-requests/

Essentially if you CTRL+F "draft" in this page, you should be able to convert this into a draft PR. Just a suggestion!

sebastianburckhardt:
This is now ready to be reviewed. It only contains the changes needed for implementing Netherite as an external NuGet package. The changes include:

- turn off message sorter logic for backend that delivers messages in order
- support poll-free version of WaitForOrchestration for backends that support client responses
- add three protected methods to DurabilityProvider that allow implementations to access functionality that is otherwise internal

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -71,6 +71,11 @@ internal class DurableClient : IDurableClient,
 
         string IDurableEntityClient.TaskHubName => this.TaskHubName;
 
+        public override string ToString()
```
davidmrdavid, line 74: I realize this might be trivial but, since this is a public method, shouldn't it at least inherit its documentation?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -530,7 +535,7 @@ async Task<CleanEntityStorageResult> IDurableEntityClient.CleanEntityStorageAsyn
                     }
 
                     if (removeEmptyEntities && !status.EntityExists && status.LockedBy == null && status.QueueSize == 0
-                        && now - state.LastUpdatedTime > TimeSpan.FromMinutes(this.config.Options.EntityMessageReorderWindowInMinutes))
+                        && now - state.LastUpdatedTime > this.config.MessageReorderWindow)
```
davidmrdavid, line 538: nitpick: since this is such a long conditional check, I would recommend splitting it into smaller, well-named, boolean variables. It would improve readability and help us detect bugs in the future

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -169,7 +169,24 @@ private static TemplateMatcher GetInstanceRaiseEventRoute()
...
+                if (client is DurableClient durableClient && durableClient.DurabilityProvider.SupportsPollFreeWait)
+                {
+                    try
+                    {
+                        var state = await durableClient.DurabilityProvider.WaitForOrchestrationAsync(instanceId, null, timeout, CancellationToken.None);
+                        status = DurableClient.ConvertOrchestrationStateToStatus(state);
+                    }
+                    catch (TimeoutException)
+                    {
+                    }
```
davidmrdavid, line 183: Why don't we take an action in case of a Timeout? It would be good to document this :)

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -169,7 +169,24 @@ private static TemplateMatcher GetInstanceRaiseEventRoute()
...
+                }
+                else
+                {
+                    status = await client.GetStatusAsync(instanceId);
+                }
+
```
davidmrdavid, line 189: Can we document / comment about the difference between these two execution paths?

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -169,7 +169,24 @@ private static TemplateMatcher GetInstanceRaiseEventRoute()
             Stopwatch stopwatch = Stopwatch.StartNew();
             while (true)
             {
-                DurableOrchestrationStatus status = await client.GetStatusAsync(instanceId);
+                DurableOrchestrationStatus status = null;
+
+                if (client is DurableClient durableClient && durableClient.DurabilityProvider.SupportsPollFreeWait)
```
davidmrdavid, line 174: The syntax of `client is DurableClient durableClient` confuses me a bit. I would have expected this to be simply `client is DurableClient`, which would be a type-check. What is this achieving?

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -704,6 +721,10 @@ private static bool TryGetIntQueryParameterValue(NameValueCollection queryString
                 {
                     return await client.WaitForCompletionOrCreateCheckStatusResponseAsync(request, id, timeout.Value, pollingInterval.Value);
                 }
+                else if (timeout.HasValue && client is DurableClient durableClient && durableClient.DurabilityProvider.SupportsPollFreeWait)
+                {
+                    return await client.WaitForCompletionOrCreateCheckStatusResponseAsync(request, id, timeout.Value, timeout.Value);
```
davidmrdavid, line 726: Can we document how this else-if is different from the pre-existing if-statement above? They seem quite similar. 

In fact, there might be a refactor we we only check for `if (timeout.HasValue)` and inside of that we switch the final parameter to `WaitForCompletionOrCreateCheckStatusResponseAsync` depending on some differentiating condition.

I do not feel strongly about a refactor like that, but I would recommend a quick comment explaining the difference. Thanks!

davidmrdavid (commented):
Looks pretty good to me, I give it my thumbs-up but I do have suggestions for improvements, please see my comments.

I'll wait for a second pair of eyes to give this the approval 😄

davidmrdavid:
Ah it also seems like there's a merge conflict with `src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs`

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
sebastianburckhardt: yes, it should. Will fix.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
sebastianburckhardt: the name 'durableClient' gets bound and is used right after the &&.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
sebastianburckhardt: Yes, I agree. I am splitting this apart now.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
sebastianburckhardt: Yes, adding comment.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
sebastianburckhardt: Yes.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
sebastianburckhardt: I changed the logic for this a bit, and added a comment.

davidmrdavid (commented):
Really happy with the latest changes. LGTM! Still waiting on a 2nd pair of eyes just in case

davidmrdavid:
Re-started the CI build checks since it appears to have been running for over an hour. Sometimes this happens, _for some reason,_ with the CI. Hoping they pass now!

src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
```diff
@@ -418,5 +433,51 @@ internal virtual bool ConnectionNameMatches(DurabilityProvider durabilityProvide
...
         }
+
+        /// <summary>
+        /// Returns the instance id of the entity scheduler for a given entity id.
+        /// </summary>
+        /// <param name="entityId">The entity id.</param>
+        /// <returns>The instance id of the scheduler.</returns>
+        protected string GetSchedulerIdFromEntityId(EntityId entityId)
+        {
+            return EntityId.GetSchedulerIdFromEntityId(entityId);
```
ConnorMcMahon, line 444: Is there a reason this is now a `protected` method, as opposed to having people just directly call `EntityId.GetScheduleIdFromEntityId`? If it's because we want to allow each provider to specify their own mapping, then shouldn't this be `virtual`?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -1,1243 +1,1251 @@
...
-                string statusStr = Enum.GetName(status.GetType(), status);
-                activity.AddTag("DurableFunctionsType", functionsType);
-                activity.AddTag("DurableFunctionsInstanceId", instanceId);
-                activity.AddTag("DurableFunctionsRuntimeStatus", statusStr);
-            }
-        }
-#endif
-    }
-}
+// Copyright (c) .NET Foundation. All rights reserved.
```
ConnorMcMahon, line 1: Would it be possible to undo the whitespace diffs here? It makes it difficult to remove the changes in this file.

src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
```diff
@@ -418,5 +433,51 @@ internal virtual bool ConnectionNameMatches(DurabilityProvider durabilityProvide
...
+        }
+
+        /// <summary>
+        /// Converts the DTFx representation of the orchestration state into the DF representation.
+        /// </summary>
+        /// <param name="orchestrationState">The orchestration state.</param>
+        /// <returns>The orchestration status.</returns>
+        protected DurableOrchestrationStatus ConvertOrchestrationStateToStatus(OrchestrationState orchestrationState)
+        {
+            return DurableClient.ConvertOrchestrationStateToStatus(orchestrationState);
```
ConnorMcMahon, line 480: Similarly here, I'm not sure I love the idea of adding protected methods to give us access to some internal static method.  Would it be easier to just to make these static methods public (and maybe put them in a more aptly named class if other storage providers need access to this method)?

ConnorMcMahon (commented):
Overall looks fairly good, just a few questions.

src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
sebastianburckhardt: It is a private method, so the point was to make it "semi-public", meaning accessible to implementors of DurabilityProvider. That way it does not pollute the public API, e.g. normal users won't see it in IntelliSense.

src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
sebastianburckhardt: I suppose I did not think of a case where anybody other than the DurabilityProvider implementor needs access.  Though I guess AzureStorage provider is an example where one may want access without going via a DurabilityProvider. Perhaps a static class would be better, such as ProviderUtils so users know not to look at it unless they implement a provider. I will do that.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
sebastianburckhardt: Not sure what is happening here. I don't see this problem on my local checkout - which shows a normal diff.

ConnorMcMahon: If that doesn't work, maybe the quickest workaround would just be to add a brief GitHub PR comment next to where the largest changes are so we know where to look.

sebastianburckhardt: I think it worked, the diff looks normal now.

src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
sebastianburckhardt: I have pushed this change also.

ConnorMcMahon (approved):
LGTM!

PR #1545, opened 2020-11-03
Added RestartAsync API to rerun existing orchestrator instances

### LINKED ISSUE #1243: Add webhook to rerun failed (or terminated) orchestrations
At this moment, there are a few operations / endpoints when using Durable Functions (IDurableOrchestrationContext):

* statusQueryGetUri

* sendEventPostUri

* terminatePostUri

* purgeHistoryDeleteUri

I would like the possibility to re-execute a workflow, and the reason is as I'm not the caller of the Starter function so I cannot provide the same input parameters. I think this feature could be useful specially when due some activity exceeds the retry count, but after some time, there are no problems anymore and I need to reprocess.

amdeel:
Added RestartAsync(string instanceId, bool restartWithNewInstanceId = true) API to rerun orchestrators. Given a valid instance Id of an existing orchestrator, a new orchestrator will be started reusing the same input.

restartWithNewInstanceId is an optional input determining if the started orchestrator instance should reuse the same instance Id or be started with a new instance Id.

resolves #1243

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -72,16 +72,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         string IDurableEntityClient.TaskHubName => this.TaskHubName;
 
         /// <inheritdoc />
-        HttpResponseMessage IDurableOrchestrationClient.CreateCheckStatusResponse(HttpRequestMessage request, string instanceId, bool returnInternalServerErrorOnFailure, bool restartWithNewInstanceId)
+        HttpResponseMessage IDurableOrchestrationClient.CreateCheckStatusResponse(HttpRequestMessage request, string instanceId, bool returnInternalServerErrorOnFailure)
         {
-            return this.CreateCheckStatusResponse(request, instanceId, this.attribute, returnInternalServerErrorOnFailure, restartWithNewInstanceId);
+            return this.CreateCheckStatusResponse(request, instanceId, this.attribute, returnInternalServerErrorOnFailure);
         }
 
         /// <inheritdoc />
-        IActionResult IDurableOrchestrationClient.CreateCheckStatusResponse(HttpRequest request, string instanceId, bool returnInternalServerErrorOnFailure, bool restartWithNewInstanceId)
+        IActionResult IDurableOrchestrationClient.CreateCheckStatusResponse(HttpRequest request, string instanceId, bool returnInternalServerErrorOnFailure)
         {
             HttpRequestMessage requestMessage = ConvertHttpRequestMessage(request);
-            HttpResponseMessage responseMessage = ((IDurableOrchestrationClient)this).CreateCheckStatusResponse(requestMessage, instanceId, returnInternalServerErrorOnFailure, restartWithNewInstanceId);
+            HttpResponseMessage responseMessage = ((IDurableOrchestrationClient)this).CreateCheckStatusResponse(requestMessage, instanceId, returnInternalServerErrorOnFailure);
             return ConvertHttpResponseMessage(responseMessage);
         }
 
@@ -791,10 +791,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             HttpRequestMessage request,
             string instanceId,
             DurableClientAttribute attribute,
-            bool returnInternalServerErrorOnFailure = false,
-            bool restartWithNewInstanceId = true)
+            bool returnInternalServerErrorOnFailure = false)
         {
-            return this.httpApiHandler.CreateCheckStatusResponse(request, instanceId, attribute, returnInternalServerErrorOnFailure, restartWithNewInstanceId);
+            return this.httpApiHandler.CreateCheckStatusResponse(request, instanceId, attribute, returnInternalServerErrorOnFailure);
         }
 
         private static void TrackNameAndScheduledTime(JObject historyItem, EventType eventType, int index, Dictionary<string, EventIndexDateMapping> eventMapper)
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -72,16 +72,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         string IDurableEntityClient.TaskHubName => this.TaskHubName;
 
         /// <inheritdoc />
-        HttpResponseMessage IDurableOrchestrationClient.CreateCheckStatusResponse(HttpRequestMessage request, string instanceId, bool returnInternalServerErrorOnFailure)
+        HttpResponseMessage IDurableOrchestrationClient.CreateCheckStatusResponse(HttpRequestMessage request, string instanceId, bool returnInternalServerErrorOnFailure, bool restartWithNewInstanceId)
         {
-            return this.CreateCheckStatusResponse(request, instanceId, this.attribute, returnInternalServerErrorOnFailure);
+            return this.CreateCheckStatusResponse(request, instanceId, this.attribute, returnInternalServerErrorOnFailure, restartWithNewInstanceId);
         }
 
         /// <inheritdoc />
-        IActionResult IDurableOrchestrationClient.CreateCheckStatusResponse(HttpRequest request, string instanceId, bool returnInternalServerErrorOnFailure)
+        IActionResult IDurableOrchestrationClient.CreateCheckStatusResponse(HttpRequest request, string instanceId, bool returnInternalServerErrorOnFailure, bool restartWithNewInstanceId)
         {
             HttpRequestMessage requestMessage = ConvertHttpRequestMessage(request);
-            HttpResponseMessage responseMessage = ((IDurableOrchestrationClient)this).CreateCheckStatusResponse(requestMessage, instanceId, returnInternalServerErrorOnFailure);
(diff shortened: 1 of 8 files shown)


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -72,16 +72,16 @@ internal class DurableClient : IDurableClient,
         string IDurableEntityClient.TaskHubName => this.TaskHubName;
 
         /// <inheritdoc />
-        HttpResponseMessage IDurableOrchestrationClient.CreateCheckStatusResponse(HttpRequestMessage request, string instanceId, bool returnInternalServerErrorOnFailure)
+        HttpResponseMessage IDurableOrchestrationClient.CreateCheckStatusResponse(HttpRequestMessage request, string instanceId, bool returnInternalServerErrorOnFailure, bool restartWithNewInstanceId)
```
ConnorMcMahon, line 75: I don't know if adding a parameter to this API and the corresponding interfaces is worth it. We did this for `returnInternalServerErrorOnFailure` to ease the transition for people who needed the legacy behavior from Functions V1, but this is a new API, and adding parameters will break customer's unit tests.

I would say we just stick to a default behavior (new instance id?), and if customers want the other behavior they can tweak the URL returned.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -323,6 +334,10 @@ public async Task<HttpResponseMessage> HandleRequestAsync(HttpRequestMessage req
                         {
                             return await this.HandleRewindInstanceRequestAsync(request, instanceId);
                         }
+                        else if (string.Equals(operation, RestartOperation, StringComparison.OrdinalIgnoreCase))
```
ConnorMcMahon, line 337: I don't love the fact that we accept GET requests for this operation. I realize we do for Terminate/Rewind, but we should arguably fix that as a bug, as [we document](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-http-api#request-6) that it requires a `POST`.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -651,6 +666,18 @@ private static bool TryGetIntQueryParameterValue(NameValueCollection queryString
             return int.TryParse(value, out intValue);
         }
 
+        private static bool TryGetTimeSpanQueryParameterValue(NameValueCollection queryStringNameValueCollection, string queryParameterName, out TimeSpan? timeSpanValue)
+        {
+            timeSpanValue = null;
+            string value = queryStringNameValueCollection[queryParameterName];
+            if (value != null)
+            {
+                timeSpanValue = TimeSpan.FromSeconds(double.Parse(value));
```
ConnorMcMahon, line 675: It's not clear to me why a TryGetMethod is better than just returning a nullable TimeSpan, as the `HasValue` on nullable types serves the same purpose as the return type from this method.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -874,6 +875,19 @@ Task<string> IDurableOrchestrationClient.StartNewAsync<T>(string orchestratorFun
...
 
+        async Task<string> IDurableOrchestrationClient.RestartAsync(string instanceId, bool restartWithNewInstanceId)
+        {
+            DurableOrchestrationStatus status = await ((IDurableOrchestrationClient)this).GetStatusAsync(instanceId, showHistory: false, showHistoryOutput: false, showInput: true);
+
+            if (status == null)
+            {
+                throw new ArgumentException($"An orchestrastion with the instanceId {instanceId} was not found.");
+            }
+
```
ConnorMcMahon, line 886: We may want to do some validation of the current status. For instance, if the current status is running, then `restartWithNewInstanceId=false` will throw an exception. 

Also, what behavior do we want for instances that are still in the running state in general? Do we want to terminate the running orchestration?

ConnorMcMahon (changes requested):
It's a good start, but there are a few design things left to consider.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
amdeel: I do think it makes sense to honor the customers setting for starting orchestrators with the same instanceId.

However, whether we want to add a terminate step is a good question. I could easily add a call to TerminateAsync in this method. I can't think of a good reason to add this though, and it feels weird to force this behavior without a reason. We can choose to wait to add a potential terminate step or any other changes and see what customers think of this API and how they use it. We could release this as "in preview" until we have more feedback so if we do make changes they wont be so unexpected. What do you think?

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
amdeel: I did this mainly to match how we are getting query parameters elsewhere. I think this method was just accidentally left behind when the TryGets were added last year.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
amdeel: I see. Should I change these all to post in this PR while I'm here?

ConnorMcMahon: If you have time to, I would say let's do it. Otherwise, feel free to punt to next release, but then we should file an issue.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
ConnorMcMahon: Fair enough. Consistency is probably more important than orthodoxy here.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
ConnorMcMahon: Eh, I think you are right that the best approach would be to not do anything automatically, and note in the C# docs that it may throw an exception if the orchestration instance is not in a state that can be restarted.

We can always harden this API later by catching any exceptions thrown by `StartNewAsync()` and give a more helpful exception message down the line.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -352,7 +361,7 @@ public async Task<HttpResponseMessage> HandleRequestAsync(HttpRequestMessage req
                             return request.CreateResponse(HttpStatusCode.NotFound);
                         }
                     }
-                    else
+                    else if (request.Method == HttpMethod.Post)
```
ConnorMcMahon, line 364: We should document this as a bug fix in the release notes.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -682,6 +695,18 @@ private static bool TryGetIntQueryParameterValue(NameValueCollection queryString
             return int.TryParse(value, out intValue);
         }
 
+        private static bool TryGetTimeSpanQueryParameterValue(NameValueCollection queryStringNameValueCollection, string queryParameterName, out TimeSpan? timeSpanValue)
+        {
+            timeSpanValue = null;
+            string value = queryStringNameValueCollection[queryParameterName];
+            if (value != null)
+            {
+                timeSpanValue = TimeSpan.FromSeconds(double.Parse(value));
```
ConnorMcMahon, line 704: We should probably handle the case where they provide a non-numerical value for the timeout/polling interval and return a 400 instead of a 500. I think that needs to be handled in other places throughout this class, so it may be worth filing a separate issue to tackle later.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -752,6 +779,46 @@ private static bool TryGetIntQueryParameterValue(NameValueCollection queryString
...
+                if (TryGetBooleanQueryParameterValue(queryNameValuePairs, RestartWithNewInstanceId, out bool restartWithNewInstanceId))
+                {
+                    newInstanceId = await client.RestartAsync(instanceId, restartWithNewInstanceId);
+                }
+                else
+                {
+                    newInstanceId = await client.RestartAsync(instanceId);
+                }
+
+                if (TryGetTimeSpanQueryParameterValue(queryNameValuePairs, TimeoutParameter, out TimeSpan? timeout)
```
ConnorMcMahon, line 802: we should consider following the new logic Sebastian implemented in the StartOrchestration API for the netherite provider.

ConnorMcMahon: Honestly, in general, I don't love exposing the `WaitForCompletionOrCreateCheckStatusResponseAsync` option on the HTTP webhook URIs, as in app service HTTP requests are limited to 230 seconds anyways. I can see the argument for leaving it in, but since this is a new API we don't _need_ to include it.

ConnorMcMahon (commented):
Getting close to ready to merge. A few nits, but some can be put off into a different issue.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
amdeel: I see. In the other query parameter value methods we just ignore incorrect input. For now I should at least fix this method to do the same.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
amdeel: I see. That makes sense. I feel like because it is in the start orchestrator method I should have it in this restart method in case a customer was using that feature in their start method. I'll add the Sebastian logic in, though.

ConnorMcMahon (approved):
LGTM!

PR #1546, opened 2020-11-03
fix NullReferenceException in task hub worker shutdown

sebastianburckhardt:
I noticed that the task hub worker does not shut down correctly if some functions are disabled with the [Disable] attribute.

The reason is that the `info` variable can be null in the following checks that happen during shutdown:

```csharp
 if (this.isTaskHubWorkerStarted &&
                    !this.knownOrchestrators.Values.Any(info => info.HasActiveListener) &&
                    !this.knownActivities.Values.Any(info => info.HasActiveListener) &&
                    !this.knownEntities.Values.Any(info => info.HasActiveListener))
```

As far as I can see this is easily fixed by ignoring null entries for the purpose of this check.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -1134,11 +1134,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             using (await this.taskHubLock.AcquireAsync())
             {
+                bool hasActiveListeners(RegisteredFunctionInfo info) 
+                    => info?.HasActiveListener ?? false; // info can be null if function is disabled via attribute
+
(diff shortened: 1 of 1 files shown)


cgillum:
Oh that's interesting. I thought we had already covered the case where functions where disabled. At least, that's what I had in this comment here:
https://github.com/Azure/azure-functions-durable-extension/blob/f1a429106980eb8b824b5a0ce6a7cbc0ac0193ad/src/WebJobs.Extensions.DurableTask/Bindings/RegisteredFunctionInfo.cs#L18-L19

So either I tested this wrong originally, or something changed, but it's very clear from the code that nulls can exist here. Thanks for submitting this fix! Maybe you can add a comment above your change explaining that `null` is expected for disabled functions?

cgillum (approved):
I like your local function approach. Very clean and easy to understand. :) LGTM as long as the CI is green.

PR #1547, opened 2020-11-05
Adding Timeout Property to DurableHttpRequest

### LINKED ISSUE #1278: Add maximum and custom polling duration to DurableHttpRequest
This issue is pulled from a comment made in the ["Allow custom polling behavior in CallHttpAsync() for long-running endpoints"](https://github.com/Azure/azure-functions-durable-extension/issues/1183) issue.

The comments say:
"Second, it would be nice to be able to specify an overall maximum polling duration, after which it could throw a time out exception.

Third, from my experiments it seems that the original HTTP request will time out in just under 90 seconds. It would be nice to be able to control that timeout duration with another property on `DurableHttpRequest`."

Here are some solutions to this issue:
An additional property (e.g. `maximumPollingDuration`) can be added to `DurableHttpRequest` to keep track of an overall maximum polling duration.
 
An additional property (e.g. `timeOutDuration`) can be added to `DurableHttpRequest` to specify if there is a custom timeout duration for the original HTTP request.

bachuv:
These changes resolve #1278 by adding a `Timeout` property to `DurableHttpRequest`. This property defines the timeout for each individual HTTP request. This timeout applies to each asynchronous polling request too.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -694,6 +694,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             }
             catch (TaskFailedException e)
             {
+                // Check to see if CallHttpAsync() threw a TimeoutException
+                // In this case, we want to throw a TimeoutException instead of a FunctionFailedException
+                if (functionName.Equals(HttpOptions.HttpTaskActivityReservedName) && e.InnerException is TimeoutException)
+                {
+                    throw e.InnerException;
+                }
+
                 exception = e;
                 string message = string.Format(
                     "The {0} function '{1}' failed: \"{2}\". See the function execution logs for additional details.",
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -306,16 +306,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         private async Task<DurableHttpResponse> ScheduleDurableHttpActivityAsync(DurableHttpRequest req)
         {
             DurableHttpResponse durableHttpResponse = await this.CallDurableTaskFunctionAsync<DurableHttpResponse>(
-            functionName: HttpOptions.HttpTaskActivityReservedName,
-            functionType: FunctionType.Activity,
-            oneWay: false,
-            instanceId: null,
-            operation: null,
-            retryOptions: null,
-            input: req,
-            scheduledTimeUtc: null);
+                functionName: HttpOptions.HttpTaskActivityReservedName,
+                functionType: FunctionType.Activity,
+                oneWay: false,
+                instanceId: null,
+                operation: null,
+                retryOptions: null,
+                input: req,
+                scheduledTimeUtc: null);
 
-            return durableHttpResponse; 
+            return durableHttpResponse;
         }
 
         private DurableHttpRequest CreateLocationPollRequest(DurableHttpRequest durableHttpRequest, string locationUri)
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -243,7 +243,21 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         async Task<DurableHttpResponse> IDurableOrchestrationContext.CallHttpAsync(DurableHttpRequest req)
         {
-            DurableHttpResponse durableHttpResponse = await this.ScheduleDurableHttpActivityAsync(req);
+            // calculate timeout expiration if DurableHttpRequest.Timeout is set
+            if (req.Timeout != null)
+            {
+                req.TimeoutExpirationDateTime = this.InnerContext.CurrentUtcDateTime + req.Timeout.Value;
+            }
+
+            DurableHttpResponse durableHttpResponse;
+            try
+            {
+                durableHttpResponse = await this.ScheduleDurableHttpActivityAsync(req);
+            }
+            catch (TimeoutException)
+            {
+                throw;
+            }
 
             HttpStatusCode currStatusCode = durableHttpResponse.StatusCode;
 
@@ -266,7 +280,18 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 }
 
                 this.IncrementActionsOrThrowException();
-                await this.InnerContext.CreateTimer(fireAt, CancellationToken.None);
+
+                if (req.Timeout == null)
+                {
+                    await this.InnerContext.CreateTimer(fireAt, CancellationToken.None);
+                }
+                else
+                {
+                    TimeSpan timeLeft = req.TimeoutExpirationDateTime - this.InnerContext.CurrentUtcDateTime;
+                    CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
+                    cancellationTokenSource.CancelAfter(timeLeft);
+                    await this.InnerContext.CreateTimer(fireAt, cancellationTokenSource.Token);
+                }
 
                 DurableHttpRequest durableAsyncHttpRequest = this.CreateLocationPollRequest(
                     req,
@@ -281,16 +306,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         private async Task<DurableHttpResponse> ScheduleDurableHttpActivityAsync(DurableHttpRequest req)
         {
             DurableHttpResponse durableHttpResponse = await this.CallDurableTaskFunctionAsync<DurableHttpResponse>(
-                functionName: HttpOptions.HttpTaskActivityReservedName,
-                functionType: FunctionType.Activity,
-                oneWay: false,
-                instanceId: null,
-                operation: null,
-                retryOptions: null,
-                input: req,
-                scheduledTimeUtc: null);
+            functionName: HttpOptions.HttpTaskActivityReservedName,
+            functionType: FunctionType.Activity,
+            oneWay: false,
+            instanceId: null,
+            operation: null,
+            retryOptions: null,
+            input: req,
+            scheduledTimeUtc: null);
 
-            return durableHttpResponse;
+            return durableHttpResponse; 
         }
 
         private DurableHttpRequest CreateLocationPollRequest(DurableHttpRequest durableHttpRequest, string locationUri)
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -243,21 +243,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         async Task<DurableHttpResponse> IDurableOrchestrationContext.CallHttpAsync(DurableHttpRequest req)
         {
-            // calculate timeout expiration if DurableHttpRequest.Timeout is set
-            if (req.Timeout != null)
-            {
-                req.TimeoutExpirationDateTime = this.InnerContext.CurrentUtcDateTime + req.Timeout.Value;
-            }
-
-            DurableHttpResponse durableHttpResponse;
-            try
-            {
-                durableHttpResponse = await this.ScheduleDurableHttpActivityAsync(req);
-            }
-            catch (TimeoutException)
-            {
-                throw;
-            }
+            DurableHttpResponse durableHttpResponse = await this.ScheduleDurableHttpActivityAsync(req);
 
             HttpStatusCode currStatusCode = durableHttpResponse.StatusCode;
 
@@ -280,18 +266,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 }
 
                 this.IncrementActionsOrThrowException();
-
-                if (req.Timeout == null)
-                {
-                    await this.InnerContext.CreateTimer(fireAt, CancellationToken.None);
-                }
-                else
-                {
-                    TimeSpan timeLeft = req.TimeoutExpirationDateTime - this.InnerContext.CurrentUtcDateTime;
-                    CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
-                    cancellationTokenSource.CancelAfter(timeLeft);
-                    await this.InnerContext.CreateTimer(fireAt, cancellationTokenSource.Token);
-                }
+                await this.InnerContext.CreateTimer(fireAt, CancellationToken.None);
 
                 DurableHttpRequest durableAsyncHttpRequest = this.CreateLocationPollRequest(
                     req,
@@ -324,7 +299,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 method: HttpMethod.Get,
                 uri: new Uri(locationUri),
                 headers: durableHttpRequest.Headers,
-                tokenSource: durableHttpRequest.TokenSource);
+                tokenSource: durableHttpRequest.TokenSource,
+                timeout: durableHttpRequest.Timeout);
 
             // Do not copy over the x-functions-key header, as in many cases, the
             // functions key used for the initial request will be a Function-level key
--- a/src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
@@ -10,6 +10,8 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using DurableTask.Core;
+using DurableTask.Core.Common;
+using DurableTask.Core.Exceptions;
 using Microsoft.Extensions.Primitives;
 using Newtonsoft.Json;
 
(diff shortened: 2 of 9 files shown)


src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -35,20 +36,59 @@ public override string Run(TaskContext context, string input)
...
+                else
+                {
+                    response = await this.httpClient.SendAsync(requestMessage, cts.Token);
+                }
+
+                DurableHttpResponse durableHttpResponse = await DurableHttpResponse.CreateDurableHttpResponseWithHttpResponseMessage(response);
+
+                return JsonConvert.SerializeObject(durableHttpResponse);
+            }
+            catch (OperationCanceledException ex) // when (cts.Token.IsCancellationRequested)
```
bachuv, line 75: To make this more accurate, it could be `catch (OperationCanceledException ex) when (cts.Token.IsCancellationRequested)` (added the commented out part). This way we know that the exception came when a cancel request was sent to `CancellationTokenSource`. The only issue with this is adding tests that mirror the action of cancelling a `CancellationTokenSource`. The `Mock<HttpMessageHandler>` Setup call takes `ItExpr.IsAny<CancellationToken>()` and the test would need the `CancellationTokenSource` to cancel the task, not the `CancellationToken`.

I saw that that `ItExpr.IsAny()` could evaluate conditions, but didn't see anything about sending in another object to manipulate. Let me know if there's a way to pass in a CancellationTokenSource using Moq.

This is how the `HttpMessageHandler` is mocked right now:
```
var handlerMock = new Mock<HttpMessageHandler>(MockBehavior.Strict);
handlerMock 
  .Protected()
  .Setup<Task<HttpResponseMessage>>("SendAsync", ItExpr.IsAny<HttpRequestMessage>(), ItExpr.IsAny<CancellationToken>())
  .ReturnsAsync(httpResponseMessage);

```

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -35,20 +36,59 @@ public override string Run(TaskContext context, string input)
...
+                    TimeSpan timeout = durableHttpRequest.TimeoutExpirationDateTime - DateTime.UtcNow;
+                    this.httpClient.Timeout = timeout;
+                    cts.CancelAfter(timeout);
+                }
+                catch (Exception)
+                {
+                    TimeSpan timeout = durableHttpRequest.Timeout.Value;
+                    this.httpClient.Timeout = timeout;
+                    cts.CancelAfter(timeout);
+                }
```
bachuv, line 56: Any suggestions on what to do if `DateTime.UtcNow` is greater than `durableHttpRequest.TimeoutExpirationDateTime`? Right now, it catches the exception that says that the `timeout` TimeSpan is invalid and resets the `timeout` value. Should it just throw a TimeoutException?

test/Common/DurableHttpTests.cs
```diff
@@ -310,6 +311,86 @@ public async Task DurableHttpAsync_SynchronousAPI_Returns200(string storageProvi
...
+                    headers: headers,
+                    timeout: TimeSpan.FromMilliseconds(5000));
+
+                string functionName = nameof(TestOrchestrations.CallHttpAsyncOrchestrator);
+                var client = await host.StartOrchestratorAsync(functionName, testRequest, this.output);
+                var status = await client.WaitForCompletionAsync(this.output, timeout: TimeSpan.FromSeconds(400));
+
+                var output = status?.Output;
+                Assert.Contains("System.TimeoutException: The operation was canceled.Reached user specified timeout", output.ToString());
+                Assert.Equal(OrchestrationRuntimeStatus.Failed, status?.RuntimeStatus);
```
bachuv, line 388: Right now, the `CallHttpAsync` API throws a `FunctionFailedException` when it reaches a timeout, so that's why these are the Assert conditions that the test is checking. The test should be checking for a `TimeoutException` using `Assert.ThrowsAsync()`. Do you have any recommendations on how to propagate the `TimeoutException`? I went through the stack trace and it looks like it fails to deserialize the exception in TaskActivityDispatcher.cs, which I can't debug.

This is an example of the exception that's thrown when the function reaches the timeout:
```
Function 'Function1 (Orchestrator)' failed with an error. Reason: Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionFailedException: The activity function 'BuiltIn::HttpActivity' failed: "Failed to deserialize exception from TaskActivity: Unhandled exception while executing task: System.TimeoutException: A task was canceled.Reached user specified timeout: 00:00:10

```

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
ConnorMcMahon: So I can see an argument for one of two approaches:

1. When the customer sets the `timeout` property, it is absolute. If it expires before we hit the activity, don't even bother with the HTTP request. The good part of this approach is that it is straightforward to explain how it works. There are a couple of downsides. The big one is that the request could never even fire successfully due to internal Durable Task latency (such as a queue backlog). The other downside I see is that it doesn't give the customer much more usefulness other than just doing a Task.WhenAny() with a Timer task.

2. We only check for the timeout before we "schedule" the HTTP call. The downside is it is unclear what to set as the request cancellation timeout token to if the timeout has expired when we enter the HttpActivityShim.

ConnorMcMahon: @cgillum, what are your thoughts here?

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -35,20 +36,59 @@ public override string Run(TaskContext context, string input)
...
+                return JsonConvert.SerializeObject(durableHttpResponse);
+            }
+            catch (OperationCanceledException ex) // when (cts.Token.IsCancellationRequested)
+            {
+                throw new TimeoutException(ex.Message + $"Reached user specified timeout: {durableHttpRequest.Timeout.Value}.");
+            }
         }
 
-        private async Task<HttpRequestMessage> ReconstructHttpRequestMessage(string serializedRequest)
+        private static DurableHttpRequest ReconstructDurableHttpRequest(string serializedRequest)
```
ConnorMcMahon, line 81: Is there a reason we split this out into a helper method?

test/Common/DurableHttpTests.cs
ConnorMcMahon: So the issue is for some reason TimeoutException is not being serialized properly... We could throw a different exception instead or handle serializing this time of exception more gracefully.

ConnorMcMahon (commented):
Answered some questions and gave some feedback, but there is still some stuff that needs to be clarified.

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
bachuv: RunAsync() needs access to the DurableHttpRequest to figure out if there is a timeout or not.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -243,7 +243,21 @@ Task<DurableHttpResponse> IDurableOrchestrationContext.CallHttpAsync(HttpMethod
...
+
+            DurableHttpResponse durableHttpResponse;
+            try
+            {
+                durableHttpResponse = await this.ScheduleDurableHttpActivityAsync(req);
+            }
+            catch (TimeoutException)
+            {
+                throw;
+            }
```
cgillum, line 260: There is no benefit to having a `catch` block that only does a `throw`. Let's remove the try/catch unless there is some way we can make it more useful.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -266,7 +280,18 @@ async Task<DurableHttpResponse> IDurableOrchestrationContext.CallHttpAsync(Durab
...
+                if (req.Timeout == null)
+                {
+                    await this.InnerContext.CreateTimer(fireAt, CancellationToken.None);
+                }
+                else
+                {
+                    TimeSpan timeLeft = req.TimeoutExpirationDateTime - this.InnerContext.CurrentUtcDateTime;
+                    CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
+                    cancellationTokenSource.CancelAfter(timeLeft);
+                    await this.InnerContext.CreateTimer(fireAt, cancellationTokenSource.Token);
```
cgillum, line 293: Rather than relying on a cancellation token, would it make sense to instead use the minimum of either `fireAt` or `req.TimeoutExpirationDateTime` as the parameter to `CreateTimer`?

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -35,20 +36,59 @@ public override string Run(TaskContext context, string input)
...
+            HttpRequestMessage requestMessage = await this.ReconstructHttpRequestMessage(durableHttpRequest);
 
-            return JsonConvert.SerializeObject(durableHttpResponse);
+            CancellationTokenSource cts = new CancellationTokenSource();
+            if (durableHttpRequest.Timeout != null)
+            {
+                try
+                {
+                    TimeSpan timeout = durableHttpRequest.TimeoutExpirationDateTime - DateTime.UtcNow;
+                    this.httpClient.Timeout = timeout;
```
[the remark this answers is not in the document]
cgillum: +1, we don't want to set the timeout on the HTTP client since that could change the behavior for other, unrelated code.

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -35,20 +36,59 @@ public override string Run(TaskContext context, string input)
...
         public async override Task<string> RunAsync(TaskContext context, string rawInput)
         {
-            HttpRequestMessage requestMessage = await this.ReconstructHttpRequestMessage(rawInput);
-            HttpResponseMessage response = await this.httpClient.SendAsync(requestMessage);
-            DurableHttpResponse durableHttpResponse = await DurableHttpResponse.CreateDurableHttpResponseWithHttpResponseMessage(response);
+            DurableHttpRequest durableHttpRequest = ReconstructDurableHttpRequest(rawInput);
+            HttpRequestMessage requestMessage = await this.ReconstructHttpRequestMessage(durableHttpRequest);
 
-            return JsonConvert.SerializeObject(durableHttpResponse);
+            CancellationTokenSource cts = new CancellationTokenSource();
```
cgillum, line 42: `CancellationTokenSource` implements `IDisposable`, so we need to wrap it in a `using` statement.

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -35,20 +36,59 @@ public override string Run(TaskContext context, string input)
...
+            if (durableHttpRequest.Timeout != null)
+            {
+                try
+                {
+                    TimeSpan timeout = durableHttpRequest.TimeoutExpirationDateTime - DateTime.UtcNow;
+                    this.httpClient.Timeout = timeout;
+                    cts.CancelAfter(timeout);
+                }
+                catch (Exception)
```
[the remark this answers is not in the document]
cgillum: +1, exceptions should only be used in exceptional cases - not for control flow.

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
cgillum: I'm not actually sure I understand the difference between those two options. However, it looks like we need to make another design decision about how the timeout should work. It seems there are two options, and maybe you were already alluding to this and I misunderstood.

1. Timeout is "absolute" in the sense that it represents a point in time. In this model, it would also include time spent in the work-item queue.
2. Timeout is relative in the sense that it does not have a specific expiration, but rather dictates how much time the HTTP client should wait for a response.

My gut tells me that (2) will be a more common scenario. In that case, we don't need to calculate any expiration time at all. We just pass the timeout value all the way to this point and use it in the call to `cts.CancelAfter()`. I worry that with (1) customers will have unpredictable and/or unexpected runtime behavior.

Thoughts?

test/Common/DurableHttpTests.cs
cgillum: Let's try to figure out why it's not being serialized properly. If it's a problem for us then it's likely also a problem for our customers when they have activity functions that throw this exception.

cgillum (changes requested):
Adding a bit of my own feedback here.

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
bachuv: I think it will be easier to document and maintain the second option (the timeout specifies the time the HTTP client should wait for a response). I agree that option 1 might be confusing for customers.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -693,6 +694,13 @@ string IDurableOrchestrationContext.StartNewOrchestration(string functionName, o
             }
             catch (TaskFailedException e)
             {
+                // Check to see if CallHttpAsync() threw a TimeoutException
+                // In this case, we want to throw a TimeoutException instead of a FunctionFailedException
+                if (functionName.Equals(HttpOptions.HttpTaskActivityReservedName) && e.InnerException is TimeoutException)
```
ConnorMcMahon, line 699: Unrelated to this PR, but can you file an in depth issue of your findings here of why we needed this workaround, so we can investigate whether this impacts other exceptions customers will throw and if we can fix it in a non-breaking way?

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -35,20 +38,49 @@ public override string Run(TaskContext context, string input)
...
         public async override Task<string> RunAsync(TaskContext context, string rawInput)
         {
-            HttpRequestMessage requestMessage = await this.ReconstructHttpRequestMessage(rawInput);
-            HttpResponseMessage response = await this.httpClient.SendAsync(requestMessage);
-            DurableHttpResponse durableHttpResponse = await DurableHttpResponse.CreateDurableHttpResponseWithHttpResponseMessage(response);
+            DurableHttpRequest durableHttpRequest = ReconstructDurableHttpRequest(rawInput);
+            HttpRequestMessage requestMessage = await this.ReconstructHttpRequestMessage(durableHttpRequest);
 
-            return JsonConvert.SerializeObject(durableHttpResponse);
+            using (CancellationTokenSource cts = new CancellationTokenSource())
```
ConnorMcMahon, line 44: small nit:

We could make this a bit cleaner by limiting the introduction of the using block, and the try-catch block for `OperationCanceledException` into the case where `durableHttpRequest.Timeout` is non-null. It moderately improves readability by keeping the more general case less indented.

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -35,20 +38,49 @@ public override string Run(TaskContext context, string input)
...
+        private static DurableHttpRequest ReconstructDurableHttpRequest(string serializedRequest)
         {
             // DeserializeObject deserializes into a List and then the first element
             // of that list is the DurableHttpRequest
             IList<DurableHttpRequest> input = JsonConvert.DeserializeObject<IList<DurableHttpRequest>>(serializedRequest);
             DurableHttpRequest durableHttpRequest = input.First();
+            return durableHttpRequest;
+        }
 
+        private async Task<HttpRequestMessage> ReconstructHttpRequestMessage(DurableHttpRequest durableHttpRequest)
```
ConnorMcMahon, line 82: nit: This could probably be renamed `ConstructHttpRequestMessage()` or `BuildHttpRequestMessage()` or `ConvertToHttpRequestMessage()`, as we never really had a `HttpRequestMessage` in the first place,

ConnorMcMahon (approved):
Looks good. I have a small nit, but not a blocker. It's up to you whether you want to fix it before merging.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
cgillum: This was my suggestion to Varshi for how to know when to throw `TimeoutException` instead of `FunctionFailedException`. We're not really working around anything AFAIK. Surfacing outer exceptions other than `FunctionFailedException` is not something we expose to our customers.

cgillum (approved):
Resetting my approval status. No blockers from me.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
ConnorMcMahon: Makes sense.

I remember this being something we discussed around the V2 timeframe, but we missed our chance to tackle it before we locked in our breaking changes.

Probably worth keeping track of as a "breaking-change" issue that we can consider looking into if we make another breaking change release in the future (a ways off hopefully!).

Issue #1538, opened 2020-10-25
[2.3.1] Possible regression - 6 day max timespan on WaitForExternalEvent Timeout.


olitomlinson:
This code was causing the orchestration to fail : 

` var result = context.WaitForExternalEvent<JObject>($"external_event_{eventId}", TimeSpan.FromDays(14), cancelToken: cts.Token);`

With the following exception : 

`The Azure Storage provider supports a maximum of 6 days for time-based delays (Parameter 'timeout')`

If I change the timeout to 6 days, it works as expected, no exceptions.

It was my assumption that the max limit on timers was lifted in 2.3.0, so this may be a regression.

ConnorMcMahon:
Hmm, I believe this is a regression. Thanks for bringing this to our attention.

I will make sure we investigate this, add a test for this case, and fix the regression in v2.4.0.

ConnorMcMahon:
Looks like this is not a regression, just a bug where we missed removing of a timer length validation check. I believe removing this no longer required validation should just make everything work, but I will write some tests to capture this.

PR #1550, opened 2020-11-05
Expand long running timer functionality to WaitForExternalEvent

ConnorMcMahon:
Internally WaitForExternalEvent with a timeout uses a timer. In v2.3.0
we added support for long-running timers, we just didn't remove the
now unecessary validation from the WaitForExternalEvent API. This PR
validates the scenario with a test and removes this validation.

Note that initially this PR attempted to expand this functionality to
RetryOptions in CallActivity/SubOrchestrationWithRetry, but
unfortunately the timer used by that retry is lower level than the
Durable Functions layer, meaning we can't use our long-running support.

Addresses #1538

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -383,26 +383,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             return result;
         }
 
-        internal Task OutOfProcCreateTimer(DurableOrchestrationContext ctx, DateTime fireAt, CancellationToken cancelToken)
-        {
-            if (!this.ValidOutOfProcTimer(fireAt, out string errorMessage))
-            {
-                throw new ArgumentException(errorMessage, nameof(fireAt));
-            }
-
-            return ((IDurableOrchestrationContext)ctx).CreateTimer(fireAt, cancelToken);
-        }
-
-        private bool ValidOutOfProcTimer(DateTime fireAt, out string errorMessage)
+        internal void ValidateOutOfProcTimer(DateTime fireAt)
         {
             this.ThrowIfInvalidAccess();
 
-            if (!this.durabilityProvider.ValidateDelayTime(fireAt.Subtract(this.InnerContext.CurrentUtcDateTime), out errorMessage))
+            if (!this.durabilityProvider.ValidateDelayTime(fireAt.Subtract(this.InnerContext.CurrentUtcDateTime), out string errorMessage))
             {
-                return false;
+                throw new ArgumentException(errorMessage, nameof(fireAt));
             }
-
-            return true;
         }
 
         /// <inheritdoc />
@@ -943,11 +931,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private Task<T> WaitForExternalEvent<T>(string name, TimeSpan timeout, Action<TaskCompletionSource<T>> timeoutAction, CancellationToken cancelToken)
         {
-            if (!this.durabilityProvider.ValidateDelayTime(timeout, out string errorMessage))
-            {
-                throw new ArgumentException(errorMessage, nameof(timeout));
-            }
-
             var tcs = new TaskCompletionSource<T>();
             var cts = CancellationTokenSource.CreateLinkedTokenSource(cancelToken);
 
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -383,7 +383,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             return result;
         }
 
-        internal void ValidateOutOfProcTimer(DateTime fireAt)
+        // We now have built in long-timer support for C#, but in some scenarios, such as out-of-proc,
+        // we may still need to enforce this limitations until the solution works there as well.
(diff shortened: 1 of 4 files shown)


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -383,26 +383,14 @@ async Task<T> IDurableOrchestrationContext.CreateTimer<T>(DateTime fireAt, T sta
...
+        internal void ValidateOutOfProcTimer(DateTime fireAt)
         {
             this.ThrowIfInvalidAccess();
 
-            if (!this.durabilityProvider.ValidateDelayTime(fireAt.Subtract(this.InnerContext.CurrentUtcDateTime), out errorMessage))
+            if (!this.durabilityProvider.ValidateDelayTime(fireAt.Subtract(this.InnerContext.CurrentUtcDateTime), out string errorMessage))
             {
-                return false;
+                throw new ArgumentException(errorMessage, nameof(fireAt));
             }
```
davidmrdavid, line 393: Why was it that before we could return `false` but now we return an exception?

Also, since this is an exception-throwing block, I would appreciate an inlined comment explaining what the error means, when it could occur, and why we need to throw it

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -127,17 +128,13 @@ private async Task ProcessAsyncActions(AsyncAction[][] actions)
...
-                                DurableOrchestrationContext ctx = this.context as DurableOrchestrationContext;
-
                                 if (ctx != null)
                                 {
-                                    tasks.Add(ctx.OutOfProcCreateTimer(ctx, action.FireAt, cts.Token));
-                                }
-                                else
-                                {
-                                    tasks.Add(this.context.CreateTimer(action.FireAt, cts.Token));
+                                    ctx.ValidateOutOfProcTimer(action.FireAt);
```
davidmrdavid, line 133: What about the timer does this validate? It would be good to add it as an inline comment :)

test/FunctionsV2/LongTimerTests.cs
```diff
@@ -103,5 +104,29 @@ public async Task EntitySignalWithLongDelay(bool extendedSessions)
...
+                var fireAt = TimeSpan.FromSeconds(90);
+                var client = await host.StartOrchestratorAsync(nameof(TestOrchestrations.ApprovalWithTimeout), (fireAt, "throw"), this.output);
+                var status = await client.WaitForCompletionAsync(this.output, timeout: TimeSpan.FromMinutes(2));
+
+                Assert.Equal(OrchestrationRuntimeStatus.Completed, status?.RuntimeStatus);
+                Assert.Equal("TimeoutException", status?.Output);
+
+                await host.StopAsync();
+            }
+        }
```
davidmrdavid, line 130: What is the "max time" being referenced in the method name? I don't think it's immediately clear from the name, it would also be good to add a short test-level comment explaining what this is testing for.

davidmrdavid (commented):
Looks mostly good to me, assuming the CI is passing!
I left some suggestions to improve readability and maintainability moving forward, mostly around adding more comments. Please review them when you get the chance ⚡

test/FunctionsV2/LongTimerTests.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Security.Cryptography;
```
bachuv, line 5: Do we need this using statement?

bachuv (approved):
Looks good to me!

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
ConnorMcMahon: I made the method name more clear, as opposed to a comment. I find comments get out of date, where method names are easier to refactor.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
ConnorMcMahon: I just simplified a bit, where before we had a helper method that returned false that the caller would use an if statement and throw the exception themselves, this method now just throws the exception directly.

I made the method a bit more self explanatory and added why we use it for out-of-proc as a comment.

test/FunctionsV2/LongTimerTests.cs
ConnorMcMahon: Each storage provider can provide a method to validate timer lengths, as Azure Storage provider sets a max for 6 days. I improve the test name, but I could only make it so long before the taskhub name got too long for Azure Storage to support.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -383,26 +383,16 @@ async Task<T> IDurableOrchestrationContext.CreateTimer<T>(DateTime fireAt, T sta
...
-            {
-                throw new ArgumentException(errorMessage, nameof(fireAt));
-            }
-
-            return ((IDurableOrchestrationContext)ctx).CreateTimer(fireAt, cancelToken);
-        }
-
-        private bool ValidOutOfProcTimer(DateTime fireAt, out string errorMessage)
+        // We now have built in long-timer support for C#, but in some scenarios, such as out-of-proc,
+        // we may still need to enforce this limitations until the solution works there as well.
```
davidmrdavid, line 387: nit: this -> these

davidmrdavid (approved):
LGTM! Found a small typo, but it's non-blocking. :shipit:

PR #1553, opened 2020-11-09
Make E2E tests more robust in CI pipeline (#1385)

ConnorMcMahon:
Our end-to-end tests currently rely on having unique taskhub names so
that tests can be run safely in parallel. This has been proven to be
hard to catch/maintain this restriction, so we add a random 4
alphanumeric suffix to each task hub to make sure that there are never
duplicate taskhub names.

The other benefit of creating non-deterministic taskhubs for each test
run is that we should no longer end up with the corrupted state problem,
where a previous test run somehow corrupted a task hub and the test now
fails until we use a new storage account or preform manual intervention.

Finally, this test adds a dummy test that deletes up to 2000 taskhubs
older than 3 days, to ensure that we don't collect to many entities in
our storage account over time.

### DIFF
--- a/test/Common/TestHelpers.cs
+++ b/test/Common/TestHelpers.cs
@@ -212,39 +212,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
             return truncatedTestName + testPropertiesSuffix + randomSuffix;
         }
 
-        internal static async Task DeleteAllElementsInStorageTaskHubAsync(string connectionString, string taskHub, int partitionCount)
-        {
-            var deletionTasks = new List<Task>();
-            CloudStorageAccount storageAccount = CloudStorageAccount.Parse(connectionString);
-
-            // Delete blobs
-            CloudBlobClient blobClient = storageAccount.CreateCloudBlobClient();
-            CloudBlobContainer leases = blobClient.GetContainerReference($"{taskHub}-leases");
-            deletionTasks.Add(leases.DeleteIfExistsAsync());
-            CloudBlobContainer largeContainers = blobClient.GetContainerReference($"{taskHub}-largemessages");
-            deletionTasks.Add(largeContainers.DeleteIfExistsAsync());
-
-            // Delete queues
-            CloudQueueClient queueClient = storageAccount.CreateCloudQueueClient();
-            CloudQueue workItemQueue = queueClient.GetQueueReference($"{taskHub}-workitems");
-            deletionTasks.Add(workItemQueue.DeleteIfExistsAsync());
-            for (int i = 0; i < partitionCount; i++)
-            {
-                string controlQueueName = $"{taskHub}-control-{i.ToString("00")}";
-                CloudQueue controlQueue = queueClient.GetQueueReference(controlQueueName);
-                deletionTasks.Add(controlQueue.DeleteIfExistsAsync());
-            }
-
-            // Delete tables
-            CloudTableClient tableClient = storageAccount.CreateCloudTableClient();
-            CloudTable historyTable = tableClient.GetTableReference($"{taskHub}History");
-            CloudTable instancesTable = tableClient.GetTableReference($"{taskHub}Instances");
-            deletionTasks.Add(historyTable.DeleteIfExistsAsync());
-            deletionTasks.Add(instancesTable.DeleteIfExistsAsync());
-
-            await Task.WhenAll(deletionTasks);
-        }
-
         public static ITypeLocator GetTypeLocator()
         {
             var types = new Type[]
--- a/test/Common/TestHelpers.cs
+++ b/test/Common/TestHelpers.cs
@@ -79,6 +79,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
                 options = new DurableTaskOptions();
             }
 
+            // Some tests require knowing the task hub that the provider uses. Because of that, they will instantiate the exact
+            // task hub name and require the usage of that task hub. Otherwise, generate a partially random task hub from the
+            // test name and properties of the test.
             options.HubName = exactTaskHubName ?? GetTaskHubNameFromTestName(testName, enableExtendedSessions);
             options.Tracing = new TraceOptions()
             {
--- a/test/Common/TestHelpers.cs
+++ b/test/Common/TestHelpers.cs
@@ -12,6 +12,9 @@ using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Microsoft.WindowsAzure.Storage;
 using Newtonsoft.Json.Linq;
+using Microsoft.WindowsAzure.Storage.Blob;
+using Microsoft.WindowsAzure.Storage.Queue;
+using Microsoft.WindowsAzure.Storage.Table;
 using Xunit;
 using Xunit.Abstractions;
 
@@ -56,7 +59,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
             bool? localRpcEndpointEnabled = false,
             DurableTaskOptions options = null,
             bool rollbackEntityOperationsOnExceptions = true,
-            int entityMessageReorderWindowInMinutes = 30)
+            int entityMessageReorderWindowInMinutes = 30,
+            string exactTaskHubName = null)
         {
             switch (storageProviderType)
             {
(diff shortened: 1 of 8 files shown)


test/Common/TestHelpers.cs
```diff
@@ -196,9 +200,48 @@ public static DurableTaskOptions GetDurableTaskOptionsForStorageProvider(string
...
+        // Create a valid task hub from the test name, and add a random suffix to avoid conflicts
         public static string GetTaskHubNameFromTestName(string testName, bool enableExtendedSessions)
         {
-            return testName.Replace("_", "") + (enableExtendedSessions ? "EX" : "") + PlatformSpecificHelpers.VersionSuffix;
+
+            string strippedTestName = testName.Replace("_", "");
+            string truncatedTestName = strippedTestName.Substring(0, Math.Min(35, strippedTestName.Length));
+            string deterministicSuffix = (enableExtendedSessions ? "EX" : "") + PlatformSpecificHelpers.VersionSuffix;
+            string randomSuffix = Guid.NewGuid().ToString().Substring(0, 4);
+            return truncatedTestName + deterministicSuffix + randomSuffix;
```
davidmrdavid, line 211: A few thoughts here:
(1) I'm a tiny bit concerned about the truncation step here. Is it necessary? I could imagine equivalent truncated names
(2) The  `deterministicSuffix` seems to be quite different from the `randomSuffix` in terms of their purpose, so I would prefer we renamed `deterministicSuffix` to `extendedSessionsSuffix` or something like that.

test/Common/TestHelpers.cs
```diff
@@ -75,7 +79,7 @@ internal static class TestHelpers
                 options = new DurableTaskOptions();
             }
 
-            options.HubName = GetTaskHubNameFromTestName(testName, enableExtendedSessions);
+            options.HubName = exactTaskHubName ?? GetTaskHubNameFromTestName(testName, enableExtendedSessions);
```
davidmrdavid, line 82: Unless I'm mistaken, this line is responsible for changing _all_ TaskHubNames, unless an exact TaskHub is specified. Since it's so consequential, I would recommend having short comment above it specifying what this line is doing, and why it's so important.

test/FunctionsV2/TestCleanup.cs
```diff
@@ -0,0 +1,89 @@
...
+        public TestCleanup(ITestOutputHelper output)
+        {
+            this.output = output;
+            this.loggerProvider = new TestLoggerProvider(output);
+        }
+
+        // This test should never fail. The sole purpose is to cleanup old taskhubs in the CI
+        // storage account to prevent clutter now that TaskHub names are non-deterministic.
+        [Fact]
+        public async Task CleanupOldAzureStorageTaskHubs()
```
davidmrdavid, line 32: Do we have a way to make sure this is run before any other tests? I'm asking because not all tests are guaranteed to run on a CI execution (for example, if another test fails first, we might not do a clean-up step), so it would be good to perform the clean-up as eagerly / early as possible.

davidmrdavid (commented):
Thank you so much for this PR, I'm looking forward to a smoother CI experience. I left some thoughts below, please let me know what you think :)

test/FunctionsV2/PlatformSpecificHelpers.FunctionsV2.cs
```diff
@@ -123,7 +130,10 @@ public Task CallAsync(string methodName, IDictionary<string, object> args)
             public Task CallAsync(MethodInfo method, IDictionary<string, object> args)
                 => this.innerWebJobsHost.CallAsync(method, args);
 
-            public void Dispose() => this.innerHost.Dispose();
+            public void Dispose()
+            {
+                this.innerHost.Dispose();
+            }
```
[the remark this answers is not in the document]
ConnorMcMahon: Ah, the original V1 PR had a different approach in the Dispose() method, which is likely why there is a spacing difference. These should be semantically identical.

test/Common/TestHelpers.cs
ConnorMcMahon: 1) Equivalent truncated names don't matter too much from a functionality standpoint, as the random suffix will cause the task hubs to be unique. The main downside is debugging failures by looking at the storage artifacts for a taskhub if there is a conflict. It will definitely be a bit more annoying to find which task hub corresponds to your test if there are two tests with equivalent names, but we need to do the truncation because taskhubs have to be less than ~50 characters due to storage naming limits.

2) `extendedSessionsSuffix` doesn't cover all that the suffix contains, because it also contains the Functions Version that we ran the test on. Maybe a `testPropertiesSuffix` would work, but I don't think that is necessarily any clearer than deterministic suffix. Essentially we are tagging on two suffixes, one which is deterministic for the test, and another that adds randomness.

test/FunctionsV2/TestCleanup.cs
ConnorMcMahon: It's not super important that that this runs every test run, as it should clean up more than the number of task hubs that it generates, and `clean-up` isn't that important. This is meant to just prevent us from ballooning 500+ taskhubs on every CI run without ever cleaning up. 

I was debating not putting this as a test and having a timer trigger perform this daily at midnight (with no limit on the number of taskhubs it deletes), but that is just more infrastructure to maintain.

test/Common/TestHelpers.cs
davidmrdavid: Thanks for clarifying. I won't block on this but I would personally prefer the `testPropertiesSuffix` name, as I think it provides more context :)

test/FunctionsV2/TestCleanup.cs
davidmrdavid: Understood and agreed, we can leave this as a "test", _for now_ ;)

test/Common/TestHelpers.cs
```diff
@@ -196,9 +202,47 @@ public static DurableTaskOptions GetDurableTaskOptionsForStorageProvider(string
...
+        {
+            var deletionTasks = new List<Task>();
+            CloudStorageAccount storageAccount = CloudStorageAccount.Parse(connectionString);
+
+            // Delete blobs
+            CloudBlobClient blobClient = storageAccount.CreateCloudBlobClient();
+            CloudBlobContainer leases = blobClient.GetContainerReference($"{taskHub}-leases");
+            deletionTasks.Add(leases.DeleteIfExistsAsync());
+            CloudBlobContainer largeContainers = blobClient.GetContainerReference($"{taskHub}-largemessages");
+            deletionTasks.Add(largeContainers.DeleteIfExistsAsync());
```
amdeel, line 225: We should also delete the {taskHub}-applease blob container.

But also, when is this method called? Can we just call a method in AzureStorageOrchestrationService instead? We would have to update this method whenever we add another container etc. Also, it's specific to AzureStorage as the backend.

ConnorMcMahon: It turns out this actually isn't in use, it is left over from my approach for Durable V1 branch. 

I'll delete this method. The actual delete logic relies on the DeleteTaskHub method on the AzureStorageOrchestrationService.

In general, I am fine with it working only for Azure Storage, as we are discussing that each storage provider supported by Durable Functions will have some way to run these tests themselves, and they will need to figure out their own cleanup strategy.

test/FunctionsV2/TestCleanup.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            this.output = output;
+            this.loggerProvider = new TestLoggerProvider(output);
+        }
+
+        // This test should never fail. The sole purpose is to cleanup old taskhubs in the CI
+        // storage account to prevent clutter now that TaskHub names are non-deterministic.
+        [Fact]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        public async Task CleanupOldAzureStorageTaskHubs()
```
amdeel, line 35: This does seem like a good addition. And I see that you're using AzureStorageOrchestratonService to delete each task hub. Should we add this logic to AzureStorageOrchestrationService to clean up all old task hubs?

ConnorMcMahon: An AzureStorageOrchestrationService is specifically tied to one task hub by definition, so I don't feel comfortable adding logic to delete other task hubs on the storage account.

amdeel (approved):

PR #1566, opened 2020-11-17
Updates to support Durable Client feature

bachuv:
These changes are for the new Durable Client feature (#1125). With the new feature, there isn't a way to create the HTTP management payloads from the lack of a webhook URL so exceptions are thrown when a user tries to call `CreateCheckStatusResponse` or `CreateHttpManagementPayload`.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -651,6 +651,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             string taskHubName,
             string connectionName)
         {
+            if (this.httpApiHandler == null)
+            {
+                throw new InvalidOperationException("CreateHttpManagementPayload not supported.");
+            }
+
             return this.httpApiHandler.CreateHttpManagementPayload(instanceId, taskHubName, connectionName);
         }
 
@@ -807,6 +812,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             DurableClientAttribute attribute,
             bool returnInternalServerErrorOnFailure = false)
         {
+            if (this.httpApiHandler == null)
+            {
+                throw new InvalidOperationException("CreateCheckStatusResponse not supported.");
+            }
+
             return this.httpApiHandler.CreateCheckStatusResponse(request, instanceId, attribute, returnInternalServerErrorOnFailure);
         }
 
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -653,7 +653,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             if (this.httpApiHandler == null)
             {
-                throw new InvalidOperationException("CreateHttpManagementPayload not supported.");
+                throw new InvalidOperationException("IDurableClient.CreateHttpManagementPayload is not supported for IDurableClient instances created outside of a Durable Functions application.");
             }
 
             return this.httpApiHandler.CreateHttpManagementPayload(instanceId, taskHubName, connectionName);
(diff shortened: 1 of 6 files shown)


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -651,6 +651,11 @@ private static async Task<OrchestrationState> GetOrchestrationInstanceStateAsync
             string taskHubName,
             string connectionName)
         {
+            if (this.httpApiHandler == null)
+            {
+                throw new InvalidOperationException("CreateHttpManagementPayload not supported.");
```
ConnorMcMahon, line 656: Consider replacing the message with something like:

"IDurableClient.CreateHttpManagementPayload is not supported for IDurableClient instances created outside of an Azure Functions application."

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -807,6 +812,11 @@ private static async Task<DurableOrchestrationStatus> GetDurableOrchestrationSta
             DurableClientAttribute attribute,
             bool returnInternalServerErrorOnFailure = false)
         {
+            if (this.httpApiHandler == null)
+            {
+                throw new InvalidOperationException("CreateCheckStatusResponse not supported.");
```
ConnorMcMahon, line 817: similar here.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClientFactory.cs
```diff
@@ -76,19 +76,12 @@ public IDurableClient CreateClient(DurableClientOptions durableClientOptions)
...
-                    return new HttpApiHandler(null, null, this.durableTaskOptions, this.logger);
-                });
-
             DurableClient client = this.cachedClients.GetOrAdd(
                 attribute,
                 attr =>
                 {
                     DurabilityProvider innerClient = this.durabilityProviderFactory.GetDurabilityProvider(attribute);
-                    return new DurableClient(innerClient, httpApiHandler, attribute, this.MessageDataConverter, this.TraceHelper, this.durableTaskOptions);
+                    return new DurableClient(innerClient, null, attribute, this.MessageDataConverter, this.TraceHelper, this.durableTaskOptions);
```
ConnorMcMahon, line 84: I believe this factory is also the implementation of  `IDurableClientFactory` for DurableTaskExtension, so I don't think we can safely set `httpApiHandler` to null here always. 

Is there a way to safely distinguish when we are running in a `DurableTaskExtension` and we can safey provide a `HttpApiHandler` instance, and when we can just specify `null`?

ConnorMcMahon (changes requested):
I think the current iteration will break clients used by Durable Functions apps.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClientFactory.cs
bachuv: Can you point to where `DurableTaskExtension` is using `IDurableClientFactory`?

`DurableClientAttribute.IsExternalClient` could be a way to check if httpApiHandler should be set to null or not.

ConnorMcMahon: Woops, I am mistaken! I assumed it got injected via dependency injection in DurableTaskExtension, but it looks like that is not the case? 

I am a bit confused of the need for `DurableClientAttrbiute.ExternalClient` then, but if I remember correctly that may be a remnant from some work @amdeel did a few releases back, not this new DurableClientFactory work...

test/Common/DurableClientBaseTests.cs
```diff
@@ -142,6 +144,91 @@ public async Task TerminateAsync_NonRunningOrchestrator_ThrowsException()
         }
 
 #if !FUNCTIONS_V1
+        [Fact]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        public async Task DurableClient_DI_StartNewAsync_ReturnsInstanceId()
```
ConnorMcMahon, line 149: nit: I don't think the names of these tests are quite accurate, as it implies we are testing the full `IDurableClientFactory` code path, which we aren't quite doing. I would maybe mark them as "DurableClient_ExternalApp_*` instead. It's not perfect, but a bit more accurate.

ConnorMcMahon (approved):
I have a small nit regarding test names, but overall I think this is ready, at least for initial release!

PR #1567, opened 2020-11-17
Update Nuget installer version in DevOps Pipeline

ConnorMcMahon:
Should fix the issue found here: https://developercommunity.visualstudio.com/content/problem/1257247/working-devops-pipeline-starts-to-fail-with-unrela.html.


amdeel (approved):
That's weird, I thought it would pick up the new YAML file in this PR. I'm ok with merging this, since this fix was recommended and it would be a good way to confirm that these PRs use the old YAML files.

PR #1571, opened 2020-11-18
Merge Correlation Branch into Dev

davidmrdavid:
Since https://github.com/Azure/durabletask/pull/422 is now merged into DTFx's `master` branch and is soon to released, it makes sense for us to start updating the correlation branch of the `durable-extension` in preparation for that dependency.

I'm opening this PR for two reasons:
(1) As an excuse to bring it's contents up to date with `dev`'s latest changes
(2) To start getting reviews out of the way so we can merge this feature to the master branch **as soon as** the upcoming DTFx release is out.

Note that the code here was not written by me, so I too will be reviewing it for the first time and proposing changes.


src/WebJobs.Extensions.DurableTask/Correlation/NoOpTelemetryChannel.cs
```diff
@@ -0,0 +1,83 @@
...
+using System;
+using Microsoft.ApplicationInsights.Channel;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Correlation
+{
+    /// <summary>
+    /// A stub of <see cref="ITelemetryChannel"/>.
+    /// </summary>
+    public sealed class NoOpTelemetryChannel : ITelemetryChannel
+    {
```
davidmrdavid, line 13: I'm not entirely sure why we need this class. Once we find out, we should add it to the class header

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
```diff
@@ -0,0 +1,99 @@
...
+        {
+            this.SetUpDistributedTracing();
+
+            this.SetUpTelemetryClient();
+            this.SetUpTelemetryCallbacks();
+        }
+
+        private void SetUpDistributedTracing()
+        {
+            DurableTaskOptions durableTaskOptions = this.options.Value;
```
davidmrdavid, line 50: Can we give this field a more informative name than `Value`? 😄

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// </summary>
+        public void Initialize()
+        {
+            this.SetUpDistributedTracing();
+
+            this.SetUpTelemetryClient();
+            this.SetUpTelemetryCallbacks();
+        }
+
+        private void SetUpDistributedTracing()
```
davidmrdavid, line 48: I feel like this whole class is setting up distributed tracing, so I think we should find a more specific name here. Maybe `LoadConfig` would be better?

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
```diff
@@ -0,0 +1,99 @@
...
+                    ? Protocol.W3CTraceContext
+                    : Protocol.HttpCorrelationProtocol;
+        }
+
+        private void SetUpTelemetryCallbacks()
+        {
+            CorrelationTraceClient.SetUp(
+                (TraceContextBase requestTraceContext) =>
+                {
+                    requestTraceContext.Stop();
```
davidmrdavid, line 64: Why is this necessary? We should comment it

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
```diff
@@ -0,0 +1,99 @@
...
+                (TraceContextBase dependencyTraceContext) =>
+                {
+                    dependencyTraceContext.Stop();
+                    var dependencyTelemetry = dependencyTraceContext.CreateDependencyTelemetry();
+                    this.telemetryClient.TrackDependency(dependencyTelemetry);
+                },
+                (Exception e) =>
+                {
+                    this.telemetryClient.TrackException(e);
+                });
```
davidmrdavid, line 78: How does the framework know to know when to call each callback?

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
```diff
@@ -0,0 +1,99 @@
...
+        {
+            TelemetryConfiguration config = TelemetryConfiguration.CreateDefault();
+            if (this.OnSend != null)
+            {
+                config.TelemetryChannel = new NoOpTelemetryChannel { OnSend = this.OnSend };
+            }
+
+            var telemetryInitializer = new DurableTaskCorrelationTelemetryInitializer();
+
+            telemetryInitializer.ExcludeComponentCorrelationHttpHeadersOnDomains.Add("127.0.0.1");
```
davidmrdavid, line 91: Why do we need this?

src/WebJobs.Extensions.DurableTask/Correlation/TraceContextBaseExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+            telemetry.Start();
+            telemetry.Duration = context.Duration;
+            telemetry.Timestamp = context.StartTime; // TimeStamp is the time of ending the Activity.
+            telemetry.Id = context.TelemetryId;
+            telemetry.Context.Operation.Id = context.TelemetryContextOperationId;
+            telemetry.Context.Operation.ParentId = context.TelemetryContextOperationParentId;
+
+            return telemetry;
+        }
+    }
```
davidmrdavid, line 49: These two methods are equivalent, with the only different being the return type. We could refactor them into a single method, with an if-statement changing the return

test/FunctionsV2/CorrelationEndToEndTests.cs
```diff
@@ -0,0 +1,253 @@
...
+using Microsoft.ApplicationInsights.Extensibility.Implementation;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests;
+using Microsoft.Azure.WebJobs.Host.TestCommon;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
+{
+    [Collection("Non-Parallel Collection")]
```
davidmrdavid, line 25: Is this just testing metadata?

samples/correlation-csharp/FunctionAppCorrelation/CorrelationIntegration.cs
```diff
@@ -0,0 +1,105 @@
...
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Azure.WebJobs;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+using Microsoft.Azure.WebJobs.Extensions.Http;
+using Microsoft.Extensions.Logging;
+
+namespace FunctionAppCorrelation
+{
+    public class CorrelationIntegration
+    {
```
davidmrdavid, line 26: I feel like a better name for this would be `CorrelationDemo`. I would also like to see this being more heavily commented

samples/correlation-csharp/FunctionAppCorrelation/FanOutFanIn.cs
```diff
@@ -0,0 +1,52 @@
...
+            [DurableClient] IDurableOrchestrationClient starter,
+            ILogger log)
+        {
+            string instanceId = await starter.StartNewAsync(nameof(FanOutFanInOrchestrator), null);
+
+            log.LogInformation($"Started orchestration with ID = '{instanceId}'.");
+            return starter.CreateCheckStatusResponse(req, instanceId);
+        }
+    }
+}
```
davidmrdavid, line 52: I see lot of samples like this one, where the orchestration does not appear to have anything `telemetry` specific. Would it be possible to re-use the already existing samples and to provide a flag to enable correlation tracing on them? I'm just trying to avoid duplicated code

samples/correlation-csharp/Readme.md
```diff
@@ -0,0 +1,64 @@
+# Distributed Tracing for Durable Functions
```
davidmrdavid, line 1: There are a few small typos in the READMEs, but we can focus on those later :)

src/WebJobs.Extensions.DurableTask/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
```diff
@@ -0,0 +1,319 @@
...
+{
+    /// <summary>
+    /// Telemetry Initializer that sets correlation ids for W3C.
+    /// This source is based on W3COperationCorrelationTelemetryInitializer.cs
+    /// 1. Modified with CorrelationTraceContext.Current
+    /// 2. Avoid to be overriden when it is RequestTelemetry
+    /// Original Source is here https://github.com/microsoft/ApplicationInsights-dotnet-server/blob/2.8.0/Src/Common/W3C/W3COperationCorrelationTelemetryInitializer.cs.
+    /// </summary>
+    internal
+        class DurableTaskCorrelationTelemetryInitializer : ITelemetryInitializer
```
davidmrdavid, line 26: so is this _only_ for the W3C protocol? If so, we could rename this class

davidmrdavid (commented):
Left some preliminary thoughts about the code. Mostly ignored going over the READMEs and dependency updates. Please let me know y'alls thoughts as well. Overall, this is already in pretty good shape, it just needs a few touch-ups before merge!

davidmrdavid:
I also expect the tests to be failing at this point, as there are missing dependencies

samples/correlation-csharp/.gitattributes
```diff
@@ -0,0 +1,63 @@
+###############################################################################
```
ConnorMcMahon, line 1: We can probably drop most of the .git* files, as we have global ones for the repo.

samples/correlation-csharp/FunctionAppCorrelation/EntityOrchestration.cs
```diff
@@ -0,0 +1,53 @@
...
+using Microsoft.Azure.WebJobs;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+using Microsoft.Azure.WebJobs.Extensions.Http;
+using Newtonsoft.Json.Linq;
+
+namespace FunctionAppCorrelation
+{
+    /// <summary>
+    /// This example is testing for Entity works on the new correlation implementation.
+    /// Distributed Tracing for Entity will be next release.
```
ConnorMcMahon, line 21: Let's delete this line, since it is this release :)

samples/correlation-csharp/FunctionAppCorrelation/EternalOrchestrations.cs
```diff
@@ -0,0 +1,70 @@
...
+                await context.CreateTimer(nextCleanup, CancellationToken.None);
+                context.ContinueAsNew(expire);
+            }
+        }
+
+        [FunctionName(nameof(CleanUpNotification))]
+        public void CleanUpNotification(
+            [ActivityTrigger] string message)
+        {
+            this.telemetryClient.TrackTrace(message);
```
ConnorMcMahon, line 48: It would be nice to comment when you want to use the telemetry client explicitly like this, vs what is handled automatically.

samples/correlation-csharp/FunctionAppCorrelation/FanOutFanIn.cs
ConnorMcMahon: We could use compilation tricks to have a common orchestration defined in one place, but include it in multiple projects. The main difference between the files appears to be the namespace, and I don't know if that matters.

ConnorMcMahon (commented):
Made it through the samples, will go through the extension source/test code changes tomorrow.

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
```diff
@@ -0,0 +1,99 @@
...
+using DurableTask.Core.Settings;
+using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.Channel;
+using Microsoft.ApplicationInsights.Extensibility;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Correlation
+{
+    /// <summary>
+    /// TelemetryActivator activates Distributed Tracing. This class only works for netstandard2.0.
```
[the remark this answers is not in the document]
ConnorMcMahon: "initializes" may be better terminology here, since this all seems setup related?

src/WebJobs.Extensions.DurableTask/Correlation/TraceContextBaseExtensions.cs
ConnorMcMahon: I'm not sure how you would do that, considering the type system in C#.

src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask.Telemetry-net461.xml
```diff
@@ -0,0 +1,3721 @@
+<?xml version="1.0"?>
```
ConnorMcMahon, line 1: The .Telemetry versions of these xmls are no longer required now that this is one single project and not a separate project.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -43,25 +43,38 @@
   
   <PropertyGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
     <DefineConstants>FUNCTIONS_V1</DefineConstants>
+    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
```
ConnorMcMahon, line 46: Not sure what this is... will want to look into it. I'm assuming it's related to the Compile Include sections below.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
-    <AssemblyName>Microsoft.Azure.WebJobs.Extensions.DurableTask</AssemblyName>
+    <AssemblyName>Microsoft.Azure.WebJobs.Extensions.DurableTask.Telemetry</AssemblyName>
```
ConnorMcMahon, line 5: We will want to revert this. The name change was to create alpha packages.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -43,25 +43,38 @@
...
 
     <!--Don't increase below versions without significantly testing on Functions V1. Increasing these versions increments
     some dependencies that have binding redirects in Functions V1.-->
     <PackageReference Include="Microsoft.Extensions.Options" Version="1.0.2" />
     <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="1.0.2" />
-    <PackageReference Include="Microsoft.AspNetCore.Routing" Version="1.0.3" />
-    <PackageReference Include="Microsoft.AspNetCore.Mvc.WebApiCompatShim" Version="1.0.3" />
+    <PackageReference Include="Microsoft.AspNetCore.Routing" Version="1.0.5" /> <!-- used to be 1.0.3 -->
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.WebApiCompatShim" Version="1.0.5" /> <!-- used to be 1.0.3 -->
+    <PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="2.2.3" />
```
ConnorMcMahon, line 63: This seems incorrect now.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -43,25 +43,38 @@
...
   <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.0'">
     <PackageReference Include="Microsoft.Azure.WebJobs" Version="2.3.0" />
 
     <!--Don't increase below versions without significantly testing on Functions V1. Increasing these versions increments
     some dependencies that have binding redirects in Functions V1.-->
     <PackageReference Include="Microsoft.Extensions.Options" Version="1.0.2" />
     <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="1.0.2" />
-    <PackageReference Include="Microsoft.AspNetCore.Routing" Version="1.0.3" />
-    <PackageReference Include="Microsoft.AspNetCore.Mvc.WebApiCompatShim" Version="1.0.3" />
+    <PackageReference Include="Microsoft.AspNetCore.Routing" Version="1.0.5" /> <!-- used to be 1.0.3 -->
```
ConnorMcMahon, line 61: Not sure why this was changed... it concerns me a bit, as Functions v1 is fairly fragile with this.

test/FunctionsV2/CorrelationEndToEndTests.cs
ConnorMcMahon: I'm pretty sure our CI requires specific collections for the tests to run.

ConnorMcMahon (commented):
Some more comments of things to address.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
davidmrdavid: @TsuyoshiUshio do you know why this changed?

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
davidmrdavid: @TsuyoshiUshio do you know why this changed?

src/WebJobs.Extensions.DurableTask/Correlation/TraceContextBaseExtensions.cs
davidmrdavid: I suppose they'd have to both inherit from the same interface, I'll resolve this one as it's more of a nit

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
davidmrdavid: @TsuyoshiUshio , do you know?

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
ConnorMcMahon: We may be able to look at the original PR that created the correlation branch, as that had several iterations of PR review and it may have came up there.

src/WebJobs.Extensions.DurableTask/Correlation/NoOpTelemetryChannel.cs
```diff
@@ -0,0 +1,83 @@
...
+        public Action OnDispose { get; set; }
+
+        /// <summary>
+        /// Implements the <see cref="ITelemetryChannel.Send"/> method by invoking the <see cref="OnSend"/> callback.
+        /// </summary>
+        public void Send(ITelemetry item)
+        {
+            if (this.ThrowError)
+            {
+                throw new Exception("test error");
```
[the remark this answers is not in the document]
davidmrdavid: Moving to ToDos issue

src/WebJobs.Extensions.DurableTask/Correlation/NoOpTelemetryChannel.cs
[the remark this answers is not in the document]
davidmrdavid: Moving to ToDos issue

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
davidmrdavid: Moving to ToDos issue

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
davidmrdavid: Moving to ToDos issue

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
davidmrdavid: Moving to ToDos issue

samples/correlation-csharp/Readme.md
davidmrdavid: Moving to ToDos issue

samples/correlation-csharp/FunctionAppCorrelation/FanOutFanIn.cs
davidmrdavid: Moving to ToDos issue

src/WebJobs.Extensions.DurableTask/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
```diff
@@ -0,0 +1,319 @@
...
+        internal static void UpdateTelemetry(ITelemetry telemetry, TraceContextBase contextBase)
+        {
+            switch (contextBase)
+            {
+                case NullObjectTraceContext nullObjectContext:
+                    return;
+                case W3CTraceContext w3cContext:
+                    UpdateTelemetryW3C(telemetry, w3cContext);
+                    break;
+                case HttpCorrelationProtocolTraceContext httpCorrelationProtocolTraceContext:
```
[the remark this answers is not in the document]
davidmrdavid: Yeah. Might just leave it as-is for now, until we get more context. Moving to ToDos issue.

src/WebJobs.Extensions.DurableTask/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
davidmrdavid: Definitely a mouthful, but that's the price to pay for abstractions :P

samples/correlation-csharp/.gitattributes
davidmrdavid: This file should have been deleted. If you resolve this issue, I believe it might go away from the UI?

davidmrdavid:
This is the ToDo file, for follow-up tasks after merging this: https://github.com/Azure/azure-functions-durable-extension/issues/1576

src/WebJobs.Extensions.DurableTask/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
```diff
@@ -0,0 +1,319 @@
...
+                if (telemetry.Context.Operation.ParentId == null)
+                {
+                    telemetry.Context.Operation.ParentId = traceParent.SpanId;
+                }
+            }
+        }
+
+        internal void SuppressTelemetry(ITelemetry telemetry)
+        {
+            // TODO For suppressing Dependency, I make the Id as suppressed. This stragey increase the number of telemetery.
```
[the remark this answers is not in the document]
TsuyoshiUshio: This method suppressing dependency call to the storage queue (https access) Application Insights Automatically Tracked a bunch of  storage queue/table call (http access) automatically, if the telemetry has the same operation id, it includes a lot of noizes to the end to end tracing view.  If that issue is already gone,  this logic don't need it.

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
TsuyoshiUshio: As I remember correctly, It is for reducing local debugging noize.  Dependency telemetry is automatically tracked by the Application Insights. That means, it automatically tracked the storage queue/table call via http. I tried to avoid it. If there is a good way, I don't matter.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
TsuyoshiUshio: I don't clearly remember it. However, Chris said that this feature (Distributed Tracing) is only for v2. (to encourage people to move v2) So that I try to make it.  It is just a matter of the compile of the library.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
TsuyoshiUshio: It should be match with the DurableTask.Core version that includes Correlation feature.

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
davidmrdavid: Added to ToDos issues

src/WebJobs.Extensions.DurableTask/Correlation/DurableTaskCorrelationTelemetryInitializer.cs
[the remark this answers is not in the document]
davidmrdavid: Moved to ToDo issues

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
davidmrdavid: I just read about it here: http://jameschambers.com/2017/11/duplicate-compile-items-included/
and here: https://docs.microsoft.com/en-us/dotnet/core/tools/csproj

In short, in response to changes to the defaults of how modern .NET builds projects, this flags prevents unnecessary files from being added to our project. Since we haven't updated our .NET version significantly, I doubt this is necessary for us. I'll remove it

davidmrdavid: Removed

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -5,8 +5,8 @@
     <AssemblyName>Microsoft.Azure.WebJobs.Extensions.DurableTask</AssemblyName>
     <RootNamespace>Microsoft.Azure.WebJobs.Extensions.DurableTask</RootNamespace>
     <MajorVersion>2</MajorVersion>
-    <MinorVersion>3</MinorVersion>
-    <PatchVersion>1</PatchVersion>
+    <MinorVersion>5</MinorVersion>
```
davidmrdavid, line 8: will remove. Accidental push

samples/correlation-csharp/FunctionAppCorrelation/EternalOrchestrations.cs
davidmrdavid: Added!

davidmrdavid: The GitHub preview UI just hasn't updated though...

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
davidmrdavid: This has been simplified in the my last few commits

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
davidmrdavid: This has been simplified in the my last few commits

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
davidmrdavid: Actually, this appears to be necessary. I think the new `samples/` project is automatically getting loaded onto the FunctionsV1 project, which is causing issues because Dist. Tracing only works for FunctionsV2. Without this line, we're getting compilation errors.

I think we should kept it and add it as a ToDo to remove this once we clean-up the samples

davidmrdavid: Added to ToDos issue: https://github.com/Azure/azure-functions-durable-extension/issues/1576

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -2674,8 +2674,8 @@ public async Task SetStatusOrchestration(bool extendedSessions, string storagePr
                 await client.WaitForCustomStatusAsync(TimeSpan.FromSeconds(10), this.output, statusValue);
 
                 // Test clearing an existing custom status
-                await client.RaiseEventAsync("UpdateStatus", null, this.output);
-                await client.WaitForCustomStatusAsync(TimeSpan.FromSeconds(10), this.output, JValue.CreateNull());
+                await client.RaiseEventAsync("UpdateStatus", "", this.output);
+                await client.WaitForCustomStatusAsync(TimeSpan.FromSeconds(30), this.output, JValue.CreateNull());
```
davidmrdavid, line 2678: Currently a hack. I found that `RaiseEventAsync` was failing when the `eventData` argument was `null`. It sent a message reading something like "value cannot null". Did something change in our implementation of `RaiseEventAsync`?

ConnorMcMahon: Yeah, there were changes with how we handled `null` objects serializing in DurableTask.Core. I'm looking into the best way to fix it now. https://github.com/Azure/durabletask/pull/465/files.

ConnorMcMahon: I am going to fix that before I merge https://github.com/Azure/azure-functions-durable-extension/pull/1577.

samples/correlation-csharp/FunctionAppCorrelation/EntityOrchestration.cs
```diff
@@ -0,0 +1,52 @@
...
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Azure.WebJobs;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+using Microsoft.Azure.WebJobs.Extensions.Http;
+using Newtonsoft.Json.Linq;
+
+namespace FunctionAppCorrelation
+{
+    /// <summary>
+    /// This example is testing for Entity works on the new correlation implementation.
```
ConnorMcMahon, line 20:
```suggestion
    /// This example is for testing that Durable Entities work with the new correlation implementation.
```

samples/correlation-csharp/FunctionAppCorrelation/SimpleCorrelationDemo.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            this.telemetryClient = new TelemetryClient(telemetryConfiguration);
+            this.httpClient = client;
+        }
+
+        [FunctionName(nameof(Orchestration_W3C))]
+        public async Task<List<string>> Orchestration_W3C(
+           [OrchestrationTrigger] IDurableOrchestrationContext context)
+        {
+            var correlationContext = CorrelationTraceContext.Current as W3CTraceContext;
```
ConnorMcMahon, line 40: This would be a bit cleaner as the following:

```csharp
if (CorrelationTraceContext.Current is W3CTraceContext correlationContext)
{
    var trace = new TraceTelemetry(
        $"Activity Id: {correlationContext.TraceParent} ParentSpanId: {correlationContext.ParentSpanId}");
    trace.Context.Operation.Id = correlationContext.TelemetryContextOperationId;
    trace.Context.Operation.ParentId = correlationContext.TelemetryContextOperationParentId;
    this.telemetryClient.Track(trace);
}

File: src/WebJobs.Extensions.DurableTask/Correlation/ITelemetryActivator.cs [hunk 29]
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Correlation
+{
+    /// <summary>
+    /// ITelemetryActivator is an interface.
+    /// </summary>
+    public interface ITelemetryActivator
```
ConnorMcMahon, line 9: If we want to rename this, we have to do it now, since it is a public interface.

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
```diff
@@ -45,6 +46,9 @@ public static IWebJobsBuilder AddDurableTask(this IWebJobsBuilder builder)
             serviceCollection.TryAddSingleton<IMessageSerializerSettingsFactory, MessageSerializerSettingsFactory>();
             serviceCollection.TryAddSingleton<IErrorSerializerSettingsFactory, ErrorSerializerSettingsFactory>();
             serviceCollection.TryAddSingleton<IApplicationLifetimeWrapper, HostLifecycleService>();
+#if !FUNCTIONS_V1
+            serviceCollection.AddSingleton<ITelemetryActivator, TelemetryActivator>();
```
ConnorMcMahon, line 50: We don't need to ever let customers inject their own custom implementation of this interface, right?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -109,7 +118,14 @@ public DurableTaskExtension()
             IDurableHttpMessageHandlerFactory durableHttpMessageHandlerFactory = null,
             ILifeCycleNotificationHelper lifeCycleNotificationHelper = null,
             IMessageSerializerSettingsFactory messageSerializerSettingsFactory = null,
+#if !FUNCTIONS_V1
+            IErrorSerializerSettingsFactory errorSerializerSettingsFactory = null,
+#pragma warning disable SA1113, SA1001, SA1115
+            ITelemetryActivator telemetryActivator = null)
```
ConnorMcMahon, line 124: From what I can tell this is never _not_ provided, at least in production, because we inject it via Dependency Injection. If it is null in our test code, let's just feed in a no-op version of the interface.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -76,6 +84,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.0.3" />
```
ConnorMcMahon, line 87: Do we need this as a part of this PR?

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -2675,8 +2675,8 @@ public async Task SetStatusOrchestration(bool extendedSessions, string storagePr
                 await client.WaitForCustomStatusAsync(TimeSpan.FromSeconds(10), this.output, statusValue);
 
                 // Test clearing an existing custom status
-                await client.RaiseEventAsync("UpdateStatus", null, this.output);
-                await client.WaitForCustomStatusAsync(TimeSpan.FromSeconds(10), this.output, JValue.CreateNull());
+                await client.RaiseEventAsync("UpdateStatus", "", this.output);
```
ConnorMcMahon, line 2678: Reminder to change this back to null.

ConnorMcMahon (commented):
Getting closer, just a few nits and new things I found + some reminders.

Let me know if you need help investigating timeouts and the emulator tests.

test/Common/DurableTaskEndToEndTests.cs
davidmrdavid: Changed!

samples/correlation-csharp/FunctionAppCorrelation/EntityOrchestration.cs
davidmrdavid: Changed! Sorry, I would have committed this suggestion directly from the UI tool, but I didn't catch that option until now :)

samples/correlation-csharp/FunctionAppCorrelation/SimpleCorrelationDemo.cs
davidmrdavid: I actually don't know why that's clearer. Judging by the names of this demo, the correlation context is always meant to be a `W3CTraceContext` so I don't see why we should replace the `as` coercion to an `if`-check.

src/WebJobs.Extensions.DurableTask/Correlation/ITelemetryActivator.cs
davidmrdavid: I don't think this is a bad name, but I don't think this needs to be `public`, right? Looks to me like it should be safe to make it `private`.

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
davidmrdavid: @TsuyoshiUshio , could you please review connor's question above ^.

As for my own guess, think the answer is *no*. I don't see the use-case.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
davidmrdavid: @TsuyoshiUshio , do you remember what you needed this dependency for?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
davidmrdavid: By "this" did you mean the `telemetryActivator` or the `errorSerializerSettingsFactory `?

davidmrdavid: I've also been playing with how to simplify this conditional compilation bit. It looks like it could be simplified since `errorSerializerSettingsFactory` is always present, but there's no clean way to add it without violating linter-rules. There are also a few places where this constructor's optional parameters are called without explicitly specifying the parameter's names, so the order of params needs to remain intact.

I would personally vote to keep this _as-is_, for now.

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
TsuyoshiUshio: No.  We don't need to let customer inject their won custom implementation. 
Some customer want open telemetry however, in this case, it should be done by PR, since there are other strategy classes are there e.g. `W3CTraceContext ...`  and `HttpCorrelationProtocol...`

src/WebJobs.Extensions.DurableTask/Correlation/ITelemetryActivator.cs
davidmrdavid: Judging by Tsuyoshi's new comments, it sounds like this should remain public if we want to support customers using alternative telemetry protocols. So, if that's a feature we want to support (and I think we should), then this can stay public.

As for the name, I really think this is _fine_. Or were you thinking of some other name?

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
davidmrdavid: Fair enough. Sounds to me then like there's merit in leaving the possibility for customers to provide their customer implementations then.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
davidmrdavid: Seems like nothing broke! Removed.

davidmrdavid: The UI just hasn't updated, but this is in fact gone.

davidmrdavid:
With the latest commit, we're finally passing the tests. I've also updated the `ToDos` issue with the latest findings. I believe this is pretty much ready to merge, but happy to take on other suggestions. Let's aim to merge this before noon PDT tomorrow.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
ConnorMcMahon: I was referring to `telemetryActivator`, but I think I am fine leaving it for now.

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
ConnorMcMahon: Let's leave it as `AddSingleton` for now, and if customers need to provide their own implementation later, we can open that up by converting this to `TryAddSingleton`

src/WebJobs.Extensions.DurableTask/Correlation/ITelemetryActivator.cs
ConnorMcMahon: I thought I remembered it being something you hadn't like, but I could be thinking of a different naming scheme. I don't have a big problem with it.

samples/correlation-csharp/FunctionAppCorrelation/SimpleCorrelationDemo.cs
ConnorMcMahon: I don't love null coalescing after an `as` conversion. It means that if for some reason the customer switched the CorrelationContext away from `W3CTraceContext` after we added support for a new protocol, then this would start failing in weird ways (specifically, a bunch of fields would silently go `null`.

I guess in that case, a preferable piece of code would look like this:

```csharp
if (!CorrelationTraceContext.Current is W3CTraceContext correlationContext)
{
    throw new InvalidOperationException($"This sample expects a correlation trace context of {nameof(W3CTraceContext)}, but the context isof type {CorrelationTraceContext.Current.GetType()}");
}

var trace = new TraceTelemetry(
    $"Activity Id: {correlationContext.TraceParent} ParentSpanId: {correlationContext.ParentSpanId}");
trace.Context.Operation.Id = correlationContext.TelemetryContextOperationId;
trace.Context.Operation.ParentId = correlationContext.TelemetryContextOperationParentId;
this.telemetryClient.Track(trace);
```

davidmrdavid: Ah yes, I see what you mean. I wasn't thinking of the full implications of the coalescing operation. I'll go for your new suggestion. Thanks!

ConnorMcMahon (approved):
LGTM!

PR #1572, opened 2020-11-19
Updated WaitForCompletionOrCreateCheckStatusResponseAsync API

### LINKED ISSUE #1471: WaitForCompletionOrCreateCheckStatusResponseAsync has no way to restore default behavior
In versions 1.x of the extension, calling the Get Status API would return a 500 if the orchestration instance had failed. To follow better API design, in 2.x of the extension, we changed the default behavior to return a 200 while simply marking the status field as "Failed". On the actual HTTP API and the `IDurableClient.CreateCheckStatusResponse` API, we have an optional parameter `returnInternalServerErrorOnFailure` to restore the 1.x behavior.

As @dixonte pointed out, `IDurableClient.WaitForCompletionOrCreateCheckStatusResponseAsync` is missing this optional parameter, leaving no way to have it return an HTTP response payload with the correct query string parameter filled out.

amdeel:
Updated WaitForCompletionOrCreateCheckStatusResponseAsync API to include the parameter returnInternalServerErrorOnFailure. This was left out when we added this parameter to the other CreateCheckStatusResponseAsync APIs.

resolves #1471

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -104,18 +104,19 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         }
 
         /// <inheritdoc />
-        async Task<HttpResponseMessage> IDurableOrchestrationClient.WaitForCompletionOrCreateCheckStatusResponseAsync(HttpRequestMessage request, string instanceId, TimeSpan? timeout, TimeSpan? retryInterval)
+        async Task<HttpResponseMessage> IDurableOrchestrationClient.WaitForCompletionOrCreateCheckStatusResponseAsync(HttpRequestMessage request, string instanceId, TimeSpan? timeout, TimeSpan? retryInterval, bool returnInternalServerErrorOnFailure)
         {
             return await this.WaitForCompletionOrCreateCheckStatusResponseAsync(
                 request,
                 instanceId,
                 this.attribute,
                 timeout ?? TimeSpan.FromSeconds(10),
-                retryInterval ?? TimeSpan.FromSeconds(1));
+                retryInterval ?? TimeSpan.FromSeconds(1),
+                returnInternalServerErrorOnFailure);
         }
 
         /// <inheritdoc />
-        async Task<IActionResult> IDurableOrchestrationClient.WaitForCompletionOrCreateCheckStatusResponseAsync(HttpRequest request, string instanceId, TimeSpan? timeout, TimeSpan? retryInterval)
+        async Task<IActionResult> IDurableOrchestrationClient.WaitForCompletionOrCreateCheckStatusResponseAsync(HttpRequest request, string instanceId, TimeSpan? timeout, TimeSpan? retryInterval, bool returnInternalServerErrorOnFailure)
         {
             HttpRequestMessage requestMessage = ConvertHttpRequestMessage(request);
             HttpResponseMessage responseMessage = await ((IDurableOrchestrationClient)this).WaitForCompletionOrCreateCheckStatusResponseAsync(requestMessage, instanceId, timeout, retryInterval);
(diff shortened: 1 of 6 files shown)


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -104,18 +104,19 @@ HttpManagementPayload IDurableOrchestrationClient.CreateHttpManagementPayload(st
         }
 
         /// <inheritdoc />
-        async Task<HttpResponseMessage> IDurableOrchestrationClient.WaitForCompletionOrCreateCheckStatusResponseAsync(HttpRequestMessage request, string instanceId, TimeSpan? timeout, TimeSpan? retryInterval)
+        async Task<HttpResponseMessage> IDurableOrchestrationClient.WaitForCompletionOrCreateCheckStatusResponseAsync(HttpRequestMessage request, string instanceId, TimeSpan? timeout, TimeSpan? retryInterval, bool returnInternalServerErrorOnFailure)
```
ConnorMcMahon, line 107: We will need to document this as a breaking change for customers. I don't anticipate it will bother many customers, since they can just add a `It.IsAny<bool>()` to their setup in their mocked function call.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -195,7 +196,7 @@ private static TemplateMatcher GetInstanceRaiseEventRoute()
                         status.RuntimeStatus == OrchestrationRuntimeStatus.Failed ||
                         status.RuntimeStatus == OrchestrationRuntimeStatus.Terminated)
                     {
-                        return await this.HandleGetStatusRequestAsync(request, instanceId);
+                        return await this.HandleGetStatusRequestAsync(request, instanceId, returnInternalServerErrorOnFailure);
                     }
                 }
```
ConnorMcMahon, line 201: It's not featured in this code segment, but I believe we need to also provide `returnInternalServerErrorOnFailure` in our call to CreateCheckStatusResponseMessage below.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -538,9 +540,12 @@ private async Task<HttpResponseMessage> HandleDeleteHistoryWithFiltersRequestAsy
                 showInput = true;
             }
 
-            if (!TryGetBooleanQueryParameterValue(queryNameValuePairs, ReturnInternalServerErrorOnFailure, out bool returnInternalServerErrorOnFailure))
+            if (returnInternalServerErrorOnFailure == false)
```
ConnorMcMahon, line 543: What if instead of making the default value for the parameter `returnInternalServerErrorOnFailure` false, we made the type `bool?` and default to `null`. Then that allows us to distinguish between the case where the value wasn't provided vs someone intentionally passing in `false`. 

We would still have the guarantee that `returnInternalServerErrorOnFailure` is non-null after this code segment due to the way we set it if the query string doesn't exist.

ConnorMcMahon (changes requested):
It's very close, but there are a few subtle behavior things I think we should address before merging.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
amdeel: Why/Where do I need to do that? Are you talking about the call 12 lines down? I actually don't need to add it there because the HttpManagementPayload with the URIs was created from GetClinetResponseLinks on line 181.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
amdeel: Ah, I like that solution.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
ConnorMcMahon: Ah, I just misunderstood that code segment. We are all good here.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
amdeel: This would actually make the type that is output from TryGetBooleanQueryParameterValue a compiler error. I would either have to change the output to bool? also or just not make this change.

ConnorMcMahon: It maybe worth setting a new `bool` variable here for `returnInternalServerErrorOnFailure` (it would need a new name obviously). I still think the nullable input to this method is important to help distinguish between `false` and the default value.

Otherwise you could get some weird unexpected behavior where we look at the query string parameters of the HTTP Trigger being able to override the value explicitly passed into `WaitForCompletionOrCreateCheckStatusResonseAsync()`, which feels incorrect to me.

amdeel: That's fair. I updated this in my latest commit.

ConnorMcMahon (approved):
I would like to see my last comment resolved before merging, but after that, it looks good!

ConnorMcMahon (approved):
LGTM!

PR #1574, opened 2020-11-19
Adding Azure Function App sample with Durable Client dependency injection

bachuv:
This sample uses the Durable Client dependency injection feature to call a Durable Function orchestrator from an Azure Function HTTP Trigger.


samples/functionapp-csharp/README.md
```diff
@@ -0,0 +1,45 @@
+# Azure Function App with Durable Client Dependency Injection
+
+This project contains an Azure Function App that calls a Durable Function through a Durable Client dependency injection.
+
+## Local setup
+
+Create a local.settings.json file and add two properties: a `Storage` property and a `TaskHub` property. Add the storage account connection string that you are using for the Durable Function. Having the same storage account allows the two function apps to communicate.
+
+```json
```
ConnorMcMahon, line 9: We should consider shipping with a default value for local.settings.json here (but have it explicitly git-ignored) so that customers just have to tweak the file rather than creating it from scratch.

ConnorMcMahon (approved):
Looking good, just one small suggestion.

ConnorMcMahon (changes requested):
This isn't an actual request for changes, just a note that this PR is likely blocked until we release v2.4.0, since it requires that nuget package explicitly.

samples/functionapp-csharp/README.md
bachuv: Should I just create a local.settings.json file with default values then? The precompiled samples folder includes a local.settings.json file so it looks like it's already included in the .gitignore.

ConnorMcMahon: Yeah, you can just copy what you have in this json blob right here and put that in the local.settings.json. Leave the placeholders as is in the json, and just remark here in the docs that customers should configure those values before running the samples.

samples/functionapp-csharp/HelloSequence.cs
```diff
@@ -0,0 +1,52 @@
...
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask.ContextImplementations;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask.Options;
+using Microsoft.Extensions.Configuration;
+
+namespace DurableClientSampleFunctionApp
+{
+    public class HelloSequence
```
ConnorMcMahon, line 17: Let's not include any orchestrations for this app so it is clear we are using exclusively as a client.

bachuv: This app contains a function with an HTTP trigger, not an orchestrator function. I can see the confusion naming the class HelloSequence though, so I renamed it to ClientFunction. Let me know if you meant something else and I'll address that!

ConnorMcMahon (approved):
LGTM!

PR #1580, opened 2020-11-23
Update azure-pipelines.yml for Azure Pipelines

davidmrdavid:
Adding the application insights environment variables to the V2 tests


davidmrdavid:
Closing so we can add this, and test this, in the correlation PR

Issue #1419, opened 2020-07-30
Unable to write test cases for IDurableEntityContext.DispatchAsync


abbasc52:
Hi i am getting an error when i mock IDurableEntityContext and call DispatchAsync

```
 Message: 
    System.InvalidCastException : Unable to cast object of type 'Castle.Proxies.IDurableEntityContextProxy' to type 'Microsoft.Azure.WebJobs.Extensions.DurableTask.DurableEntityContext'.
```

On looking at the source i found this which forces the input to be of type DurableEntityContext:
https://github.com/Azure/azure-functions-durable-extension/blob/affadf71b71d3711c3f57c9420b7e81a6e9ede1f/src/WebJobs.Extensions.DurableTask/EntityScheduler/TypedInvocationExtensions.cs#L78

Is there any way i can get around this?

cgillum:
Hmm...I'm not sure if our current design allows this method to be mocked.

/cc @sebastianburckhardt

abbasc52:
Can we replace `GetStateWithInjectedDependencies` with `GetState` ? The code is more or less same and can solve this issue with few minor changes.
I can help with raising a Pull request if needed.

ConnorMcMahon:
Unfortunately this was missed in our attempt to move away from extension methods in 2.3.0.

It is something that should be fairly straightforward to move from an extension method to a direct interface method I believe.

@cgillum, should we consider this for our next release?

ConnorMcMahon:
I think I may be able to fit this into v2.4.0.

PR #1581, opened 2020-11-25
Add IDurableEntityContext.DispatchAsync() directly to interface

ConnorMcMahon:
In a similar vein to PR #1422, this PR aims to transfer an extension
method to a direct interface implementation. This allows the method to
be directly mockable via tooling like Moq.

Addresses #1419


test/Common/DurableClientBaseTests.cs
```diff
@@ -166,7 +166,7 @@ public async void DurableClient_ExternalApp_GetStatusAsync_ReturnsStatus()
             orchestrationServiceClientMock.Setup(x => x.GetOrchestrationStateAsync(It.IsAny<string>(), It.IsAny<bool>()))
                 .ReturnsAsync(GetInstanceState(OrchestrationStatus.Running));
 
-            var durableOrchestrationClient = this.GetDurableClient(orchestrationServiceClientMock.Object);
```
davidmrdavid, line 169: Surprised we had some whitespace in the codebase. Don't we have linter rules against it?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
```diff
@@ -415,12 +417,94 @@ string IDurableEntityContext.StartNewOrchestration(string functionName, object i
...
+            var interfaces = type.GetInterfaces();
+            const BindingFlags bindingFlags = BindingFlags.IgnoreCase | BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance;
+
+            var method = type.GetMethod(context.OperationName, bindingFlags);
+            if (interfaces.Length == 0 || method != null)
+            {
+                return method;
+            }
+
+            return interfaces.Select(i => i.GetMethod(context.OperationName, bindingFlags)).FirstOrDefault(m => m != null);
```
davidmrdavid, line 505: So this gets us the first result of a potential list of matching methods, right? under what conditions would we see a  result list with more than 1 element? Is that even possible for our use-case? 

I think it would be good to add a comment answering the questions above :)

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
```diff
@@ -415,12 +417,94 @@ string IDurableEntityContext.StartNewOrchestration(string functionName, object i
             return instanceId;
         }
 
+        async Task IDurableEntityContext.DispatchAsync<T>(params object[] constructorParameters)
```
davidmrdavid, line 420: Should there be a `\\\ <inheritdoc/>` (inherit documentation) comment over this?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
```diff
@@ -415,12 +417,94 @@ string IDurableEntityContext.StartNewOrchestration(string functionName, object i
...
+                {
+                    Entity.Current.DeleteState();
+                    return;
+                }
+                else
+                {
+                    throw new InvalidOperationException($"No operation named '{context.OperationName}' was found.");
+                }
+            }
+
```
davidmrdavid, line 437: I think this block could have a small clarification comment. Is it expected that `"delete"` won't be found via `FindMethodForContext` or is that only under some edge-case? Also, do we expect to grow this if-statement with other special operations?

I would be good to add a clarifying comment answering the questions above :) ^

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityContext.cs
```diff
@@ -169,5 +171,25 @@ public interface IDurableEntityContext
         /// </exception>
         /// <returns>The instance id of the new orchestration.</returns>
         string StartNewOrchestration(string functionName, object input, string instanceId = null);
+
+        /// <summary>
+        /// Dynamically dispatches the incoming entity operation using reflection.
```
davidmrdavid, line 176: Just for my own understanding, we _have_ to use reflection because the entity is not known statically, right?

davidmrdavid (approved):
It looks good to me. I did leave some suggestions for documentation improvements, but this seems reasonable. I'm _assuming_ that we *have* to use reflection for this, right?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
ConnorMcMahon: I'm not sure. I copied this from the existing extension method code.

My guess is that we automatically have functionality for "Delete" built in, so the customer doesn't need their own implementation.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityContext.cs
ConnorMcMahon: Not sure what you mean by this. 

Essentially we use reflection to create the object given the type they provide, because that is the toolset that C# gives us to create objects given a specific type.

test/Common/DurableClientBaseTests.cs
ConnorMcMahon: We do, but they are warnings, and warnings currently only fail in dev, because some of our tests purposely have bad practices.

It may be worth explicitly ignoring those warning, and making warnings errors in test code as well.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
davidmrdavid: Fair enough. If you don't mind, it might be good to have that in the code itself, but I wont' block for it

test/Common/DurableClientBaseTests.cs
ConnorMcMahon: Filed issue for it here: https://github.com/Azure/azure-functions-durable-extension/issues/1583

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableEntityContext.cs
davidmrdavid: That answers my question :) 
To answer yours, what I meant by "not known statically" is simply that `T` can only be known at runtime, which you're saying is true!

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
ConnorMcMahon: Actually, the way this is implemented means that `<inheritdoc/>` isn't useful.

Essentially we implemented these interface methods explicitly, meaning that if someone has the object cast as `DurableEntityContext`, they can't actually call this method. They have to first cast it to `IDurableEntityContext` before they use the method, meaning the IDE will already be providing the docs.

I'm not in love with explicit interface method implementations, but they are spread throughout the codebase, so I keep it this way for consistency.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
ConnorMcMahon: I think this is a bug/edge case in the code, and I am a bit hesitant to tackle it in the scope of this PR.

This code assumes that the interfaces customers code against have no method overloads. However, I think we could theoretically support an interface like

```csharp
interface TestInterface 
{
   void TestMethod();

   void TestMethod(string parameter1);
}
```

We could theoretically look at the types of the argument provided to us in the case where the customer has multiple overloads for a single operation, and use that to determine which method overload to call. This does strike me as adding a bit more complicated logic for little gain, but the reflection would not be that challenging.

If we decide this is a scenario we don't want to support, then we should just throw and exception as soon as we detect multiple overloads for a method on an interface, instead of just picking the first, as that would lead to some strange exceptions.

ConnorMcMahon: @sebastianburckhardt , thoughts?

davidmrdavid: I see what you mean. I'm ok with moving this to a separate follow-up issue and unblocking this PR

amdeel (approved):

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
amdeel: Interesting case. If entity methods are meant to be different operations, it doesn't make sense to me to have more than one method with the same name. In the switch version of an entity this wouldn't be possible. I don't think we documented, though, that this was a violation. I think it makes sense to enforce this constraint, personally. If we do, I could add an analyzer for this as well.

ConnorMcMahon: I think it technically _is_ possible in the `switch` entities, you would just have a secondary set of `if-else` statements based on what arguments are present on the operation.

Now whether that is a good design we should encourage is a completely separate discussion...

ConnorMcMahon: Tracking this here: https://github.com/Azure/azure-functions-durable-extension/issues/1584.

PR #1586, opened 2020-11-25
Changed restartUri to restartPostUri

amdeel:
Changed restartUri to restartPostUri to make it clear that we require http post.


ConnorMcMahon (approved):
LGTM!

PR #1593, opened 2020-12-04
Remove ApplicationInsights.DependencyCollector package

shibayan:
The `Microsoft.ApplicationInsights.DependencyCollector` package is not used, but it will be included in the artifacts and unnecessarily increases the artifacts size

The tests were run locally. cc @TsuyoshiUshio @cgillum


davidmrdavid:
@shibayan, thank you for this contribution! I'm a bit hesitant to merge this just by the passing tests alone so I'll defer my approval until I test this on the Azure portal. I'll need a few days to get to it. It would also help me if you can confirm that this isn't breaking the distributed tracing visualization on the azure portal for you.

That aside, I suspect this PR is right: we're not using that package. It appears that we're _manually_ creating dependency telemetry instead. We're doing that via this method: https://github.com/Azure/azure-functions-durable-extension/blob/2b5e58229d829828368094c27bb80bd4808f78e9/src/WebJobs.Extensions.DurableTask/Correlation/TraceContextBaseExtensions.cs#L31-L49

I'll perform some end-to-end validation of this in the coming days and look to merging it next week! ⚡ ⚡

PR #1598, opened 2020-12-09
Fix the warning message for unsuccessful termination

cgillum:
I found this while investigating an unrelated customer issue. The `Status` message represents custom status, which is not what we intended to log (it's also dangerous to log this since it could include sensitive customer data). We should actually be logging the `OrchestrationStatus` value - i.e. "Running", "Completed", "Failed", etc.


amdeel (approved):

ConnorMcMahon (approved):
LGTM!

PR #1621, opened 2020-12-21
Reduce number of entity deserializations

sebastianburckhardt:
The current implementation re-constructs the entity state (i.e. constructs the object and populates the fields by deserializing the json) prior to processing each operation (if RollbackEntityOperationsOnExceptions=true, default) or each operation batch (if RollbackEntityOperationsOnExceptions=false). This happens even if that entity state is still in memory.

This seems wasteful; in this PR I modified the state management slightly so it can reuse the already existing object.


ConnorMcMahon (approved):
LGTM!

PR #1641, opened 2021-01-13
Replace the impl of Syscall.rename and remove Mono.Posix.NETStandard

### LINKED ISSUE #1597: Mono.Posix.NETStandard is making the deployment package size excessively large

`Mono.Posix.NETStandard` was included since Durable Functions v2.3.1, which increased the size of the deployment package by nearly three times.

Especially in the case of Windows, Mono.Posix is not necessary, only the size of the package will increase.

`Mono.Posix.NETStandard` is used to perform rename in the `LinuxAppServiceFileLogger`, an implementation that can be replaced with a couple of lines of P/Invoke.

**test branch**
https://github.com/Azure/azure-functions-durable-extension/compare/dev...shibayan:remove-mono-posix

You can reduce the package size by downgrading to Durable Functions v2.3.0.

**Additional context**

**raw directory size**

image

**zipped package size**

image

davidmrdavid:
Mono.Posix.NETStandard was included since Durable Functions v2.3.1, which increased the size of the deployment package by nearly three times.

This is a contribution by @shibayan :) 

And I have verified that this implementation still raises the correct `inotify` signal, so our telemetry pipeline should continue to work with it

https://github.com/Azure/azure-functions-durable-extension/issues/1597

resolves #1597 

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/LinuxAppServiceFileLogger.cs
+++ b/src/WebJobs.Extensions.DurableTask/LinuxAppServiceFileLogger.cs
@@ -6,12 +6,9 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
-
-#if !FUNCTIONS_V1
-using Mono.Unix.Native;
-#endif
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 {
@@ -173,9 +170,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             // Rename current file to older file.
 
 #if !FUNCTIONS_V1
-            Syscall.rename(this.logFilePath, this.archiveFilePath);
+            rename(this.logFilePath, this.archiveFilePath);
 #endif
 
         }
+
+#if !FUNCTIONS_V1
+        [DllImport("libc", SetLastError = true)]
+        private static extern int rename(string oldPath, string newPath);
+#endif
     }
 }
--- a/src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
+++ b/src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
@@ -70,7 +70,6 @@
     <PackageReference Include="Microsoft.Extensions.Http" Version="2.2.0" />
     <PackageReference Include="Microsoft.ApplicationInsights" Version="2.12.2" />
     <PackageReference Include="Microsoft.ApplicationInsights.DependencyCollector" Version="2.12.2" />
-    <PackageReference Include="Mono.Posix.NETStandard" Version="1.0.0" />
   </ItemGroup>
 
   <ItemGroup>


davidmrdavid:
As for adding unit tests for this, we don't have a linux enviroment in our durable-extension pipeline, do we? If we did, I could add an inotify test

ConnorMcMahon:
We could add a step to the smoke test validation step for release to make sure that high volume traffic on Linux doesn't miss data? Not sure how easy that would be and how much value it would have, but without substantial work, we definitely can't unit test this right now (we use windows workers on our CI).

src/WebJobs.Extensions.DurableTask/LinuxAppServiceFileLogger.cs
```diff
@@ -173,9 +170,14 @@ private void RollFiles()
...
 
 #if !FUNCTIONS_V1
-            Syscall.rename(this.logFilePath, this.archiveFilePath);
+            rename(this.logFilePath, this.archiveFilePath);
 #endif
 
         }
+
+#if !FUNCTIONS_V1
+        [DllImport("libc", SetLastError = true)]
```
ConnorMcMahon, line 179: How does `[DllImport]` work if the library is not available (for instance, on Windows)?

Does this dynamically try to import this functionality when `rename` is actually called? In that case, this likely isn't an issue since this logger should never actually be used on Windows.

ConnorMcMahon, line 179: Small nit: it looks like most of the resources I see that use P/Invoke with libc specifically callout "libc.so" instead of just "libc".

ConnorMcMahon (commented):
I am pretty sure this looks good, I just want to fully understand what is going on in the Windows (or even running locally on Mac) case, where "libc" is not available.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceFileLogger.cs
davidmrdavid: My understanding is that `DllImport` will not actually seek to invoke any system-level libraries until `rename` is called, so this shouldn't be an issue.

From this SO answer: https://stackoverflow.com/questions/39144458/what-happens-if-i-have-dllimport-c-net-for-an-api-not-supported-by-the-opera

davidmrdavid: I would agree with that, but `Mono` actually drops the `so` as seen here: https://github.com/mono/mono/blob/da11592cbea4269971f4b1f9624769a85cc10660/mcs/class/Mono.Posix/Mono.Posix/Syscall.cs#L323

Since we're trying to drop the `Mono` dependency (which we know does the right thing) I think there's good reason to implement this "Mono-style" :)

ConnorMcMahon: Interestingly, the source I got this impression from is [Mono's documentation](https://www.mono-project.com/docs/advanced/pinvoke/). Source code always wins though!

ConnorMcMahon: In that case, we should decide on whether in the degenerate case this logger class is used on a non-Linux platform if we fail fast with an exception, or we should wrap the call to `rename()` in a try-catch.

@cgillum, what are your thoughts in this scenario? I normally lean towards just letting the exception propagate, but since this is logging code I'm never quite sure of proper error handling.

davidmrdavid: I +1 failing fast if we're somehow using the logger in windows

davidmrdavid:
There are also some style-cop errors here. I'm thinking I'll fix them in an immediate follow-up PR since this is merging from a fork that we don't own.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceFileLogger.cs
cgillum: Don't we have code that always ensures we're only running this logger in Linux? I'm fine with letting it fail on Windows, though I recommend we log the failure before rethrowing any exception so that someone debugging will have a bit more context.

davidmrdavid: So the suggestion is to: catch, then log, then re-throw? Got it :)

davidmrdavid:
Merging this into `dajusto/fix-rename` where I'll make sure this passes the linter and that the new suggestions are incorporated before merging to `dev`. _Thank you for your contribution!_

PR #1663, opened 2021-01-26
Update new-release-template.md

amdeel:
I've expanded on some release tasks and reordered some as I saw necessary.

### DIFF
--- a/.github/ISSUE_TEMPLATE/new-release-template.md
+++ b/.github/ISSUE_TEMPLATE/new-release-template.md
@@ -20,7 +20,7 @@ _Due: <2-business-days-before-release>_
 - [ ] Update Durable Functions references (Analyzer? DTFx?) and check current version.
 - [ ] Add Durable Functions package to myget staging feed.
 - [ ] Check for package size, make sure it's not surprisingly heavier than a previous release.
-- [ ] Merge dev into main (Merge NOT Squash merge).
+- [ ] Merge dev into main. Person performing validation must approve PR. Important: Merge NOT Squash merge.
 
 **Validation**
 _Due: <1-business-days-before-release>_
--- a/.github/ISSUE_TEMPLATE/new-release-template.md
+++ b/.github/ISSUE_TEMPLATE/new-release-template.md
@@ -9,11 +9,11 @@ assignees: comcmaho, amdeel, davidmrdavid, bachuv
 
 **Prep DTFx Release: (assigned to:)**
 _Due: <2-3-business-days-before-release>_
-- [ ] Increment updated DTFx packages (either DT-AzureStorage only or DT-AzureStorage, DT-Core, and DT-Redis)
-- [ ] Delete DTFx test packages from myget.
+- [ ] Increment updated DTFx packages (either DT-AzureStorage only or DT-AzureStorage, DT-Core, DT-Emulator, and DT-Redis)
+- [ ] Delete DTFx test packages from staging myget.
 - [ ] Update OneBranch durabletask repo, run signing pipeline (using these notes:  [Durable Task Framework Build Pipeline] (https://microsoft.sharepoint.com/teams/AzureWebjobs/_layouts/OneNote.aspx?id=%2Fteams%2FAzureWebjobs%2FSiteAssets%2FAzureWebjobs%20Notebook&wd=target%28Planning%2FFunctions%20Post-GA.one%7CA43CF112-7272-481A-B23E-9AA5CA8EEE06%2FDurable%20Task%20Framework%20Build%20Pipeline%7CD0946823-6FB0-44E3-A57F-E252617B69CD%2F%29
 onenote:https://microsoft.sharepoint.com/teams/AzureWebjobs/SiteAssets/AzureWebjobs%20Notebook/Planning/Functions%20Post-GA.one#Durable%20Task%20Framework%20Build%20Pipeline&section-id={A43CF112-7272-481A-B23E-9AA5CA8EEE06}&page-id={D0946823-6FB0-44E3-A57F-E252617B69CD}&end) ).
-- [ ] Publish DTFx packages to myget for testing. (either DT-AzureStorage only or DT-AzureStorage, DT-Core, DT-Redis, and DT-Emulator)
+- [ ] Publish DTFx packages to staging myget for testing. (either DT-AzureStorage only or DT-AzureStorage, DT-Core, DT-Redis, and DT-Emulator)
 
 **Prep Release (assigned to: )**
 _Due: <2-business-days-before-release>_
--- a/.github/ISSUE_TEMPLATE/new-release-template.md
+++ b/.github/ISSUE_TEMPLATE/new-release-template.md
@@ -7,30 +7,41 @@ assignees: comcmaho, amdeel, davidmrdavid, bachuv
 
 ---
 
-**Prep Release: (assigned to:)**
-_Due: <3-business-days-before-release>_
-- [ ] Merge all features and fixes into the `dev` branch
-- [ ] Publish updated versions of DurableTask.Core and DurableTask.AzureStorage to app-service MyGet feed
-- [ ] Update dependencies and increment extension version in the `dev` branch.
-- [ ] Publish signed version of Microsoft.Azure.WebJobs.Extensions.DurableTask to app-service MyGet feed
-- [ ] Close out or punt remaining GitHub issues for the current milestone
+**Prep DTFx Release: (assigned to:)**
+_Due: <2-3-business-days-before-release>_
+- [ ] Increment updated DTFx packages (either DT-AzureStorage only or DT-AzureStorage, DT-Core, and DT-Redis)
+- [ ] Delete DTFx test packages from myget.
+- [ ] Update OneBranch durabletask repo, run signing pipeline (using these notes:  [Durable Task Framework Build Pipeline] (https://microsoft.sharepoint.com/teams/AzureWebjobs/_layouts/OneNote.aspx?id=%2Fteams%2FAzureWebjobs%2FSiteAssets%2FAzureWebjobs%20Notebook&wd=target%28Planning%2FFunctions%20Post-GA.one%7CA43CF112-7272-481A-B23E-9AA5CA8EEE06%2FDurable%20Task%20Framework%20Build%20Pipeline%7CD0946823-6FB0-44E3-A57F-E252617B69CD%2F%29
+onenote:https://microsoft.sharepoint.com/teams/AzureWebjobs/SiteAssets/AzureWebjobs%20Notebook/Planning/Functions%20Post-GA.one#Durable%20Task%20Framework%20Build%20Pipeline&section-id={A43CF112-7272-481A-B23E-9AA5CA8EEE06}&page-id={D0946823-6FB0-44E3-A57F-E252617B69CD}&end) ).
+- [ ] Publish DTFx packages to myget for testing. (either DT-AzureStorage only or DT-AzureStorage, DT-Core, DT-Redis, and DT-Emulator)
 
-
-**Validation (assigned to: )**
+**Prep Release (assigned to: )**
 _Due: <2-business-days-before-release>_
-- [ ] Run private performance tests and ensure no regressions
-- [ ] Smoke test Functions V1 and Functions V3 .NET apps
-- [ ] Smoke test JavaScript and Python apps
-- [ ] Check for package size, make sure it's not surprisingly heavier than a previous release
+- [ ] Update Durable Functions references (Analyzer? DTFx?) and increment version.
+- [ ] Add Durable Functions package to myget staging feed.
+- [ ] Check for package size, make sure it's not surprisingly heavier than a previous release.
+- [ ] Merge dev into main (Merge NOT Squash merge).
 
+**Validation**
+_Due: <1-business-days-before-release>_
+- [ ] Run private performance tests and ensure no regressions. **(assigned to: )**
+- [ ] Smoke test Functions V1, Functions V2, and Functions V3 .NET apps. **(assigned to: )**
+- [ ] Smoke test JavaScript and Python apps. **(assigned to: )**
 
-** Release Completion (assigned to: )**:
+**DTFx Release Completion (assigned to: )**
 _Due: <release-deadline>_
-- [ ] Push staged package on MyGet to Nuget.org (for Durable Task Framework, you may need to manually update them)
-- [ ] Create a PR in the [Azure Functions templates repo](https://github.com/Azure/azure-functions-templates) to update templates to the latest version
-- [ ] Create a PR in the [Azure Functions bundles repo](https://github.com/Azure/azure-functions-extension-bundles) to update bundles to the latest version
-- [ ] Merge all pending PR docs from `pending_docs.md`
+- [ ] Publish DTFx packages to nuget (directly to nuget.org).
+- [ ] Publish release notes for DTFx.
+
+**Release Completion (assigned to: )**
+_Due: <release-deadline>_
+- [ ] Delete Durable Functions packages from myget.
+- [ ] Run Durable Functions release pipeline.
+- [ ] Push myget package to nuget (nuget.org extensions package option).
+- [ ] Create a PR in the [Azure Functions templates repo](https://github.com/Azure/azure-functions-templates) to update templates to the latest version.
+- [ ] Create a PR in the [Azure Functions bundles repo](https://github.com/Azure/azure-functions-extension-bundles) to update bundles to the latest version.
+- [ ] Merge all pending PR docs from `pending_docs.md.`
 - [ ] Reset `pending_docs.md` and `release_notes.md` in the `dev` branch. You will want to save `release_notes.md` somewhere for when you publish release notes.
-- [ ] Merge `dev` into `main`
-- [ ] Publish release notes from the pre-reset `release_notes.md`
-- [ ] Post announcement on [App Service Announcements GitHub repo](https://github.com/Azure/app-service-announcements) and Twitter.
+- [ ] Merge `dev` into `main.`
+- [ ] Publish release notes from the pre-reset `release_notes.md.`
+- [ ] Post announcement on [App Service Announcements GitHub repo](https://github.com/Azure/app-service-announcements) and Twitter (Chris).
--- a/.github/ISSUE_TEMPLATE/new-release-template.md
+++ b/.github/ISSUE_TEMPLATE/new-release-template.md
@@ -9,15 +9,15 @@ assignees: comcmaho, amdeel, davidmrdavid, bachuv
 
 **Prep DTFx Release: (assigned to:)**
 _Due: <2-3-business-days-before-release>_
-- [ ] Increment updated DTFx packages (either DT-AzureStorage only or DT-AzureStorage, DT-Core, DT-Emulator, and DT-Redis)
+- [ ] Check DTFx package versions (either DT-AzureStorage only or if there were Core changes DT-AzureStorage, DT-Core, DT-Emulator, and DT-Redis)
 - [ ] Delete DTFx test packages from staging myget.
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/new-release-template.md
```diff
@@ -7,30 +7,41 @@ assignees: comcmaho, amdeel, davidmrdavid, bachuv
...
+- [ ] Update Durable Functions references (Analyzer? DTFx?) and increment version.
+- [ ] Add Durable Functions package to myget staging feed.
+- [ ] Check for package size, make sure it's not surprisingly heavier than a previous release.
+- [ ] Merge dev into main (Merge NOT Squash merge).
+
+**Validation**
+_Due: <1-business-days-before-release>_
+- [ ] Run private performance tests and ensure no regressions. **(assigned to: )**
+- [ ] Smoke test Functions V1, Functions V2, and Functions V3 .NET apps. **(assigned to: )**
+- [ ] Smoke test JavaScript and Python apps. **(assigned to: )**
```
amdeel, line 29: These tasks could just be assigned to the same person. However they can be done in parallel so I gave the option to individually assign, in case people want help.

.github/ISSUE_TEMPLATE/new-release-template.md
```diff
@@ -7,30 +7,41 @@ assignees: comcmaho, amdeel, davidmrdavid, bachuv
...
 
+**DTFx Release Completion (assigned to: )**
+_Due: <release-deadline>_
+- [ ] Publish DTFx packages to nuget (directly to nuget.org).
+- [ ] Publish release notes for DTFx.
 
-** Release Completion (assigned to: )**:
+**Release Completion (assigned to: )**
```
amdeel, line 36: I grouped tasks together, but as you can see, there are more than 3/4 roles. Instead of assigning one person each role, we can assign ourselves as much as we're comfortable taking. Some might be on call or have as especially busy after-servicing week (or be OOF), while others might be more free.

Also, we will be not always have a DTFx release accompanying our Durable Functions extension release.

.github/ISSUE_TEMPLATE/new-release-template.md
amdeel: Also, I wanted to note that these steps will become obsolete with the test bot that Varshi has been working on.

.github/ISSUE_TEMPLATE/new-release-template.md
```diff
@@ -7,30 +7,41 @@ assignees: comcmaho, amdeel, davidmrdavid, bachuv
...
-**Prep Release: (assigned to:)**
-_Due: <3-business-days-before-release>_
-- [ ] Merge all features and fixes into the `dev` branch
-- [ ] Publish updated versions of DurableTask.Core and DurableTask.AzureStorage to app-service MyGet feed
-- [ ] Update dependencies and increment extension version in the `dev` branch.
-- [ ] Publish signed version of Microsoft.Azure.WebJobs.Extensions.DurableTask to app-service MyGet feed
-- [ ] Close out or punt remaining GitHub issues for the current milestone
+**Prep DTFx Release: (assigned to:)**
+_Due: <2-3-business-days-before-release>_
+- [ ] Increment updated DTFx packages (either DT-AzureStorage only or DT-AzureStorage, DT-Core, and DT-Redis)
```
davidmrdavid, line 12: So it's either just 
(A) AzureStorage or
(B) AzureStorage + Core + Redis all at once?

It might be good to clarify in the notes themselves, as well as how to determine which approach to take :)

.github/ISSUE_TEMPLATE/new-release-template.md
```diff
@@ -7,30 +7,41 @@ assignees: comcmaho, amdeel, davidmrdavid, bachuv
...
+- [ ] Delete DTFx test packages from myget.
```
davidmrdavid, line 13: Sometimes a DTFx package might be on myget for a customer, shouldn't we have an exception for that case? Maybe we can delete the myget packages post-nuget release

.github/ISSUE_TEMPLATE/new-release-template.md
```diff
@@ -7,30 +7,41 @@ assignees: comcmaho, amdeel, davidmrdavid, bachuv
...
+- [ ] Update OneBranch durabletask repo, run signing pipeline (using these notes:  [Durable Task Framework Build Pipeline] (https://microsoft.sharepoint.com/teams/AzureWebjobs/_layouts/OneNote.aspx?id=%2Fteams%2FAzureWebjobs%2FSiteAssets%2FAzureWebjobs%20Notebook&wd=target%28Planning%2FFunctions%20Post-GA.one%7CA43CF112-7272-481A-B23E-9AA5CA8EEE06%2FDurable%20Task%20Framework%20Build%20Pipeline%7CD0946823-6FB0-44E3-A57F-E252617B69CD%2F%29
```
davidmrdavid, line 14: are we good to share one-notes on the public repo? Just checking!

.github/ISSUE_TEMPLATE/new-release-template.md
```diff
@@ -7,30 +7,41 @@ assignees: comcmaho, amdeel, davidmrdavid, bachuv
...
-**Validation (assigned to: )**
+**Prep Release (assigned to: )**
 _Due: <2-business-days-before-release>_
-- [ ] Run private performance tests and ensure no regressions
-- [ ] Smoke test Functions V1 and Functions V3 .NET apps
-- [ ] Smoke test JavaScript and Python apps
-- [ ] Check for package size, make sure it's not surprisingly heavier than a previous release
+- [ ] Update Durable Functions references (Analyzer? DTFx?) and increment version.
+- [ ] Add Durable Functions package to myget staging feed.
+- [ ] Check for package size, make sure it's not surprisingly heavier than a previous release.
```
davidmrdavid, line 22: We should probably provide an expected package size to the notes

davidmrdavid (commented):
Left some thoughts, mostly questions

.github/ISSUE_TEMPLATE/new-release-template.md
ConnorMcMahon: It should be AzureStorage, or AzureStorage, Core, Emulator, and Redis if Core was touched.

.github/ISSUE_TEMPLATE/new-release-template.md
ConnorMcMahon: We should be using staging for release prep, and the main appservice feed for test packages for customers.

.github/ISSUE_TEMPLATE/new-release-template.md
ConnorMcMahon: I think permissions should block people who shouldn't have access from access, but it's still a valid point.

.github/ISSUE_TEMPLATE/new-release-template.md
ConnorMcMahon: Let's be explicit about which myget feed we delete/upload to.

.github/ISSUE_TEMPLATE/new-release-template.md
```diff
@@ -7,30 +7,41 @@ assignees: comcmaho, amdeel, davidmrdavid, bachuv
...
-
-**Validation (assigned to: )**
+**Prep Release (assigned to: )**
 _Due: <2-business-days-before-release>_
-- [ ] Run private performance tests and ensure no regressions
-- [ ] Smoke test Functions V1 and Functions V3 .NET apps
-- [ ] Smoke test JavaScript and Python apps
-- [ ] Check for package size, make sure it's not surprisingly heavier than a previous release
+- [ ] Update Durable Functions references (Analyzer? DTFx?) and increment version.
+- [ ] Add Durable Functions package to myget staging feed.
```
ConnorMcMahon, line 21: Shouldn't this step happen after we merge dev into main? It is done via just launching our release pipeline, which will deploy to myget.

.github/ISSUE_TEMPLATE/new-release-template.md
davidmrdavid: Thanks! Good to know. I would still recommend to have it in the notes though, but no strong feelings about this

.github/ISSUE_TEMPLATE/new-release-template.md
ConnorMcMahon: I personally still have a preference for 3-4 roles that we rotate between. It doesn't matter if the roles are perfectly balanced in terms of work, as a consistent rotation should even things out over time.

While it is true we may have Durable extension releases without a DTFx release, that is fairly rare.

.github/ISSUE_TEMPLATE/new-release-template.md
```diff
@@ -7,30 +7,41 @@ assignees: comcmaho, amdeel, davidmrdavid, bachuv
...
 
-** Release Completion (assigned to: )**:
+**Release Completion (assigned to: )**
 _Due: <release-deadline>_
-- [ ] Push staged package on MyGet to Nuget.org (for Durable Task Framework, you may need to manually update them)
-- [ ] Create a PR in the [Azure Functions templates repo](https://github.com/Azure/azure-functions-templates) to update templates to the latest version
-- [ ] Create a PR in the [Azure Functions bundles repo](https://github.com/Azure/azure-functions-extension-bundles) to update bundles to the latest version
-- [ ] Merge all pending PR docs from `pending_docs.md`
+- [ ] Delete Durable Functions packages from myget.
+- [ ] Run Durable Functions release pipeline.
```
ConnorMcMahon, line 39: If the bits are already on myget from the release pipeline earlier, then we can just push them from myget to nuget.org.

.github/ISSUE_TEMPLATE/new-release-template.md
cgillum: I'm not sure if this has changed, but the MyGet *staging* feed had some automatic cleanup processes before based on Functions release schedules. Given we don't release at the same cadence as the Functions team, we may want to be careful about taking a dependency on their staging feed.

.github/ISSUE_TEMPLATE/new-release-template.md
cgillum: As an aside, it might be worth considering redoing the DTFx release pipeline to something simpler, like what we do for the Durable extension w/Azure DevOps. The whole thing with using the msazure OneBranch and sub-modules (thus requiring all these steps in the OneNote) seems unnecessarily complicated. Just my $.02.

.github/ISSUE_TEMPLATE/new-release-template.md
```diff
@@ -7,30 +7,41 @@ assignees: comcmaho, amdeel, davidmrdavid, bachuv
...
 
-
-**Validation (assigned to: )**
+**Prep Release (assigned to: )**
 _Due: <2-business-days-before-release>_
-- [ ] Run private performance tests and ensure no regressions
-- [ ] Smoke test Functions V1 and Functions V3 .NET apps
-- [ ] Smoke test JavaScript and Python apps
-- [ ] Check for package size, make sure it's not surprisingly heavier than a previous release
+- [ ] Update Durable Functions references (Analyzer? DTFx?) and increment version.
```
cgillum, line 20: Another thing we could consider is incrementing the version at the *start* of every milestone. This would make it easier to do nightly releases that have sensible version numbers and is more consistent with build systems I've seen used around the company.

.github/ISSUE_TEMPLATE/new-release-template.md
```diff
@@ -7,30 +7,41 @@ assignees: comcmaho, amdeel, davidmrdavid, bachuv
...
+- [ ] Create a PR in the [Azure Functions templates repo](https://github.com/Azure/azure-functions-templates) to update templates to the latest version.
+- [ ] Create a PR in the [Azure Functions bundles repo](https://github.com/Azure/azure-functions-extension-bundles) to update bundles to the latest version.
+- [ ] Merge all pending PR docs from `pending_docs.md.`
 - [ ] Reset `pending_docs.md` and `release_notes.md` in the `dev` branch. You will want to save `release_notes.md` somewhere for when you publish release notes.
-- [ ] Merge `dev` into `main`
-- [ ] Publish release notes from the pre-reset `release_notes.md`
-- [ ] Post announcement on [App Service Announcements GitHub repo](https://github.com/Azure/app-service-announcements) and Twitter.
+- [ ] Merge `dev` into `main.`
+- [ ] Publish release notes from the pre-reset `release_notes.md.`
+- [ ] Post announcement on [App Service Announcements GitHub repo](https://github.com/Azure/app-service-announcements) and Twitter (Chris).
```
cgillum, line 47: We haven't really been consistent about using the Announcements GitHub repo. Does anyone have opinions on whether we're adding sufficient value by adding our announcements here? Are there any other announcement avenues we should consider, like internal emails that include CSS, etc.?

.github/ISSUE_TEMPLATE/new-release-template.md
ConnorMcMahon: We could accomplish that by moving this step until after we merge dev to main.

.github/ISSUE_TEMPLATE/new-release-template.md
amdeel: No, I don't mean run signing pipeline, I mean just manually upload it to the staging myget. This step is optional, but necessary if we are each taking a smoke test/performance test role. It is just for a test package for local testing, not meant to be a signed package.

.github/ISSUE_TEMPLATE/new-release-template.md
amdeel: I understand having 3-4 roles makes things easy to assign, but we could accomplish the same thing by assigning the first two to the same person and the second two to the same person if you'd rather we do it that way. I just kept the breakdown in categories to make it easy to see which parts are left out when not doing a DTFx release and just easy to read in general.

.github/ISSUE_TEMPLATE/new-release-template.md
amdeel: This is just the increment step. Technically you increment Emulator at the same time as Core, so I just left it out to avoid confusion. I can update it if you have a preference and add "if changes were made to Core"

.github/ISSUE_TEMPLATE/new-release-template.md
amdeel: @cgillum Good point. Can we create our own myget feed? Or should we just talk to the functions team about leaving our packages out of their automatic cleanup (if possible)?

.github/ISSUE_TEMPLATE/new-release-template.md
amdeel: This is for another thread, but I want to discuss nightly releases further, I don't really understand them, how that would work with DTFx and the extension, and if they would be beneficial.

ConnorMcMahon: @amdeel, the goal of nightly releases would be that as soon as we merge a fix, there would be a build available for customers to get access to that fix in an unofficial capacity without us uploading the build to myget manually. It makes it much easier for customers to validate fixes before our official releases.

We could eventually look at having test bot run our smoke/load tests against the nightly release bits so we identify regressions earlier in the release process as well.

.github/ISSUE_TEMPLATE/new-release-template.md
ConnorMcMahon: If we are going through the process of uploading bits to myget and running smoke tests/perf tests I think we might as well use the signed release bits, and the pipeline will load it onto MyGet for us with a single button anyways, so it may even be less work. It means the exact same bits that we validated actually get loaded up to nuget.org.

.github/ISSUE_TEMPLATE/new-release-template.md
davidmrdavid: My only concern is that eagerly incrementing the version number might complicate making major-version releases. For example, what happens if the PR that justified the major-version upgrade isn't complete by the milestone's end date? Do we delay the entire release? Or do we revert the major-version upgrade and merge a minor-version upgrade instead to unblock the release?

.github/ISSUE_TEMPLATE/new-release-template.md
davidmrdavid: @amdeel I would prefer the extra detail, yes :) Just to avoid confusion

.github/ISSUE_TEMPLATE/new-release-template.md
amdeel: I don't think we would commit to a major version release if we weren't sure about it. The issue would more likely be that we increment to 2.4.2 when we never intend to release 2.4.2. Taking nightly releases out of the conversation, we can increment the version again to 2.5.0 when we commit on having that as our next release.

ConnorMcMahon: That sounds reasonable to me. Most of the time we have a _fairly_ good idea if we are doing a patch or minor release next, and there is minimal overhead to change it if we change our mind a couple weeks later.

.github/ISSUE_TEMPLATE/new-release-template.md
ConnorMcMahon: I'll talk with Fabio's team regarding the deletion. I believe it is manual cleanup, but I can verify.

.github/ISSUE_TEMPLATE/new-release-template.md
amdeel: People can just check the previous package size.

.github/ISSUE_TEMPLATE/new-release-template.md
amdeel: Then we would have to wait to sign off on the merge dev into main PR, run the pipeline, and solve potential roadblocks in the pipeline just to start running our first smoke test. Which might show us a new error. Then, we would be working out of dev again and merging into main etc.

This testing phase will be obsolete when the test bot is done, so we can take this out entirely when that happens and use the automated process instead.

You're saying that we are testing the same code that we are releasing, which makes sense. But we're also using the same GitHub commit so we can be sure it's the same code already.

I don't think it makes sense to require the package to be signed before running tests on it.

.github/ISSUE_TEMPLATE/new-release-template.md
ConnorMcMahon: I'm 100% on board for simplifying the DTFx release pipeline. @cgillum, I may look into what you did for your sql provider repo to see if we can make something simpler for the broader DTFx repo.

.github/ISSUE_TEMPLATE/new-release-template.md
ConnorMcMahon: Our current package size is under ~470KB. I'm not necessarily opposed to setting some arbitrary limit (say 500 or 600KB) for now so that we make sure that we have some consistent definition of "suprisingly heavier".

.github/ISSUE_TEMPLATE/new-release-template.md
ConnorMcMahon: I'd personally prefer to smoke testing against signed bits and deal with having to remerge into `main` if the smoke tests fail than have something subtle pop up in the difference between unsighed vs signed bits.

For instance, in the last release, we would have caught that we needed to publish a new version of DTFx.Redis + DTFx.Emulator earlier if we were using signed bits, as I believe that compiler issue only pops up with signed packages/bits.

In general, smoke tests failures should be exceedingly rare, and so I'd rather just to all of our validation on the exact package in myget so we can just do the "push package to nuget" step once validation is complete.

cgillum: > For instance, in the last release, we would have caught that we needed to publish a new version of DTFx.Redis + DTFx.Emulator earlier if we were using signed bits, as I believe that compiler issue only pops up with signed packages/bits.

What was the issue you encountered? There shouldn't be any difference at all between signed bits and unsigned bits, assuming we're talking about ESRP authenticode signing. My only concern about testing with signed bits is that it further increases the time to complete release validation. I don't think we've ever done testing with signed bits.

ConnorMcMahon: I think the issue here was more related to release vs debug builds, vs signed vs unsighed builds now that I think about it. I just know it is not super straightforward to get local Release builds from the DTFx repo, but as long as we test against Release, I'm willing to budge on the signed bits.

ConnorMcMahon (approved):
I would say that the doc is in a fairly good state at this point, and we can try our next release with this template, and continue to iterate as we find gaps.

PR #1685, opened 2021-02-13
Improve CI reliability and performance

### LINKED ISSUE #1602: Invest in CI reliability
Our CI has been failing somewhat frequently recently.

We should investigate why, and continue to invest until our CI until we hit the below reliability criteria:

- Less than 1 in 20 CI runs that would have succeeded fail due to test flakiness.
- Our CI pipeline can tolerate multiple concurrent queued runs without increasing the likelihood of false negatives.

ConnorMcMahon:
Set of test improvements

- The most common CI failure is related to when we try to listen to event source. I have attempted to fix this by both removing the lookup by event source name (instead defaulting to GUID), as well as updating the underlying dependency.
- Reduced test length for some of our longer running tests.
- Flaky tests are bad for both CI reliability and CI latency, as rerunning tests seems to add 1+ minutes per test to the CI. This PR tried to fix as many flaky tests as possible.
- Reduced dotnet restore/build time by only building relevant projects for tests
- Removed unnecessary publish step

After this PR, the CI runs in 8-10 minutes assuming no flakey tests fail.

resolves #1602 

### Pull request checklist


src/WebJobs.Extensions.DurableTask/DurableHttpMessageHandlerFactory.cs
```diff
@@ -2,6 +2,8 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System.Net.Http;
+using System.Threading;
```
ConnorMcMahon, line 5: This can be deleted.

azure-pipelines.yml
```diff
@@ -118,37 +122,4 @@ jobs:
         rerunMaxAttempts: 2
       env:
         AzureWebJobsStorage: $(AzureWebJobsStorage)
-        APPINSIGHTS_INSTRUMENTATIONKEY: $(APPINSIGHTS_INSTRUMENTATIONKEY) 
-    
-  - job: PublishPipelineArtifact
```
ConnorMcMahon, line 123: @amdeel, I think this is safe to remove, as we only need publishing on our releases, not our CI, but I'd like your input.

azure-pipelines.yml
```diff
@@ -61,22 +62,24 @@ jobs:
         feedsToUse: 'config'
         nugetConfigPath: '.nuget/nuget.config'
 
-    - task: VSBuild@1
+    - task: DotNetCoreCLI@2
```
davidmrdavid, line 65: Why is this task name change necessary?

azure-pipelines.yml
```diff
@@ -49,7 +50,7 @@ jobs:
     variables:
       solution: 'WebJobs.Extensions.DurableTask.sln'
       buildPlatform: 'Any CPU'
-      buildConfiguration: 'Debug'
+      buildConfiguration: 'Release'
```
davidmrdavid, line 53: Why is this change necessary?

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -63,9 +63,11 @@ private void StartLogCapture()
         {
             if (this.useTestLogger)
             {
+                // Use GUID for eventsource, as TraceEventProviders.GetProviderGuidByName() is causing
+                // the CI to abort runs.
```
davidmrdavid, line 67: Interesting, do you know why looking up EventSource events by name would lead to an aborted run? Or is this just from our experience so far?

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -4621,6 +4624,7 @@ public async Task DurableEntity_CleanEntityStorage_Many(string storageProvider)
                 var client = await host.StartOrchestratorAsync(nameof(TestOrchestrations.CreateEmptyEntities), entityIds, this.output);
                 var status = await client.WaitForCompletionAsync(this.output);
 
+                // TODO: try removing
```
davidmrdavid, line 4627: could you please clarify/ expand on this comment?

test/FunctionsV1/PlatformSpecificHelpers.FunctionsV1.cs
```diff
@@ -40,6 +40,12 @@ public static class PlatformSpecificHelpers
             var loggerFactory = new LoggerFactory();
             loggerFactory.AddProvider(loggerProvider);
 
+            // Unless otherwise specified, use legacy partition management for tests as it makes them run faster.
```
davidmrdavid, line 43: Should we be concerned about using a legacy strategy as the default for our tests? If not, I would prefer if the comments should explain this

test/FunctionsV2/TestCleanup.cs
```diff
@@ -34,7 +34,9 @@ public TestCleanup(ITestOutputHelper output)
         [Trait("Category", PlatformSpecificHelpers.TestCategory)]
         public async Task CleanupOldAzureStorageTaskHubs()
         {
-            TimeSpan oldTaskHubDeletionThreshold = TimeSpan.FromMinutes(5);
+            // The CI should run at least once a day, so this timespan should avoid deleting
+            // deterministic taskhubs
+            TimeSpan oldTaskHubDeletionThreshold = TimeSpan.FromHours(25);
```
davidmrdavid, line 39: I'm a bit confused by these "deterministic taskhubs" we're trying to avoid deleting. Are those taskhubs of currently running tests or something else?

davidmrdavid (commented):
Thank you for this work, especially good call with the new `WaitUntilTrue` helper.
Left some comments, mostly questions

test/Common/DurableTaskEndToEndTests.cs
ConnorMcMahon: Not sure, but no one else has experience the stack trace. My guess is something to do with concurrency. I tried revving the dependency that code lives in to see if they have fixed the bug, but it happened once, so I am trying to lookup by GUID instead to see if that fixes the issue.

test/Common/DurableTaskEndToEndTests.cs
ConnorMcMahon: Woops, I meant to try removing that as a part of this PR. I'm just trying to remove delays unless they are really necessary.

test/FunctionsV1/PlatformSpecificHelpers.FunctionsV1.cs
ConnorMcMahon: I'm not super concerned. We test the new strategy logic in the durable task repo, and the main downside to the new strategy is that it can take moderately longer to start up a taskhubworker, and that perf hit adds up over 100s of tests. None of these tests have multiple taskhubworkers (or very few do), so changing the partition management strategy should be irrelevant.

test/FunctionsV2/TestCleanup.cs
ConnorMcMahon: Yeah, I ran into an issue where tests would fail because we try to start a taskhub while this test was trying to delete that taskhub.

azure-pipelines.yml
ConnorMcMahon: This version allows you to run arbitrary `dotnet` cli commands. This allows me to build individual projects instead of the whole solution, which saves about 2 minutes per run.

azure-pipelines.yml
ConnorMcMahon: Release bits run faster, plus we fail release builds on warnings, so this will prevent warnings from slipping through PRs.

azure-pipelines.yml
```diff
@@ -61,22 +62,24 @@ jobs:
...
+        projects: 'test/FunctionsV2/*.csproj'
+        arguments: '--configuration $(buildConfiguration)'
+      displayName: 'dotnet build $(buildConfiguration)'
 
     - task: VSTest@2
       inputs:
         testSelector: 'testAssemblies'
         testAssemblyVer2: |
-          **/*tests.v2.dll
+          **\\bin\\**\\*tests.v2.dll
```
cgillum, line 76: nit: here and in line 117, I see you switched from forward slashes to backslashes. I think it's better if we standardize on forward slashes for consistency.

azure-pipelines.yml
amdeel: This is just to publish the nuget package of this build in the artifacts dropdown, in case someone wants a quick nuget package for testing. It's an insignificant performance cost to have this job. Admittedly though, I've never used this. And if I'm testing my own changes I'll just make my own nuget package. I assume if we are testing nuget packages as a team we will likely upload our nuget package to myget. So I'm not opposed to removing this.

ConnorMcMahon: Since it has the "dependsOn" clause, it runs only after all of the tests have passed, meaning even though the task itself only takes a few minutes, it adds time until the CI is marked as green.

test/FunctionsV1/PlatformSpecificHelpers.FunctionsV1.cs
davidmrdavid: Sounds good. I would suggest adding this rationale in the comments themselves, as I could imagine this question coming up again :)

davidmrdavid (approved):
Looks good to me. Looking forward to a speedy CI 🐰

Issue #24, opened 2017-07-25
Should throw exception when a called function does not exist.


cgillum:
Today if a user calls `DurableOrchestrationClient.StartNewAsync` or `DurableOrchestrationContext.CallFunctionAsync` and specifies a function name which doesn't exist, the call fails silently.

Instead, these methods should throw an exception that can be handled/logged by user code.

cgillum:
This is fixed in the latest release. If you try to call a function that doesn't exist, an exception will be thrown and the message will include all the active function names, making it easier to figure out what went wrong.

PR #1692, opened 2021-02-19
Add a mechanism for providers to override the scale monitor

sebastianburckhardt:
This allows backends to plug in their own autoscaler for use in elastic premium deployments.
 
- Checks if durability provider implements `IProviderWithAutoScaling`; if so, calls it to construct a ScaleMonitor
- If provider does not provide a scale monitor, constructs a pseudo-scale monitor to avoid exceptions when runtime scale is disabled
- Allows the provider to trace scaling decisions in the same way as currently done by Azure Table Scaling monitor

### Pull request checklist


src/WebJobs.Extensions.DurableTask/ProviderUtils.cs
```diff
@@ -16,6 +19,22 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
     /// </summary>
     public static class ProviderUtils
     {
+#if !FUNCTIONS_V1
+        /// <summary>
+        /// Providers can implement this interface to override the autoscaling.
+        /// </summary>
+        public interface IProviderWithAutoScaling
```
ConnorMcMahon, line 26: `DurabilityProvider` is itself a class with virtual methods, so I think we can substantially simplify the design by just making `TryGetScaleMonitor` a virtual method that by default returns `PsuedoScaleMonitor`, and `DurabilityProvider` implementations that want to support runtime driven scale just override that method. That way we don't need to worry about this optional interface on the implementations or when retrieving scale monitors.

src/WebJobs.Extensions.DurableTask/ProviderUtils.cs
```diff
@@ -16,6 +19,22 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
...
+        {
+            /// <summary>
+            /// Tries to obtain a scale monitor for autoscaling.
+            /// </summary>
+            /// <param name="extension">The extension, which is used for tracing.</param>
+            /// <param name="scaleMonitor">The scale monitor.</param>
+            /// <returns>True if autoscaling is supported, false otherwise.</returns>
+            bool TryGetScaleMonitor(DurableTaskExtension extension, out IScaleMonitor scaleMonitor);
```
ConnorMcMahon, line 34: I notice that the Azure Storage implementation of IScaleMonitor takes in a FunctionId and a FunctionName to generate a scale monitor descriptor. I'm not sure how necessary it is to have a unique ScaleMonitorDescriptor per trigger, especially if all Durable Triggers for an app should use similar scale logic, but if we do need that, we should put those as parameters on this method.

ConnorMcMahon (changes requested):
I proposedan alternative design that I think is a bit simpler, but I'm open to the current design if you have strong reasons for going with it.

I may consult with Graham to understand if we really need unique descriptors for each trigger or not.

src/WebJobs.Extensions.DurableTask/ProviderUtils.cs
sebastianburckhardt: I agree that it makes more sense. The problem was that in order to treat AzureStorage the same as other providers, I have to pass a whole lot of parameters into the call that seem superfluous (e.g. taskhub name, connections string) because the provider should already know these (but AzureStorageDurabilityProvider does not). I wanted to change existing code as little as possible.

sebastianburckhardt: If you have some spare cycles, feel free to clean this up.

bachuv:
I worked on cleaning this PR up by making `TryGetScaleMonitor` a virtual method in `DurabilityProvider` and it was mostly looking good except for the `TraceHelper` argument that's needed for creating the `DurableTaskScaleMonitor` if the provider is Azure Storage. I changed the `TryGetScaleMonitor` method signature to be this: 

```
public virtual bool TryGetScaleMonitor(
    string functionId,
    string functionName,
    string hubName,
    string storageConnectionString,
    EndToEndTraceHelper traceHelper,
    DurableTaskExtension extension,
    out IScaleMonitor scaleMonitor)
```

and the problem with this method signature is that `EndToEndTraceHelper` has an internal access modifier so it's more restrictive than the public `TryGetScaleMonitor` method. 

One option to get around this could be to leave the "if the provider is Azure Storage then create a `DurableTaskScaleMonitor`" case in `DurableTaskExtension` so there's access to the `traceHelper`, but still make `TryGetScaleMonitor` a virtual method in `DurabilityProvider`. This would at least get rid of the `IProviderWithAutoScaling` interface, but it wouldn't be as straightforward to understand in the future. @ConnorMcMahon Thoughts?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -1267,6 +1268,66 @@ internal static void TagActivityWithOrchestrationStatus(OrchestrationRuntimeStat
...
+                return new PseudoScaleMonitor($"{functionId}-DurableTaskTrigger-{this.Options.HubName}".ToLower());
+            }
+        }
+
+        /// <summary>
+        /// A placeholder scale monitor, can be used by durability providers that do not support runtime scaling.
+        /// This is required to allow operation of those providers even if runtime scaling is turned off
+        /// see discussion https://github.com/Azure/azure-functions-durable-extension/pull/1009/files#r341767018.
+        /// </summary>
+        private sealed class PseudoScaleMonitor : IScaleMonitor
```
cgillum, line 1303: nit: I'd avoid calling it a "pseudo" scale monitor - my first impression is that it's something for testing. Maybe `NullScaleMonitor` or `NoOpScaleMonitor`?

cgillum (commented):
I don't feel great about having custom scale monitors writing to the extension logs using the generic "informational" and "warning" traces. I suggest either have them use their own internal logging mechanism or give them a pointer to our `ILoggerFactory` and have them write whatever logs they want to write using that.

cgillum:
Taking a step back, have we or should we consider an alternate design where the `IScaleProvider` is injected via Dependency Injection into the Functions host instead of using an interface/virtual method in the durability provider? It seems like this would be valuable outside of just Durable Functions. I worry if we go down this path then we could run into conflicts if the Functions team decides to add a similar abstraction down the road.

ConnorMcMahon:
@bachuv, I think I agree with @cgillum's sentiment that we should just give them a hook into ILoggerFactory to create their own logging infrastructure. This will ensure logging decisions for scaling are logged into the FunctionsLogs table I believe, which feels right.

We would want to test this in the public cloud though to verify.

sebastianburckhardt:
The reason why I included those tracing hooks are so I could generate the same traces as are emitted by the current code, e.g. [this is what the existing code calls](https://github.com/Azure/azure-functions-durable-extension/blob/2ac3761767c437ac792c4f72fb3cfe9f0328cd5d/src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs#L176)
```csharp
 this.traceHelper.ExtensionInformationalEvent(
                                    this.hubName,
                                    string.Empty,
                                    this.functionName.Name,
                                    $"Durable Functions Trigger Scale Decision: {scaleStatus.Vote.ToString()}, Reason: {scaleRecommendation?.Reason}",
                                    writeToUserLogs: writeToUserLogs);
```

Of course I can generate my own tracing events (which is easy) but it means our tooling will not work across backends.

Perhaps a better way to do that would be to create specific trace events for scaling decisions, and then make those events accessible to providers somehow.

bachuv:
I found this code in the azure-functions-host repo that logs a scale vote. Is this the scale logging we are looking for to replace the trace helper logs?

https://github.com/Azure/azure-functions-host/blob/14adf6d336db5da0feb0221cc4fe0a40bc14edb0/src/WebJobs.Script/Scale/FunctionsScaleManager.cs#L71

sebastianburckhardt:
We probably need to preserve logging on at least 2 levels. There are 3 levels where stuff is happening currently.
1. The FunctionsScaleManager logs decisions based on all triggers. I see no need to change anything there.
2. DurableFunctionsExtension currently emits scale vote information on behalf of the AzureStorageBackend-specific scaler.
3. Netherite/SQL emit scale vote information through their own specific events. No need to change this either.

I see two directions we can go:
a. "every backend for itself":  remove the tracing 2. and have AzureStorageBackend instead emit scaling information on its own.
b. "common events for all backends": make the events in 2. be general events that are emittable by all backends. The advantage is that we can build uniform monitoring for all backends.

I don't think it's necessarily worth spending a lot of time on this upfront, so I would probably recommend a., or leave it the way it is right now. I don't really see a big problem with that.

ConnorMcMahon:
+1 to Sebastian's option A, where we let the FunctionsScaleManager log scale decisions for all triggers, and then have each backend emit logs as they desire to ILogger and/or EventSource as they please.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
```diff
@@ -173,12 +174,13 @@ private ScaleStatus GetScaleStatusCore(int workerCount, DurableTaskTriggerMetric
...
-            this.traceHelper.ExtensionInformationalEvent(
-                                    this.hubName,
-                                    string.Empty,
-                                    this.functionName.Name,
-                                    $"Durable Functions Trigger Scale Decision: {scaleStatus.Vote.ToString()}, Reason: {scaleRecommendation?.Reason}",
-                                    writeToUserLogs: writeToUserLogs);
+            if (writeToUserLogs)
+            {
+                this.logger.LogInformation(
+                $"Durable Functions Trigger Scale Decision: {scaleStatus.Vote.ToString()}, Reason: {scaleRecommendation?.Reason}",
```
ConnorMcMahon, line 180: nit: Fix indentation.

src/WebJobs.Extensions.DurableTask/ProviderUtils.cs
```diff
@@ -61,5 +61,32 @@ public static DurableOrchestrationStatus ConvertOrchestrationStateToStatus(Orche
...
+        /// <summary>
+        /// Trace an informational event for the extension.
+        /// </summary>
+        /// <param name="extension">The durable task extension.</param>
+        /// <param name="hubName">The name of the task hub.</param>
+        /// <param name="instanceId">The instance id.</param>
+        /// <param name="functionName">The function name.</param>
+        /// <param name="message">The message.</param>
+        /// <param name="writeToUserLogs">Whether to write this event to the user logs.</param>
+        public static void TraceInformationalEvent(this DurableTaskExtension extension, string hubName, string instanceId, string functionName, string message, bool writeToUserLogs)
```
ConnorMcMahon, line 74: These don't seem to be used anymore?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -1267,6 +1269,61 @@ internal static void TagActivityWithOrchestrationStatus(OrchestrationRuntimeStat
                 activity.AddTag("DurableFunctionsRuntimeStatus", statusStr);
             }
         }
+
+        internal IScaleMonitor GetScaleMonitor(string functionId, FunctionName functionName, string storageConnectionString)
+        {
+            if (this.defaultDurabilityProvider.TryGetScaleMonitor(
+                functionId,
```
ConnorMcMahon, line 1276: nit: indent one more tab.

ConnorMcMahon (approved):
LGTM, just a couple of nits.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -1267,6 +1269,61 @@ internal static void TagActivityWithOrchestrationStatus(OrchestrationRuntimeStat
...
         }
+
+        internal IScaleMonitor GetScaleMonitor(string functionId, FunctionName functionName, string storageConnectionString)
+        {
+            if (this.defaultDurabilityProvider.TryGetScaleMonitor(
+                functionId,
+                functionName.Name,
+                this.Options.HubName,
+                storageConnectionString,
+                this.logger,
```
cgillum, line 1280: I feel like it's not a good idea to pass the `ILogger` used by the extension to the 3rd party durability provider. That basically gives it the ability to emit logs that look like they are coming from the extension when really they are not. Instead, I would prefer we either pass in the `ILoggerFactory` so that the durability provider can generate it's own `ILogger` with its own category, OR create a new `ILogger` with a category name we define, based on the name of the durability provider.

Thoughts?

ConnorMcMahon: That's fair. In fact, we may not even need this as a parameter, because the IDurabilityProviderFactory implementation can just access ILoggerFactory by adding it as a parameter on its constructor.

bachuv: That makes sense. We should keep it separate. Taking Azure Storage as an example, `AzureStorageDurabilityProviderFactory` has an `ILoggerFactory` constructor parameter. Is it fine to create an `ILogger` in `GetDurabilityProvider()` and then pass that `ILogger` when creating an `AzureStorageDurabilityProvider`?

Example in `GetDurabilityProvider()` in `AzureStorageDurabilityProviderFactory`:
```
ILogger logger = this.loggerFactory.CreateLogger("AzureStorage");
this.defaultStorageProvider = new AzureStorageDurabilityProvider(
    defaultService,
    this.defaultConnectionName,
    this.azureStorageOptions,
    logger);
```

Then, `TryGetScaleMonitor()` would use the `ILogger` that's passed into `AzureStorageDurabilityProvider` and the `ILogger` parameter in `TryGetScaleMonitor()` can be removed.

ConnorMcMahon: Yeah, that makes the most sense I think.

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
```diff
@@ -11,6 +11,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 {
     internal class AzureStorageDurabilityProviderFactory : IDurabilityProviderFactory
     {
+        private const string LoggerName = "AzureStorage";
```
cgillum, line 14: Should we use a logger name that is consistent with the existing Azure Functions logger names? I worry that "AzureStorage" will be confusing for users since it doesn't have "Functions" anywhere in the name. Ideally it's consistent with our other log category names. How about "Host.Triggers.DurableTask.AzureStorage" so that it's effectively a subsection of the existing "Host.Triggers.DurableTask" category?

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
```diff
@@ -178,5 +185,25 @@ internal static OrchestrationInstanceStatusQueryCondition ConvertWebjobsDurableC
...
         }
+
+#if !FUNCTIONS_V1
+        /// <inheritdoc/>
+        public override bool TryGetScaleMonitor(
+            string functionId,
+            string functionName,
+            string hubName,
+            string storageConnectionString,
+            DurableTaskExtension extension,
```
cgillum, line 196: Can we remove `DurableTaskExtension extension` from the list of parameters? Seems like it's not actually being used, and I won't think we necessarily want to give 3rd parties access to this.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -74,6 +75,7 @@ public class DurableTaskExtension :
         private IDurabilityProviderFactory durabilityProviderFactory;
         private INameResolver nameResolver;
         private ILoggerFactory loggerFactory;
+        private ILogger logger;
```
cgillum, line 78: Do we still need this change, or can it be reverted? Looks like it's not actually being used.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
```diff
@@ -86,7 +87,7 @@ public async Task<DurableTaskTriggerMetrics> GetMetricsAsync()
             }
             catch (StorageException e)
             {
-                this.traceHelper.ExtensionWarningEvent(this.hubName, this.functionName.Name, string.Empty, e.ToString());
+                this.logger.LogWarning(e.ToString(), this.hubName, this.functionName);
```
cgillum, line 90: What's this log message expected to look like? I thought `LogWarning` was supposed to take a template as the first parameter and arguments for that template as the subsequent parameters.

src/WebJobs.Extensions.DurableTask/ProviderUtils.cs
```diff
@@ -61,5 +61,18 @@ public static DurableOrchestrationStatus ConvertOrchestrationStateToStatus(Orche
...
+
+        /// <summary>
+        /// Trace a warning event for the extension.
+        /// </summary>
+        /// <param name="extension">The durable task extension.</param>
+        /// <param name="hubName">The name of the task hub.</param>
+        /// <param name="functionName">The function name.</param>
+        /// <param name="instanceId">The instance id.</param>
+        /// <param name="message">The message.</param>
+        public static void TraceWarningEvent(this DurableTaskExtension extension, string hubName, string functionName, string instanceId, string message)
```
cgillum, line 73: Is this still being used or can it be reverted?

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
bachuv: I kept it in because it looks like the corresponding PR in Netherite that implements `TryGetScaleMonitor()` uses `extension`.

PR - [Support autoscaling on elastic premium plans #24](https://github.com/microsoft/durabletask-netherite/pull/24/)

[TryGetScaleMonitor implementation in that PR](https://github.com/microsoft/durabletask-netherite/pull/24/files#diff-f47141027cfddf5df3bff1836fc7931ca6c1622f56eb3f60cf192b4d2850461dR105)

ConnorMcMahon: It looks like it was just using it for logging purposes, so they can move to using a logger they create themselves.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
bachuv: `LogWarning` is supposed to take a template and then parameters. I updated it to that format. Thanks for pointing that out!

src/WebJobs.Extensions.DurableTask/ProviderUtils.cs
bachuv: I actually don't see anything in ProviderUtils.cs being used anywhere so I'm going to delete the file.

cgillum (approved):
Thanks for making these changes! I look forward to updating the SQL backend to take a dependency on this work.

ConnorMcMahon (approved):
LGTM!

PR #1702, opened 2021-03-01
Adding Storage Backend Selection in host.json

### LINKED ISSUE #1666: Ability to select alternate storage backend in host.json
Soon we will be ready to release new backends for Durable Functions: a Microsoft SQL Database backend and a Netherite backend. To make it easier for customers to opt-into these new backends, we want to include them in the default Azure Functions extension bundles. However, the backend extensibility design does not currently allow you to choose the backend you want using configuration. Rather, it relies on the user to selectively load the backend extension and this won't work well if we need to ship all backends together in the same extension bundle.

We would like a way to select which Durable backend to use in host.json, with the default being the Azure Storage backend. For example, it could be as simple as a `type` field under `durableTask/storageProvider`:

```json
{
  "version": "2.0",
  "extensions": {
    "durableTask": {
      "storageProvider": {
        "type": "MicrosoftSQL",
        "connectionStringName": "SQLDB_Connection"
      }
    }
  }
}
```

If a backend provider is registered with type "MicrosoftSQL", that provider will be loaded. If no provider is found with the specified name, then an exception is thrown. If no type is specified, the Azure Storage backend is used.

Alternatively we could require that alternate backends get deployed to separate bundles. However, this requires the bundles infrastructure and related tools to be updated to support alternate backends, which would be a time-consuming project.

bachuv:
These changes allow customers to configure which storage backend they want their app to use (e.g. Azure Storage, MicrosoftSQL, or Netherite) with a new `type` attribute under `durableTask/storageProvider` in host.json (example below). The extension will register all storage backend services and then choose which one to use in the `DurableTaskExtension` constructor. Adding the ability to register all backends and then choose a storage backend allows us to deploy all storage backend packages through extension bundles.

The main changes in this PR are allowing the extension to register a collection of services through dependency injection, adding a `Name` property in `IDurabilityProviderFactory`, and choosing the `DurabilityProvider` by comparing a customers configured `type` with the `Name` property in the available `IDurabilityProviderFactory` implementations.

```
{
  "version": "2.0",
  "extensions": {
    "durableTask": {
      "storageProvider": {
        "type": "MicrosoftSQL",
        "connectionStringName": "SQLDB_Connection"
      }
    }
  }
}
```

Resolves #1666 

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
+++ b/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
@@ -41,6 +41,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.defaultConnectionName = this.azureStorageOptions.ConnectionStringName ?? ConnectionStringNames.Storage;
         }
 
+        public string Name => "Azure Storage";
+
         internal string GetDefaultStorageConnectionString()
         {
             return this.connectionStringResolver.Resolve(this.defaultConnectionName);
--- a/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
+++ b/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
@@ -11,6 +11,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 {
     internal class AzureStorageDurabilityProviderFactory : IDurabilityProviderFactory
     {
+        internal const string ProviderName = "AzureStorage";
+
         private readonly DurableTaskOptions options;
         private readonly AzureStorageOptions azureStorageOptions;
         private readonly IConnectionStringResolver connectionStringResolver;
@@ -41,7 +43,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.defaultConnectionName = this.azureStorageOptions.ConnectionStringName ?? ConnectionStringNames.Storage;
         }
 
-        public string Name => "AzureStorage";
+        public string Name => ProviderName;
 
         internal string GetDefaultStorageConnectionString()
         {
--- a/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
+++ b/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
@@ -41,7 +41,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.defaultConnectionName = this.azureStorageOptions.ConnectionStringName ?? ConnectionStringNames.Storage;
         }
 
-        public string Name => "Azure Storage";
+        public string Name => "AzureStorage";
 
         internal string GetDefaultStorageConnectionString()
         {
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -101,7 +101,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// <param name="options">The configuration options for this extension.</param>
         /// <param name="loggerFactory">The logger factory used for extension-specific logging and orchestration tracking.</param>
         /// <param name="nameResolver">The name resolver to use for looking up application settings.</param>
-        /// <param name="orchestrationServiceFactory">The factory used to create orchestration service based on the configured storage provider.</param>
+        /// <param name="orchestrationServiceFactories">The factories used to create orchestration service based on the configured storage provider.</param>
         /// <param name="durableHttpMessageHandlerFactory">The HTTP message handler that handles HTTP requests and HTTP responses.</param>
         /// <param name="hostLifetimeService">The host shutdown notification service for detecting and reacting to host shutdowns.</param>
         /// <param name="lifeCycleNotificationHelper">The lifecycle notification helper used for custom orchestration tracking.</param>
@@ -113,7 +113,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             IOptions<DurableTaskOptions> options,
             ILoggerFactory loggerFactory,
             INameResolver nameResolver,
-            IDurabilityProviderFactory orchestrationServiceFactory,
+            IEnumerable<IDurabilityProviderFactory> orchestrationServiceFactories,
             IApplicationLifetimeWrapper hostLifetimeService,
             IDurableHttpMessageHandlerFactory durableHttpMessageHandlerFactory = null,
             ILifeCycleNotificationHelper lifeCycleNotificationHelper = null,
@@ -137,7 +137,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
             this.TraceHelper = new EndToEndTraceHelper(logger, this.Options.Tracing.TraceReplayEvents);
             this.LifeCycleNotificationHelper = lifeCycleNotificationHelper ?? this.CreateLifeCycleNotificationHelper();
-            this.durabilityProviderFactory = orchestrationServiceFactory;
+            this.durabilityProviderFactory = this.GetDurabilityProviderFactory(logger, orchestrationServiceFactories);
             this.defaultDurabilityProvider = this.durabilityProviderFactory.GetDurabilityProvider();
             this.isOptionsConfigured = true;
 
@@ -170,11 +170,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             IOptions<DurableTaskOptions> options,
             ILoggerFactory loggerFactory,
             INameResolver nameResolver,
-            IDurabilityProviderFactory orchestrationServiceFactory,
+            IEnumerable<IDurabilityProviderFactory> orchestrationServiceFactories,
             IConnectionStringResolver connectionStringResolver,
             IApplicationLifetimeWrapper shutdownNotification,
             IDurableHttpMessageHandlerFactory durableHttpMessageHandlerFactory)
-            : this(options, loggerFactory, nameResolver, orchestrationServiceFactory, shutdownNotification, durableHttpMessageHandlerFactory)
+            : this(options, loggerFactory, nameResolver, orchestrationServiceFactories, shutdownNotification, durableHttpMessageHandlerFactory)
         {
             this.connectionStringResolver = connectionStringResolver;
         }
@@ -234,6 +234,29 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             isDefault = errorSerializerSettingsFactory is ErrorSerializerSettingsFactory;
 
             return new MessagePayloadDataConverter(errorSerializerSettingsFactory.CreateJsonSerializerSettings(), isDefault);
+        }
+
+        private IDurabilityProviderFactory GetDurabilityProviderFactory(ILogger logger, IEnumerable<IDurabilityProviderFactory> orchestrationServiceFactories)
+        {
+            bool storageTypeIsConfigured = this.Options.StorageProvider.TryGetValue("type", out object storageType);
+            if (!storageTypeIsConfigured)
+            {
+                logger.LogInformation("Using the default storage provider: Azure Storage.");
+                return orchestrationServiceFactories.FirstOrDefault(f => f.Name.Equals("Azure Storage"));
+            }
+
+            if (orchestrationServiceFactories.FirstOrDefault(f => f.Name.Equals(storageType)) == null)
+            {
+                IList<string> factoryNames = orchestrationServiceFactories.Select(f => f.Name).ToList();
+                logger.LogWarning($"Storage provider type ({storageType}) was not found. " +
+                    $"Available storage providers: {string.Join(", ", factoryNames)}. " +
+                    $"Using the default storage provider: Azure Storage.");
+
+                return orchestrationServiceFactories.FirstOrDefault(f => f.Name.Equals("Azure Storage"));
+            }
+
+            logger.LogInformation($"Using the {storageType} storage provider.");
+            return orchestrationServiceFactories.FirstOrDefault(f => f.Name.Equals(storageType));
         }
 
         internal string GetBackendInfo()
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -187,7 +187,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             IPlatformInformationService platformInformationService)
 #pragma warning restore CS0612 // Type or member is obsolete
 
-            : this(options, loggerFactory, nameResolver, orchestrationServiceFactory, shutdownNotification, durableHttpMessageHandlerFactory)
+            : this(options, loggerFactory, nameResolver, orchestrationServiceFactories, shutdownNotification, durableHttpMessageHandlerFactory)
         {
             this.connectionStringResolver = connectionStringResolver;
             this.platformInformationService = platformInformationService;
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         INameVersionObjectManager<TaskOrchestration>,
         INameVersionObjectManager<TaskActivity>
     {
-        private const string DefaultProvider = "AzureStorage";
+        private const string DefaultProvider = AzureStorageDurabilityProviderFactory.ProviderName;
 
         internal static readonly string LoggerCategoryName = LogCategories.CreateTriggerCategory("DurableTask");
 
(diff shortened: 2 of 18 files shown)


src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -236,6 +236,29 @@ private MessagePayloadDataConverter CreateErrorDataConverter(IErrorSerializerSet
...
+                IList<string> factoryNames = orchestrationServiceFactories.Select(f => f.Name).ToList();
+                logger.LogWarning($"Storage provider type ({storageType}) was not found. " +
+                    $"Available storage providers: {string.Join(", ", factoryNames)}. " +
+                    $"Using the default storage provider: Azure Storage.");
+
+                return orchestrationServiceFactories.FirstOrDefault(f => f.Name.Equals("Azure Storage"));
+            }
+
+            logger.LogInformation($"Using the {storageType} storage provider.");
+            return orchestrationServiceFactories.FirstOrDefault(f => f.Name.Equals(storageType));
```
cgillum, line 259: It looks like you're doing `orchestrationServiceFactories.FirstOrDefault(f => f.Name.Equals(storageType))` multiple times. Let's reduce this to just once:
```csharp
// Lookup is case-insensitive
IDurabilityProviderFactory selectedFactory = orchestrationServiceFactories.FirstOrDefault(
    f => f.Name.Equals(storageType, StringComparison.OrdinalIgnoreCase));
if (selectedFactory == null)
{
    // ...
}

return selectedFactory;
```

test/Common/DurableClientBaseTests.cs
```diff
@@ -335,7 +335,7 @@ private static DurableTaskExtension GetDurableTaskConfig()
                 wrappedOptions,
                 new LoggerFactory(),
                 nameResolver,
-                serviceFactory,
+                new[] { serviceFactory }.ToList(),
```
cgillum, line 338: No need for `.ToList()` here.
```suggestion
                new[] { serviceFactory },
```

test/Common/DurableTaskLifeCycleNotificationTest.cs
```diff
@@ -1243,15 +1243,16 @@ public void EventGridApiConfigureCheck()
...
             var extension = new DurableTaskExtension(
                 wrappedOptions,
                 new LoggerFactory(),
                 mockNameResolver.Object,
-                new AzureStorageDurabilityProviderFactory(
-                    wrappedOptions,
-                    connectionStringResolver,
-                    mockNameResolver.Object,
-                    NullLoggerFactory.Instance),
+                new[] { serviceFactory }.ToList(),
```
cgillum, line 1255: No need for `.ToList()` here.
```suggestion
                new[] { serviceFactory },
```

test/Common/DurableTaskLifeCycleNotificationTest.cs
```diff
@@ -1289,15 +1290,16 @@ public void CustomHelperTypeActivationFailed()
 
             var wrappedOptions = new OptionsWrapper<DurableTaskOptions>(options);
             var nameResolver = new SimpleNameResolver();
+            var serviceFactory = new AzureStorageDurabilityProviderFactory(
+                    wrappedOptions,
+                    new TestConnectionStringResolver(),
+                    nameResolver,
+                    NullLoggerFactory.Instance);
```
cgillum, line 1297: minor nit: use 4 spaces for indenting instead of 8 spaces.

test/Common/DurableTaskLifeCycleNotificationTest.cs
```diff
@@ -1289,15 +1290,16 @@ public void CustomHelperTypeActivationFailed()
...
             var extension = new DurableTaskExtension(
                 wrappedOptions,
                 new LoggerFactory(),
                 nameResolver,
-                new AzureStorageDurabilityProviderFactory(
-                    wrappedOptions,
-                    new TestConnectionStringResolver(),
-                    nameResolver,
-                    NullLoggerFactory.Instance),
+                new[] { serviceFactory }.ToList(),
```
cgillum, line 1302:
```suggestion
                new[] { serviceFactory },
```

test/Common/DurableTaskLifeCycleNotificationTest.cs
```diff
@@ -1317,15 +1319,16 @@ public async Task CustomHelperTypeDependencyInjection()
...
```
cgillum, line 1331:
```suggestion
                new[] { serviceFactory },
```

test/Common/HttpApiHandlerTests.cs
```diff
@@ -1389,11 +1389,14 @@ public MockDurableTaskExtension(DurableTaskOptions options)
...
                     TestHelpers.GetTestNameResolver(),
-                    new AzureStorageDurabilityProviderFactory(
+                    new[]
+                    {
+                        new AzureStorageDurabilityProviderFactory(
                         new OptionsWrapper<DurableTaskOptions>(options),
                         new TestConnectionStringResolver(),
                         TestHelpers.GetTestNameResolver(),
                         NullLoggerFactory.Instance),
+                    }.ToList(),
```
cgillum, line 1399:
```suggestion
                    },
```

test/FunctionsV1/PlatformSpecificHelpers.FunctionsV1.cs
```diff
@@ -50,7 +51,7 @@ public static class PlatformSpecificHelpers
                 options,
                 loggerFactory,
                 nameResolver,
-                orchestrationServiceFactory,
+                new[] { orchestrationServiceFactory }.ToList(),
```
cgillum, line 54:
```suggestion
                new[] { orchestrationServiceFactory },
```

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
```diff
@@ -41,6 +41,8 @@ internal class AzureStorageDurabilityProviderFactory : IDurabilityProviderFactor
             this.defaultConnectionName = this.azureStorageOptions.ConnectionStringName ?? ConnectionStringNames.Storage;
         }
 
+        public string Name => "Azure Storage";
```
cgillum, line 44: It feels a little strange having spaces in a symbolic name like this that needs to be referenced in configuration. I wonder if we should use "AzureStorage" instead of "Azure Storage". I fear that customers may accidentally choose the former when trying to be explicit in their configuration.

cgillum (approved):
Looks good to me! Just a few nitpicks from my POV.

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
ConnorMcMahon: Definitely agree on having types with no space characters. Feels more inline with standard configuration enumerations.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -236,6 +236,29 @@ private MessagePayloadDataConverter CreateErrorDataConverter(IErrorSerializerSet
             return new MessagePayloadDataConverter(errorSerializerSettingsFactory.CreateJsonSerializerSettings(), isDefault);
         }
 
+        private IDurabilityProviderFactory GetDurabilityProviderFactory(ILogger logger, IEnumerable<IDurabilityProviderFactory> orchestrationServiceFactories)
+        {
+            bool storageTypeIsConfigured = this.Options.StorageProvider.TryGetValue("type", out object storageType);
```
ConnorMcMahon, line 241: I would almost prefer if the `DurableTaskOptions` was directly passed into a parameter for this method. I realize that we have already set `this.Options` has already been set at this point, but it's possible during a refactor that this method gets called when this is null, resulting in a hard to diagnose NullReferenceException.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -236,6 +236,29 @@ private MessagePayloadDataConverter CreateErrorDataConverter(IErrorSerializerSet
...
+            if (!storageTypeIsConfigured)
+            {
+                logger.LogInformation("Using the default storage provider: Azure Storage.");
+                return orchestrationServiceFactories.FirstOrDefault(f => f.Name.Equals("Azure Storage"));
+            }
+
+            if (orchestrationServiceFactories.FirstOrDefault(f => f.Name.Equals(storageType)) == null)
+            {
+                IList<string> factoryNames = orchestrationServiceFactories.Select(f => f.Name).ToList();
+                logger.LogWarning($"Storage provider type ({storageType}) was not found. " +
```
ConnorMcMahon, line 251: Do we want to select the AzureStorage provider when the storageType provided doesn't actually exist, or should we throw an exception? I'm just afraid of cases where a customer accidentally typos their provider name but the program relatively silently succeeds, just with a different backend than intended.

ConnorMcMahon (commented):
A few nits, but overall looks great!

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -236,6 +236,29 @@ private MessagePayloadDataConverter CreateErrorDataConverter(IErrorSerializerSet
...
+                    $"Available storage providers: {string.Join(", ", factoryNames)}. " +
+                    $"Using the default storage provider: Azure Storage.");
+
+                return orchestrationServiceFactories.FirstOrDefault(f => f.Name.Equals("Azure Storage"));
```
sebastianburckhardt, line 255: I would probably use a global definition here, instead of a string literal, e.g. 
```csharp
const string DefaultProvider = "AzureStorage"
```

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
sebastianburckhardt: +1 for throwing an exception if the configuration contains an error.

sebastianburckhardt: +1 for throwing an exception if the configuration contains an error.

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
```diff
@@ -41,7 +41,7 @@ internal class AzureStorageDurabilityProviderFactory : IDurabilityProviderFactor
             this.defaultConnectionName = this.azureStorageOptions.ConnectionStringName ?? ConnectionStringNames.Storage;
         }
 
-        public string Name => "Azure Storage";
+        public string Name => "AzureStorage";
```
cgillum, line 44: To avoid duplication, I wonder if either this property should reference DurableTaskExtension.DefaultProvider or (perhaps slightly better) you define an internal const in this class and have both the `Name` property here and the DurableTaskExtension code both reference that const.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -236,27 +237,27 @@ private MessagePayloadDataConverter CreateErrorDataConverter(IErrorSerializerSet
...
+        private IDurabilityProviderFactory GetDurabilityProviderFactory(DurableTaskOptions options, ILogger logger, IEnumerable<IDurabilityProviderFactory> orchestrationServiceFactories)
         {
-            bool storageTypeIsConfigured = this.Options.StorageProvider.TryGetValue("type", out object storageType);
+            bool storageTypeIsConfigured = options.StorageProvider.TryGetValue("type", out object storageType);
+
             if (!storageTypeIsConfigured)
             {
                 logger.LogInformation("Using the default storage provider: Azure Storage.");
-                return orchestrationServiceFactories.FirstOrDefault(f => f.Name.Equals("Azure Storage"));
+                return orchestrationServiceFactories.FirstOrDefault(f => f.Name.Equals(DefaultProvider));
```
cgillum, line 247: I recognize that this should never happen, but I wonder if we should try to fail here if `DefaultProvider` isn't found. Perhaps we can change `FirstOrDefault()` to just `First()`? An exception thrown here might be easier to debug that a null-reference exception thrown by the caller.

cgillum (approved):
This looks good to me. Just a couple non-blocking comments.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
bachuv: That makes sense. I changed both `FirstOrDefault()` calls in this method to `First()` to keep it consistent.

ConnorMcMahon (approved):
LGTM!

test/FunctionsV2/StorageProviderSelectionTests.cs
```diff
@@ -0,0 +1,80 @@
...
+        private static IEnumerable<IDurabilityProviderFactory> EmptyDurabilityProviderFactoriesList { get; }
+
+        public static IEnumerable<object[]> ValidStorageProviderTypesData { get; }
+
+        [Theory]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        [MemberData(nameof(ValidStorageProviderTypesData))]
+        public void SelectingAvailableStorageProvider(string storageProvider, IDurabilityProviderFactory expectedFactory)
+        {
+            IDurabilityProviderFactory defaultFactory = AllDurabilityProviderFactories.First(f => f.Name.Equals(storageProvider));
```
ConnorMcMahon, line 59: I'm not seeing much value from these unit tests, because it is not actually running through production code, it is instead running against code that is used in production code.

This has two main disadvantages.
1. If the production code changes (and regresses), these tests will still pass, because they are out of sync with actual production code.
2. They don't test end-to-end behavior.

Instead, consider creating an end-to-end run of a JobHost (similar to DurableTaskEndToEndTests) that takes an IEnumerable<IDurabilityProviderFactory> as a parameter fed to it. Feed in the mocks you created above, and then you can call `.Verify()` on those mocks to see which of the IDurabilityProviderFactory objects was used to create the durability provider.

Note, that since you may not be returning an actual Durability provider, you may run into an exception at some point during the `JobHost.Start()` process, but you can just catch that exception gracefully in your test code, and then do the verification.

ConnorMcMahon (changes requested):
I think the current tests need some tweaking.

test/Common/TestHelpers.cs
```diff
@@ -194,6 +195,33 @@ internal static class TestHelpers
                 onSend: onSend);
         }
 
+#if !FUNCTIONS_V1
+        public static ITestHost GetJobHostWithMultipleDurabilityProviders(
+        DurableTaskOptions options = null,
```
ConnorMcMahon, line 200: small nit: properly indent these parameters.

test/Common/TestHelpers.cs
```diff
@@ -20,6 +20,7 @@
 using Microsoft.WindowsAzure.Storage.Table;
 using Xunit;
 using Xunit.Abstractions;
+using static Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests.PlatformSpecificHelpers;
```
ConnorMcMahon, line 23: nit: this may lead to unused using statements for FUNCTIONS_V1. Consider just prepending your static calls below with PlatformSpecificHelpers.

test/FunctionsV2/StorageProviderSelectionTests.cs
```diff
@@ -4,77 +4,174 @@
...
-            ValidStorageProviderTypesData = new List<object[]>
+            using (ITestHost host = TestHelpers.GetJobHostWithMultipleDurabilityProviders(
+                options: options,
+                durabilityProviderFactories: durabilityProviderFactories))
             {
-                new object[] { "AzureStorage", AzureStorageDurabilityProviderFactoryMock.Object },
-                new object[] { "MicrosoftSQL", MicrosoftSQLDurabilityProviderFactoryMock.Object },
-                new object[] { "Netherite", NetheriteProviderDurabilityProviderFactoryMock.Object },
-            };
+                azureStorageMock.Verify(a => a.GetDurabilityProvider(), Times.Once());
```
ConnorMcMahon, line 38: Yep, this is exactly what I had in mind!

test/FunctionsV2/StorageProviderSelectionTests.cs
```diff
@@ -4,77 +4,174 @@
...
-        {
-            AzureStorageDurabilityProviderFactoryMock = new Mock<IDurabilityProviderFactory>();
-            AzureStorageDurabilityProviderFactoryMock.Setup(f => f.Name).Returns("AzureStorage");
+        public StorageProviderSelectionTests() { }
 
-            MicrosoftSQLDurabilityProviderFactoryMock = new Mock<IDurabilityProviderFactory>();
-            MicrosoftSQLDurabilityProviderFactoryMock.Setup(f => f.Name).Returns("MicrosoftSQL");
+        [Theory]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        [InlineData("AzureStorage")]
```
ConnorMcMahon, line 21: If you are only going to use one piece of data, don't parameterize the method and just use a constant.

ConnorMcMahon: That being said, most of the code is identical for these tests, so I think you can parameterize them.

All you need to do is modify your verify statements to look something like this.

```csharp
azurestorageMock.Verify(a => a.GetDurabilityProvider(), string.Equals(storageProvider, "AzureStorage") ? Times.Once() : Times.Never())
```

test/FunctionsV2/StorageProviderSelectionTests.cs
```diff
@@ -4,77 +4,174 @@
...
+        public void SelectingAzureStorageStorageProvider(string storageProvider)
```
ConnorMcMahon, line 22: nit: If you parameterize the method, a more specific name is helpful. Maybe something like `StorageProviderTypeSpecified_CorrectStorageProviderFactoryUsed)`

test/FunctionsV2/StorageProviderSelectionTests.cs
```diff
@@ -4,77 +4,174 @@
...
+                microsoftSQLMock.Verify(m => m.GetDurabilityProvider(), Times.Once());
+            }
         }
 
-        [Theory]
+        [Fact]
         [Trait("Category", PlatformSpecificHelpers.TestCategory)]
-        [InlineData("AzureStorage")]
-        public void SelectingFromEmptyFactoryListThrowsException(string storageProvider)
+        public void SelectingFromEmptyFactoryListThrowsException()
```
ConnorMcMahon, line 96: Maybe we could generalize this test to `NoProviderSpecified_AzureStorageFactoryNotRegisterd()`, and then parameterize it with both the empty list and a list with netherite and sql?

ConnorMcMahon (approved):
The PR content is solid at this point! Just address some of the stylistic and naming suggestions before merging.

PR #1706, opened 2021-03-03
Apply new consumption defaults in AzureStorageDurabilityProvider

### LINKED ISSUE #1646: Reliability: Improve concurrency defaults for Consumption plan hosting
Some of the default concurrency settings do not work well for the Consumption plan. In particular, it's very easy to exceed TCP connection limits on Windows due to large numbers of concurrent Azure Storage queue operations. See [this customer support incident](https://portal.microsofticm.com/imp/v3/incidents/details/221648212/home) (requires MSFT credentials) as a reference.

My specific recommendation:
* Reduce the default value of `controlQueueBufferThreshold` from 256 to **32**. This ensures we don't do too many concurrent delete operations at once.
* Reduce the default value of `maxConcurrentOrchestratorFunctions` to **5**. This ensures fan-out operations do not create too many concurrent network connections.
* Reduce the default value of `MaxStorageOperationConcurrency` to **25**. It's currently derived from the CPU count, which is misleading in the Consumption plan. This setting may not be exposed to developers, but it probably should be.

The current defaults are fine for non-Consumption plans. These recommendations are specifically for Consumption because of tighter network resource restrictions. The extension code therefore needs to be aware of what the current plan is before making these settings changes (I assume this is possible at runtime, but not 100% sure).

davidmrdavid:
Addresses: https://github.com/Azure/azure-functions-durable-extension/issues/1646

This PR applies the new concurrency defaults outlined in the issue above. This all done in the `AzureStorageDurabilityProviderFactory`, where defaults are explicitly overridden after inspecting if the application is running in the consumption plan.

This PR also introduces a new interface: `IPlatformInformationService`. This interface serves to abstract over the minutia of inspecting environment vars to determine the underlying app service plan and instead exposes self-descriptive methods like `InLinuxConsumption` and `InWindowsConsumption`, etc. In the future, I hope this interface will also provide info about the user-facing PL, and other information that can help us guide optimizations and defaults.

Currently, this interface contains only 1 implementation: `DefaultPlatformInformationProvider`. It uses the same `INameResolver` injected by DI to look up environment variables and config setting values. This `DefaultPlatformInformationProvider` is also injected via DI to the `DurableTaskExtension`.

The rest of changes are all in the tests. Since the `DurableTaskExtension` constructor now takes an extra argument, a bunch of tests had to be refactored to account for the extra parameter. The guiding strategy here was to creater a new `TestHelper` which provides a `Moq.Mock` of the `IPlatformInformationService` interface. With it, we can easily construct a dummy instance of the interface for platform-agnostic tests. Additionally, this mock is parametrizable and would allow for platform-specific tests moving forward! 🚀 

**Some Open Questions**
* Is the "PlatformInformation" portion of `IPlatformInformationService` really the best descriptor? I want this interface to expose info about the OS, underlying SKU, and, in the future, also the user's PL. Alternatives could be "ApplicationContext", "UserContext", etc.

**Remaining ToDos**
* Potentially modify the upcoming backends to account for this change
* Add one or two new tests
* Test this on a few OS+SKU combinations
* Complete PR checklist

resolves #1646 

### Pull request checklist

    * [x] Otherwise: Documentation PR is ready to merge and referenced in `pending_docs.md`


src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
```diff
@@ -27,12 +28,21 @@ internal class AzureStorageDurabilityProviderFactory : IDurabilityProviderFactor
             IOptions<DurableTaskOptions> options,
             IConnectionStringResolver connectionStringResolver,
             INameResolver nameResolver,
-            ILoggerFactory loggerFactory)
+            ILoggerFactory loggerFactory,
+            PlatformInformation platformInfo = null)
```
ConnorMcMahon, line 32: You should likely make PlatformInformation a service, and then inject that service via dependency injection (see [this class](https://github.com/Azure/azure-functions-durable-extension/blob/e600de93b78eb45335947741232292ea60c1530e/src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs#L32) and its methods called by the Functions runtime to see how we add services).

That will allow any of the other services injected (such as this one), to access your service just by adding it as a parameter in the constructor. You just need to make sure there are no circular dependencies.

This gives us flexibility in how the other backends' implementations of IDurabilityProviderFactory consume platform information.

ConnorMcMahon: Note, the DI patterns used in .NET Core generally requires you to define an interface and then make a concrete implementation of that interface. You then program against the interface.

davidmrdavid: Perfect, thanks! After reading up a bit more on how we use DI in our codebase, and DI in .NET more broadly, I have a very clear idea of what you mean now. Should be an easy refactor!

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
```diff
@@ -27,12 +28,21 @@ internal class AzureStorageDurabilityProviderFactory : IDurabilityProviderFactor
...
             INameResolver nameResolver,
-            ILoggerFactory loggerFactory)
+            ILoggerFactory loggerFactory,
+            IPlatformInformationService platformInfo = null)
         {
             this.options = options.Value;
             this.nameResolver = nameResolver;
             this.loggerFactory = loggerFactory;
             this.azureStorageOptions = new AzureStorageOptions();
+            this.inConsumption = platformInfo != null ? platformInfo.InConsumption() : false;
```
davidmrdavid, line 38: Alternatively, we could just fail fast when a platformInfo object isn't provided. Looking back, I would prefer this alternative. What do others think?

src/WebJobs.Extensions.DurableTask/DefaultPlatformInformationProvider.cs
```diff
@@ -0,0 +1,61 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    internal class DefaultPlatformInformationProvider : IPlatformInformationService
```
davidmrdavid, line 8: I forgot to add a header comment to this class. Will do later.
Essentially, this class looks up env-var values using the DI-provided nameResolver

davidmrdavid:
I'm asking for a review now to make sure we feel comfortable with the current approach, I do realize there are still a few leftovers, which I list above :) . Thanks everyone

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
ConnorMcMahon: I second fail fast. It should never be null, since we provide the implementation.

src/WebJobs.Extensions.DurableTask/IPlatformInformationService.cs
```diff
@@ -0,0 +1,58 @@
...
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    /// <summary>
+    /// Interface for accessing the AppService plan information,
+    /// the OS, and user-facing PL.
+    ///
+    /// Note: The functionality is currently limited, but will grow
+    /// along with the pursuit of more platform-specific defaults.
+    /// </summary>
+    public interface IPlatformInformationService
```
ConnorMcMahon, line 13: I think we may be able to clean this up a little bit in a way that would appeal to the functional programmer in you.

Imagine we have `IPlatformInformationService` have a single method `PlatformInformation GetPlatformInformation()`.

Then you can have various subclasses of PlatformInformation, such as AppServiceWindows, AppServiceWindowsConsumption, AppServiceLinux, AppServiceLinuxConsumption, Unknown, etc. 

That type hierarchy alone will allow you to eliminate many of these methods.

Then, you can put the information that is only specific to windows or linux only on the appropriate subclasses.

You can then use pattern matching for these various cases, as opposed to a series of if-statements.

Overall, this is more of a matter of preference if you prefer a sequence of if-statements vs a type based approach + pattern matching, but given your background in PL, I thought I would mention this option.

test/Common/TestHelpers.cs
```diff
@@ -192,6 +193,24 @@ internal static class TestHelpers
                 onSend: onSend);
         }
 
+        public static IPlatformInformationService GetMockPlatformInformationService(
```
ConnorMcMahon, line 196: Instead of mocking IPlatformInformationService, it may be better to mock the underlying INameResolver, as that mimics production slightly better.

Considering how slim of a wrapper IPlatformInformationService is it may not make a huge difference, but it's worth considering.

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
```diff
@@ -155,6 +165,11 @@ private AzureStorageDurabilityProvider GetAzureStorageStorageProvider(DurableCli
                 UseLegacyPartitionManagement = this.azureStorageOptions.UseLegacyPartitionManagement,
             };
 
+            if (this.inConsumption)
```
ConnorMcMahon, line 168: I'm guessing the reason we do this here as opposed to above is because this is not exposed on the host.json?

It's also worth noting that I'm not sure this is any different that the default value for this setting anyways.

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
```diff
@@ -27,12 +28,21 @@ internal class AzureStorageDurabilityProviderFactory : IDurabilityProviderFactor
...
             this.options = options.Value;
             this.nameResolver = nameResolver;
             this.loggerFactory = loggerFactory;
             this.azureStorageOptions = new AzureStorageOptions();
+            this.inConsumption = platformInfo != null ? platformInfo.InConsumption() : false;
+
+            if (this.inConsumption)
+            {
+                this.azureStorageOptions.ControlQueueBufferThreshold = 32;
+                this.options.MaxConcurrentOrchestratorFunctions = 5;
```
ConnorMcMahon, line 43: Other settings we may want to edit include `this.options.MaxConcurrentActivityFunctions`.

It's also worth adjusting `this.azureStorageOptions.ControlQueueBatchSize`. Currently that defaults to 32, which means we are going to get a very start/stop pre-fetch behavior given the current control queue buffer threshold. Maybe the batch size should be some fraction of the controlqueuebuffertreshold by default?

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
```diff
@@ -27,12 +28,21 @@ internal class AzureStorageDurabilityProviderFactory : IDurabilityProviderFactor
...
         {
             this.options = options.Value;
             this.nameResolver = nameResolver;
             this.loggerFactory = loggerFactory;
             this.azureStorageOptions = new AzureStorageOptions();
+            this.inConsumption = platformInfo != null ? platformInfo.InConsumption() : false;
+
+            if (this.inConsumption)
+            {
+                this.azureStorageOptions.ControlQueueBufferThreshold = 32;
```
ConnorMcMahon, line 42: 32 may be a bit aggressive for all consumption customers, at least for C#. It may be different when we incorporate out-of-proc as well...

I would love to see some fan-out, fan-in perf numbers with both our current default and these new defaults on consumption plan (maybe with 1000 and then 10000?) just to make sure that the new default experience isn't a huge perf drop?

ConnorMcMahon (commented):
Overall, good initial stab at this. 

I propose a more pattern matchy approach to platform information that may or may not be a good idea, along with some specific nits about the new defaults.

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
davidmrdavid: Will do, I was planning to provide those before merging.
Also, for clarity, this PR will not consider OOProc performance tuning, but that's coming shortly

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
davidmrdavid: I think you're right. It's `5 * ProcessorCount` which I think is 1 in the consumption plan.

https://github.com/Azure/durabletask/blob/d3c53b69259ea401839a8dd2aa21f96857a79314/src/DurableTask.AzureStorage/AzureStorageOrchestrationServiceSettings.cs#L114

I just went off the literal recommendation here: https://github.com/Azure/azure-functions-durable-extension/issues/1646

@cgillum, any thoughts on whether or not this setting's default value should be tweaked?
Another alternative is that I can leave it as-is and simply expose it for customers to tweak

test/Common/TestHelpers.cs
davidmrdavid: That's true, but I'm not sure that we gain that much benefit from the cost of making the mock a little bit more low-level. Let me get back to this suggestion. I'll want to first prioritize the other bits :)

src/WebJobs.Extensions.DurableTask/IPlatformInformationService.cs
davidmrdavid: Haha, I've been nerd-snipped.
Yeah I see what you mean and I do like it. I can certainly see this being a clean solution for serving OS-specific information. I suppose I could also bundle PL-specific data in there, I still need to think about that.

I'll get back to this. I think there's still some contention about how many settings to optimize and what values to give them so let me focus on that first since we're time-constrained. There will definitely be a follow-up PR to this for OOProc and we can use that to finalize the design of this utility if we don't get to it now.

davidmrdavid:
**Performance data for Baseline (Durable-Extension v.2.4.1)**

_Note:_ The Duration column is calculated as the time difference between the orchestrator's `FunctionStarting` to `FunctionCompleted` events.

_Time to completion_

| Benchmark ID | Description | Duration |
| ------------ |------------ | -------- |
|  A | Fanning over 1k activities | 20ms |
|  B | Fanning over 10k activities |  ~3min + 10ms | 
|  C | Fanning over 100 sun-orchestrators w/ 100 sequential activities each | ~7min |
|  D | Fanning over 100 sub-orchestrators fanning over 10k activities each | ~13h and 28min |

_InstanceIDs to review raw data_

| Benchmark ID | InstanceID | 
| ------------ |------------ | 
|  A | abea706d76c045f890398a81c6ed807a |
|  B | 463638f2003e4243aa256736849ea4e5 | 
|  C | 69e8ce5f3acd406b84641ab67ac574e2 |
|  D | d901db064ab4458ab8cf8c2a11bd7526 |

davidmrdavid:
**Performance data for commit 7a22217 (this PR)**

_Note:_ The Duration column is calculated as the time difference between the orchestrator's `FunctionStarting` to `FunctionCompleted` events.

_Time to completion_

| Benchmark ID | Description | Duration |
| ------------ |------------ | -------- |
|  A | Fanning over 1k activities | ~37ms  |
|  B | Fanning over 10k activities |  ~5min and 10ms |
|  C | Fanning over 100 sun-orchestrators w/ 100 sequential activities each | ~2min 30s |
|  D | Fanning over 100 sub-orchestrators fanning over 10k activities each | ~7h |

_InstanceIDs to review raw data_

| Benchmark ID | InstanceID | 
| ------------ |------------ | 
|  A | 71f1fec6ee0f425c865d936ce5e09782 |
|  B | 7ca022b927c443f9823d0998fecf499e | 
|  C | e6de2507482747c7a79d6ac9f1d682b0 |
|  D | 3a56f220f08841e0845b2e2a937779d2 |

davidmrdavid:
Ping to @ConnorMcMahon and @cgillum to get their thoughts on the performance comparison above. In particular, I'm curious if this suggests any further config changes. Happy to iterate further.

From what I see, this PR performs relatively similar to the baseline in the "easy" benchmarks which are A and B. But, in the heavier benchmarks like C and D, it appears that the changes in this PR clearly outperform the baseline.

I did find it noteworthy that benchmark D took so many hours to complete. Note that benchmark D is essentially 100 concurrent instances of benchmark B. Considering that benchmark B takes on average 5 minutes to run, running 100 of them in sequence would have taken 5minutes * 100 =~ 8 hours. So, benchmark D, which parallelizes the previous scenario, didn't seem to have an effect to the average total time. I would love it if there was anything we could do at the default-settings level to improve this particular benchmark.

Thanks!

cgillum:
@davidmrdavid you had me at "the changes in this PR clearly outperform the baseline". :)

7h definitely seems problematic for case D, but I'm happy to see that these changes noticeably improve over the baseline for both cases C and D. I think the next step for investigating case D is to take a look and see whether there are problems related to not scaling out or reliability (process crashes, etc.) that might be causing things to take so long.

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
cgillum: Consumption VMs on Windows can have as many as 4 vCPUs but the app is only allowed to use one of them, which is why we need special handling (in other words, I expecte `Environment.ProcessorCount` may return 4 and therefore cause us to increase network concurrency more than we want to.

davidmrdavid:
**Performance data for commit `c35c4b8` in this PR / PRv2**
In this PR, in addition to the original changes introduced, we additionally make the following change:

```C#
this.azureStorageOptions.ControlQueueBatchSize = this.azureStorageOptions.ControlQueueBufferThreshold / 4
```

For clarity's sake, we now refer to the original results for "this PR" (correspond to commit 7a22217) as "PRv1" and this set of results is "PRv2".

_Note:_ The Duration column is calculated as the time difference between the orchestrator's `FunctionStarting` to `FunctionCompleted` events.

_Time to completion_

| Benchmark ID | Description | Duration |
| ------------ |------------ | -------- |
|  A | Fanning over 1k activities | 26ms |
|  B | Fanning over 10k activities | ~15min |
|  C | Fanning over 100 sub-orchestrators w/ 100 sequential activities each | ~2min 30s |
|  D | Fanning over 100 sub-orchestrators fanning over 10k activities each | TBD  |

_InstanceIDs to review raw data_

| Benchmark ID | InstanceID | 
| ------------ |------------ | 
|  A | 8909f37b60ea4e90841404adce4aa1e0 |
|  B | 225bd3c5088348269eb97de619a34a67 | 
|  C | 63abc9977a1e4acfbb6b9eae3ef8e411 |
|  D | e1792d065b17405fa1969137aefc17e8 |

_Comments_
The results for benchmark A and C seem on-par with PRv1. However, with these changes, we also see the worst performance for benchmark B, which is twice as long as for the baseline and about 3 times longer than for PRv1. Finally, the results for benchmark D are still TBD.

davidmrdavid:
FYI, I updated my results for PRv1. Apparently I made a typo: the duration for benchmark B was ~15min instead of ~5.

davidmrdavid:
**Performance data for commit `815043c` in this PR / PRv3**
In this PR, in addition to the original changes introduced in PRv1, we additionally make the following change:

```C#
this.options.MaxConcurrentActivityFunctions = 10;
```

_Note:_ The Duration column is calculated as the time difference between the orchestrator's `FunctionStarting` to `FunctionCompleted` events.

_Time to completion_

| Benchmark ID | Description | Duration |
| ------------ |------------ | -------- |
|  A | Fanning over 1k activities | 26ms |
|  B | Fanning over 10k activities | ~15min |
|  C | Fanning over 100 sub-orchestrators w/ 100 sequential activities each | ~2min 30s |
|  D | Fanning over 100 sub-orchestrators fanning over 10k activities each | TBD  |

_InstanceIDs to review raw data_

| Benchmark ID | InstanceID | 
| ------------ |------------ | 
|  A | 8909f37b60ea4e90841404adce4aa1e0 |
|  B | 225bd3c5088348269eb97de619a34a67 | 
|  C | 63abc9977a1e4acfbb6b9eae3ef8e411 |
|  D | e1792d065b17405fa1969137aefc17e8 |

test/Common/TestHelpers.cs
davidmrdavid: Given the timeline, I'm gonna opt to wait until the OOProc performance tuning PR to consider this change.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -4,7 +4,7 @@
     <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
     <AssemblyName>Microsoft.Azure.WebJobs.Extensions.DurableTask</AssemblyName>
     <RootNamespace>Microsoft.Azure.WebJobs.Extensions.DurableTask</RootNamespace>
-    <MajorVersion>2</MajorVersion>
+    <MajorVersion>101</MajorVersion>
```
davidmrdavid, line 7: typo, will remove

src/WebJobs.Extensions.DurableTask/IPlatformInformationService.cs
davidmrdavid: I'll opt to postpone this refactor until the OOProc performance tuning PR

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
davidmrdavid: I've included performance data based on setting `MaxConcurrentActivityFunctions` to10 and `ControlQueueBatchSize` to 0.25*`ControlQueueBufferThreshold`. These results are named PRv3 and PRv2 respectively. From what I gather so far (at the time of writing the results for benchmark D are still to come) these didn't make a big difference in performance.

Since I didn't see a noticeable effect by changing those settings, I'm thinking it might be a good idea to revert back to the settings of PRv1 and to reconsider further tweaking of `MaxConcurrentActivityFunctions` and `ControlQueueBatchSize` for a future PR. What do you think, @ConnorMcMahon?

We can re-consider our decision once the results for benchmark D arrive in a few hours, but I have no reason to think they'll be surprising.

ConnorMcMahon:
I vote we leave in `this.options.MaxConcurrentActivityFunctions = 10;`. The reason is less for performance, and more for consistency. Currently, consumption customers could land on a small worker, where this value would be 1*10, a medium worker, where this value would be 2*10, or a large worker, where this value would be 4*10.

Hardcoding it to 10 for consumption apps eliminates this inconsistency, which matters more than absolute performance I would say.

davidmrdavid:
Just confirmed that the current approach **does not** override user-provided defaults on `host.json`. Instead, it just serves as new defaults when the corresponding `host.json` value isn't provided.

I did this by inspecting the values in the `settings` variable found in the snippet below under different `host.json` configs.

https://github.com/Azure/azure-functions-durable-extension/blob/3bdd5704f8ac41b4c549d6e0d811a6428d9c4c1f/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs#L116-L157

davidmrdavid:
I also just updated the numbers for benchmark D for PRv2 and PRv3 above ^. Before this, they were TBD.

davidmrdavid:
**Performance data for PRv4**
In this experiment, we keep the same config as in PRv3 but increase the control buffer size to be

```C#
 this.azureStorageOptions.ControlQueueBufferThreshold = 96; // was 32 in PRv1, PRv2, and PRv3
```

_Note:_ The Duration column is calculated as the time difference between the orchestrator's `FunctionStarting` to `FunctionCompleted` events.

_Time to completion_

| Benchmark ID | Description | Duration |
| ------------ |------------ | -------- |
|  A | Fanning over 1k activities | 11s |
|  B | Fanning over 10k activities | ~6min |
|  C | Fanning over 100 sub-orchestrators w/ 100 sequential activities each | ~3min  |
|  D | Fanning over 100 sub-orchestrators fanning over 10k activities each | ~3h 12 min |

_InstanceIDs to review raw data_

| Benchmark ID | InstanceID | 
| ------------ |------------ | 
|  A | 6bd8a138a1d647e8b31d077faeb8b696 |
|  B | d0e223b60a584179b450fef8ba9c4662 | 
|  C | 6004a761ca4044a99102cae87c8f5d8c|
|  D | e07a9f848db443d88269a59db40c0731|

Comments: This seems comparable to the baseline in benchmarks A and B and C, but its 4 times faster than the baseline for benchmark D: a drop from 12 hours to just 3!. It is also faster in benchmark D for PRv1, PRv2, and PRv3, which average about 8 hours. Finally, this version is also faster than PRv1, PRv2, and PRv3 on benchmark B by a factor of 3. I think this is our best configuration so far.

ConnorMcMahon:
We may also need to update our documentation regarding defaults.

That may be tricky if we soon end up with multiple variable playing into these defaults. Maybe we don't document the default values, and say we make a best effort based on platform/language to select intelligent defaults? @cgillum, thoughts about how we document this?

cgillum:
> In this experiment, we keep the same config as in PRv3 but increase the control buffer size to be 96

I worry about this change for languages like Python. If all Python threads are blocked waiting for long-running executions to complete, we'll be sitting on these buffered messages and the dequeue counts will increase because of the expirations. It's also not possible for these messages to be load balanced to other instances. It would be good to understand why this change had such an impact to see whether it was a coincidence or whether it really is an impactful change.

> Maybe we don't document the default values, and say we make a best effort based on platform/language to select intelligent defaults?

I would agree if the platform were able to adjust per-instance concurrency dynamically, but that's just not the case today. I worry that if we don't document these values then we'll be in trouble because of how critically important this information is to blocking Python workloads.

davidmrdavid:
I'm already working on a documentation update that will show the alternative defaults in this page: https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-bindings#durable-functions-1x

As for @cgillum's concern over PRv4's larger control-queue buffer size, I have two comments. First, the current global default is of 256 so irrespective of us choosing 32 (PRv1, PRv2, PRv3) or 96 (PRv4), we'd be improving the status quo for python one way of another. I'm happy to go with 32 just for safety's sake, and we can revisit this configuration once we perform OOProc-specific experiments. I just wanted to post PRv4's results anyways since they were so dramatic 😄

davidmrdavid:
@ConnorMcMahon,
with respect to docs: I also think we will need a restructuring of our docs once we have PL-aware defaults. However, for this PR alone, I don't think that's necessary.

As for whether OOProc performance tuning is in the scope of 2.4.2: _I actually remember discussing having OOProc defaults not be a part of 2.4.2._  Otherwise I would have already opened a PR for that already. If you feel strongly about having this in the next release, I would need to have until next Wednesday to reasonably test PL-aware defaults 😄 . That being said, I would prefer `to give OOProc performance tuning enough time to do a proper exploration of it.

ConnorMcMahon:
Probably a miscommunication on my part. In that case, let's push it to 2.5.0, but go with the safer defaults for out-of-proc (i.e. 32 instead of 96).

cgillum:
> the current global default is of 256 so irrespective of us choosing 32 (PRv1, PRv2, PRv3) or 96 (PRv4), we'd be improving the status quo for python one way of another by lowering the current default size.

Ah, good point! I forgot that we were defaulting to 256 (for some reason I mistakenly thought the default was already 32)! As long as we're improving the status quo, I have no objections. :)

src/WebJobs.Extensions.DurableTask/Options/AzureStorageOptions.cs
```diff
@@ -222,6 +222,11 @@ public void Validate()
             {
                 throw new InvalidOperationException($"{nameof(this.ControlQueueBufferThreshold)} must be between 1 and 1000.");
             }
+
+            if (this.ControlQueueBatchSize > this.ControlQueueBufferThreshold)
+            {
+                throw new InvalidOperationException($"{nameof(this.ControlQueueBatchSize)} cannot be larger than {nameof(this.ControlQueueBufferThreshold)}");
```
ConnorMcMahon, line 228: Not sure if we should explicitly throw an error because that may break customers.

I would say we implicitly change it, or we allow it and log a warning (or change it and log a warning). I don't have a strong preference for which approach we take.

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
```diff
@@ -27,12 +28,26 @@ internal class AzureStorageDurabilityProviderFactory : IDurabilityProviderFactor
...
             INameResolver nameResolver,
-            ILoggerFactory loggerFactory)
+            ILoggerFactory loggerFactory,
+            IPlatformInformationService platformInfo)
         {
             this.options = options.Value;
             this.nameResolver = nameResolver;
             this.loggerFactory = loggerFactory;
             this.azureStorageOptions = new AzureStorageOptions();
+            this.inConsumption = platformInfo.InConsumption();
```
ConnorMcMahon, line 38: I think if we moved this logic down into the GetAzureStorageOrchestrationServiceSettings, it may make certain things cleaner (for instance, we wouldn't need to check for the nullability of this.options.MaxConuccrent*Functions because we know they are explicitly not null.

It is also just nice to have all of the logic regarding settings in one place.

That being said, that may delay the validation logic, so consider this a soft suggestion.

src/WebJobs.Extensions.DurableTask/IPlatformInformationService.cs
ConnorMcMahon, line 13: Part of me wants to leave an [Obsolete] attribute on this, with an explanation that this is meant for internal purposes only and may change in the near future. Considering we are going to be adding methods at the very least, if not fully redesigning this interface in a subsequent release, I don't want customers to take a dependency on this via dependency injection.

Even once this class is finalized, I'm not convinced we want customers to inject their own value for this ever?

ConnorMcMahon (commented):
A few small suggestions and then I think we are good to go here.

src/WebJobs.Extensions.DurableTask/Options/AzureStorageOptions.cs
davidmrdavid: Hmm, I'm conflicted here as well. My personal preference is to fail fast because otherwise we can cause confusion, especially if they're trying to do some performance tuning themselves.

That being said, I'm ok with logging a warning and leaving their config as is. How do you feel about including a blurb in the warning about DF, in the next major release, potentially dropping "support" for misconfigurations like this one?

ConnorMcMahon: In general, fail fast is a good policy for bad configurations, but when they are more awkward than actively harmful, and we have allowed them for a long time, I consider that a breaking change, which I am hesitant to take on, especially in  patch release.

src/WebJobs.Extensions.DurableTask/IPlatformInformationService.cs
davidmrdavid: I agree. I actually completely forgot that this was public. Happy to add `[Obsolete]` as the design is still WIP and also we never want customers to add DI here.

Would making this `internal` help prevent customer DI?

src/WebJobs.Extensions.DurableTask/Options/AzureStorageOptions.cs
ConnorMcMahon: Let's go with the log a warning approach.

src/WebJobs.Extensions.DurableTask/IPlatformInformationService.cs
ConnorMcMahon: It would, but there are issues making these internal, because the constructor of DurableTaskExtension has to be public I believe (not sure if this is a Functions extension limitation or a DI restriction). And you can't put an internal type on a public constructor :(.

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
davidmrdavid: I'm a little concerned about delaying the validation step. I do want to follow-up on this suggestion, but I'll defer making a decision on it just yet because I want to test what happens when we move things around. There'll be another opportunity with the OOProc PR :)

ConnorMcMahon (approved):
LGTM!

PR #1710, opened 2021-03-05
Adding event source name property to DurabilityProvider

bachuv:
These changes add an event source name property to `DurabilityProvider` and remove the hard coded event source names for different storage backends in `OnEventSourceCreated()`.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.innerService = service ?? throw new ArgumentNullException(nameof(service));
             this.innerServiceClient = serviceClient ?? throw new ArgumentNullException(nameof(serviceClient));
             this.connectionName = connectionName ?? throw new ArgumentNullException(connectionName);
-            this.name = eventSourceName ?? throw new ArgumentNullException(nameof(eventSourceName));
+            this.eventSourceName = eventSourceName ?? throw new ArgumentNullException(nameof(eventSourceName));
         }
 
         /// <summary>
--- a/src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
@@ -28,7 +28,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         private readonly IOrchestrationService innerService;
         private readonly IOrchestrationServiceClient innerServiceClient;
         private readonly string connectionName;
-        private readonly string eventSourceName;
 
         /// <summary>
         /// Creates the default <see cref="DurabilityProvider"/>.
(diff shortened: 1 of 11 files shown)


test/Common/DurableClientBaseTests.cs
```diff
@@ -39,7 +39,7 @@ public async Task StartNewAsync_InvalidInstanceId_ThrowsException(string instanc
         {
             var orchestrationServiceClientMock = new Mock<IOrchestrationServiceClient>();
             orchestrationServiceClientMock.Setup(x => x.GetOrchestrationStateAsync(It.IsAny<string>(), It.IsAny<bool>())).ReturnsAsync(GetInvalidInstanceState());
-            var storageProvider = new DurabilityProvider("test", new Mock<IOrchestrationService>().Object, orchestrationServiceClientMock.Object, "test");
+            var storageProvider = new DurabilityProvider("test", new Mock<IOrchestrationService>().Object, orchestrationServiceClientMock.Object, "test", "test");
```
davidmrdavid, line 42: Could we differentiate between these two "test" arguments? Just a different name like "testDurabilityProvider" should do

test/Common/DurableClientBaseTests.cs
```diff
@@ -214,7 +214,7 @@ public void DurableClient_ExternalApp_CreateHttpManagementPayload_ThrowsExceptio
 
         private IDurableOrchestrationClient GetDurableClient(IOrchestrationServiceClient orchestrationServiceClientMockObject)
         {
-            var storageProvider = new DurabilityProvider("test", new Mock<IOrchestrationService>().Object, orchestrationServiceClientMockObject, "test");
+            var storageProvider = new DurabilityProvider("test", new Mock<IOrchestrationService>().Object, orchestrationServiceClientMockObject, "test", "test");
```
davidmrdavid, line 217: I realize this was here from before, but this line seems to be repeated, without modification, in a lot of tests. I wonder if we should track some refactoring item to abstract this repeated scaffolding logic

davidmrdavid (approved):
It looks good to me, left some non-blocking comments. Would love to hear other's opinions before we merge though!

test/Common/DurableClientBaseTests.cs
bachuv: I was following the other two "test" arguments to keep the test string consistent.

src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
```diff
@@ -38,12 +39,14 @@ public class DurabilityProvider : IOrchestrationService, IOrchestrationServiceCl
...
         /// <param name="connectionName">The name of the app setting that stores connection details for the storage provider.</param>
-        public DurabilityProvider(string storageProviderName, IOrchestrationService service, IOrchestrationServiceClient serviceClient, string connectionName)
+        /// <param name="eventSourceName">The name of the event source provider.</param>
+        public DurabilityProvider(string storageProviderName, IOrchestrationService service, IOrchestrationServiceClient serviceClient, string connectionName, string eventSourceName)
         {
             this.name = storageProviderName ?? throw new ArgumentNullException(nameof(storageProviderName));
             this.innerService = service ?? throw new ArgumentNullException(nameof(service));
             this.innerServiceClient = serviceClient ?? throw new ArgumentNullException(nameof(serviceClient));
             this.connectionName = connectionName ?? throw new ArgumentNullException(connectionName);
+            this.eventSourceName = eventSourceName ?? throw new ArgumentNullException(nameof(eventSourceName));
```
ConnorMcMahon, line 49: Not sure this needs to be a required field.

src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
```diff
@@ -38,12 +39,14 @@ public class DurabilityProvider : IOrchestrationService, IOrchestrationServiceCl
         /// <param name="serviceClient">The internal <see cref="IOrchestrationServiceClient"/> that provides functionality
         /// for this classes implementions of <see cref="IOrchestrationServiceClient"/>.</param>
         /// <param name="connectionName">The name of the app setting that stores connection details for the storage provider.</param>
-        public DurabilityProvider(string storageProviderName, IOrchestrationService service, IOrchestrationServiceClient serviceClient, string connectionName)
+        /// <param name="eventSourceName">The name of the event source provider.</param>
+        public DurabilityProvider(string storageProviderName, IOrchestrationService service, IOrchestrationServiceClient serviceClient, string connectionName, string eventSourceName)
```
ConnorMcMahon, line 43: Instead of feeding this in as a constructor parameter, maybe we should just make EventSourceName a virtual property that subclasses can override.

src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
```diff
@@ -82,6 +85,11 @@ public DurabilityProvider(string storageProviderName, IOrchestrationService serv
         /// </summary>
         public virtual TimeSpan LongRunningTimerIntervalLength { get; set; }
 
+        /// <summary>
+        /// Event source name (e.g. DurableTask-AzureStorage).
+        /// </summary>
+        public string EventSourceName => this.eventSourceName;
```
ConnorMcMahon, line 91: Not sure if this is overoptimizing for a situation that might not happen, but maybe instead of a single EventSourcName, what if we supported an IEnumerable<string> of event source names. I don't think any providers currently need more than one, but it's theoretically possible in the future?

Not sold on needing this change.

ConnorMcMahon (commented):
Overall looks pretty good, just a few nitpicks/suggestions.

ConnorMcMahon (approved):
LGTM!

PR #1716, opened 2021-03-09
Ensure using latest webhook url for HTTP management APIs

ConnorMcMahon:

Previously, we store the value of the webhook URL during extension
initialization with the assumption that this webhook URL never changes
after startup. This stored value is used for various IDurableClient
methods, along with the out-of-proc clients when local RPC endpoints are
disabled.

However, it turns out that in the case of slot-swap operations,
WEBSITE_HOSTNAME uses the staging slot hostname at site startup, and
only after the slot swap does the hostname change. The function host
listens for a new header value sent as a part of some ping, updating the
source of ExtensionConfigContext.GetWebhookUrl() with the new value once
the slot swap is complete. This PR makes sure we always fetch the value
from the ExtensionConfigContext instead of caching it to ensure we have
the most up-to-date value.

### Pull request checklist

    * [x] Otherwise: Backport tracked by issue/PR #1717

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -106,6 +106,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// <param name="hostLifetimeService">The host shutdown notification service for detecting and reacting to host shutdowns.</param>
         /// <param name="lifeCycleNotificationHelper">The lifecycle notification helper used for custom orchestration tracking.</param>
         /// <param name="messageSerializerSettingsFactory">The factory used to create <see cref="JsonSerializerSettings"/> for message settings.</param>
+        /// <param name="webhookProvider">A delegate to fetch the webhook URL.</param>
         /// <param name="errorSerializerSettingsFactory">The factory used to create <see cref="JsonSerializerSettings"/> for error settings.</param>
         /// <param name="telemetryActivator">The activator of DistributedTracing. .netstandard2.0 only.</param>
 #pragma warning restore CS1572
@@ -265,13 +266,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             // Throw if any of the configured options are invalid
             this.Options.Validate(this.nameResolver, this.TraceHelper);
 
-            // For 202 support
-            if (this.Options.NotificationUrl == null)
-            {
 #pragma warning disable CS0618 // Type or member is obsolete
-                this.Options.NotificationUrl = context.GetWebhookHandler();
-#pragma warning restore CS0618 // Type or member is obsolete
+
+            // Invoke webhook handler to make functions runtime register extension endpoints.
+            context.GetWebhookHandler();
+
+            if (this.Options.TestWebhookUri == null)
+            {
+                this.HttpApiHandler.RegisterWebhookProvider(() => context.GetWebhookHandler());
             }
+#pragma warning restore CS0618 // Type or member is obsolete
 
             this.TraceConfigurationSettings();
 
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -273,6 +273,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
             if (this.Options.TestWebhookUri == null)
             {
+                // This line ensure every time we need the webhook URI, we get it directly from the
+                // function runtime, which has the most up-to-date knowledge about the site hostname.
                 this.HttpApiHandler.RegisterWebhookProvider(() => context.GetWebhookHandler());
             }
 #pragma warning restore CS0618 // Type or member is obsolete
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -319,7 +319,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             // function runtime, which has the most up-to-date knowledge about the site hostname.
             this.HttpApiHandler.RegisterWebhookProvider(
                 this.Options.WebhookUriProviderOverride ??
-                (() => context.GetWebhookHandler()));
+                context.GetWebhookHandler);
 #pragma warning restore CS0618 // Type or member is obsolete
 
             this.TraceConfigurationSettings();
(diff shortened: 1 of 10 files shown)


src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -210,6 +199,10 @@ public string HubName
         // Used for mocking the lifecycle notification helper.
         internal HttpMessageHandler NotificationHandler { get; set; }
 
+        // This is purely a way for tests to overwrite the webhook url, since there is no easyway
+        // to mock the value from ExtensionConfigContext.
```
davidmrdavid, line 203: minor nit:
```suggestion
        // This is just a way for tests to overwrite the webhook url, since there is no easy way
        // to mock the value from ExtensionConfigContext.
```

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -264,13 +275,17 @@ public async Task<HttpResponseMessage> HandleRequestAsync(HttpRequestMessage req
...
                 {
-                    throw new InvalidOperationException($"Don't know how to handle request to {request.RequestUri}.");
+                    Uri webhookUrl = this.GetWebhookUri();
+                    if (webhookUrl != null)
+                    {
+                        basePath = webhookUrl.AbsolutePath;
+                    }
+                    else
+                    {
+                        throw new InvalidOperationException($"Don't know how to handle request to {request.RequestUri}.");
```
davidmrdavid, line 287: Could we provide a more descriptive error? For example, isn't the problem here that the webhook isn't available?
Additionally, since GetWebhookUri throws an exception, would we ever reach this else statement?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -265,13 +266,16 @@ void IExtensionConfigProvider.Initialize(ExtensionConfigContext context)
...
 #pragma warning disable CS0618 // Type or member is obsolete
-                this.Options.NotificationUrl = context.GetWebhookHandler();
-#pragma warning restore CS0618 // Type or member is obsolete
+
+            // Invoke webhook handler to make functions runtime register extension endpoints.
+            context.GetWebhookHandler();
+
+            if (this.Options.TestWebhookUri == null)
+            {
+                this.HttpApiHandler.RegisterWebhookProvider(() => context.GetWebhookHandler());
```
davidmrdavid, line 276: This line does quite a lot of heavy lifting: by registering the provider like this, we make sure `context.WebhookHandler` gets called every time we need to URI. Since this line is so crucial, can we comment it with a note on what this achieves?

test/Common/HttpApiHandlerTests.cs
```diff
@@ -62,6 +62,45 @@ public async Task WaitForCompletionOrCreateCheckStatusResponseAsync_Throws_Excep
...
+            httpApiHandler.RegisterWebhookProvider(() => webhookProvider.WebhookUri);
+
+            AssertApisUsingCorrectWebhookUri(httpApiHandler, TestConstants.NotificationUrlBase);
+
+            string newWebhookUri = TestConstants.NotificationUrl.Replace("localhost:7071", "localhost:5050");
+            string newBaseUri = TestConstants.NotificationUrlBase.Replace("localhost:7071", "localhost:5050");
+            webhookProvider.WebhookUri = new Uri(newWebhookUri);
+
+            AssertApisUsingCorrectWebhookUri(httpApiHandler, newBaseUri);
+        }
```
davidmrdavid, line 80: nit: The double use of "Update" throws me off here. Can we rename it?
Maybe change the 2nd "Update" with "New"

test/Common/HttpApiHandlerTests.cs
```diff
@@ -62,6 +62,45 @@ public async Task WaitForCompletionOrCreateCheckStatusResponseAsync_Throws_Excep
...
+            Assert.StartsWith(expectedBaseUri, managementPayload.TerminatePostUri);
+
+            string baseUri = httpApiHandler.GetBaseUrl();
+            Assert.Equal(expectedBaseUri, baseUri);
+
+            HttpCreationPayload creationPayload = httpApiHandler.GetInstanceCreationLinks();
+            Assert.StartsWith(expectedBaseUri, creationPayload.CreateNewInstancePostUri);
+            Assert.StartsWith(expectedBaseUri, creationPayload.CreateAndWaitOnNewInstancePostUri);
+        }
+
```
davidmrdavid, line 103: Is this actually testing the behavior introduced in this PR or is this just a missing test? I'm asking because I don't see any URI updating in this test.

davidmrdavid (approved):
Overall, it looks good to me. Left some suggestions, mostly nits, and a question

test/Common/HttpApiHandlerTests.cs
ConnorMcMahon: This is a helper method for the test that updates the URI, immediately above this method.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
ConnorMcMahon: You are right that it is no longer necessary, since GetWebhookUri already throws by this point.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -210,6 +199,10 @@ public string HubName
         // Used for mocking the lifecycle notification helper.
         internal HttpMessageHandler NotificationHandler { get; set; }
 
+        // This is just a way for tests to overwrite the webhook url, since there is no easy way
+        // to mock the value from ExtensionConfigContext. It should not be used in production code paths.
+        internal Uri TestWebhookUri { get; set; }
```
cgillum, line 204: It feels weird to have something named "Test" in non-test code.  I wonder if we can fix the name and make this a bit more extensible by doing something like this:

 ```suggestion
        public Func<Uri> WebhookUriProviderOverride { get; set; }
```

Then, in DurableTaskExtension we could do something like this:

```csharp
this.HttpApiHandler.RegisterWebhookProvider(
    this.Options.WebhookUriProviderOverride ?? context.GetWebhookHandler);
```

ConnorMcMahon:
I have linked #1717 to backport this fix to Durable 1.x, as this may impact out-of-proc customers at a higher rate.

I also need to test this end-to-end in prod before merging.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -83,6 +84,11 @@ internal class HttpApiHandler : IDisposable
             this.durableTaskOptions = durableTaskOptions;
             this.traceHelper = traceHelper;
 
+            if (this.durableTaskOptions.WebhookUriProviderOverride != null)
```
cgillum, line 87: nit: Seems like this null check doesn't serve any purpose?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -265,13 +265,17 @@ void IExtensionConfigProvider.Initialize(ExtensionConfigContext context)
...
-                this.Options.NotificationUrl = context.GetWebhookHandler();
+
+            // Invoke webhook handler to make functions runtime register extension endpoints.
+            context.GetWebhookHandler();
+
+            // This line ensure every time we need the webhook URI, we get it directly from the
+            // function runtime, which has the most up-to-date knowledge about the site hostname.
+            this.HttpApiHandler.RegisterWebhookProvider(
+                this.Options.WebhookUriProviderOverride ??
+                (() => context.GetWebhookHandler()));
```
cgillum, line 277: Can this expression be simplified to this?
```suggestion
                context.GetWebhookHandler);
```

cgillum (approved):

PR #1718, opened 2021-03-10
Handling HttpRequestException in CallHttpAsync()

### LINKED ISSUE #1669: CallHttpAsync() throws a serialization exception if the target endpoint doesn't exist
### Description
@robs reported this issue in a comment of #697.

We will need to investigate/reproduce this a bit more to flesh out this issue.

### Expected behavior
An HTTPException (or some other reasonable exception) should be thrown.

### Actual behavior
An exception referencing exception serialization is thrown.

bachuv:
These changes add a way to handle when  `this.httpClient.SendAsync()` throws an `HttpRequestException` and propagate the exception to customer code. Before this change, the extension would throw an exception about serialization instead of `HttpRequestException`. By handling the `HttpRequestException` in `TaskHttpActivityShim`, an `HttpRequestException` is thrown with a descriptive message.

Resolves #1669

### Pull request checklist

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -1,2 +1,5 @@
 ## Bug fixes
 - Fix CallHttpAsync() to throw an HttpRequestException instead of a serialization exception if the target endpoint doesn't exist (#1718)
+
+## Breaking changes
+- Fix CallHttpAsync() to throw an HttpRequestException instead of a serialization exception if the target endpoint doesn't exist (#1718). This is a breaking change if you were handling `HttpRequestException`s by catching `FunctionFailedException`s
--- a/release_notes.md
+++ b/release_notes.md
@@ -1,1 +1,2 @@
-
+## Bug fixes
+- Fix CallHttpAsync() to throw an HttpRequestException instead of a serialization exception if the target endpoint doesn't exist (#1718)
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -690,8 +690,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             {
                 // Check to see if CallHttpAsync() threw a TimeoutException
                 // In this case, we want to throw a TimeoutException instead of a FunctionFailedException
-                if (functionName.Equals(HttpOptions.HttpTaskActivityReservedName) && e.InnerException is TimeoutException)
+                if (functionName.Equals(HttpOptions.HttpTaskActivityReservedName) &&
+                    (e.InnerException is TimeoutException || e.InnerException is HttpRequestException))
                 {
+                    if (e.InnerException is HttpRequestException)
+                    {
+                        throw new HttpRequestException(e.Message);
+                    }
+
                     throw e.InnerException;
                 }
 
--- a/src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
@@ -42,32 +42,40 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             HttpRequestMessage requestMessage = await this.ConvertToHttpRequestMessage(durableHttpRequest);
 
             HttpResponseMessage response;
-            if (durableHttpRequest.Timeout == null)
+            try
             {
-                response = await this.httpClient.SendAsync(requestMessage);
-            }
-            else
-            {
-                try
+                if (durableHttpRequest.Timeout == null)
                 {
-                    using (CancellationTokenSource cts = new CancellationTokenSource())
+                    response = await this.httpClient.SendAsync(requestMessage);
+                }
+                else
+                {
+                    try
                     {
-                        cts.CancelAfter(durableHttpRequest.Timeout.Value);
-                        response = await this.httpClient.SendAsync(requestMessage, cts.Token);
+                        using (CancellationTokenSource cts = new CancellationTokenSource())
+                        {
+                            cts.CancelAfter(durableHttpRequest.Timeout.Value);
+                            response = await this.httpClient.SendAsync(requestMessage, cts.Token);
+                        }
+                    }
+                    catch (OperationCanceledException ex)
+                    {
+                        TimeoutException e = new TimeoutException(ex.Message + $" Reached user specified timeout: {durableHttpRequest.Timeout.Value}.");
+
+                        string details = Utils.SerializeCause(e, this.config.ErrorDataConverter);
+                        throw new TaskFailureException(e.Message, e, details);
                     }
                 }
-                catch (OperationCanceledException ex)
-                {
-                    TimeoutException e = new TimeoutException(ex.Message + $" Reached user specified timeout: {durableHttpRequest.Timeout.Value}.");
 
-                    string details = Utils.SerializeCause(e, this.config.ErrorDataConverter);
-                    throw new TaskFailureException(e.Message, e, details);
-                }
+                DurableHttpResponse durableHttpResponse = await DurableHttpResponse.CreateDurableHttpResponseWithHttpResponseMessage(response);
+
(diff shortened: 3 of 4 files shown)


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -690,8 +690,14 @@ string IDurableOrchestrationContext.StartNewOrchestration(string functionName, o
             {
                 // Check to see if CallHttpAsync() threw a TimeoutException
                 // In this case, we want to throw a TimeoutException instead of a FunctionFailedException
-                if (functionName.Equals(HttpOptions.HttpTaskActivityReservedName) && e.InnerException is TimeoutException)
+                if (functionName.Equals(HttpOptions.HttpTaskActivityReservedName) &&
+                    (e.InnerException is TimeoutException || e.InnerException is HttpRequestException))
                 {
+                    if (e.InnerException is HttpRequestException)
+                    {
+                        throw new HttpRequestException(e.Message);
```
bachuv, line 698: I tried to follow the same pattern as the `TimeoutException` but the `Message` property in `HttpRequestException` doesn't serialize the same way and the value doesn't get passed. That's why I chose to create a new `HttpRequestException` in this method instead of in `RunAsync()` in TaskHttpActivityShim.cs.

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -42,32 +42,40 @@ public async override Task<string> RunAsync(TaskContext context, string rawInput
...
-                    TimeoutException e = new TimeoutException(ex.Message + $" Reached user specified timeout: {durableHttpRequest.Timeout.Value}.");
+                    try
+                    {
+                        using (CancellationTokenSource cts = new CancellationTokenSource())
+                        {
+                            cts.CancelAfter(durableHttpRequest.Timeout.Value);
+                            response = await this.httpClient.SendAsync(requestMessage, cts.Token);
+                        }
+                    }
+                    catch (OperationCanceledException ex)
```
cgillum, line 61: Can we just do the `catch (HttpRequestException)` here at the same level as `catch (OperationCanceledException)` instead of putting it into a new try/catch block?

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
```diff
@@ -42,32 +42,40 @@ public async override Task<string> RunAsync(TaskContext context, string rawInput
...
 
-            DurableHttpResponse durableHttpResponse = await DurableHttpResponse.CreateDurableHttpResponseWithHttpResponseMessage(response);
+                DurableHttpResponse durableHttpResponse = await DurableHttpResponse.CreateDurableHttpResponseWithHttpResponseMessage(response);
 
-            return JsonConvert.SerializeObject(durableHttpResponse);
+                return JsonConvert.SerializeObject(durableHttpResponse);
+            }
+            catch (HttpRequestException ex)
+            {
+                string details = Utils.SerializeCause(ex, this.config.ErrorDataConverter);
```
cgillum, line 76: You mentioned in the description that `HttpRequestException` doesn't serialize correctly. How is it that we don't run into the same problem here?

cgillum (commented):
A couple questions:

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
bachuv: The `catch (OperationCanceledException)` is specifically for the case where `durableHttpRequest.Timeout` is set and the `catch (HttpRequestException)` is applicable to whether or not the Timeout property is set. The `catch(OperationCanceledException)` could move one level out to be consistent with `catch (HttpRequestException)`.

cgillum: I'm just wondering why you couldn't put `catch (HttpRequestException)` on line 68. That would be a smaller change and I don't think that exception can be thrown anywhere else.

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
bachuv: The details on line 76 are correctly serialized with the right information, but the `TaskFailureException` that's created doesn't serialize the `Message` property for `ex`, the InnerException.

For example, on line 77, `ex.Message` is "No such host is known.", but then in `StartNewOrchestration` in `DurableOrchestrationContext`, the `e.InnerException.Message` turns out to be "'System.Net.Http.HttpRequestException' was thrown."

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
davidmrdavid: Interesting! Can you clarify how is the serialization different?

davidmrdavid: ah, I see you clarify this in a later comment. Nevermind

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
davidmrdavid: Should we track a fix to that serialization problem?

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
bachuv: I don't think that would cover the `await this.httpClient.SendAsync(requestMessage)` call on line 49 if it throws an HttpRequestException.

test/Common/DurableHttpTests.cs
```diff
@@ -1509,6 +1547,21 @@ private static HttpMessageHandler MockSynchronousHttpMessageHandlerWithTimeoutEx
             return handlerMock.Object;
         }
 
+        private static HttpMessageHandler MockSynchronousHttpMessageHandlerWithHttpRequestException()
+        {
+            HttpResponseMessage httpResponseMessage = CreateTestHttpResponseMessage(HttpStatusCode.OK);
```
davidmrdavid, line 1552: Is there a reason why we reply status-code "OK" even though the message failed? I realize this is just a test, but it did throw me off :)

davidmrdavid (commented):
Looks pretty good, but I would like for us to reach consensus on the try-catch re-arrangement suggestions before merge :)

src/WebJobs.Extensions.DurableTask/Listener/TaskHttpActivityShim.cs
cgillum: Oh! I totally missed that other send on line 49. What you have makes sense then. However, it makes me wonder if we should consider also handing `OperationCanceledException` in the outer catch block as well. We might not handle it exactly the same since there isn't a user-specified timeout in the case of line 49, but it might be a good idea so that there is a consistent experience for HTTP timeouts, whether it's the default timeout or a user-provided timeout.

test/Common/DurableHttpTests.cs
bachuv: There's no particular reason, it's just there as a dummy value to create a `HttpResponseMessage`. The important part of this test is on line 1554 where the content stores an `HttpRequestException`. I do see how it can be confusing though so I can change it to `NotFound`.

release_notes.md
```diff
@@ -1 +1,2 @@
-
+## Bug fixes
+- Fix CallHttpAsync() to throw an HttpRequestException instead of a serialization exception if the target endpoint doesn't exist (#1718)
```
cgillum, line 2: Should we also mention that this is a breaking change, since we're changing the type of exception being thrown?

cgillum (approved):

ConnorMcMahon (approved):
LGTM!

release_notes.md
bachuv: That's a good point! I'll add it to the release notes.

PR #1719, opened 2021-03-11
Fix Local RPC race condition on app startup

### LINKED ISSUE #1273: Local HTTP listener causing startup issues

ConnorMcMahon:
Our previous attempt to dynamically select ports for the local RPC endpoint fixed many issues customers had, but there still exists a race condition when multiple hosts start up at the same time. To completely resolve this issue, this PR makes the following changes:

- Move the code that finds a free port closer to the code that attempts to listen on that port.
- Add a retry layer if the free port has been taken by another worker between when we found the free port and when we tried to listen to that port.
- If we somehow hit this race condition 3 times in a row, even though it should be more rare now, we gracefully fallback to using the public endpoints for out-of-process clients instead of failing to start up the host.

resolves #1273

### Pull request checklist

    * [X] Otherwise: I've added my notes to `release_notes.md`
    * [X] Otherwise: Backport tracked by issue/PR #1723
* [X] I have added all required tests (Unit tests, E2E tests)

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
+++ b/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
@@ -1140,15 +1140,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         internal bool TryGetRpcBaseUrl(out Uri rpcBaseUrl)
         {
-            if (this.durableTaskOptions.LocalRpcEndpointEnabled != false
-                && this.localHttpListener.IsListening)
+            if (this.durableTaskOptions.LocalRpcEndpointEnabled != false)
             {
                 rpcBaseUrl = this.localHttpListener.InternalRpcUri;
                 return true;
             }
 
-            // The app owner explicitly disabled the local RPC endpoint, or we were unable to
-            // successfully find a port after several tries.
+            // The app owner explicitly disabled the local RPC endpoint
             rpcBaseUrl = null;
             return false;
         }
--- a/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
+++ b/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
@@ -1140,13 +1140,15 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         internal bool TryGetRpcBaseUrl(out Uri rpcBaseUrl)
         {
-            if (this.durableTaskOptions.LocalRpcEndpointEnabled != false)
+            if (this.durableTaskOptions.LocalRpcEndpointEnabled != false
+                && this.localHttpListener.IsListening)
             {
                 rpcBaseUrl = this.localHttpListener.InternalRpcUri;
                 return true;
             }
 
-            // The app owner explicitly disabled the local RPC endpoint.
+            // The app owner explicitly disabled the local RPC endpoint, or we were unable to
+            // successfully find a port after several tries.
             rpcBaseUrl = null;
             return false;
         }
@@ -1156,14 +1158,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             if (!this.localHttpListener.IsListening)
             {
+                await this.localHttpListener.StartAsync();
+
                 this.traceHelper.ExtensionInformationalEvent(
                     this.durableTaskOptions.HubName,
                     instanceId: string.Empty,
                     functionName: string.Empty,
-                    message: $"Opening local RPC endpoint: {this.localHttpListener.InternalRpcUri}",
+                    message: $"Opened local RPC endpoint: {this.localHttpListener.InternalRpcUri}",
                     writeToUserLogs: true);
-
-                await this.localHttpListener.StartAsync();
             }
         }
 
--- a/src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
@@ -22,10 +22,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
     {
         private const int DefaultPort = 17071;
 
-        private readonly IWebHost localWebHost;
         private readonly Func<HttpRequestMessage, Task<HttpResponseMessage>> handler;
         private readonly EndToEndTraceHelper traceHelper;
         private readonly DurableTaskOptions durableTaskOptions;
+        private IWebHost localWebHost;
 
         public LocalHttpListener(
             EndToEndTraceHelper traceHelper,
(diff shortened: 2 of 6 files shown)


src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -1140,13 +1140,15 @@ private void ThrowIfWebhooksNotConfigured()
 
         internal bool TryGetRpcBaseUrl(out Uri rpcBaseUrl)
         {
-            if (this.durableTaskOptions.LocalRpcEndpointEnabled != false)
+            if (this.durableTaskOptions.LocalRpcEndpointEnabled != false
+                && this.localHttpListener.IsListening)
```
davidmrdavid, line 1144: nit: I would prefer if we didn't introduce the double negative with "!= false"
```suggestion
            if (this.durableTaskOptions.LocalRpcEndpointEnabled
                && this.localHttpListener.IsListening)

```

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -1140,13 +1140,15 @@ private void ThrowIfWebhooksNotConfigured()
...
+            if (this.durableTaskOptions.LocalRpcEndpointEnabled != false
+                && this.localHttpListener.IsListening)
             {
                 rpcBaseUrl = this.localHttpListener.InternalRpcUri;
                 return true;
             }
 
-            // The app owner explicitly disabled the local RPC endpoint.
+            // The app owner explicitly disabled the local RPC endpoint, or we were unable to
+            // successfully find a port after several tries.
```
davidmrdavid, line 1151: I don't see the "several tries" logic in here. If it's elsewhere, could we document that?

src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
```diff
@@ -66,8 +55,35 @@ public async Task StartAsync()
...
+                        .UseKestrel()
+                        .UseUrls(listenUri.OriginalString)
+                        .Configure(a => a.Run(this.HandleRequestAsync))
+                        .Build();
+
+                    await this.localWebHost.StartAsync();
+                    this.IsListening = true;
+                    break;
+                }
+                catch
```
davidmrdavid, line 76: Should we try to catch a more specific exception?

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -4796,35 +4796,43 @@ public async Task CallActivity_Like_From_Azure_Portal()
...
-            {
-                await host2.StartAsync();
-            }
-            catch (Exception)
-            {
-                Assert.True(false, "Could not start up two hosts on the same device in parallel");
-            }
-            finally
+            // This test wants to be sure there are no race conditions while starting up multiple hosts in parallel,
+            // so attempt various times to make sure to hit race condition if it exists.
```
davidmrdavid, line 4806: nit: replace "make sure" with "try" or "attempt"
```suggestion
            // so attempt various times to try to hit race condition if it exists.

```

davidmrdavid (approved):
Looks good to me, left some stylistic requests and some questions. Some comments were not associated with the review itself but please give those a look as well. Thanks!

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
cgillum: Unfortunately the two expressions are not equivalent because `LocalRpcEndpointEnabled` is not a bool, but a bool? (i.e. `Nullable<Boolean>`). In our case, `true`, `false`, and `null` all have different semantic meanings.

davidmrdavid: Ah, that makes sense. Thanks for clarifying!

src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
```diff
@@ -36,27 +36,16 @@ internal class LocalHttpListener : IDisposable
...
-#endif
         }
 
-        public Uri InternalRpcUri { get; }
+        public Uri InternalRpcUri { get; private set; }
 
         public bool IsListening { get; private set; }
 
-        public void Dispose() => this.localWebHost.Dispose();
+        public void Dispose() => this.localWebHost?.Dispose();
```
cgillum, line 48: It doesn't look like `this.localWebHost` can ever be since we assign it in the constructor. We also never explicitly assign it to null. Is this a leftover artifact of an earlier iteration?

src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
```diff
@@ -36,27 +36,16 @@ internal class LocalHttpListener : IDisposable
...
-                .UseKestrel()
-                .UseUrls(listenUri.OriginalString)
-                .Configure(a => a.Run(this.HandleRequestAsync))
-                .Build();
-#else
-            // Just use default port for internal Uri. No need to check for port availability since
-            // we won't be listening to this endpoint.
-            this.InternalRpcUri = new Uri($"http://127.0.0.1:{DefaultPort}/durabletask/");
+            // Set to a non null value
+            this.InternalRpcUri = new Uri($"http://127.0.0.1:{DefaultPort}/defaultvalue/");
```
cgillum, line 40: What's the reasoning behind changing "/durabletask/" to "/defaultvalue/" here?

If this is intended to be a dummy value, I would suggest we make that more explicit - e.g. http://uninitialized.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -4796,35 +4796,43 @@ public async Task CallActivity_Like_From_Azure_Portal()
...
+            int numAttempts = 5;
+            for (int attempt = 0; attempt < numAttempts; attempt++)
             {
-                await host1.StopAsync();
-                host1.Dispose();
-                await host2.StopAsync();
-                host2.Dispose();
+                int numHost = 10;
```
cgillum, line 4810: naming nit: consider `threadCount` as the name of this local variable. It's a better match for it's intended use.

src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
```diff
@@ -66,8 +55,35 @@ public async Task StartAsync()
...
+                {
+                    this.traceHelper.ExtensionWarningEvent(
+                        this.durableTaskOptions.HubName,
+                        functionName: string.Empty,
+                        instanceId: string.Empty,
+                        message: $"Failed to open local socket {availablePort}. This was attempt #{numAttempts} to open a local port.");
+                    numAttempts++;
+                }
+            }
+            while (numAttempts <= 3);
```
cgillum, line 86: I wonder if we should try more times, like 10. I also wonder if we should have a small sleep in-between retries. I'm not sure how random the port selection is, so it would be nice if we could mitigate the case where two hosts are starting up at the same time and selecting the same random sequence of available port numbers.

cgillum (commented):
Mostly nits. My only real concern is what to do if we fail to find a free port. I'm not super comfortable to falling back to public endpoints because that could put a customer into an unpredictable state - i.e. it works on some machines but not on others - and diagnosing such issues will be a lot harder. I'm also not sure what this means when someone is running outside of the Azure Functions hosted service (K8s, WebJobs, etc.) where they may not have any public HTTP endpoints. Throwing an exception might actually be safer.

src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
ConnorMcMahon: Indeed. Good catch.

src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
ConnorMcMahon: Makes sense. if we are worried about them being in sync, would doing a semi-random amount of sleep time help (i.e. between 50-100 ms)? That way we keep the two hosts out of lockstep.

ConnorMcMahon:
@cgillum, I am a bit afraid about just throwing an exception, because I don't know if that shuts down the whole host. I know some customers who hit the current race condition seem to hit to be stuck in a bad state. Would a failfast be better?

cgillum:
Oy...if an exception here doesn't restart the host then that would be really bad. I hate to resort to FailFast because that's a pretty severe hammer, but I guess I don't have any better ideas than this...

ConnorMcMahon:
I'll see if I can find an example of what happens when an exception is thrown. Overall, I expect this to happen incredibly rarely, especially if we up the retry count to 10 and add some randomized delays, as we have already made the race condition less likely by reducing the time window between port selection and listening on that port.

src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
ConnorMcMahon: Added a dynamic sleep between 0 and 1 seconds here. Let me know if that's what you had in mind.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
ConnorMcMahon: I just removed that, because now this code shouldn't fail if we failed to start up, since we are throwing an exception.

cgillum (approved):
LGTM!

PR #1721, opened 2021-03-11
Transition to JSON logging for the Linux Dedicated plan

davidmrdavid:
Previously, logs on Linux Dedicated were interfacing with the legacy regex endpoint. This meant that some of our logs required extra post-processing for them to behave as expected. With this change, we'll be writing to the new JSON endpoint which should _just work_.

_Also, this change will **need** to be backported_

N/A

### Pull request checklist

    * [x] Otherwise: Backport tracked by issue/PR #1733

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
+++ b/src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
@@ -22,81 +22,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
     {
         private const string ConsolePrefix = "MS_DURABLE_FUNCTION_EVENTS_LOGS";
 
-        // The ordered list of regex columns in our legacy logging strategy for linux dedicated.
-        private readonly string[] orderedRegexCol = new string[]
-        {
-            "Account",
-            "ActiveActivities",
-            "ActiveOrchestrators",
-            "Age",
-            "AppName",
-            "ContinuedAsNew",
-            "CreatedTimeFrom",
-            "CreatedTimeTo",
-            "DequeueCount",
-            "Details",
-            "Duration",
-            "ETag",
-            "Episode",
-            "EventCount",
-            "EventName",
-            "EventType",
-            "Exception",
-            "ExceptionMessage",
-            "ExecutionId",
-            "ExtensionVersion",
-            "FromWorkerName",
-            "FunctionName",
-            "FunctionState",
-            "FunctionType",
-            "Input",
-            "InstanceId",
-            "IsCheckpointComplete",
-            "IsExtendedSession",
-            "IsReplay",
-            "LastCheckpointTime",
-            "LatencyMs",
-            "MessageId",
-            "MessagesRead",
-            "MessagesSent",
-            "MessagesUpdated",
-            "NewEventCount",
-            "NewEvents",
-            "NextVisibleTime",
-            "OperationId",
-            "OperationName",
-            "Output*",
-            "PartitionId",
-            "PendingOrchestratorMessages",
-            "PendingOrchestrators",
-            "Reason",
-            "RelatedActivityId",
-            "RequestCount",
-            "RequestId",
-            "RequestingExecutionId",
-            "RequestingInstance",
-            "RequestingInstanceId",
-            "Result",
-            "RuntimeStatus",
-            "SequenceNumber",
-            "SizeInBytes",
-            "SlotName",
-            "StatusCode",
-            "StorageRequests",
-            "Success",
-            "TableEntitiesRead",
-            "TableEntitiesWritten",
-            "TargetExecutionId",
-            "TargetInstanceId",
-            "TaskEventId",
-            "TaskHub",
-            "Token",
-            "TotalEventCount",
-            "Version",
-            "VisibilityTimeoutSeconds",
-            "WorkerName",
-        };
-
         // variable below is internal static for testing and other convenient purposes
         // we need to be able to change the logging path for a windows-based CI
         // the logger being internal static is convenient for flushing it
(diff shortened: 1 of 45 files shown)


test/Common/TestHelpers.cs
```diff
@@ -257,6 +256,26 @@ public static DurableTaskOptions GetDurableTaskOptionsForStorageProvider(string
             }
         }
 
+        /// <summary>
+        /// Read a file's contents, line by line, even if another process is currently writing to it.
```
davidmrdavid, line 260: Otherwise we get an exception about not being able to read the log-file's contents because another process is using it

davidmrdavid:
I tested this on a linux dedicated instance already. Will test again early next week so we can merge this. I believe this needs to be backported to V1 as well though

ConnorMcMahon (commented):
LGTM

test/Common/TestHelpers.cs
davidmrdavid: It's a tricky one :) . Before, I was using `File.ReadAllLines` which apparently attempts to open the file using a restrictive, read-only, file permission. This is incompatible with our file logger's write permissions, so it errors out. With this, we are opening the file using a "read-write" permission so it works just fine :)

davidmrdavid: I got the description, **and the method itself,** from this SO answer: https://stackoverflow.com/questions/12744725/how-do-i-perform-file-readalllines-on-a-file-that-is-also-open-in-excel

ConnorMcMahon: Let's directly link to that SO answer in our comment.

davidmrdavid:
Hey @ConnorMcMahon, in my latest commit I refactored the JSON-logging tests to use the `WaitUntilTrue` approach and also simplify some of their shared logic. I'd love to get a final LGTM on it, thanks!

davidmrdavid:
Also, since this needs to be backported to V1, should we wait on the merge until we have a V1 PR? Just to make sure we don't forget about it.

ConnorMcMahon:
For V1, just open a ticket, and reference the ticket in this PR (and mention this PR in the ticket). Then you can check off the sub-box in that category.

test/Common/TestHelpers.cs
davidmrdavid: I didn't see this before, I'll link it

ConnorMcMahon (approved):
LGTM!

If we have tested this end-to-end before we start the validation step, I am 100% on board for merging this in for v2.4.2.

davidmrdavid:
Thanks! I don't plan to modify the code further, but this could use just a tad more testing so I won't be merging this for 2.4.2 :)

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -685,85 +636,71 @@ public async Task OutputsValidJSONLogs()
...
-                }
+                        // (2) Contains minimal expected fields
+                        TestHelpers.IsValidJSONLog(json);
 
-                if (keys.Contains("RelatedActivityId"))
-                {
-                    foundRelatedActivityId = true;
-                }
+                        // recording EventSource providers seen
+                        int eventId = (int)json.GetValue("EventId");
```
ConnorMcMahon, line 662: Is EventId the best way of identifying provider? I get that these event ids are unique across providers (at least I believe so), but there should be a Provider field that is more conclusive right?

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -685,85 +636,71 @@ public async Task OutputsValidJSONLogs()
...
-            // Validating JSON outputs
-            foreach (string line in lines)
-            {
-                // (1) Ensure it can be parsed to JSON
-                JObject json = JObject.Parse(line);
+            await TestHelpers.WaitUntilTrue(
+                predicate: () =>
+                {
+                    bool foundEtwEventSourceLog, foundAzureStorageLog, foundDurableTaskCoreLog, foundActivityId, foundRelatedActivityId;
+                    foundEtwEventSourceLog = foundAzureStorageLog = foundDurableTaskCoreLog = foundActivityId = foundRelatedActivityId = false;
```
ConnorMcMahon, line 648: Is there a reason we prioritized fields like activity ID and related activity ID over other fields like "InstanceId", "Partition", "MessageId", "Message", etc?

Also, it may be slightly more performant (and imo easier to read) if you perform checks one by one.

For instance:

```csharp
var jsonLines = TestHelpers.WriteSafeReadAllLines(LinuxAppServiceLogger.LogginPath).Select(line => JObject.Parse(line));

if (jsonLines.Any(json => !TestHelpers.IsValidJson(json))
{
    return false;
}

if (!jsonLines.Any(json => json.GetValue<string>("ProviderName") == "DurableTask.AzureStorage"))) {
    return false;
}

//etc
```

ConnorMcMahon (approved):
Nits on tests, but otherwise looks good.

test/Common/DurableTaskEndToEndTests.cs
davidmrdavid: I think the only reason why I prioritized the ActivityID fields is that, in the past, during development, we had received faulty logs which missed them. I realize all fields are relevant, this test is just particularly informed by a previous bug. To summarize: it was an arbitrary decision.

I also find it surprising that doing a `.Any()` test might be faster than the current approach, since with the current approach we only iterate through the data once. Regardless, since this is a test, I do think readability beats performance, so I'll incorporate your suggestion :)

ConnorMcMahon: It's hard to say whether it is more performant or not.

My expectation is right now we are doing ~5x checks over 1 iteration, and we have to complete the entire iteration before returning.

With this approach, while we are iterating 5 times, which introduces overhead, we are only doing 1x check per iteration, so in the case where one of the earlier cases fails, we catch it earlier and don't have to do subsequent iterations/checks. 

.Any() also shortcircuits as soon as any value matches that condition, so if the logs have 50 lines, and all the conditions are met by the first 10 lines, then it should in theory be faster.

Overall, like we said though, this is a test, so perf isn't the top priority (at least not in the order of a few million CPU cycles).

davidmrdavid:
@ConnorMcMahon, only requesting a re-review since I changed the testing code after your last approval. A quick skim should do.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
```diff
@@ -22,86 +22,11 @@ internal class LinuxAppServiceLogger
...
-            "VisibilityTimeoutSeconds",
-            "WorkerName",
-        };
-
         // variable below is internal static for testing and other convenient purposes
         // we need to be able to change the logging path for a windows-based CI
         // the logger being internal static is convenient for flushing it
 #pragma warning disable SA1401 // Fields should be private
-        internal static string LoggingPath = "/var/log/functionsLogs/durableevents.log";
+        internal static string LoggingPath = "/var/log/functionsLogs/durableeventsJSON.log";
```
ConnorMcMahon, line 29: Before we merge, we should just 100% verify that this is indeed the file path listened to by linux app services.

ConnorMcMahon (approved):
LGTM!

src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs
davidmrdavid: Confirmed and e2e tested

PR #1726, opened 2021-03-12
Allow use of encoded characters in http apis

### LINKED ISSUE #1124: Entity REST API won't accept URL encoded keys
Understand if this is unsupported by design, but was hoping to create an entity for each Azure Resource I'm having durable monitor.  My hope was to have the entity key be the resource ID, which has the format:

/subscriptions/<sub-id>/resourceGroups/<resource-group>

But when trying to invoke with the REST API, it throws a 404 when I encode it:

> POST http://localhost:7071/runtime/webhooks/durabletask/entities/AzureResource/%2Fsubscriptions%2F123-345%2FresourceGroups%2FmyRg?op=MonitorResource

Same call works fine when I use a key that doesn't have URL encoding (e.g. just a alpha-numeric string)

amdeel:
Work done in this PR https://github.com/Azure/durabletask/pull/481 allows us to use characters that are restricted in Azure Storage for partition keys. This helps us remove these restrictions from the Durable extension side and let each backend focus on handling its own individual restrictions.

This PR adds the ability to use encoded characters in our Http Apis to allow customers to use instance ids with characters that were previously Azure Storage character restrictions, like slash for example. After our http api route is matched, we decode the parameters to ensure the intended representation is stored as the instance id.

This PR also removes the exceptions we added when a customer attempts to use an entity key or instance id with one of these restricted characters.

resolves #1124 

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
+++ b/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
@@ -279,8 +279,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 var routeValues = new RouteValueDictionary();
                 if (StartOrchestrationRoute.TryMatch(pathString, routeValues))
                 {
-                    string functionName = this.GetDecodedRouteParameter(routeValues, FunctionNameRouteParameter);
-                    string instanceId = this.GetDecodedRouteParameter(routeValues, InstanceIdRouteParameter);
+                    string functionName = GetDecodedRouteParameter(routeValues, FunctionNameRouteParameter);
+                    string instanceId = GetDecodedRouteParameter(routeValues, InstanceIdRouteParameter);
                     if (request.Method == HttpMethod.Post)
                     {
                         return await this.HandleStartOrchestratorRequestAsync(request, functionName, instanceId);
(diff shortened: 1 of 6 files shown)


src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -414,6 +414,12 @@ public async Task<HttpResponseMessage> HandleRequestAsync(HttpRequestMessage req
             }
         }
 
+        private string GetDecodedRouteParameter(RouteValueDictionary routeValues, string routeParameter)
```
cgillum, line 417: It looks like this method doesn't reference any instance variables or methods, so let's make it `static`. It's a good habit to get into because it makes the code a bit more efficient and it helps the reader quickly understand that a method doesn't depend on any external state (instance variables, etc.).

cgillum (approved):
The fix looks good to me. Should we add a test to verify that we can successfully use these characters?

ConnorMcMahon (approved):
LGTM!

Did we test this end-to-end? The reason I ask is because some of the HTTP pipeline doesn't happen in our tests code because it is provided by functions runtime.

torosent:

PR #1727, opened 2021-03-12
Register IPlatformInformationService in AddDurableClientFactory()

bachuv:
These changes incorporate the changes from #1706 and register `IPlatformInformationService` in `AddDurableClientFactory()`. Without registering `IPlatformInformationService` , creating a DurableClient with Azure Storage fails because it can't find an implementation of `IPlatformInformationService` in `AzureStorageDurabilityProviderFactory`. This PR also includes a test in DurableClientBaseTests to help us catch if any new services need to be registered in `AddDurableClientFactory()` in the future.

### Pull request checklist

### DIFF
--- a/test/Common/DurableClientBaseTests.cs
+++ b/test/Common/DurableClientBaseTests.cs
@@ -7,8 +7,6 @@ using System.Linq;
 using System.Net.Http;
 using System.Threading.Tasks;
 using DurableTask.Core;
-using FluentAssertions;
-using Microsoft.Azure.WebJobs.Host.TestCommon;
 #if !FUNCTIONS_V1
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
@@ -28,14 +26,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
 {
     public class DurableClientBaseTests
     {
-        private readonly ITestOutputHelper output;
-        private readonly TestLoggerProvider loggerProvider;
-
-        public DurableClientBaseTests(ITestOutputHelper output)
-        {
-            this.output = output;
-            this.loggerProvider = new TestLoggerProvider(output);
-        }
+        public DurableClientBaseTests() { }
 
         [Theory]
         [Trait("Category", PlatformSpecificHelpers.TestCategory)]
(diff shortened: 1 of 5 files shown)


test/FunctionsV2/PlatformSpecificHelpers.FunctionsV2.cs
```diff
@@ -157,6 +215,21 @@ private static IWebJobsBuilder AddMultipleDurabilityProvidersDurableTask(this IW
             return builder;
         }
 
+        private static IWebJobsBuilder AddDurableClientFactoryDurableTask(this IWebJobsBuilder builder, IOptions<DurableTaskOptions> options, IEnumerable<IDurabilityProviderFactory> durabilityProviderFactories = null)
```
davidmrdavid, line 218: Could we please add a short method-header comment to this?

davidmrdavid (approved):
Looks good to me, thank you for cleaning up my post-merge mess 😄 !
Left a non-blocking comment

test/FunctionsV2/PlatformSpecificHelpers.FunctionsV2.cs
bachuv: Is there anything in particular that you think would be helpful to add in the comment? I'm just asking since the other methods in TestHelpers.cs don't have comments.

davidmrdavid: Hmm, it's hard to say, **I would be good with any blurb about when to invoke this method and for why it's needed**.
The reason why I ask is that this method is written at a very high level of abstraction (as it should due to interfacing w/ DI and factories) but that can make it difficult, for me, to know where and how to invoke it.

In short, I'm not too picky so any short comment about what this does and when to use it would do it for me :)

test/Common/DurableClientBaseTests.cs
```diff
@@ -212,6 +228,30 @@ public void DurableClient_ExternalApp_CreateHttpManagementPayload_ThrowsExceptio
...
+        /// End to end test that ensures that DurableClientFactory is set up correctly
+        /// (i.e. the correct services are injected through dependency injection
+        /// and AzureStorageDurabilityProvider is created).
+        /// </summary>
+        [Fact]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        public async Task DurableClient_AzureStorage_SuccessfulSetup()
+        {
+            string orchestratorName = nameof(TestOrchestrations.SayHelloInline);
+            using (ITestHost host = TestHelpers.GetJobHost(
```
ConnorMcMahon, line 241: Just to make sure, does this path fail if `DurableTaskJobHostConfigurationExtensions` changes aren't present?

ConnorMcMahon, line 241: In the long run, we may be able to simplify this test (and the additional helper methods we added) by making a generic .NET Core host that calls `builder.Services.AddDurableClientFactory();`. That is more inline with what end customers would be doing, and then you don't need all of the changes to TestHelpers to accommodate wiring this into the JobHost.

In the interest of time and the code freeze today, I am fine postponing this.

ConnorMcMahon (approved):
Approved with suggestions.

test/Common/DurableClientBaseTests.cs
```diff
@@ -20,12 +21,27 @@
...
 using static Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests.HttpApiHandlerTests;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
 {
     public class DurableClientBaseTests
     {
+        private readonly ITestOutputHelper output;
+
+#if !FUNCTIONS_V1
+        private readonly TestLoggerProvider loggerProvider;
```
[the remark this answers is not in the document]
bachuv: No it's not. I looked again and `TestLoggerProvider` needed `using Microsoft.Azure.WebJobs.Host.TestCommon;` but that was in a `#if !FUNCTIONS_V1` block so it wasn't showing up. I moved the using statement out of that block and now I'm able to take the conditions out from around `TestLoggerProvider`.

test/Common/DurableClientBaseTests.cs
bachuv: Yes, this test fails with an `InvalidOperationException` saying it's unable to resolve a service.

amdeel (changes requested):

amdeel (approved):
Looks great!

PR #1740, opened 2021-03-18
Fix bug in new Webhook URL fetch logic

ConnorMcMahon:
The context.GetWebhookUrl() method only works in the context of the
Initialize() method. In order to properly handle this case, we must use
the direct IWebhookProvider made available in FunctionsV2+, and in
FunctionsV1 we have to live with the fact that we can't dynamically
determine the current webhook URI.

### Pull request checklist

* [X] My changes **do not** require documentation changes
* [X] My changes **should not** be added to the release notes for the next release
* [X] My changes **do not** need to be backported to a previous version
* [X] I have added all required tests (Unit tests, E2E tests)

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -3,7 +3,7 @@
 - Improved concurrency defaults for the App Service Consumption plan (https://github.com/Azure/azure-functions-durable-extension/pull/1706)
 
 ## Bug Fixes:
-- Properly used update management API URLs after a successful slot swap (#1716)
+- Properly used update management API URLs after a successful slot swap on Functions V3 (#1716)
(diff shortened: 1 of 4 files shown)


davidmrdavid (approved):
Seems like a simple fix, it looks good to me!
I'm good to merge so long as the CI goes green and this is validated across V1,V2,V3

davidmrdavid:
I also wonder: should we document this V1 limitation? Maybe at least create a tracking ticket to write it down somewhere?

ConnorMcMahon:
This may honestly be more of a Functions V1 limitation than a Durable Functions running on V1 limitation. I'm not quite sure where we would even document this behavior?

ConnorMcMahon:
The only test failure was with the clean up storage queues, so scheduling another run.

PR #1741, opened 2021-03-18
Patch PlatformInfoProvider exception in Functions V1

davidmrdavid:
As in title.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -74,7 +74,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         private readonly bool isOptionsConfigured;
         private readonly IApplicationLifetimeWrapper hostLifetimeService = HostLifecycleService.NoOp;
 #pragma warning disable CS0612 // Type or member is obsolete
-        private readonly IPlatformInformationService platformInformationService;
+        private IPlatformInformationService platformInformationService;
 #pragma warning restore CS0612 // Type or member is obsolete
         private IDurabilityProviderFactory durabilityProviderFactory;
         private INameResolver nameResolver;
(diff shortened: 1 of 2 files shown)


src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -202,6 +202,11 @@ public DurableTaskExtension()
             : this(options, loggerFactory, nameResolver, orchestrationServiceFactories, shutdownNotification, durableHttpMessageHandlerFactory)
         {
             this.connectionStringResolver = connectionStringResolver;
+            if (platformInformationService == null)
```
ConnorMcMahon, line 205: Why don't we move this code (205-210) into the main constructor, and then eliminate the new code you added in AzureStorageDurabilityProviderFactory?

davidmrdavid: Not exactly sure what you mean by "main" constructor. Do you mean the DurableTaskExtension constructor in line 115? Why would that remove the need for the check in AzureStorageDurabilityProviderFactory? As I understand it, the durability provider in that constructor is still generated via DI because of this line: https://github.com/Azure/azure-functions-durable-extension/blob/95e677686000ce036f5f7c4d040f2702c8e73990/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs#L153

So I'm not convinced that making any changes in this file would prevent the durability provider from error'ing out

davidmrdavid:
I did some testing and found out that we were not using this constructor in V1:
https://github.com/Azure/azure-functions-durable-extension/blob/95e677686000ce036f5f7c4d040f2702c8e73990/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs#L189-L206

Instead, we're using this:
https://github.com/Azure/azure-functions-durable-extension/blob/95e677686000ce036f5f7c4d040f2702c8e73990/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs#L87-L100

which means that the "main" constructor here:
https://github.com/Azure/azure-functions-durable-extension/blob/95e677686000ce036f5f7c4d040f2702c8e73990/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs#L118-L148

never gets called in V1. Instead, all the variable initialization is happening here:
https://github.com/Azure/azure-functions-durable-extension/blob/95e677686000ce036f5f7c4d040f2702c8e73990/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs#L496-L520

FYI @ConnorMcMahon. This means that changing the exception-throwing behavior in the "main constructor" would not serve as an initialization path for Functions V1 as that constructor does not get called in V1. My latest commit reflects this but still removes the need to have null-checking in the durability provider

ConnorMcMahon (approved):
LGTM! Good investigation here. Merge after a clean CI + once we have end-to-end confidence.

PR #1742, opened 2021-03-18
Update nuget license URL to our MIT license

ConnorMcMahon:
Point to hardcoded commit of our LICENSE.md file, so that the license
for this release remains static.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
+++ b/src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
@@ -96,7 +96,7 @@
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <PackageTags>Microsoft;Azure;WebJobs;Durable;Extension;Orchestration;Workflow;Functions</PackageTags>
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
-    <PackageLicenseUrl>https://raw.githubusercontent.com/Azure/azure-functions-durable-extension/7e01577f57fdfdb15d20fa66b09c3714926e0aea/LICENSE</PackageLicenseUrl>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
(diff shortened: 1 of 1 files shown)


davidmrdavid (commented):
A bit surprised that we need to hardcode a commit. Why not just point to the file in `main` or `dev`?

ConnorMcMahon:
@davidmrdavid, those are changing branches, meaning that if we ever made changes to the LICENSE.md file, it would retroactively change the license that older releases point to, which feels like a huge problem to me. I don't see us changing our LICENSE in the future, but on the principle of the matter, it should point to a static file.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -96,7 +96,7 @@
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <PackageTags>Microsoft;Azure;WebJobs;Durable;Extension;Orchestration;Workflow;Functions</PackageTags>
     <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
-    <PackageLicenseUrl>https://go.microsoft.com/fwlink/?linkid=2028464</PackageLicenseUrl>
+    <PackageLicenseUrl>https://raw.githubusercontent.com/Azure/azure-functions-durable-extension/7e01577f57fdfdb15d20fa66b09c3714926e0aea/LICENSE</PackageLicenseUrl>
```
cgillum, line 99: FWIW, this property has been deprecated for a while. The recommendation is to use `<PackageLicenseExpression>` instead. Not sure how much we want to take on for this release, but here is an example of how the SQL backend project does this: https://github.com/microsoft/durabletask-mssql/blob/4ff7b2f5364aadf0b8d322a4eeb86df153ff746e/src/common.props#L31

ConnorMcMahon: Honestly, that seems simpler than my fix, and since it will get rid of the nuget.org warning as well, I'll go with that.

davidmrdavid (approved):
LGTM

PR #1745, opened 2021-03-19
Add smoke tests, one per functions version

davidmrdavid:
Adding easy-to-access smoke tests, to accelerate the release process, and for the test-bot to leverage.
It's just a leaner version of the precompiled samples, but with a different csproj per functions version.

I was additionally thinking of adding the perf tests to the repo, how do we feel about that?


test/SmokeTests/SmokeTestsV1/VSSample.csproj
```diff
@@ -0,0 +1,21 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net461</TargetFramework>
+    <AzureFunctionsVersion>v1</AzureFunctionsVersion>
+  </PropertyGroup>
+  <!-- Common packages for all targets -->
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="2.4.1" />
```
cgillum, line 8: To make this process even smoother, add a project reference directly to the extension project. That way you don't need to deal with nuget packages at all. It makes it waaay easier to just quickly F5 and go.

davidmrdavid:
Note to self: also add smoke tests for ooproc!

cgillum:
I went ahead and submitted some suggested updates for the smoke tests. Feel free to take these, revert them, whichever you'd prefer. :)

Quick summary of the changes:
* Replaced nuget references with project references so that we can easily do F5 w/out updating nuget package version
* Simplified HTTP triggers to hardcode the orchestration name. This allows us to just click links to verify correctness w/out typing orchestrator names.
* Added/simplified local.settings.json files, pointing them to the local Azure Storage emulator
* Updated host.json files for more consistency and task hub isolation
* Added smoke test projects to root .sln file to make sure they are all included in the build and for easy access in Visual Studio

ConnorMcMahon:
@bachuv, do you have a link to how TestBot can deploy these smoke tests? 

I'm curious if we can introduce some conditional compilation logic in the .csproj to add support for smoke test runs with the new backends?

bachuv:
@ConnorMcMahon  `DeployToFunctionApp` is the Durable PowerShell orchestrator function that handles the deployment.

https://github.com/Azure/azure-functions-durable-extension/blob/f9b445cf8e89e1537b5907f58d9a1d563dc963c5/samples/testbot/deployment-service/DeployToFunctionApp/run.ps1

davidmrdavid:
I think this is ready to be merged. There's a pending conversation about automating the testing of this via TestBot, but I don't think that should be a blocker for this PR :)

ConnorMcMahon (commented):
LGTM!

PR #1787, opened 2021-04-08
Emit warning if APPINSIGHTS_INSTRUMENTATIONKEY isn't set, but Distributed Tracing is enabled

### LINKED ISSUE #1694: Distrubted tracing fails on local debug if instrumentation key not set
With the following host.json settings the function fails during startup for local debug because the Instrumentation Key isn't defined.  In all other function experiences running without an instrumentation key just doesn't emit logs, but doesn't block startup.  Maybe this should be a warning instead of an error?

```json
{
    "version": "2.0",
    "extensions": {
        "durableTask": {
            "tracing": {
                "distributedTracingEnabled": true,
                "distributedTracingProtocol": "W3CTraceContext"
            }
        }
    },
    "logging": {
        "applicationInsights": {
            "samplingSettings": {
                "isEnabled": false
            },
            "httpAutoCollectionOptions": {
                "enableW3CDistributedTracing": true
            }
        }
    }
}
```

bachuv:
Currently, if a customer enables distributed tracing in the app's host.json, but doesn't set `APPINSIGHTS_INSTRUMENTATIONKEY` as an environment variable then startup will fail saying that `APPINSIGHTS_INSTRUMENTATIONKEY` isn't set. 

This PR allows startup to succeed if Distributed Tracing is enabled, but `APPINSIGHTS_INSTRUMENTATIONKEY` isn't set. If `APPINSIGHTS_INSTRUMENTATIONKEY` isn't set then the extension will emit this warning: `'APPINSIGHTS_INSTRUMENTATIONKEY' isn't defined in the current environment variables, but Distributed Tracing is enabled. Please set 'APPINSIGHTS_INSTRUMENTATIONKEY' to use Distributed Tracing.`.

Resolves #1694 

### Pull request checklist


src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
```diff
@@ -93,7 +94,14 @@ private void SetUpTelemetryClient()
...
             config.TelemetryInitializers.Add(telemetryInitializer);
 
-            config.InstrumentationKey = Environment.GetEnvironmentVariable("APPINSIGHTS_INSTRUMENTATIONKEY");
+            if (Environment.GetEnvironmentVariable("APPINSIGHTS_INSTRUMENTATIONKEY") != null)
+            {
+                config.InstrumentationKey = Environment.GetEnvironmentVariable("APPINSIGHTS_INSTRUMENTATIONKEY");
+            }
+            else
+            {
+                logger.LogWarning("'APPINSIGHTS_INSTRUMENTATIONKEY' isn't defined in the current environment variables, but Distributed Tracing is enabled. Please set 'APPINSIGHTS_INSTRUMENTATIONKEY' to use Distributed Tracing.");
```
cgillum, line 103: I wonder if it would be better to use `EndToEndTraceHelper.ExtensionWarningEvent` instead of `ILogger.LogWarning`. That way we can also see this warning in Kusto. The problem, however, is that `EndToEndTraceHelper` is an internal class and we can't simply pass it to public methods on public classes. If we agree that `EndToEndTraceHelper` is a better solution for logging, then maybe we can make `Initialize()` internal, or have a separate internal `Initialize()` method if it's not practical to change the existing one?

test/FunctionsV2/CorrelationEndToEndTests.cs
```diff
@@ -155,6 +155,62 @@ public async Task AllOrchestrationActivityActions(bool extendedSessions, string
...
+
+            using (var host = TestHelpers.GetJobHost(
+                this.loggerProvider,
+                "SingleOrchestration",
+                extendedSessions,
+                options: options))
+            {
+                string warningMessage = "'APPINSIGHTS_INSTRUMENTATIONKEY' isn't defined in the current environment variables, but Distributed Tracing is enabled. Please set 'APPINSIGHTS_INSTRUMENTATIONKEY' to use Distributed Tracing.";
+                var warningLogMessage = this.loggerProvider.GetAllLogMessages().ToList().FindAll(l => l.FormattedMessage.Equals(warningMessage));
+                Assert.True(warningLogMessage.Count != 0);
```
cgillum, line 181: I think xunit should normally issue warnings about this, but you should generally avoid using `Assert.True(...)` for things like this and instead use more specific asserts, like `Assert.NotEqual(0, warningLogMessage.Count)` or `Assert.NotEmpty(warningLogMessage)`. One of the benefits is that you get better error messages if the check fails. If you want to get even fancier, you could use:
```csharp
Assert.Single(this.loggerProvider.GetAllLogMessages().Where(l => l.FormatMessage.Equals(warningMessage));
```

test/FunctionsV2/CorrelationEndToEndTests.cs
```diff
@@ -155,6 +155,62 @@ public async Task AllOrchestrationActivityActions(bool extendedSessions, string
...
+
+            using (var host = TestHelpers.GetJobHost(
+                this.loggerProvider,
+                "SingleOrchestration",
+                extendedSessions,
+                options: options))
+            {
+                string warningMessage = "'APPINSIGHTS_INSTRUMENTATIONKEY' isn't defined in the current environment variables, but Distributed Tracing is enabled. Please set 'APPINSIGHTS_INSTRUMENTATIONKEY' to use Distributed Tracing.";
+                var warningLogMessage = this.loggerProvider.GetAllLogMessages().ToList().FindAll(l => l.FormattedMessage.Equals(warningMessage));
+                Assert.True(warningLogMessage.Count == 0);
```
cgillum, line 210: Same comment as above about avoiding `Assert.True()`.

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
bachuv: It would be nice to use `EndToEndTraceHelper` so we have this information in Kusto. If we go with this approach then making `Initialize()` internal doesn't look like it would work. I tried it out and an error showed up saying that internal isn't a valid access modifier for `Initialize()` in C# 7.3. It's allowed in C# 8.0 though.

ConnorMcMahon: This problem comes up pretty frequently with types that use our dependency injection framework. Essentially, the Functions DI model requires our constructor is public, which means all the interfaces (and all of their parameters) need to be public. Options:

1. Don't use DI for instantiating ITelemetryActivator, and just instantiate it in the constructor directly. We may lose some testability benefits here, but it looks like we don't allow customers to customize this via DI anyways, so we wouldn't lose those advantages.
2. EndToEndTraceHelper essentially just takes a logger, so you could just reinstantiate one in your `Initialize(ILogger logger)` method directly.
3. We try and come up with a more general solution for this problem. We could see if Functions V2/V3 allows us to provide a constructor that just takes the IServiceCollection, and we pull the services we want from that IServiceCollection in the constructor implementation. This would prevent us from having to make all of the services we inject public, but I'm not sure if this works without some modifications to the Functions/WebJobs layer...

bachuv: @ConnorMcMahon thanks for listing out some different options! With option 1, we probably don't want to take away from any testing benefits. Option 2 might be the simplest option since it would reinstantiate `EndToEndTraceHelper` in the scope of `TelemetryActivator`. Is there an option that you are leaning more towards using?

ConnorMcMahon: I realize that thumbs up didn't answer your question. I think option 2 is the easiest approach to get to where we need.

bachuv: Sounds good, I'll implement option 2.

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
```diff
@@ -93,7 +96,18 @@ private void SetUpTelemetryClient()
             telemetryInitializer.ExcludeComponentCorrelationHttpHeadersOnDomains.Add("127.0.0.1");
             config.TelemetryInitializers.Add(telemetryInitializer);
 
-            config.InstrumentationKey = Environment.GetEnvironmentVariable("APPINSIGHTS_INSTRUMENTATIONKEY");
+            if (Environment.GetEnvironmentVariable("APPINSIGHTS_INSTRUMENTATIONKEY") != null)
```
ConnorMcMahon, line 99: If you inject `INameResolver` into the constructor of this service, you can use that instead of `Environment.GetEnvironmentVariable()`.

That makes mocking environment variables for the tests possible.

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
```diff
@@ -93,7 +96,18 @@ private void SetUpTelemetryClient()
...
+            {
+                config.InstrumentationKey = Environment.GetEnvironmentVariable("APPINSIGHTS_INSTRUMENTATIONKEY");
```
ConnorMcMahon, line 101: I recommend fetching the value for `APPINSIGHTS_INSTRUMENTATIONKEY` before null checking, so that we only need to access the environment variable once.

test/FunctionsV2/CorrelationEndToEndTests.cs
```diff
@@ -155,6 +155,50 @@ public async Task AllOrchestrationActivityActions(bool extendedSessions, string
...
+            TraceOptions traceOptions = new TraceOptions()
+            {
+                DistributedTracingEnabled = true,
+                DistributedTracingProtocol = "W3CTraceContext",
+            };
+
+            DurableTaskOptions options = new DurableTaskOptions();
+            options.Tracing = traceOptions;
+
+            Environment.SetEnvironmentVariable("APPINSIGHTS_INSTRUMENTATIONKEY", keyIsSet ? "test key" : null);
```
ConnorMcMahon, line 180: As mentioned in the other comments, we should mock INameResolver for this test.

ConnorMcMahon (changes requested):
A couple of small changes, but otherwise looks good!

ConnorMcMahon (approved):
LGTM!

PR #1800, opened 2021-04-26
Change local RPC port selection logic

### LINKED ISSUE #1798: Avoid listening to ephemeral ports for local RPC
### Description
The full details are too complicated to spell out here, but the usage of ephemeral ports on hosted Function apps running on Windows results in an edge case where outbound requests use the local RPC port, and cause the sandbox to think the port is closed, rejecting all future requests to that port despite it still being open.

It looks [like Windows](https://www.ncftp.com/ncftpd/doc/misc/ephemeral_ports.html#Windows) does not start this range until 49152, so using a random port between 1024-32768 seems like a reasonable approach to fix this.

ConnorMcMahon:
Current RPC port selection logic favors ephemeral ports. This leads to
issues with the web apps sandbox.

Instead, we now use randomized port selection from an approved range of
ports.

resolves #1798

### Pull request checklist

    * [x] Otherwise: Backport tracked by issue/PR #1801

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
@@ -49,6 +49,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.InternalRpcUri = new Uri($"http://uninitialized");
             this.localWebHost = new NoOpWebHost();
             this.portGenerator = new Random();
+            this.attemptedPorts = new HashSet<int>();
         }
 
         public Uri InternalRpcUri { get; private set; }
--- a/src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
@@ -22,17 +22,19 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
     {
         private const int DefaultPort = 17071;
 
-        // Minimum port value is 1024, as the app service sandbox blocks ports below that.
-        private const int MinPort = 1024;
-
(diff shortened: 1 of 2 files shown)


davidmrdavid (approved):
LGTM!

src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
```diff
@@ -80,11 +90,8 @@ public async Task StartAsync()
                         this.durableTaskOptions.HubName,
                         functionName: string.Empty,
                         instanceId: string.Empty,
-                        message: $"Failed to open local socket {availablePort}. This was attempt #{numAttempts} to open a local port.");
+                        message: $"Failed to open local port {availablePort}. This was attempt #{numAttempts} to open a local port.");
                     numAttempts++;
-                    var random = new Random();
```
ConnorMcMahon, line 85: This randomized delay is no longer necessary as we don't have a semi-deterministic mechanism for selecting ports anymore, so it doesn't matter if multiple apps are port-searching at the same time.

src/WebJobs.Extensions.DurableTask/LocalHttpListener.cs
```diff
@@ -59,7 +67,9 @@ public async Task StartAsync()
             int numAttempts = 1;
             do
             {
-                int availablePort = this.GetAvailablePort();
+                int availablePort = numAttempts == 1
+                    ? DefaultPort
+                    : this.portGenerator.Next(MinPort, MaxPort);
```
stekou, line 72: Kind of nitpicky of me, but this might select the same port multiple times.  A better way would be to generate a random permutation and then iterate through it; then you'd be trying each port exactly once.

cgillum (approved):

Issue #1733, opened 2021-03-15
Backport JSON logging for linux dedicated to V1


davidmrdavid:
As in title, a follow-up to: https://github.com/Azure/azure-functions-durable-extension/pull/1721

PR #1802, opened 2021-04-27
Backporting Transition to JSON logging to V1

davidmrdavid:
Tackles: https://github.com/Azure/azure-functions-durable-extension/issues/1733
Backports: https://github.com/Azure/azure-functions-durable-extension/pull/1721

resolves https://github.com/Azure/azure-functions-durable-extension/issues/1733

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -1,0 +1,7 @@
+## New Features
+- Improved supportability logs on Linux Dedicated plans (https://github.com/Azure/azure-functions-durable-extension/pull/1721)
+
+## Bug fixes
+- Fix issue with local RPC endpoint used by non-.NET languages on Windows apps (#1800)
+
+## Breaking Changes
--- a/release_notes.md
+++ b/release_notes.md
@@ -1,7 +1,2 @@
 ## New Features
-- Improved supportability logs on Linux Dedicated plans (https://github.com/Azure/azure-functions-durable-extension/pull/1721)
-
-## Bug fixes
-- Fix issue with local RPC endpoint used by non-.NET languages on Windows apps (#1800)
-
-## Breaking Changes
+- Improved supportability logs on Linux Dedicated plans (https://github.com/Azure/azure-functions-durable-extension/pull/1802)
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
@@ -34,7 +34,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// </summary>
         public TraceOptions Tracing { get; set; } = new TraceOptions();
 
-
         /// <summary>
         /// Gets or sets the number of messages to pull from the control queue at a time.
         /// </summary>
--- a/src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
@@ -39,6 +39,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// <param name="eventSource">An instance of EventSource.</param>
         protected override void OnEventSourceCreated(EventSource eventSource)
         {
+            // "7DA4779A-152E-44A2-A6F2-F80D991A5BEE" is the old DurableTask-Core event source,
+            // so we provide extra logic to ignore it.
             if ((eventSource.Name == "DurableTask-Core"
                    && eventSource.Guid != new Guid("7DA4779A-152E-44A2-A6F2-F80D991A5BEE")) ||
                 eventSource.Name == "DurableTask-AzureStorage" ||
--- a/src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
@@ -43,6 +43,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             // so we provide extra logic to ignore it.
             if ((eventSource.Name == "DurableTask-Core"
                   && eventSource.Guid != new Guid("7DA4779A-152E-44A2-A6F2-F80D991A5BEE")) ||
+                eventSource.Name == "DurableTask-AzureStorage" ||
                 eventSource.Name == "WebJobs-Extensions-DurableTask" ||
                 eventSource.Name == "DurableTask-SqlServer")
             {
--- a/src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
@@ -43,7 +43,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             // so we provide extra logic to ignore it.
             if ((eventSource.Name == "DurableTask-Core"
                   && eventSource.Guid != new Guid("7DA4779A-152E-44A2-A6F2-F80D991A5BEE")) ||
-                eventSource.Name == "DurableTask-AzureStorage" ||
                 eventSource.Name == "WebJobs-Extensions-DurableTask" ||
                 eventSource.Name == "DurableTask-SqlServer")
             {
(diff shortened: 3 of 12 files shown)


release_notes.md
```diff
@@ -0,0 +1,7 @@
+## New Features
+- Improved supportability logs on Linux Dedicated plans (https://github.com/Azure/azure-functions-durable-extension/pull/1721)
+
+## Bug fixes
+- Fix issue with local RPC endpoint used by non-.NET languages on Windows apps (#1800)
+
+## Breaking Changes
```
davidmrdavid, line 7: I'll remove this

src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
```diff
@@ -39,10 +39,8 @@ public EventSourceListener(LinuxAppServiceLogger logger, bool enableVerbose, End
         /// <param name="eventSource">An instance of EventSource.</param>
         protected override void OnEventSourceCreated(EventSource eventSource)
         {
-            // "7DA4779A-152E-44A2-A6F2-F80D991A5BEE" is the old DurableTask-Core event source,
-            // so we provide extra logic to ignore it.
-            if ((eventSource.Name == "DurableTask-Core"
-                  && eventSource.Guid != new Guid("7DA4779A-152E-44A2-A6F2-F80D991A5BEE")) ||
+
+            if (eventSource.Name == "DurableTask-Core" ||
```
davidmrdavid, line 43: So, for the code in `dev`, we explicitly filter out the "DurableTask-Core" provider of GUID "7DA4779A-152E-44A2-A6F2-F80D991A5BEE" since it "is the old DurableTask-Core event source". From my testing, in V1, it is also the _only_ DurableTask-Core event source, so filtering it out means we get no Core logs. Does that ring true to you all?

davidmrdavid:
This still needs E2E testing, but I'll complete that prior to a merge

davidmrdavid:
I seem to recall E2E testing this, but it's been a while, so I can do it again. Still, I'd appreciate a conditional approval :)

src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
```diff
@@ -34,7 +34,6 @@ public class DurableTaskOptions
         /// </summary>
         public TraceOptions Tracing { get; set; } = new TraceOptions();
 
-
```
ConnorMcMahon, line 37: We may want to revert this change.

src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
ConnorMcMahon: I don't believe it is the only DurableTask-Core implementation, because we update v1's version of DurableTask.AzureStorage to get the latest changes. That will inherently increase our DurableTask.Core version to one that has the new event source.

ConnorMcMahon (changes requested):
I think we want to filter out the old ones like we do in dev, but double check the version of DurableTask.Core to see if it is the version that has both event sources.

src/WebJobs.Extensions.DurableTask/DurableTaskOptions.cs
davidmrdavid: Hmm, but I think the code wouldn't compile with two blank lines in a row, right?

ConnorMcMahon: Ah, that makes sense.

src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
davidmrdavid: I trust that's correct but somehow my tests were missing logs when I had that filter enabled. I'll add the filter back and try again, it's been a while since I validated this.

davidmrdavid: Yeah, in my tests, the filtere'd out GUID is the only ETW event with name "DurableTask-Core".
I'll sync with you about this internally early next week

cgillum:
I'm not sure if this has been covered already in other PR comments, but I noticed that the v1 branch points to an old version of DurableTask.AzureStorage:

```xml
<PackageReference Include="Microsoft.Azure.DurableTask.AzureStorage" Version="1.7.7" />
```

Can we update this to point to the latest (currently **1.8.5**)?

src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
cgillum: We need to keep this filter as-is and change our DurableTask.AzureStorage dependency to the latest version.

davidmrdavid: @cgillum, yup, turned out that was the culprit :) . Thanks for pointing that out!

ConnorMcMahon:
@davidmrdavid, we ready to do final validation on this with 1.8.5 of DurableTask.AzureStorage now that #1838 was merged?

davidmrdavid:
@ConnorMcMahon, yes but I'll need to pull in some changes and re-push. I'm responding to feedback in a different PR, but I'll get back to this shortly, I'll send you a re-review request once this PR is ready to be validated once more.

test/Common/TestHelpers.cs
```diff
@@ -111,6 +113,98 @@ public static string GetTaskHubNameFromTestName(string testName, bool enableExte
...
+                "EventPrimaryStampName",
+                "ProviderName",
+                "TaskName",
+                "EventId",
+                "EventTimestamp",
+                "Tenant",
+                "Pid",
+                "Tid",
+            };
+            List<string> keys = json.Properties().Select(p => p.Name).ToList();
```
cgillum, line 135: nit: `HashSet<string>` would be more appropriate than `List<string>`. That or just do `json.TryGetValue(expectedKey, out _)` and avoid the extra data structure altogether.

test/Common/TestHelpers.cs
```diff
@@ -111,6 +113,98 @@ public static string GetTaskHubNameFromTestName(string testName, bool enableExte
...
+            using (var fileStream = new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.ReadWrite))
+            using (var streamReader = new StreamReader(fileStream, Encoding.Default))
+            {
+                List<string> file = new List<string>();
+                while (!streamReader.EndOfStream)
+                {
+                    file.Add(streamReader.ReadLine());
+                }
+
+                return file.ToArray();
```
cgillum, line 169: nit: This isn't important for test code, but it would be more efficient just to return `file` rather than call `ToArray()` on it (which copies all the elements).

test/Common/TestHelpers.cs
```diff
@@ -111,6 +113,98 @@ public static string GetTaskHubNameFromTestName(string testName, bool enableExte
...
+        public static bool IsValidJSONLog(JObject json)
+        {
+            List<string> expectedKeys = new List<string>
+            {
+                "EventStampName",
+                "EventPrimaryStampName",
+                "ProviderName",
+                "TaskName",
+                "EventId",
+                "EventTimestamp",
```
cgillum, line 130: Probably out of scope for this PR, but with this change (or with changes that preceded it) will we get fractional seconds just like we do with Windows? Not being able to properly sort by timestamps has been killing me when doing debugging on Linux. :)

cgillum (approved):
No blockers from me - but it's probably best to get a final sign-off from Connor. Just a few small nits than you can ignore plus a question.

ConnorMcMahon (approved):
Chris raised some valid minor critiques, other than that, ready to merge.

test/Common/TestHelpers.cs
davidmrdavid: Ah yes, that's killing me as well. Unfortunately, the answer is no, and I explain more about why in here: https://github.com/Azure/azure-functions-durable-extension/issues/1843

Let's use that issue to track and discuss this work-item :)

davidmrdavid:
@cgillum, thank you for your suggestions. I've applied them here, and also in V2 via this PR: https://github.com/Azure/azure-functions-durable-extension/pull/1844

davidmrdavid:
Just e2e tested this again and it works. Merging!

PR #1836, opened 2021-05-13
Support alternative OOProc replay mechanism

davidmrdavid:
We recently discovered a handful of telemetry mishaps that occur only to OOProc SDKs. It was determined that these were due to our existing replay logic, which can be imprecise in a few cases. Furthermore, there's evidence suggesting that our replay logic _might_ be the cause for runtime errors as well, which motivates the need to mitigate this issue further.

This PR aims to mitigate these replay imprecisions in a backwards compatible way.

First, some definitions:

- **Compound task**: A task of tasks, either: Task.WhenAll and Task.WhenAny.
- **Atomic task:** A singleton task,  i.e a Task that does not contain other tasks. For example: callActivityAsync.
- **Deep compound tasks**: A compound task containing at least 1 task that is also **compound**,  potentially  **deep** as well. For example: Task.WhenAll( [ Task.WhenAny( TaskA, TaskB), TaskC ] )

With that, here are the changes.

First, this PR introduces a `SchemaVersion` enum to track OOProc return payload schemas versions. We use this to determine whether or not to use the legacy (current) replay mechanism or if to use the one developed in this PR.

Second, this PR introduces a new method `ProcessAsyncActionsV2` (name open to change) which represents a more accurate replay procedure than `ProcessAsyncActions`. The key difference between these two methods is that `ProcessAsyncActionsV2` properly replays **deep compound tasks**. We can do this because we augment the existing OOProc output payload to include extra fields that allow us to determine
(1) if an atomic task is part of a compound task  [tracked by the Depth field]
(2) If the surrounding compound task is a WhenAny or WhenAll [tracked by the TaskGrouping field]
(3) how many parent compound tasks deep we are from the top-level compound task [tracked by the Depth field]

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
@@ -239,6 +239,27 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             }
         }
 
+        public void ProcessingOutOfProcPayload(
+            string taskHub,
+            string instanceId,
+            string details,
+            string functionType,
+            bool isReplay)
+        {
+                EtwEventSource.Instance.ProcessingOutOfProcPayload(
+                    taskHub,
+                    LocalAppName,
+                    LocalSlotName,
+                    details,
+                    instanceId,
+                    functionType.ToString(),
+                    ExtensionVersion,
+                    isReplay);
+
+                this.logger.LogInformation("TBD");
+
+        }
+
         public void FunctionTerminated(
             string hubName,
             string functionName,
--- a/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
@@ -242,22 +242,17 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         public void ProcessingOutOfProcPayload(
             string taskHub,
             string instanceId,
-            string details,
-            string functionType,
-            bool isReplay)
+            string details)
         {
                 EtwEventSource.Instance.ProcessingOutOfProcPayload(
                     taskHub,
                     LocalAppName,
                     LocalSlotName,
-                    details,
                     instanceId,
-                    functionType.ToString(),
-                    ExtensionVersion,
-                    isReplay);
+                    details,
+                    ExtensionVersion);
 
                 this.logger.LogInformation("TBD");
-
         }
 
         public void FunctionTerminated(
--- a/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
@@ -254,8 +254,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 details,
                 ExtensionVersion);
 
-            this.logger.LogInformation(
-                "{instanceId}: Function '{functionName} ({functionType})' is being replayed. Details: {details}. : {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
+            this.logger.LogDebug(
+                "{instanceId}: Function '{functionName} ({functionType})' returned the following OOProc orchestration state: {details}. : {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
                 instanceId, functionName, FunctionType.Orchestrator, details, taskHub, LocalAppName, LocalSlotName, ExtensionVersion, this.sequenceNumber++);
         }
 
--- a/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
@@ -245,18 +245,18 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             string instanceId,
             string details)
         {
-                EtwEventSource.Instance.ProcessingOutOfProcPayload(
-                    functionName,
-                    taskHub,
-                    LocalAppName,
-                    LocalSlotName,
-                    instanceId,
-                    details,
-                    ExtensionVersion);
+            EtwEventSource.Instance.ProcessingOutOfProcPayload(
+                functionName,
+                taskHub,
+                LocalAppName,
+                LocalSlotName,
+                instanceId,
+                details,
+                ExtensionVersion);
 
-                this.logger.LogInformation(
-                    "{instanceId}: Function '{functionName} ({functionType})' is being replayed. Details: {details}. : {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
-                    instanceId, functionName, FunctionType.Orchestrator, details, taskHub, LocalAppName, LocalSlotName, ExtensionVersion, this.sequenceNumber++);
+            this.logger.LogInformation(
+                "{instanceId}: Function '{functionName} ({functionType})' is being replayed. Details: {details}. : {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
+                instanceId, functionName, FunctionType.Orchestrator, details, taskHub, LocalAppName, LocalSlotName, ExtensionVersion, this.sequenceNumber++);
         }
 
         public void FunctionTerminated(
--- a/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
@@ -240,11 +240,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         }
 
         public void ProcessingOutOfProcPayload(
+            string functionName,
             string taskHub,
             string instanceId,
             string details)
         {
                 EtwEventSource.Instance.ProcessingOutOfProcPayload(
+                    functionName,
                     taskHub,
                     LocalAppName,
                     LocalSlotName,
@@ -252,7 +254,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     details,
                     ExtensionVersion);
 
-                this.logger.LogInformation("TBD");
+                this.logger.LogInformation(
+                    "{instanceId}: Function '{functionName} ({functionType})' is being replayed. Details: {details}. : {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
+                    instanceId, functionName, FunctionType.Orchestrator, details, taskHub, LocalAppName, LocalSlotName, ExtensionVersion, this.sequenceNumber++);
         }
 
         public void FunctionTerminated(
--- a/src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
+++ b/src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
@@ -401,7 +401,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.WriteEvent(224, TaskHub, AppName, SlotName, FunctionName, InstanceId, Reason, FunctionType, ExtensionVersion, IsReplay);
         }
 
-        [Event(225, Level = EventLevel.Informational)]
+        [Event(225, Level = EventLevel.Verbose)]
         public void ProcessingOutOfProcPayload(
             string FunctionName,
             string TaskHub,
--- a/src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
+++ b/src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
@@ -403,6 +403,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         [Event(225, Level = EventLevel.Informational)]
         public void ProcessingOutOfProcPayload(
+            string FunctionName,
             string TaskHub,
             string AppName,
             string SlotName,
@@ -410,7 +411,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             string Details,
             string ExtensionVersion)
         {
-            this.WriteEvent(225, TaskHub, AppName, SlotName, InstanceId, Details, FunctionType.Orchestrator, ExtensionVersion);
+            this.WriteEvent(225, FunctionName, TaskHub, AppName, SlotName, InstanceId, Details, FunctionType.Orchestrator, ExtensionVersion);
         }
 #pragma warning restore SA1313 // Parameter names should begin with lower-case letter
     }
--- a/src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
+++ b/src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
@@ -400,6 +400,20 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             this.WriteEvent(224, TaskHub, AppName, SlotName, FunctionName, InstanceId, Reason, FunctionType, ExtensionVersion, IsReplay);
         }
+
+        [Event(225, Level = EventLevel.Informational)]
+        public void ProcessingOutOfProcPayload(
+            string TaskHub,
+            string AppName,
+            string SlotName,
+            string InstanceId,
+            string Details,
+            string FunctionType,
+            string ExtensionVersion,
+            bool IsReplay)
+        {
+            this.WriteEvent(225, TaskHub, AppName, SlotName, InstanceId, Details, FunctionType, ExtensionVersion, IsReplay);
+        }
 #pragma warning restore SA1313 // Parameter names should begin with lower-case letter
     }
 }
--- a/src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
+++ b/src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
@@ -408,11 +408,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             string SlotName,
             string InstanceId,
             string Details,
-            string FunctionType,
-            string ExtensionVersion,
-            bool IsReplay)
+            string ExtensionVersion)
         {
-            this.WriteEvent(225, TaskHub, AppName, SlotName, InstanceId, Details, FunctionType, ExtensionVersion, IsReplay);
+            this.WriteEvent(225, TaskHub, AppName, SlotName, InstanceId, Details, FunctionType.Orchestrator, ExtensionVersion);
         }
 #pragma warning restore SA1313 // Parameter names should begin with lower-case letter
     }
--- a/src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
     {
         private readonly IDurableOrchestrationContext context;
 
-        private EndToEndTraceHelper TraceHelper;
+        private EndToEndTraceHelper traceHelper;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="OutOfProcOrchestrationShim"/> class.
@@ -31,11 +31,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         public OutOfProcOrchestrationShim(IDurableOrchestrationContext context)
         {
             this.context = context ?? throw new ArgumentNullException(nameof(context));
-        }
-
-        public void SetTraceHelper(EndToEndTraceHelper tracehelper)
-        {
-            this.TraceHelper = tracehelper;
         }
 
         /// <summary>
@@ -64,6 +59,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             WhenAll = 12,
         }
 
+        public void SetTraceHelper(EndToEndTraceHelper tracehelper)
+        {
+            this.traceHelper = tracehelper;
+        }
+
         // Handles replaying the Durable Task APIs that the out-of-proc function scheduled
         // with user code.
         public async Task HandleDurableTaskReplay(OrchestrationInvocationResult executionJson)
@@ -85,7 +85,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             {
                 try
                 {
-                    this.TraceHelper.ProcessingOutOfProcPayload("", "", jsonText, "", this.context.IsReplaying);
+                    this.traceHelper.ProcessingOutOfProcPayload(
+                        taskHub: "TBD",
+                        instanceId: this.context.InstanceId,
+                        details: jsonText);
+
                     jObj = JObject.Parse(jsonText);
                 }
                 catch
@@ -133,7 +137,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// <returns>If the API returns a task, the DF task corresponding to the input action. Else, null.</returns>
         private Task InvokeAPIFromAction(AsyncAction action)
         {
-
             Task fireAndForgetTask = Task.CompletedTask;
             Task task = null;
             switch (action.ActionType)
--- a/src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Threading;
 using System.Threading.Tasks;
 using DurableTask.Core.Exceptions;
@@ -27,6 +28,26 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         public OutOfProcOrchestrationShim(IDurableOrchestrationContext context)
         {
             this.context = context ?? throw new ArgumentNullException(nameof(context));
+        }
+
+        /// <summary>
+        /// Identifiers for each OOProc Schema Version.
+        /// </summary>
+        private enum SchemaVersion
+        {
+            Original = 0,
+            ReplayPatchV1 = 1,
+        }
+
+        /// <summary>
+        /// Identifiers for each kind of action object.
+        /// </summary>
+        private enum TaskGroupingType
+        {
+            Unknown = -1,
+            Atomic = 0,
+            Any = 1,
+            All = 2,
         }
 
         private enum AsyncActionType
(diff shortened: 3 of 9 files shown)


src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -104,6 +125,195 @@ internal async Task<bool> ScheduleDurableTaskEvents(OrchestrationInvocationResul
...
+                AsyncAction topLevelAction = actions[0];
+
+                if (actions.Length == 1)
+                {
+                    newTask = this.InvokeAPIFromAction(topLevelAction);
+                }
+                else
+                {
+                    // If there's more than one action, we know we're looking at a compound Task / a task of tasks
+                    // such as Task.WhenAll or Task.WhenAny. Recursively, we construct that compound task.
```
cgillum, line 228: Is this always true? Will this logic work if someone writes the following orchestrator logic?
```python
task1 = context.call_activity("foo")
task2 = context.call_activity("bar")
yield task1 # two activities were scheduled, but we're only waiting for one of them
yield task2
```

davidmrdavid: Each `yield`ed statement creates its own action array. So the snippet of code you provided would look like:

```
[ 
   [ CallActivityAction.Foo ],
   [ CallActivityAction.Bar ]
]
```
and so each of them will be detected as a single element `actions` array :)

cgillum: Perfect, thanks! :)

davidmrdavid:
Ah, glad to report this appears to fix the "missing logs" behavior we were seeing in our benchmarks. I suspect it's because `isReplaying` now can't go "out of sync" as we're more accurately representing the JS orchestration in C# :)

davidmrdavid:
I just wrapped up some e2e testing and it appears this new replay logic mitigates the "missing logs" problem I was referencing. In particular, we were prone to missing the "FunctionCompleted" log for the orchestration function, but not any more. 

I'll list the result of my experiments below.

| App Name | InstanceID | Start timestamp UTC | Replay logic | Is the FunctionCompleted log for the orchestrator missing |
| -------- | --------| -------- | -------- | -------- |
| dajustojsconsbench1 |  0d739db38a064c0d84a331c3c01256e0 | 2021-05-14T17:32:07Z |  this PR | No |
| dajustojsconsbench5 |  d50a448d362f4abfa32ec75ab75aee9a | 2021-05-14T18:27:50Z |  this PR  | No |
| dajustojsconsbench3 | 4b45525a091b41e5841d6dae7f879657  | 2021-05-14T17:54:49Z |  legacy | Yes |
| dajustojsconsbench6 |  5673ffa9a3914252896684492952d05d  | 2021-05-14T19:57:36Z |  legacy | Yes |

davidmrdavid:
I'm thinking that a good unit test would be one that checks if `isReplay` could ever be "out of sync" in an OOProc replay scenario. I could do this by capturing the ETW output as I already do in my linux logging tests and simply making sure that the "FunctionCompleted" log is present for orchestrations in a fanning scenario.

I'll get to that next

davidmrdavid:
I'll also be submitting a PR for the complementing TypeScript implementation shortly.

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -104,6 +125,195 @@ internal async Task<bool> ScheduleDurableTaskEvents(OrchestrationInvocationResul
...
+
+        /// <summary>
+        /// Recursively construct a compound Task (a task of tasks) from a list of actions.
+        /// </summary>
+        /// <param name="actions">The OOProc actions payload.</param>
+        /// <param name="index">The index at which to start processing the actions.</param>
+        /// <param name="depth">The current task depth, equal to the number of outer / parent tasks.</param>
+        /// <param name="kind">The kind of parent task: WhenAll or WhenAny.</param>
+        /// <returns>A compound task, either: WhenAll or WhenAny.</returns>
+        private (Task, int) GetCompoundTaskFromActions(AsyncAction[] actions, int index, int depth, TaskGroupingType kind)
```
ConnorMcMahon, line 248: Overall, I am struggling a bit to understand the algorithm without a concrete set of examples of walking through the algorithm and the expected output, along with the algorithm on the SDK side of how orchestrations would generate an AsyncAction[] of this format.

It seems to me that a simpler implementation for both the SDKs and the extension side would be to embed the task structure directly in the data than a flat list with metadata.

For instance, instead of each compound action being an AsyncAction[] that needs to be recursively built into a single Task, why don't we have a single `AsyncAction` with all of the current atomic types that we return, along with two new types `AnyAction` and `AllAction` which have an additional field which is an `AsyncAction[]`. If we are already having a new return schema type, this seems far simpler to both construct on the SDK side, as well as process on the extension side.

ConnorMcMahon (changes requested):
I think I need more examples of the algorithm (both on the expected inputs and how you expect to generate those inputs).

Alternatively, take a look at an alternative design I propose in the comments.

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -104,6 +125,195 @@ internal async Task<bool> ScheduleDurableTaskEvents(OrchestrationInvocationResul
...
+        /// <summary>
+        /// Invokes a DF API based on the input action object.
+        /// Most APIs will return a Task to be awaited, but others such as SignalEntity do not.
+        /// </summary>
+        /// <param name="action">An OOProc action object representing a DF task.</param>
+        /// <returns>If the API returns a task, the DF task corresponding to the input action. Else, null.</returns>
+        private Task InvokeAPIFromAction(AsyncAction action)
+        {
+            Task task = null;
+            switch (action.ActionType)
```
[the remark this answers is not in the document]
davidmrdavid: It is almost certainly identical. I did not want to modify the pre-existing `ProcessAsyncActions` to avoid accidentally breaking backwards compatibility, but you're right in that there's a wasted opportunity for code re-use in here.

I'll look into that refactor.

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
davidmrdavid: I see what you mean, and I did think about it 😄 . 
The idea you suggest, with new Action types for `WhenAny` and `WhenAll,` I believe is a better approach if we didn't have to worry about backwards compatibility.

I ended up going against this idea because introducing new Action types would require the OOProc SDKs to know what version of the durable-extension they're communicating with, and I didn't know how to _easily_ have the extension communicate the OOProc schemas it supports.

The approach I prototyped does not require the OOProc SDK to know which OOProc schemas are supported by the extension, the approach is fully backwards compatible with pre-existing logic as we're just adding metadata.

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -104,6 +125,195 @@ internal async Task<bool> ScheduleDurableTaskEvents(OrchestrationInvocationResul
...
+                    // we recurse down, processing its sequence of tasks to obtain the singular, top-level, compound task that correspond to our depth.
+                    (newTask, index) = this.GetCompoundTaskFromActions(actions, index, action.Depth, action.TaskGrouping);
+                }
+                else
+                {
+                    break;
+                }
+
+                // check for null, for when an action is fire-and-forget
+                if (newTask != null)
```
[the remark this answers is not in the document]
davidmrdavid: I'm pretty certain those methods never return tasks, that would be a bug on our end anyways, and an incorrect usage of the API

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -104,6 +116,149 @@ internal async Task<bool> ScheduleDurableTaskEvents(OrchestrationInvocationResul
...
+                    task = this.context.CallHttpAsync(action.HttpRequest);
+                    break;
+                case AsyncActionType.WhenAll:
+                    task = Task.WhenAll(mapInvokeAPIFromAction.Invoke(action.CompoundActions));
+                    break;
+                case AsyncActionType.WhenAny:
+                    task = Task.WhenAny(mapInvokeAPIFromAction.Invoke(action.CompoundActions));
+                    break;
+                default:
+                    break;
```
ConnorMcMahon, line 216: I know that this is existing from before, but at some point we may want to reconsider our silent handling of the default case.

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -104,6 +116,149 @@ internal async Task<bool> ScheduleDurableTaskEvents(OrchestrationInvocationResul
...
+                Task[] tasks = new Task[actions.Length];
+                for (var i = 0; i < actions.Length; i++)
+                {
+                    tasks[i] = this.InvokeAPIFromAction(actions[i]);
+                }
+
+                return tasks;
+            };
+
+            Task fireAndForgetTask = returnfireAndForgetTask ? Task.CompletedTask : null;
```
ConnorMcMahon, line 140: Instead of selectively returning null vs Task.CompletedTask depending on the schema, consider always returning Task.CompletedTask. I understand the desire to keep the SchemaVersion.Original behavior the same, but you can accomplish the same thing by just changing your null check to a `!= Task.CompletedTask` check. Since Task.CompletedTask is a static singleton, this will filter out the fire and forget events from the `Task.WhenAny()` performed in the legacy algorithm.

In general, without nullable references, I feel a strong aversion to returning Task or null in the same function, because whenever I see `Task` as the return type my instinct is to immediately `await` and not bother null checking.

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -104,6 +116,149 @@ internal async Task<bool> ScheduleDurableTaskEvents(OrchestrationInvocationResul
...
+        /// Invokes a DF API based on the input action object.
+        /// Most APIs will return a Task to be awaited, but others such as SignalEntity do not.
+        /// </summary>
+        /// <param name="action">An OOProc action object representing a DF task.</param>
+        /// <param name="returnfireAndForgetTask">Whether to return a `CompletedTask` to represent fire-and-forget APIs. Defaults to `true`.</param>
+        /// <returns>If the API returns a task, the DF task corresponding to the input action. Else, null.</returns>
+        private Task InvokeAPIFromAction(AsyncAction action, bool returnfireAndForgetTask = true)
+        {
+            // Helper function to obtain a list of tasks for compoundTasks.
+            Func<AsyncAction[], Task[]> mapInvokeAPIFromAction = (actions) =>
```
ConnorMcMahon, line 129: Consider moving this out into a static method, or even just replace it with a simple `.Select(this.InvokeApiFromAction)` linq query. The current approach creates `mapInvokeApiFromAction` per method execution, and considering this is a recursive function, it may be called quite often.

ConnorMcMahon (commented):
Overall, this approach seems much cleaner to me on the extension side, and I imagine it is easier to construct nested actions than the metadata approach on the SDK side as well.

The overall logic looks solid to me, just make sure to send over the preferred SchemaVersion via the orchestration input, and the actual schema version via the OutOfProcOrchestratorState. This gives us lots of flexibility in the future with Response schema changes.

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
davidmrdavid: Sweet! I'm always glad to use LINQ and write neat functional code :)

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
davidmrdavid: Good call, I actually would like to fix it _now_, I want to avoid having silent errors moving forward, as part of improving confidence in our ability to reason about OOProc behavior. I'll push something.

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -65,6 +85,11 @@ internal async Task<bool> ScheduleDurableTaskEvents(OrchestrationInvocationResul
             {
                 try
                 {
+                    this.traceHelper.ProcessingOutOfProcPayload(
+                        taskHub: "TBD",
```
davidmrdavid, line 89: I wasn't sure how to access the taskhub within the OOProcShim. I figured it's a necessity for identifying the orchestration instance in kusto, right?

src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask.xml
```diff
@@ -3451,6 +3451,36 @@
             </summary>
             <param name="context">The orchestration execution context.</param>
         </member>
+        <member name="T:Microsoft.Azure.WebJobs.Extensions.DurableTask.OutOfProcOrchestrationShim.SchemaVersion">
```
cgillum, line 3454: These are not public APIs, right? It's unfortunate that these comments potentially are getting shipped out with our public API documentation.

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -239,6 +239,26 @@ public void ExtensionWarningEvent(string hubName, string functionName, string in
...
+                    LocalAppName,
+                    LocalSlotName,
+                    instanceId,
+                    details,
+                    ExtensionVersion);
+
+                this.logger.LogInformation(
+                    "{instanceId}: Function '{functionName} ({functionType})' is being replayed. Details: {details}. : {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
+                    instanceId, functionName, FunctionType.Orchestrator, details, taskHub, LocalAppName, LocalSlotName, ExtensionVersion, this.sequenceNumber++);
+        }
```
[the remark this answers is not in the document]
davidmrdavid: I've indented them back :)

src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
```diff
@@ -400,6 +400,19 @@ private EtwEventSource()
         {
             this.WriteEvent(224, TaskHub, AppName, SlotName, FunctionName, InstanceId, Reason, FunctionType, ExtensionVersion, IsReplay);
         }
+
+        [Event(225, Level = EventLevel.Informational)]
```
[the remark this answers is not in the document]
davidmrdavid: Yes, that's right: I want to capture this in linux and, by default, we don't capture verbose logs in there.

Another option is: we could make an exception for this event in our "Verbose-filtering" logic for the linux-logger.

However, especially that this event occurs only once per replay, I'd argue it is more fitting for "informational" than "verbose"

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -104,6 +123,136 @@ internal async Task<bool> ScheduleDurableTaskEvents(OrchestrationInvocationResul
...
+            switch (schema)
+            {
+                case SchemaVersion.ReplayPatchV1:
+                    // In this schema, action arrays should be 1 dimensional (1 action per yield), but due to legacy behavior they're nested within a 2-dimensional array.
+                    if (actions.Length != 1)
+                    {
+                        throw new ArgumentException($"With OOProc schema {schema}, expected actions array to be of length 1 in outer layer but got size: {actions.Length}");
+                    }
+
+                    await this.ProcessAsyncActionsV2(actions[0]);
```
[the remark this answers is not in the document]
davidmrdavid: I've renamed `ReplayPatchV1` to just read as `V2`, for consistency

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -134,6 +134,25 @@ public override async Task<string> Execute(OrchestrationContext innerContext, st
             return serializedOutput;
         }
 
+        private void TraceOutOfProcReplay(object result)
+        {
+            string details;
+            if (result is JObject)
+            {
+                details = ((JObject)result).ToString();
```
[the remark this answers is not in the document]
davidmrdavid: I see what you mean, but we were already calling `.ToString()`  at least once per replay due to the code in "ParseOOProcResult", which preceeds this change. This change, at most calls it one more time :)

What I can do is to refactor this code such that we don't add any _extra_ calls to `toString()`. Let me get to that now

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -104,6 +123,136 @@ internal async Task<bool> ScheduleDurableTaskEvents(OrchestrationInvocationResul
...
+                    break;
+                default: // same as: case SchemaVersion.Original
+                    await this.ProcessAsyncActions(actions);
```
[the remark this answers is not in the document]
davidmrdavid: Done!

src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask.xml
davidmrdavid: None of my additions should be public, they're "internal" at most. Would those also affect public API docs? I suspect they shouldn't, but maybe I'm mistaken

src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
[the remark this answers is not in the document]
cgillum: More than just how frequently telemetry is written, the spirit of "information" is that it contains traces that tell the customer _what is happening_. This goes beyond that by dumping a bunch of internal protocol data that is intended only for us. This is why I feel like "Informational" is not really appropriate. I think it would be better to have the Linux logger handle this case specifically. However, that doesn't necessarily have to be part of this PR.

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -254,11 +248,56 @@ private void TraceAndSendExceptionNotification(string exceptionDetails)
             }
         }
 
+        private OrchestrationInvocationResult GetInvocationResult(object ooprocResult, Exception ex = null)
+        {
+            JObject resultJObject = this.ParseOOProcResult(ooprocResult);
+            string resultJSONString = resultJObject.ToString();
```
cgillum, line 254: Sometimes the OOProc result is already a string, right? Could we optimize this further by having `ParseOOProcResult` return both the string and JObject representation, and doing only the minimum amount of conversion necessary?

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -71,33 +71,9 @@ public async Task HandleDurableTaskReplay(OrchestrationInvocationResult executio
...
-            }
-
-            return jObj;
-        }
-
         internal async Task<bool> ScheduleDurableTaskEvents(OrchestrationInvocationResult result)
         {
-            JObject jObj = this.ParseOOProcResult(result);
-            var execution = JsonConvert.DeserializeObject<OutOfProcOrchestratorState>(jObj.ToString());
+            var execution = JsonConvert.DeserializeObject<OutOfProcOrchestratorState>(result.JsonString);
```
cgillum, line 76: Since we've already spent CPU cycles converting a string into a `JObject`, could we replace this redundant deserialization with a simpler call that converts the `JObject` directly into an `OutOfProcOrchestratorState`? For example:

```suggestion
            var execution = result.Json.ToObject<OutOfProcOrchestratorState>();
```

This, of course, assumes that `OrchestrationInvocationResult` has a `Json` property which is of type `JObject`.

src/WebJobs.Extensions.DurableTask/EtwEventSource.cs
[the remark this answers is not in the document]
davidmrdavid: I see what you mean, I've changed it to `Verbose` and I'll handle this special case with the linux logger in another PR

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -239,6 +239,26 @@ public void ExtensionWarningEvent(string hubName, string functionName, string in
...
+            EtwEventSource.Instance.ProcessingOutOfProcPayload(
+                functionName,
+                taskHub,
+                LocalAppName,
+                LocalSlotName,
+                instanceId,
+                details,
+                ExtensionVersion);
+
+            this.logger.LogInformation(
```
cgillum, line 257: This should also be updated to verbose.

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -239,6 +239,26 @@ public void ExtensionWarningEvent(string hubName, string functionName, string in
...
+                "{instanceId}: Function '{functionName} ({functionType})' is being replayed. Details: {details}. : {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
```
cgillum, line 258: I would change this message a bit. Instead of  "XXX is being replayed", change it to something like "XXX returned the following orchestrator actions: "

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -104,7 +99,135 @@ internal async Task<bool> ScheduleDurableTaskEvents(OrchestrationInvocationResul
...
+                case SchemaVersion.V2:
+                    // In this schema, action arrays should be 1 dimensional (1 action per yield), but due to legacy behavior they're nested within a 2-dimensional array.
+                    if (actions.Length != 1)
+                    {
+                        throw new ArgumentException($"With OOProc schema {schema}, expected actions array to be of length 1 in outer layer but got size: {actions.Length}");
+                    }
+
+                    await this.ProcessAsyncActionsV2(actions[0]);
+                    break;
+                default: // same as: case SchemaVersion.Original
```
cgillum, line 224: I wonder if you should treat `default` as an error case and have `SchemaVersion.Original` be handled explicitly. That way if we release a V3 schema and someone tries to run it on an old extension, we can correctly catch the problem and throw a meaningful exception.

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -134,6 +134,25 @@ public override async Task<string> Execute(OrchestrationContext innerContext, st
             return serializedOutput;
         }
 
+        private void TraceOutOfProcReplay(object result)
```
cgillum, line 137: Is this method being called anywhere? A quick search using Ctrl+F in my browser didn't find anything. I also noticed that  we have another call to TraceHelper.ProcessingOutOfProcPayload inside the GetInvocationResult method, and that version seems to have the optimizations I was asking for.

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -234,11 +248,66 @@ private void TraceAndSendExceptionNotification(string exceptionDetails)
...
+
         internal class OrchestrationInvocationResult
         {
             public object ReturnValue { get; set; }
 
             public Exception Exception { get; set; }
+
+            public JObject Json { get; set; }
+
+            public string JsonString { get; set; }
```
cgillum, line 310: Tip: I recommend making all these properties read-only (removing the setters) and have it get fully initialized in a constructor. That way we can leverage the compiler to help us guarantee that these last two properties are never null. Immutable types FTW!

cgillum (commented):
This is looking great. Thank you for considering my feedback. A few other small things and I think I'm good to sign off.

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
davidmrdavid: Ah yes, sorry I missed this, too much multi-tasking. Fixed!

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
davidmrdavid: I went for "returned the following OOProc orchestration state" because we're not just logging the actions, that would be more expensive that simply outputting the orchestration state, which only contains ~3 more fields in addition to the actions array

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
davidmrdavid: Good call, done!

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
davidmrdavid: You're right, this wasn't being used anymore. I thought Visual Studio would have complained about this, interesting 🤔

src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
davidmrdavid: Of course ❤️ ! I made them part of a constructor, but I couldn't get the project to compile with public-readonly fields: Visual Studio really wanted them to be private if I added the `readonly` modifier, which meant I would need to add some boilerplate by creating getters for each property. 

So instead I went for public fields with private setters, let me know what you think!

cgillum: This is the syntax for read-only properties in C# (you just need to remove `set`):
```csharp
public JObject Json { get; }

public string JsonString { get; }
```

This is even better than private setters since not even internal code can modify them later.

davidmrdavid: Oh I see, my mistake was to additionally add the `readonly` keyword. I'll refactor this now

ConnorMcMahon: You may want to do some final validation here end-to-end. I don't think Newtonsoft can deserialize fields that have no setter method...

ConnorMcMahon: [This](https://stackoverflow.com/questions/4066947/private-setters-in-json-net) may be the best guidance. If we could use C# 9, you would want init-setters, but I don't think we can get away with that currently.

Maybe a combination of read-only properties with a constructor with the correct parameter names would work?

cgillum: Are these objects ever deserialized? It looked to me like they're only created in code.

ConnorMcMahon: Ah good call, I was mixing this object up with `OutOfProcOrchestratorState`

ConnorMcMahon (approved):
LGTM!

PR #1838, opened 2021-05-17
Upgrade AzureStorage to 1.8.5 in V1

davidmrdavid:
Unblocks: https://github.com/Azure/azure-functions-durable-extension/pull/1802

This PR simply upgrades the V1 dependency on AzureStorage from 1.7.7 to 1.8.1.
I decided against upgrading to version 1.8.5 because, starting from version 1.8.2, there are serialization changes in azureStorage that make some of our unit tests to fail. Upgrading to version 1.8.5 will require more careful cherry-picking of commits, as well as expertise in the serialization changes that took place.

### Issue describing the changes in this PR: N/A

resolves N/A

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableActivityContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableActivityContext.cs
@@ -82,6 +82,11 @@ namespace Microsoft.Azure.WebJobs
             }
 
             JToken jToken = this.GetInputAsJson();
+            if (jToken == null)
+            {
+                return null;
+            }
+
             var value = jToken as JValue;
             if (value != null)
             {
--- a/src/WebJobs.Extensions.DurableTask/MessagePayloadDataConverter.cs
+++ b/src/WebJobs.Extensions.DurableTask/MessagePayloadDataConverter.cs
@@ -50,6 +50,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// </summary>
         public string Serialize(object value, int maxSizeInKB)
         {
+            if (value == null)
+            {
+                return null;
+            }
+
             string serializedJson;
 
             JToken json = value as JToken;
--- a/src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
+++ b/src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
@@ -52,7 +52,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.DurableTask.AzureStorage" Version="1.8.1" />
+    <PackageReference Include="Microsoft.Azure.DurableTask.AzureStorage" Version="1.8.5" />
   </ItemGroup>
 
   <!-- NuGet Publishing Metadata -->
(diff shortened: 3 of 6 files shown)


cgillum:
Can you elaborate on this?
> starting from version 1.8.2, there are serialization changes in azureStorage that make some of our unit tests to fail

ConnorMcMahon:
I believe [these](https://github.com/Azure/azure-functions-durable-extension/commit/9bf54d002fca9bd85ca35e99a18a0d43de35e585#diff-600040e5159d2e1fcc9eb7497eb0f7f20da4d4fc72993711a33d50634234a8eb) are the changes required to work with 1.8.2/1.8.3. I would prefer to take on these changes, as 1.8.2-1.8.7 have some reliability improvements it would be nice if we could have on the 1.x side.

davidmrdavid:
@cgillum, I observed in https://github.com/Azure/azure-functions-durable-extension/pull/1802 that upgrading to AzureStorage 1.8.5 caused the CI to fail. One of the tests that was failing was `SetStatusOrchestration` [here](https://github.com/Azure/azure-functions-durable-extension/blob/babe19e13a76178b5d7440ea64951d67924c6447/test/Common/DurableTaskEndToEndTests.cs#L2655). I suspect other tests fail as well, but this one consistently fails starting in AzureStorage 1.8.2

_Why are the tests failing?_
It appears there are some null-serialization changes in AzureStorage 1.8.2 which, in durable-extension V2, expect some serialization utilities and scenarios that haven't been ported yet to V1. The test for `SetStatusOrchestration` directly tests this `null` serialization scenario, and so now there's some faulty behavior.

@ConnorMcMahon, I did try cherry-picking that PR but, while the PR diff itself seems minimal, when cherry-picking it to v1, the entire `DurableActivityContext.cs` file gets overriden, because the changes are so great. Even after merging, the code will not compile because there are utilities like the `MessageDataConverter` and `ErrorDataConverter` [here](https://github.com/Azure/azure-functions-durable-extension/blob/1e5d87f714f0d6ad01cbac7d8a9ac3c58aebf1fb/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs#L165-L166) which do not exist in V1. After that, I looked at the git blame annotations for the relevant missing classes and interfaces and it lead me down a rabbit whole of PRs which will take some time to sort out, in my opinion. I'm happy to be shown otherwise though :)

davidmrdavid:
@cgillum: what I meant was that, in https://github.com/Azure/azure-functions-durable-extension/pull/1802, I noticed that upgrading AzureStorage to v1.8.5 caused a few test cases to fail. An example is the test "SetStatusOrchestration", which tests null serialization logic. The reason for the new failures were that we handle `null` values differently starting from AzureStorage v1.8.3 or so.

@ConnorMcMahon, before originally opening this PR, I did try cherry-picking that PR you linked. However, doing so led me to discovering that there are very many serialization utilities that were not ported to V1, as the code did not easily compile, and those utilities were not made available after cherry-picking that commit.  That, and trying to follow the trail of  PRs around data serialization in V2, made me believe that it would take quite a bit of effort to get the new serialization logic working with V1.

However, I then decided to try **not** to cherry pick the PR you linked and instead **manually** apply the change myself, and it worked! So, to summarize, the utilities that handle serialization in V1 and V2 have differed greatly, making cherry-picking merges difficult. However, there were still enough similarities to allow me to manually patch the right places to make azureStorage v1.8.5 work.

ConnorMcMahon (approved):
LGTM!

PR #1848, opened 2021-05-25
Fix custom backend provider logging on Linux

### LINKED ISSUE #1796: Fix support for ETW tracing on linux and custom backends
### Description

There's a bug in the linux trace collection which means ETW event sources inside custom backends are not collected.

The problem is in EventSourceListener.cs. Specifically, OnEventSourceCreated is called before the constructor EventSourceListener is called, thus the field durabilityProviderEventSourceName has not been initialized yet. So the comparison always fails.

ConnorMcMahon:
If a custom durability provider uses its ETW event source before we
instantiate the Linux logging architecture (like Netherite's ETW event
for instantiating its orchestration service), then the
OnEventSourceCreated method will call before our constructor finishes
instantiating the event source name for the custom provider. This
happens due to the ordering of constructor calls in .NET.

To work around this, we temporarily store all event sources before our
constructor establishes the durability provider event source name so
that we can enable the correct provider when we do establish the provider name.

resolves #1796

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.durabilityProviderEventSourceName = durabilityProviderEventSourceName;
 
             // Check to see if any event sources were created before we knew the event source
-            //// name for the durability provider and enable that provider.
+            // name for the durability provider and enable that provider.
             var eventSourcesToEnable = this.pendingEventSources.Where(eventSource => eventSource.Name == this.durabilityProviderEventSourceName);
             foreach (var eventSource in eventSourcesToEnable)
             {
--- a/src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
@@ -2,7 +2,9 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics.Tracing;
+using System.Linq;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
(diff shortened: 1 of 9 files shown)


src/WebJobs.Extensions.DurableTask/EventSourceListener.cs
```diff
@@ -32,6 +36,16 @@ public EventSourceListener(LinuxAppServiceLogger logger, bool enableVerbose, End
             this.disableVerbose = !enableVerbose; // We track the opposite value ro simplify logic later
             this.traceHelper = traceHelper;
             this.durabilityProviderEventSourceName = durabilityProviderEventSourceName;
+
+            // Check to see if any event sources were created before we knew the event source
+            //// name for the durability provider and enable that provider.
```
ConnorMcMahon, line 41: woops, extra comment `//` here.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -299,6 +300,55 @@ public async Task AzureStorageEmittingLogsWithEventSourceName()
...
+                    var client = await host.StartOrchestratorAsync(orchestratorName, input: "World", this.output);
+                    var status = await client.WaitForCompletionAsync(this.output);
+                    await host.StopAsync();
+                }
+
+                string consoleOutput = sw.ToString();
+
+                // Validate that the JSON has DurableTask-AzureStorage fields
+                string[] lines = consoleOutput.Split('\n');
+                var azureStorageLogLines = lines.Where(l => l.Contains("DurableTask-CustomSource") && l.StartsWith(prefix));
```
ConnorMcMahon, line 347: should change name of variable here.

davidmrdavid (commented):
Thank you for this change! Seems like there are a few minor typos to address and I also left some comments.

Also, a question: Wouldn't it be possible to read the value for the durability provider from the app configuration itself inside of `OnEventSourceCreated`? 
Say, would it be possible to read it as an environment variable when `this.durabilityProviderEventSourceName` is `null`?

I think that would be a cleaner approach.

davidmrdavid (approved):
LGTM so long as CI goes 🟢

bachuv (approved):

PR #1850, opened 2021-05-25
Fix code-gen intialization in Functions V1

ConnorMcMahon:
TypedCodeProvider was not properly initialized in FunctionsV1, leading to a null reference exception for FunctionsV1 scenarios. This PR properly initializes this type for FunctionsV1.


cgillum:
This is a little confusing. Currently in the constructor we have code that initializes the TypedCodeProvider. In that same constructor, we also have some code that does checks for the `FUNCTIONS_V1` compiler constant. Does Functions V1 not actually hit that constructor ever (in which case those compiler flag checks are misleading)?

ConnorMcMahon:
@cgillum 

From what I can tell, the constructor under the compilation flag `FUNCTIONS_V1` is not used in Functions V1, making me think it is dead code. I tested this by putting a break point in the constructor while running the Functions V1 smoke tests. 

The code that is actually being run is the `[Obsolete]` constructor and the InitializeForFunctionsV1 codepath. The Obsolete attribute and comment is actually incorrect there.

We should definitely fix this, though I think tackling that post 2.5.0 release (along with some tests that actually go through this code path, would be a great way to give us more confidence we aren't breaking V1 scenarios.

cgillum (approved):

PR #1854, opened 2021-05-26
Disable OOProc State Transfer Logging

davidmrdavid:
Follow-up to: https://github.com/Azure/azure-functions-durable-extension/pull/1836
Disables OOProc state transfer logging until we find a less verbose alternative.


src/WebJobs.Extensions.DurableTask/Listener/TaskOrchestrationShim.cs
```diff
@@ -230,11 +230,12 @@ private void TraceAndSendExceptionNotification(string exceptionDetails)
         private async Task TraceAndReplay(object result, Exception ex = null)
         {
             var invocationResult = new OrchestrationInvocationResult(result, ex);
-            this.Config.TraceHelper.ProcessingOutOfProcPayload(
+            /* this.Config.TraceHelper.ProcessingOutOfProcPayload(
```
ConnorMcMahon, line 233: Let's delete it, and we can always re-add it later.

ConnorMcMahon (approved):

PR #1889, opened 2021-07-09
Adding Name to IDurableActivityContext

adstep:
Bringing ```IDurableActivityContext``` in parity with ```IDurableOrchestrationContext``` by adding a ```Name``` property to track the current executing function's name.


cgillum (approved):
I'm good with this addition. @ConnorMcMahon?

ConnorMcMahon:
I definitely like the idea of adding this field to IDurableActivityContext. I am a little hesitant on the timing, because it technically is a breaking change for customers who mock IDurableActivityContext without using Moq. This is against our official recommendations for testing, so it's uncommon, but I generally like to reserve new properties/methods on interfaces to minor releases because of this.

PR #1890, opened 2021-07-13
Dynamic concurrency support for Durable Functions

cgillum:
This PR adds dynamic concurrency support for the Durable Functions extension. It leverages new extensibility being added into the WebJobs SDK that enables throttling when VM resources like CPU or memory reach low levels.

More details to be provided later (e.g. how to enable it, etc.) once those details are available.

This PR requires changes from both the WebJobs SDK and an unreleased version of **Microsoft.Azure.DurableTask.Core** (see https://github.com/Azure/durabletask/pull/564).

FYI @mathewc


cgillum:
Closing as this isn't being actively worked on and because more design work is needed.

PR #1895, opened 2021-07-22
Allow custom connection string names when creating a DurableClient in an ASP.NET Core app

### LINKED ISSUE #1656: DurableClientOptions not honoured if using custom connection string name in aspnet core app when using AddDurableClientFactory extention method
What I am trying to achieve is to use a custom connection string name and taskhub name.

So I want the DurableClientFactory.CreateClient() method to honor both of these values.

I have tried to configure this in startup as follows:

`services.AddDurableClientFactory((config)=> {
                config.ConnectionName = "MyCustomConnectionStringName";
                config.TaskHub = "MyTaskHubName";
            });`

In this scenario the taskhub name is honoured but the connection is not.  I created my own implementation of IConnectionStringResolver and was able to see the ConnectionStringName which was passed to this service was named "Storage".

After a little digging through the source code I tried the following:

`            services.Configure<DurableTaskOptions>(config => {
                config.StorageProvider.Add("ConnectionStringName", "MyCustomConnectionStringName");
            });`

After doing this the connection string is now honoured but is requiring a TaskHub name to be configured.  

I cannot no matter what I try get both of the values to be configured with my custom values.  

I have even tried passing values directly to the CreateClient method on the DurableClientFactory class:

`_durableClientOptions = new DurableClientOptions
  {
      ConnectionName = "MyCustomConnectionStringName",
      TaskHub = "MyTaskHubName"
  };
`
` var durableClient = _durableClientFactory.CreateClient(_durableClientOptions);`

In this case again the ConnectionName is ignored and replaced with "Storage".

----------------------------

After pulling down the source code and debugging through it appears that this function in AzureStorageDurabilityProviderFactory.cs

```
public virtual DurabilityProvider GetDurabilityProvider()
        {
            this.EnsureInitialized();
            if (this.defaultStorageProvider == null)
            {
                var defaultService = new AzureStorageOrchestrationService(this.defaultSettings);
                this.defaultStorageProvider = new AzureStorageDurabilityProvider(
                    defaultService,
                    this.defaultConnectionName,
                    this.azureStorageOptions);
            }
return this.defaultStorageProvider;
}
```

invokes `this.EnsureInitialized();`

which in turn invokes the following method in the same class.
`                this.defaultSettings = this.GetAzureStorageOrchestrationServiceSettings();`

As you can see the method takes in no parameters and the method itself is described as follows:

```       
internal AzureStorageOrchestrationServiceSettings GetAzureStorageOrchestrationServiceSettings(
            string connectionName = null,
            string taskHubNameOverride = null)
        {
            connectionName = connectionName ?? this.defaultConnectionName;

            string resolvedStorageConnectionString = this.connectionStringResolver.Resolve(connectionName);

            if (string.IsNullOrEmpty(resolvedStorageConnectionString))
            {
                throw new InvalidOperationException("Unable to find an Azure Storage connection string to use for this binding.");
            }
......
```

As you can see this will always throw an error if there is no default connection string name "Storage" which is the value set in this.defaultConnectionName.

In addition when consuming this service via non web job (standard aspnet application) the extention method defaults to using the implementation of IConnectionStringProvider as StandardConnectionStringProvider.

After looking the source code it appears that the standard implementation would be better suited using WebJobsConnectionStringProvider this will locate connection string via both local.settings.json:Values:ConnString in webjob and appsettings.json:ConnectionStrings:ConnString in standard ASP.NET application and therefore feels like the naming convention of these implementations are reversed.

bachuv:
This PR allows customers to use custom app setting names to create a DurableClient in an ASP.NET Core app.
 
```
_client = clientFactory.CreateClient(new DurableClientOptions
          {
              ConnectionName = "Storage",
              TaskHub = configuration["TaskHub"]
          });
```
Currently, customers have to set `ConnectionName` to "Storage" in `DurableClientOptions` because of [this line](https://github.com/Azure/azure-functions-durable-extension/blob/cc7f8b20844a07b0729a3ae66bcc5dcf156baaf3/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs#L158) in AzureStorageDurabilityProviderFactory.cs. `connectionName` is always set to null which means that it's always set to the `this.defaultConnectionName`.  `this.defaultConnectionName` defaults to using `ConnectionStringNames.Storage` which is "Storage" because this.azureStorageOptions.ConnectionStringName` is null in this flow.
 
This PR passes the `ConnectionName` and `TaskHub` name from `DurableClientOptions` to `GetAzureStorageOrchestrationServiceSettings()` to create an `AzureStorageDurabilityProvider` using the configured values instead of default values.
 
Resolves #1656
 
Note: This is a draft PR because the tests are not fully setup yet.
 
### Pull request checklist
 
### DIFF
--- a/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
+++ b/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
@@ -83,7 +83,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         // This method should not be called before the app settings are resolved into the options.
         // Because of this, we wait to validate the options until right before building a durability provider, rather
         // than in the Factory constructor.
-        private void EnsureInitialized(string connectionName = null, string taskHubNameOverride = null)
+        private void EnsureInitialized(bool isExternalClient = false)
         {
             if (!this.hasValidatedOptions)
             {
@@ -96,7 +96,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     this.options.SetDefaultHubName(sanitizedHubName);
                 }
 
-                this.defaultSettings = this.GetAzureStorageOrchestrationServiceSettings(connectionName, taskHubNameOverride);
+                if (!isExternalClient)
+                {
+                    this.defaultSettings = this.GetAzureStorageOrchestrationServiceSettings();
+                }
+
                 this.hasValidatedOptions = true;
             }
         }
@@ -120,7 +124,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         public virtual DurabilityProvider GetDurabilityProvider(DurableClientAttribute attribute)
         {
-            this.EnsureInitialized(attribute.ConnectionName, attribute.TaskHub);
+            this.EnsureInitialized(attribute.ExternalClient);
             return this.GetAzureStorageStorageProvider(attribute);
         }
 
@@ -130,9 +134,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             AzureStorageOrchestrationServiceSettings settings = this.GetAzureStorageOrchestrationServiceSettings(connectionName, attribute.TaskHub);
 
             AzureStorageDurabilityProvider innerClient;
-            if (string.Equals(this.defaultSettings.TaskHubName, settings.TaskHubName, StringComparison.OrdinalIgnoreCase) &&
-                string.Equals(this.defaultSettings.StorageConnectionString, settings.StorageConnectionString, StringComparison.OrdinalIgnoreCase) &&
-                this.defaultStorageProvider != null)
+            if (!attribute.ExternalClient &&
+                string.Equals(this.defaultSettings?.TaskHubName, settings.TaskHubName, StringComparison.OrdinalIgnoreCase) &&
+                string.Equals(this.defaultSettings?.StorageConnectionString, settings.StorageConnectionString, StringComparison.OrdinalIgnoreCase))
             {
                 // It's important that clients use the same AzureStorageOrchestrationService instance
                 // as the host when possible to ensure we any send operations can be picked up
--- a/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
+++ b/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
@@ -83,7 +83,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         // This method should not be called before the app settings are resolved into the options.
         // Because of this, we wait to validate the options until right before building a durability provider, rather
         // than in the Factory constructor.
-        private void EnsureInitialized(bool isExternalClient = false)
+        private void EnsureDefaultClientSettingsInitialized()
         {
             if (!this.hasValidatedOptions)
             {
@@ -96,18 +96,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     this.options.SetDefaultHubName(sanitizedHubName);
                 }
 
-                if (!isExternalClient)
-                {
-                    this.defaultSettings = this.GetAzureStorageOrchestrationServiceSettings();
-                }
-
+                this.defaultSettings = this.GetAzureStorageOrchestrationServiceSettings();
                 this.hasValidatedOptions = true;
             }
         }
 
         public virtual DurabilityProvider GetDurabilityProvider()
         {
-            this.EnsureInitialized();
+            this.EnsureDefaultClientSettingsInitialized();
             if (this.defaultStorageProvider == null)
             {
                 var defaultService = new AzureStorageOrchestrationService(this.defaultSettings);
@@ -124,7 +120,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         public virtual DurabilityProvider GetDurabilityProvider(DurableClientAttribute attribute)
         {
-            this.EnsureInitialized(attribute.ExternalClient);
+            if (!attribute.ExternalClient)
+            {
+                this.EnsureDefaultClientSettingsInitialized();
+            }
+
             return this.GetAzureStorageStorageProvider(attribute);
         }
 
@@ -134,9 +134,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             AzureStorageOrchestrationServiceSettings settings = this.GetAzureStorageOrchestrationServiceSettings(connectionName, attribute.TaskHub);
 
             AzureStorageDurabilityProvider innerClient;
-            if (!attribute.ExternalClient &&
-                string.Equals(this.defaultSettings?.TaskHubName, settings.TaskHubName, StringComparison.OrdinalIgnoreCase) &&
-                string.Equals(this.defaultSettings?.StorageConnectionString, settings.StorageConnectionString, StringComparison.OrdinalIgnoreCase))
+
+            // Need to check this.defaultStorageProvider != null for external clients that call GetDurabilityProvider(attribute)
+            // which never initializes the defaultStorageProvider.
+            if (string.Equals(this.defaultSettings?.TaskHubName, settings.TaskHubName, StringComparison.OrdinalIgnoreCase) &&
+                string.Equals(this.defaultSettings?.StorageConnectionString, settings.StorageConnectionString, StringComparison.OrdinalIgnoreCase) &&
+                this.defaultStorageProvider != null)
             {
                 // It's important that clients use the same AzureStorageOrchestrationService instance
                 // as the host when possible to ensure we any send operations can be picked up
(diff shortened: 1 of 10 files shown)


test/Common/ClientFunctions.cs
```diff
@@ -45,6 +45,26 @@ internal static class ClientFunctions
                 instanceCreationTime);
         }
 
+        // NOTE: The TestStorageConnectionString app setting adn TestTaskHub app setting name must exist
```
ConnorMcMahon, line 48: nit: "and" instead of "adn"

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
```diff
@@ -130,8 +134,9 @@ private AzureStorageDurabilityProvider GetAzureStorageStorageProvider(DurableCli
             AzureStorageOrchestrationServiceSettings settings = this.GetAzureStorageOrchestrationServiceSettings(connectionName, attribute.TaskHub);
 
             AzureStorageDurabilityProvider innerClient;
-            if (string.Equals(this.defaultSettings.TaskHubName, settings.TaskHubName, StringComparison.OrdinalIgnoreCase) &&
-                string.Equals(this.defaultSettings.StorageConnectionString, settings.StorageConnectionString, StringComparison.OrdinalIgnoreCase))
+            if (!attribute.ExternalClient &&
```
ConnorMcMahon, line 137: I now remember that some customers use the ExternalClient attribute for external clients that use the DurableClient input binding, not just the factory. In that case, they technically don't need to provider a TaskHub name and/or storage connection string, as we just use the default values.

I think we are OK with not using the default storage provider in that case however, as the comment below really just wants us to use the defaultStorageProvider for clients that are running on the same instance as the triggers to bypass queue polling.

test/FunctionsV2/TestCustomConnectionsStringResolver.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+
+namespace WebJobs.Extensions.DurableTask.Tests.V2
```
ConnorMcMahon, line 9: How does this differ from our default connection string resolver injected by our tests?

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
```diff
@@ -41,7 +41,7 @@ public static IServiceCollection AddDurableClientFactory(this IServiceCollection
             }
 
             serviceCollection.TryAddSingleton<INameResolver, DefaultNameResolver>();
-            serviceCollection.TryAddSingleton<IConnectionStringResolver, StandardConnectionStringProvider>();
+            serviceCollection.TryAddSingleton<IConnectionStringResolver, WebJobsConnectionStringProvider>();
```
ConnorMcMahon, line 44: I am a little concerned about changing this, just because some customers will be running in non-webjobs scenario.

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
```diff
@@ -83,7 +83,7 @@ internal string GetDefaultStorageConnectionString()
         // This method should not be called before the app settings are resolved into the options.
         // Because of this, we wait to validate the options until right before building a durability provider, rather
         // than in the Factory constructor.
-        private void EnsureInitialized()
+        private void EnsureInitialized(bool isExternalClient = false)
```
ConnorMcMahon, line 86: What are the values of this.options and this.azureStorageOptions when we are using external clients? I'm just wondering if we even need to try and validate those for external client scenarios.

One other niche edge case about setting `this.hasValidationOptions=true` for external clients is the case where the customer has both external clients and a a default durability provider used in the same app. In that case, if the external client is created first (which may not be possible?), we would never instantiate this.defaultSettings. This wouldn't likely have a huge impact, but we may just be better renaming this method `EnsureDefaultClientSettingsInitialized()` and then only call it if ExternalClient is false.

ConnorMcMahon (commented):
I think this is very close.

I have a couple of minor production code suggestions, and a few questions/suggestions about the tests.

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
bachuv: They have default values. For example, `this.options.HubName` is "TestTaskHub" and `this.azureStorageOptions.ConnectionStringName` is `null`. `EnsureInitialized()` doesn't seem to have an impact on external clients because there's already a check for a task hub name in `CreateClient()` from `DurableClientFactory`. The storage connection string name check happens in `GetAzureStorageOrchestrationServiceSettings()` though.

I see, I'll rename the method and add a check in `GetDurabilityProvider(attribute)` to only call `EnsureDefaultClientSettingsInitialized()` if it's not an external client.

test/FunctionsV2/TestCustomConnectionsStringResolver.cs
bachuv: The `TestConnectionStringResolver` always returns `TestHelpers.GetStorageConnectionString()` regardless of the `connectionStringName` that's passed in. This new one stores a dictionary of connection strings.

https://github.com/Azure/azure-functions-durable-extension/blob/cc7f8b20844a07b0729a3ae66bcc5dcf156baaf3/test/Common/TestConnectionStringResolver.cs#L10

ConnorMcMahon: Can we just replace TestConnectionStringResolver()'s logic with this implementation (with a fallback to TestHlperes.GetStorageConnectionString() if the app setting doesn't exist?

bachuv: For this scenario, the code path expects `null` when an app setting doesn't exist. It then throws an `InvalidOperationException` which fails the test. If the connection string resolver returned `TestHelpers.GetStorageConnectionString()` as a default then the test would still pass because it got a connection string, but not the right one (passed through app settings).

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs
bachuv: Would it be better to go back to checking if `defaultStorageProvider` is not null instead of checking `attribute.ExternalClient`?

ConnorMcMahon: in general, they should be synonymous, but it is probably more robust to check for the nullness of `defaultStorageProvider`. Let's consider moving it back to checking if `defaultStorageProvider` is null, with a comment explaining when it would be null and why it is ok to not return it in those scenarios.

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
bachuv: I initially changed it because of this comment in the [linked issue](https://github.com/Azure/azure-functions-durable-extension/issues/1656#issue-790126163):

> In addition when consuming this service via non web job (standard aspnet application) the extention method defaults to using the implementation of IConnectionStringProvider as StandardConnectionStringProvider.
> 
> After looking the source code it appears that the standard implementation would be better suited using WebJobsConnectionStringProvider this will locate connection string via both local.settings.json:Values:ConnString in webjob and appsettings.json:ConnectionStrings:ConnString in standard ASP.NET application and therefore feels like the naming convention of these implementations are reversed.

I can change it back to `StandardConnectionStringProvider`. I looked through `StandardConnectionStringProvider` and `WebJobsConnectionStringProvider` to understand this comment and the only difference I see is that `WebJobsConnectionStringProvider` evaluates `configuration.GetConnectionString(connectionName)` first before `configuration[connectionName]`. I'm not exactly clear on what `configuration.GetConnectionString(connectionName)` does/where it looks for connection strings. Let me know if you think `WebJobsConnectionStringProvider` is fine otherwise I'll change it to `StandardConnectionStringProvider`

https://github.com/Azure/azure-functions-durable-extension/blob/cc7f8b20844a07b0729a3ae66bcc5dcf156baaf3/src/WebJobs.Extensions.DurableTask/StandardConnectionStringProvider.cs#L31

https://github.com/Azure/azure-functions-durable-extension/blob/cc7f8b20844a07b0729a3ae66bcc5dcf156baaf3/src/WebJobs.Extensions.DurableTask/WebJobsConnectionStringProvider.cs#L29

https://github.com/Azure/azure-webjobs-sdk/blob/4a16c273b40ee47ec318bb6d7aa12e0075f30402/src/Microsoft.Azure.WebJobs.Host/Extensions/IConfigurationExtensions.cs#L79

ConnorMcMahon (approved):
LGTM!

PR #1914, opened 2021-08-11
Refactor PlatformInformationProvider

davidmrdavid:
Refactors the `PlatformInformationProvider` to utilize `Enum`s more heavily to represent the enviroment. This allows for a pattern-matching-style approach to platform-aware configuration, and requires fewer methods to maintain.

N/A

### Pull request checklist

* [] My changes **do not** need to be backported to a previous version

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DefaultPlatformInformationProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/DefaultPlatformInformationProvider.cs
@@ -20,18 +20,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.nameResolver = nameResolver;
         }
 
-        public bool InConsumption()
-        {
-            return this.InLinuxConsumption() | this.InWindowsConsumption();
-        }
-
-        public bool InWindowsConsumption()
+        private bool InWindowsConsumption()
         {
             string value = this.nameResolver.Resolve("WEBSITE_SKU");
             return string.Equals(value, "Dynamic", StringComparison.OrdinalIgnoreCase);
         }
 
-        public bool InLinuxConsumption()
+        private bool InLinuxConsumption()
         {
             string containerName = this.GetContainerName();
             string azureWebsiteInstanceId = this.nameResolver.Resolve("WEBSITE_INSTANCE_ID");
@@ -40,13 +35,67 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             return inLinuxConsumption;
         }
 
-        public bool InLinuxAppService()
+        private bool InLinuxAppService()
         {
             string azureWebsiteInstanceId = this.nameResolver.Resolve("WEBSITE_INSTANCE_ID");
             string functionsLogsMountPath = this.nameResolver.Resolve("FUNCTIONS_LOGS_MOUNT_PATH");
             bool inAppService = !string.IsNullOrEmpty(azureWebsiteInstanceId);
             bool inLinuxDedicated = inAppService && !string.IsNullOrEmpty(functionsLogsMountPath);
             return inLinuxDedicated;
+        }
+
+        public OperatingSystem GetOperatingSystem()
+        {
+            if (this.InLinuxConsumption() || this.InLinuxAppService())
+            {
+                return OperatingSystem.Linux;
+            }
+
+            return OperatingSystem.Windows;
+        }
+
+        public AppServicePlan GetAppSevicePlan()
+        {
+            if (this.InLinuxConsumption() | this.InWindowsConsumption())
+            {
+                return AppServicePlan.Consumption;
+            }
+
+            return AppServicePlan.AppService;
+        }
+
+        public ProgLanguage GetProgLanguage()
+        {
+            string workerRuntime = this.nameResolver.Resolve("FUNCTIONS_WORKER_RUNTIME");
+            if (workerRuntime == "python")
+            {
+                return ProgLanguage.Python;
+            }
+            else if (workerRuntime == "javascript")
+            {
+                return ProgLanguage.JavaScript;
+            }
+            else if (workerRuntime == "powershell")
+            {
+                return ProgLanguage.PowerShell;
+            }
+            else if (workerRuntime == "csharp")
+            {
+                return ProgLanguage.Csharp;
+            }
+
+            throw new Exception("Could not determine user-level programming language.");
+        }
+
+        public bool IsOutOfProc()
+        {
+            ProgLanguage progLanguage = this.GetProgLanguage();
+            if (progLanguage != ProgLanguage.Csharp)
+            {
+                return true;
+            }
+
+            return false;
         }
 
         public string GetLinuxTenant()
@@ -62,11 +111,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         public string GetContainerName()
         {
             return this.nameResolver.Resolve("CONTAINER_NAME");
-        }
-
-        public bool IsPython()
-        {
-            return this.nameResolver.Resolve("FUNCTIONS_WORKER_RUNTIME") == "python";
         }
     }
 }
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -74,7 +74,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         private readonly bool isOptionsConfigured;
         private readonly IApplicationLifetimeWrapper hostLifetimeService = HostLifecycleService.NoOp;
 #pragma warning disable CS0612 // Type or member is obsolete
-        private IPlatformInformationService platformInformationService;
+#pragma warning disable SA1401 // Fields should be private
+        internal IPlatformInformationService PlatformInformationService;
+#pragma warning restore SA1401 // Fields should be private
 #pragma warning restore CS0612 // Type or member is obsolete
         private IDurabilityProviderFactory durabilityProviderFactory;
         private INameResolver nameResolver;
@@ -143,7 +145,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.Options = options?.Value ?? new DurableTaskOptions();
             this.nameResolver = nameResolver ?? throw new ArgumentNullException(nameof(nameResolver));
             this.loggerFactory = loggerFactory ?? throw new ArgumentNullException(nameof(loggerFactory));
-            this.platformInformationService = platformInformationService ?? throw new ArgumentNullException(nameof(platformInformationService));
+            this.PlatformInformationService = platformInformationService ?? throw new ArgumentNullException(nameof(platformInformationService));
             this.ResolveAppSettingOptions();
 
             ILogger logger = loggerFactory.CreateLogger(LoggerCategoryName);
@@ -311,7 +313,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 #if !FUNCTIONS_V1
             // .NET461 is not supported in linux, so this is conditionally compiled
             // We initialize linux logging early on in case any initialization steps below were to trigger a log event.
-            this.InitializeLinuxLogging();
+            if (this.PlatformInformationService.GetOperatingSystem() == OperatingSystem.Linux)
+            {
+                this.InitializeLinuxLogging();
+            }
 #endif
 
             ConfigureLoaderHooks();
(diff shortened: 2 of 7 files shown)


src/WebJobs.Extensions.DurableTask/DefaultPlatformInformationProvider.cs
```diff
@@ -49,6 +44,60 @@ public bool InLinuxAppService()
...
+        {
+            if (this.InLinuxConsumption() || this.InLinuxAppService())
+            {
+                return OperatingSystem.Linux;
+            }
+
+            return OperatingSystem.Windows;
+        }
+
+        public AppServicePlan GetAppSevicePlan()
```
cgillum, line 57: The name "AppServicePlan" is misleading because it has a very specific meaning in the context of Azure (it refers to a _name_ of a plan, and excludes Consumption and ElasticPremium plan types, which are technically distinct from "App Service plans"). Consider "PlanType" instead for both the name of this enum and the name of this method.

src/WebJobs.Extensions.DurableTask/DefaultPlatformInformationProvider.cs
```diff
@@ -49,6 +44,60 @@ public bool InLinuxAppService()
...
+        {
+            if (this.InLinuxConsumption() | this.InWindowsConsumption())
```
cgillum, line 59: Seems like `||` (boolean operator with shortcircuiting) would be more appropriate than `|` here.

```suggestion
            if (this.InLinuxConsumption() || this.InWindowsConsumption())
```

src/WebJobs.Extensions.DurableTask/IPlatformInformationService.cs
```diff
@@ -5,6 +5,64 @@
...
+        /// <summary>
+        /// Windows OS
+        /// </summary>
+        Windows,
+    }
+
+    /// <summary>
+    /// Representation of supported Programming Languages.
+    /// </summary>
+    public enum ProgLanguage
```
cgillum, line 43: The name `ProgLanguage` is misleading because the programming language itself is irrelevant. What matters is the stack (.NET, Node.js, Python, PowerShell).

src/WebJobs.Extensions.DurableTask/DefaultPlatformInformationProvider.cs
```diff
@@ -49,6 +44,60 @@ public bool InLinuxAppService()
...
+                return ProgLanguage.Csharp;
+            }
+
+            throw new Exception("Could not determine user-level programming language.");
+        }
+
+        public bool IsOutOfProc()
+        {
+            ProgLanguage progLanguage = this.GetProgLanguage();
+            if (progLanguage != ProgLanguage.Csharp)
```
cgillum, line 93: .NET Isolated is also out-of-proc.

src/WebJobs.Extensions.DurableTask/DefaultPlatformInformationProvider.cs
```diff
@@ -49,6 +44,60 @@ public bool InLinuxAppService()
...
+                return AppServicePlan.Consumption;
+            }
+
+            return AppServicePlan.AppService;
+        }
+
+        public ProgLanguage GetProgLanguage()
+        {
+            string workerRuntime = this.nameResolver.Resolve("FUNCTIONS_WORKER_RUNTIME");
+            if (workerRuntime == "python")
```
cgillum, line 70: Use case-insensitive string comparison.

cgillum (commented):
Some feedback, mostly in regards to naming.

src/WebJobs.Extensions.DurableTask/DefaultPlatformInformationProvider.cs
davidmrdavid: I think I'll remove this method in that case, as I can just do a more enum comparison instead

davidmrdavid:
In my latest commit, in addition to addressing the feedback so far, I also rename the `PlatformInformationProvider` class itself.

In particular, I remove the `Provider` part and I also removed the `Service` suffix from the interface it implements. I do this because `Provider` has a very specific connotation as a design pattern, which this code does not adhere to. Similarly, the word `Service` doesn't really add any extra meaning to the interface name.

src/WebJobs.Extensions.DurableTask/DefaultPlatformInformation.cs
```diff
@@ -49,6 +44,44 @@ public bool InLinuxAppService()
...
+
+        public bool IsInConsumptionPlan()
+        {
+            return this.InLinuxConsumption() || this.InWindowsConsumption();
+        }
+
+        public WorkerRuntimeType GetWorkerRuntimeType()
+        {
+            string workerRuntime = this.nameResolver.Resolve("FUNCTIONS_WORKER_RUNTIME");
+            if (string.Compare(workerRuntime, "python", ignoreCase: true) == 0)
```
ConnorMcMahon, line 65: nit: it is more common to see `string.Equals(workerRuntime, "python", StringComparison.OrdinalIgnoreCase)`, but this seems to work too.

src/WebJobs.Extensions.DurableTask/DefaultPlatformInformation.cs
```diff
@@ -10,28 +10,23 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
...
         {
             string value = this.nameResolver.Resolve("WEBSITE_SKU");
             return string.Equals(value, "Dynamic", StringComparison.OrdinalIgnoreCase);
         }
 
-        public bool InLinuxConsumption()
+        private bool InLinuxConsumption()
         {
             string containerName = this.GetContainerName();
             string azureWebsiteInstanceId = this.nameResolver.Resolve("WEBSITE_INSTANCE_ID");
```
ConnorMcMahon, line 32: Consider using `Lazy<string>` variables here so that each environment variable is only fetched once, regardless of how many times methods needing this environment variable are called.

src/WebJobs.Extensions.DurableTask/IPlatformInformation.cs
```diff
@@ -6,45 +6,74 @@
...
+    public enum WorkerRuntimeType
     {
         /// <summary>
-        /// Determines if the application is running on a Consumption plan,
-        /// irrespective of OS.
+        /// C-Sharp.
         /// </summary>
-        /// <returns>True if running in Consumption. Otherwise, False.</returns>
-        bool InConsumption();
+        Csharp,
```
ConnorMcMahon, line 32: rename from `Csharp` to `DotNetInProcess` for the eventual support for the out of process .NET worker.

src/WebJobs.Extensions.DurableTask/DefaultPlatformInformation.cs
ConnorMcMahon: This entire if-else block could probably be represented with [Enum.TryParse](https://docs.microsoft.com/en-us/dotnet/api/system.enum.tryparse?view=net-5.0)

ConnorMcMahon (commented):
Mostly looks good, provided some tips for maintainability and performance.

src/WebJobs.Extensions.DurableTask/DefaultPlatformInformation.cs
```diff
@@ -49,6 +44,44 @@ public bool InLinuxAppService()
             return inLinuxDedicated;
         }
 
+        public OperatingSystem GetOperatingSystem()
+        {
+            if (this.InLinuxConsumption() || this.InLinuxAppService())
```
amdeel, line 49: Is there no single way to determine if something is Linux or Windows?

src/WebJobs.Extensions.DurableTask/DefaultPlatformInformation.cs
```diff
@@ -40,7 +35,7 @@ public bool InLinuxConsumption()
             return inLinuxConsumption;
         }
 
-        public bool InLinuxAppService()
+        private bool InLinuxAppService()
         {
             string azureWebsiteInstanceId = this.nameResolver.Resolve("WEBSITE_INSTANCE_ID");
```
amdeel, line 40: You could pull this out in another method to make it clear what azureWebsiteInstanceId and inAppService are used for. I see you use them in both of these Linux methods.

And if this variable determines if it's consumption or not, why do you need to call InLinuxConsumption and InWindowsConsumption in the method IsInConsumption? It seems like you're using this variable to determine that already.

src/WebJobs.Extensions.DurableTask/IPlatformInformation.cs
```diff
@@ -6,45 +6,74 @@
...
+        /// <returns>An OperatingSystem enum.</returns>
+        OperatingSystem GetOperatingSystem();
 
         /// <summary>
-        /// Determines if the application is running in a Linux AppService plan.
+        /// Determine the underlying programming language.
         /// </summary>
-        /// <returns>True if running in Linux AppService. Otherwise, False.</returns>
-        bool InLinuxAppService();
+        /// <returns>A ProgLanguage enum.</returns>
```
amdeel, line 75: WorkerRuntimeType. Don't forget to update these.

src/WebJobs.Extensions.DurableTask/DefaultPlatformInformation.cs
```diff
@@ -0,0 +1,106 @@
...
+        public WorkerRuntimeType GetWorkerRuntimeType()
+        {
+            string workerRuntime = this.ReadEnviromentVariable("FUNCTIONS_WORKER_RUNTIME");
+            WorkerRuntimeType workerRuntimeType;
+            if (Enum.TryParse(workerRuntime, ignoreCase: true, out workerRuntimeType))
+            {
+                return workerRuntimeType;
+            }
+
+            throw new Exception("Could not determine worker runtime type.");
```
ConnorMcMahon, line 88: It may be safer to assume in process if we don't recognize the enum, and just log a warning if we don't recognize the enum. I'm concerned that we may not have 100% uniform enumerations for this environment variable.

If we feel confident in throwing an exception, then a nit would be to make the exception more specific, maybe `InvalidOperationException`.

ConnorMcMahon (changes requested):
A few small concerns, but otherwise looks good.

src/WebJobs.Extensions.DurableTask/DefaultPlatformInformation.cs
```diff
@@ -0,0 +1,106 @@
...
+        public bool IsInConsumptionPlan()
+        {
+            return this.IsInLinuxConsumption() || this.IsInWindowsConsumption();
+        }
+
+        public WorkerRuntimeType GetWorkerRuntimeType()
+        {
+            string workerRuntime = this.ReadEnviromentVariable("FUNCTIONS_WORKER_RUNTIME");
+            WorkerRuntimeType workerRuntimeType;
+            if (Enum.TryParse(workerRuntime, ignoreCase: true, out workerRuntimeType))
```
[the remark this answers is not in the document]
davidmrdavid: My newly added test ("WorkerRuntimeTypeFollowsSpec") tests against every worker-runtime value.

I also decided to rename the enum `DotNetInProcess` into simply `dotnet` as I don't think we currently have a clear strategy for how dotnet-in-proc and dotnet-ooproc will be differentiated.

My current preference would be not to couple getting the worker-runtime with determining whether a PL is OOProc or not. Instead, once we have full support for the ooproc dotnet worker, we should figure out how to differentiate between in-proc and ooproc in the dotnet case and to use that to write an `IsOOProc` method in this class.

src/WebJobs.Extensions.DurableTask/DefaultPlatformInformation.cs
davidmrdavid: My latest changes include this feedback :)

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -422,20 +427,19 @@ void IExtensionConfigProvider.Initialize(ExtensionConfigContext context)
...
 
             // If running in linux, initialize the EventSource listener with the appropiate logger.
             LinuxAppServiceLogger linuxLogger = null;
-            if (inLinuxDedicated)
+            if (!inConsumption)
             {
                 linuxLogger = new LinuxAppServiceLogger(writeToConsole: false, containerName, tenant, stampName);
             }
-            else if (inLinuxConsumption)
+            else if (inConsumption)
```
ConnorMcMahon, line 442: nit: the if clause in this else-if is not necessary. 

In addition, below this, we have a guarantee that linuxLogger is initialized so we don't need the null check.

src/WebJobs.Extensions.DurableTask/IPlatformInformation.cs
```diff
@@ -6,45 +6,79 @@
...
+    public enum WorkerRuntimeType
     {
         /// <summary>
-        /// Determines if the application is running on a Consumption plan,
-        /// irrespective of OS.
+        /// dotnet
         /// </summary>
-        /// <returns>True if running in Consumption. Otherwise, False.</returns>
-        bool InConsumption();
+        Dotnet,
```
ConnorMcMahon, line 32: nit: prefer casing DotNet.

test/FunctionsV2/OutOfProcTests.cs
```diff
@@ -371,5 +372,42 @@ public async Task InvokeLocalRpcEndpoint()
...
+                {
+                    { "FUNCTIONS_WORKER_RUNTIME", workerRuntime },
+                });
+
+            ILoggerFactory loggerFactory = new LoggerFactory();
+            var platformInformation = new DefaultPlatformInformation(nameResolver, loggerFactory);
+
+            var runtimeType = platformInformation.GetWorkerRuntimeType();
+
+            if (workerRuntime == "haskell")
```
ConnorMcMahon, line 400: This makes this test fragile. I'm curious if we should just have a HashSet (initialize with StringComparison.OrdinalIgnoreCase) of the approved list of runtimes, and then make this check if the workerRuntime does not exist there.

Issue #1826, opened 2021-05-06
Function App instance with a single Durable Entity in it produced tons of 'EventRaised' records in its *History table


scale-tone:
My test instance of [this code](https://github.com/scale-tone/sftp-watcher) went into a weird state.
There's one single `SftpToStorageQueueWatcherEntity` entity in it, that entity is periodically signalled by a [TimerTrigger](https://github.com/scale-tone/sftp-watcher/blob/main/TimerTrigger.cs).
According to logs, the TimerTrigger always succeeds (aka successfully sends a signal to the entity). But the entity doesn't get that signal, according to logs. And its state isn't updated.

And the underlying *History table now contains > 300K records that look like this:

> @sftptostoragequeuewatcherentity@bfkrmsiwbkubo.northeurope.azurecontainer.io-*.txt,0000000000050088,2021-05-06T22:08:09.076Z,-1,Edm.Int32,EventRaised,Edm.String,1552010a9ddc4f02954f3b3be4802126,Edm.String,false,Edm.Boolean,2021-05-06T20:47:25.005Z,Edm.DateTime,"{""op"":""Watch"",""signal"":true,""input"":""\""bfkrmsiwbkubo.northeurope.azurecontainer.io/*.txt\"""",""id"":""0226d4b7-9ec2-41e1-aab9-9bea19b70326"",""Timestamp"":""0001-01-01T00:00:00"",""Predecessor"":""0001-01-01T00:00:00""}",Edm.String,op,Edm.String,,,,

, which keep coming.

I can also see lots of weird logs like this in the AppInsights 'traces' table:

> @sftptostoragequeuewatcherentity@bfkrmsiwbkubo.northeurope.azurecontainer.io-*.txt: Function 'sftptostoragequeuewatcherentity (Entity)' completed 'Watch' operation 8a49fc8c-ef15-42ec-a598-5350ed7fab0a in 2060.9566ms. IsReplay: False. Input: (1612 bytes). Output: (null). HubName: sftpw2mkeaxozmpyvmfunction. AppName: sftpw2mkeaxozmpyvm-function. SlotName: Production. ExtensionVersion: 2.4.0. SequenceNumber: 393.
> 
[...]

PR #1917, opened 2021-08-13
Handle entity operation timeouts and support specification of maximal batch size

sebastianburckhardt:
As observed in #1826, entity timeouts are currently not handled correctly because of a bug which can lead to unbounded history growth. This is a very real risk on consumption plans, in particular since batch sizes are not bounded, so timeouts can trigger even if individual entity operations are well below the time limit.

This PR addresses this issues in several ways:

1. When encountering timeouts, it is recorded as an operation failure in the same way as any other unhandled exception in user code. In particular, if the operation was called from an orchestrator that is awaiting its result, a timeout exception is rethrown on the orchestrator.
2. Adds a new optional `int? MaxEntityOperationBatchSize` parameter to `DurableTaskOptions`. This parameter makes it possible to reduce or entirely disable batching of entity operations, which can help to avoid timeouts on consumption plans. 

This PR is ready other than adding testing.

### Pull request checklist

* [x] create docs PR 
* [x] I've added my notes to `release_notes.md`


test/TimeoutTests/Python/proxies.json
```diff
@@ -0,0 +1,4 @@
+{
+  "$schema": "http://json.schemastore.org/proxies",
+  "proxies": {}
+}
```
cgillum, line 4: Are you using the proxies feature for something?

sebastianburckhardt: Not sure why this is there or if it is needed. I just did a copy/paste of the durablePy project inside SmokeTests, which seemed convenient.

src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask.xml
```diff
@@ -2855,7 +2855,7 @@
         </member>
         <member name="T:Microsoft.Azure.WebJobs.Extensions.DurableTask.RequestMessage">
             <summary>
-            A message that represents an operation request or a lock request.
+            A message send to an entity, such as operation, signal, lock, release, or continue messages.
```
cgillum, line 2858: This might not intentionally be part of your PR, but it looks like "send" should be "sent" here.

sebastianburckhardt: Indeed.

test/TimeoutTests/Python/requirements.txt
```diff
@@ -0,0 +1,6 @@
+# DO NOT include azure-functions-worker in this file
+# The Python Worker is managed by Azure Functions platform
+# Manually managing azure-functions-worker may cause unexpected issues
+
+azure-functions
+azure-functions-durable==1.1.3
```
sebastianburckhardt, line 6: This was not meant to be committed. It should not specify the specific version.

test/TimeoutTests/Python/proxies.json
davidmrdavid: This is automatically added by the scaffolding in VSCode. I think it can be safely removed

src/WebJobs.Extensions.DurableTask/Listener/WrappedFunctionResult.cs
```diff
@@ -17,9 +17,11 @@ internal class WrappedFunctionResult
...
         internal enum FunctionResultStatus
         {
-            Success = 0,
-            UserCodeError = 1,
-            FunctionsRuntimeError = 2,
+            Success = 0, // the function executed successfully.
+            UserCodeError = 1, // the user code had an unhandled exception
+            FunctionsRuntimeError = 2, // the runtime could not execute the user code for some reason; considered transient and retried
+            FunctionTimeoutError = 3, // execution timed out; treated as a user error
+            FunctionsHostStoppingError = 4, // host was shutting down; treated as a functions runtime error
```
cgillum, line 24: Why distinguish `FunctionsHostStoppingError` from the existing `FunctionsRuntimeError`? It looks to me like you're checking for both of these together each time.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -122,6 +122,18 @@ public string HubName
...
+        /// Gets or sets the maximum number of entity operations that are processed as a single batch.
+        /// </summary>
+        /// <remarks>
+        /// Reducing this number can help to avoid timeouts on consumption plans. If set to 1, batching is disabled, and each operation
+        /// message executes and is billed as a separate function invocation.
+        /// </remarks>
+        /// <value>
+        /// A positive integer configured by the host.
+        /// </value>
+        public int? MaxEntityOperationBatchSize { get; set; } = null;
```
cgillum, line 135: Can we also create PRs to update the documentation for this (both host.json reference and the entities documentation)?

src/WebJobs.Extensions.DurableTask/Listener/FunctionExecutionHelper.cs
```diff
@@ -26,27 +27,40 @@ internal static class FunctionExecutionHelper
             try
             {
                 context.ExecutorCalledBack = false;
+                var timeoutSource = new System.Threading.Tasks.TaskCompletionSource<Exception>();
+                shim.TimeoutTask = timeoutSource.Task;
```
cgillum, line 31: Any reason not to always assign `TimeoutTask` when the shim object is created? I'm wondering if we need to be careful about potential null-ref exceptions for consumers of this property since it's not immediately obvious when this property gets assigned (this is a situation where nullable reference type annotations in C# can be really helpful).

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -188,14 +206,6 @@ public override async Task<string> Execute(OrchestrationContext innerContext, st
                 innerContext.ErrorDataConverter = this.errorDataConverter;
             }
 
-            this.Config.TraceHelper.FunctionStarting(
```
cgillum, line 191: Can you tell us more about the change to remove these log statements? Is it because they're overly verbose and/or not meaningful?

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -227,62 +237,90 @@ public override async Task<string> Execute(OrchestrationContext innerContext, st
...
+                    var stopwatch = new Stopwatch();
+                    stopwatch.Start();
+
+                    // call the function once per operation in the batch
+                    for (this.BatchPosition = 0; this.BatchPosition < this.operationBatch.Count; this.BatchPosition++)
+                    {
+                        // first, check if we should stop here and not execute the rest of the batch
+                        if (onHostStopping.IsCancellationRequested // host is shutting down
+                            || this.TimeoutTask.IsCompleted // we have timed out
+                            || stopwatch.Elapsed > TimeSpan.FromMinutes(1)) // we have spent significant time in this batch
```
cgillum, line 278: This 1-minute limitation might also be good to include in documentation since your changes could impact the throughput of existing entity code.

test/TimeoutTests/Python/.gitignore
```diff
@@ -0,0 +1,130 @@
+# Byte-compiled / optimized / DLL files
```
cgillum, line 1: Should this .gitignore be included at a higher-level, instead of being specific to just this test app?

test/TimeoutTests/Python/proxies.json
cgillum: Let's remove this. It might actually break something down the road if we leave it here.

cgillum (changes requested):

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
sebastianburckhardt: These events are very useful and the verbosity is appropriate.

I simply moved them into the EntityMiddleware method so they are emitted at a point of time that makes more sense. In particular, FunctionStarted is now emitted before we execute the operations, not after we already executed them.

src/WebJobs.Extensions.DurableTask/Listener/WrappedFunctionResult.cs
sebastianburckhardt: You are correct, they are not currently handled any differently. 

I just found it somewhat helpful to break this down so the code is explaining itself a bit more. It is conceivable we might use this information somewhere.

test/TimeoutTests/Python/.gitignore
sebastianburckhardt: I think it is fine to have it for the test app specifically since the test app is a different type of project (python) than the rest of the code.

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
sebastianburckhardt: yes, I will add it. 

I don't think it should reduce throughput significantly. The idea is that the interruption should be relatively brief (processing continues after saving state to disk and receiving the continuation signal) so I would think it gets amortized sufficiently if it happens once a minute only. Though I should probably test that.

src/WebJobs.Extensions.DurableTask/Listener/FunctionExecutionHelper.cs
sebastianburckhardt: Makes sense, I will change that.

cgillum (approved):

sebastianburckhardt:
Now have a docs PR for this as well: MicrosoftDocs/azure-docs-pr#185116

PR #1936, opened 2021-08-30
Spelling fix in RetryOptions.cs

kendaleiv:


cgillum (approved):
Thanks for catching this and for the PR!

Issue #1559, opened 2020-11-13
Expose setting retry options for CallHttpAsync method


gled4er:
**Is your feature request related to a problem? Please describe.**
We have an use case where the last step of an orchestration is PATCH request to a HTTP endpoint. We would like to set retry options to the `CallHttpAsync` method to ensure that the HTTP endpoint is called several times if it is not available. 

Currently the `CallHttpAsync` does not expose way to set retry options and for that we will need to use dedicated activity function to configure the retry options. 

**Describe the solution you'd like**
We would prefer `CallHttpAsync ` to expose configuring retry options as part of the method or new dedicated CallHttpWithRetryAsync

**Describe alternatives you've considered**
The only alternative we considered is to use dedicated activity function that will wrap the PATCH request.

brandonh-msft:
<details><summary>Internal tracking</summary>
405440
</details>

PR #1948, opened 2021-09-13
Modifications to the RetryOptions fix put in for #1559 

brandonh-msft:
Modifying previous fix for #1559 to be more wholesome; also retrying if any non-success HTTP status codes are received by the request instead of just doing retries for `TimeoutException` and `HttpException`.

We can discuss the correctness of removing `Handler` from the `RetryOptions` object inside DF in  this PR as well. It was necessary to allow for serialization of the new property off `DurableHttpRequest`.

### Pull request checklist


brandonh-msft:
@ConnorMcMahon here's the content of the changes we were talking about late last week

brandonh-msft:
image

@cgillum I took your changes to `TestDurableHttpRequest` and my changes for the expected look of the serialized JSON and applied both here.

src/WebJobs.Extensions.DurableTask/SerializableRetryOptions.cs
```diff
@@ -0,0 +1,105 @@
...
+
+using System;
+using DurableTaskCore = DurableTask.Core;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    /// <summary>
+    /// Defines retry policies that can be passed as parameters to various operations.
+    /// </summary>
+    public class SerializableRetryOptions
```
ConnorMcMahon, line 12: Since all current `RetryOptions` are also `SerializableRetryOptions`, I feel like this is an internal implementation detail and there is no reason to make this public.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
```diff
@@ -114,8 +114,9 @@ public interface IDurableOrchestrationContext
         /// <param name="method">HttpMethod used for api call.</param>
         /// <param name="uri">uri used to make the HTTP call.</param>
         /// <param name="content">Content passed in the HTTP request.</param>
+        /// <param name="retryOptions">The retry option for the HTTP task.</param>
         /// <returns>A <see cref="Task{DurableHttpResponse}"/>Result of the HTTP call.</returns>
-        Task<DurableHttpResponse> CallHttpAsync(HttpMethod method, Uri uri, string content = null);
+        Task<DurableHttpResponse> CallHttpAsync(HttpMethod method, Uri uri, string content = null, SerializableRetryOptions retryOptions = null);
```
ConnorMcMahon, line 119: In the interest of reducing the breaking change impact and keeping our API surface fairly simple, I feel like I'd be fine leaving this simple `CallHttpAsync()` without the RetryOptions, and saying that RetryOptions is only available via `DurableHttpRequest.FailedRequestRetryOptions`.

If we do decide to include retryOptions here, we should rename it to `failedRequestRetryOptions` and add the C# docs from the DurableHttpRequest property to the param docs for this method.

ConnorMcMahon (commented):
I think this covers the gaps that we discussed offline, so thanks for this second iteration!

I have a few minor concerns, but I think the overall approach will definitely be a win for customers.

src/WebJobs.Extensions.DurableTask/SerializableRetryOptions.cs
brandonh-msft: If the `CallHttpAsync` takes `RetryOptions` instead of `SerializableRetryOptions`, we'll get a circular reference error when it attempts to serialize the `DurableHttpRequest` into `rawInput` for the underlying task. 

... but... 🤔

Could we _internally_ cast a `RetryOptions` object to a `SerializableRetryOptions` and thereby "lop off" the `Handle` property which causes the circular ref... I'll give this a shot and see what happens but I have some reservations:
1. I think this will still have to be `public` as long as `RetryOptions` inherits from it, because `RetryOptions` is `public`.
2. If somebody were to think they can set `Handle` on a `RetryOptions` object for `CallHttpAsync`, they'd get unexpected behavior when we (internally) chop that off so the object is serializable...

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
brandonh-msft: Fair enough; the reason I went with this approach is because the majority of the time I've used `CallHttpAsync` it's been through the `HttpMethod, Uri` overload. Any chance we have metrics/data (or can get some) on which of the overloads is used most often?

src/WebJobs.Extensions.DurableTask/SerializableRetryOptions.cs
ConnorMcMahon: I think #2 is a problem regardless due to RetryOptions also being a SerializableRetryOptions, so nothing is stopping customers from just passing in a RetryOptions object and getting the Handle lopped off.

In general, is there a reason we don't just stick a `[JsonIgnore]` on the `Handle` property and just document on that property that it will be ignored in cases where the options are serialized?

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
ConnorMcMahon: Unfortunately, I don't believe we log the usage of CallHttpAsync(), and if we do, definitely not the distinction between these two method overloads, so I think we are going to have to go based on gut calls here.

In general, I have a preference towards keeping the simple overload simple and then having the `DurableHttpRequest` overload as the escape hatch for advanced functionality, but I'm open to the other option. @cgillum  and @bachuv, what are your thoughts?

src/WebJobs.Extensions.DurableTask/SerializableRetryOptions.cs
brandonh-msft: yeah my thought 1 up there is correct.

Do you have any different suggestions on what to do to remove `.Handle` off the object while _also_ conveying to a user that it can't be used (because internally we can't serialize it)? This was the best way I could figure to do both.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
brandonh-msft: Based on [a GitHub code search](https://github.com/search?q=.Callhttpasync%28&type=Code) it definitely seems like the `HttpMethod, Uri` overload is the one users gravitate toward... still want to nix it?

src/WebJobs.Extensions.DurableTask/SerializableRetryOptions.cs
brandonh-msft: I considered the `[JsonIgnore]` approach as well but really wanted to avoid somebody thinking they could pass something with a `.Handle` defined and have it utilized. Understood about `RetryOptions` inheriting, but I think at least tries to guide the developer in the correct direction...

Totally fine to nix it all and go w/ an Ignored property + documentation. I just like to avoid having to crack docs whenever possible ;)

ConnorMcMahon: If we want to enforce it, I think would be best to extract RetryOptions into a base class that has common logic between RetryOptions and SerializableRetryOptions as an internal implementation detail, and then have them be two distinct types so they can't be used interchangeably by accident. We can still have the option to translate SerializableRetryOptions into RetryOptions in our internal code.

I figure if we are going to expose a new type, we might as well have it enforce restrictions rather than just give friendly nudges.

brandonh-msft: Roger that - how would we keep the "internal implementation detail" bit of what you're looking for from bleeding out into the public API, though? If, say, we created `BaseRetryOptions` as a wrapper around `Core.RetryOptions`, then our current `RetryOptions` would be `BaseRetryOptions` + `.Handle` and `SerializableRetryOptions` would be exactly `BaseRetryOptions`, right? There'd be no way to keep any of that `internal` due to the `public` classes inheriting from it...

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
cgillum: I’m not sure the best approach here. The advantage of adding this to the simple overload of that it’s more discoverable. If we think that it will be common for people configure retries, then making it more discoverable would be helpful. It’s also less code churn for existing codebases to adopt it. The disadvantage is that it’s technically a breaking change for anyone unit testing using their own implementation of this interface. That said, I’m leaning towards adding it to the simple overload because I feel like retries are a first class feature that should be configured broadly and isn’t niche.

src/WebJobs.Extensions.DurableTask/SerializableRetryOptions.cs
cgillum: My vote would be to either reuse RetryOptions and *throw* if `Handle` is not null or create a new `HttpRetryOptions` that’s unrelated to the existing RetryOptions. An HttpRetryOptions could replace `Handle` with something more appropriate for retrying HTTP calls, like an HTTP status code filter.

brandonh-msft: Ooo... definitely don't hate either of those. Our devs are already used to having runtime exceptions for things like `await` in an orchestrator, etc. We could also write an Analyzer that calls it out/throws compilation error.

For the Status Code filter that's a decent idea as well! something like `List<HttpStatusCode> RetriableStatusCodes` and then have our implementation code explicitly check for those if they exist, **or** if none were specified do what we are doing now with `EnsureSuccess....`

brandonh-msft:
Have a look at this go-round, @cgillum and @ConnorMcMahon

I pivoted to a disconnected object called `HttpRetryOptions` where I removed `.Handle` (thereby making it serializable) and added `IList<HttpStatusCode> StatusCodesToRetry`. The logic then checks a few things:
1. Is that status code we got back from the request in our list to retry? If so, throw `HttpRequestException` with a related message to trigger the underlying DTFx retry logic
2. Did the user specify retry logic, but no status codes? If so, `EnsureSuccess`
3. If user did neither of the above, and `HttpClient` didn't time out or except out, succeed.

PR #1963, opened 2021-09-29
Draft proposal for "Laguna Gold" - Lightweight SDKs

davidmrdavid:
Just checkpointing my hackathon prototype. On the DF layer, we're essentially using the extra fields added to HistoryEvents in the durableTask layer (https://github.com/Azure/durabletask/pull/615) in order to determine which user-scheduled tasks have completed. We then send that as part of the extension's payload to the OOProc SDK.

This PR also disables a handful of "safety checks" that make it easier to test for long timers and the polling pattern in CallHttp.

This implementation supports:
* CallActivity
* CallActivityWithRetry
* Timers: long and short
* CallHttp with polling

This is by no means "clean code" nor an optimal way to implement this. I'm mostly creating this PR for communication and visibility purposes.

### DIFF
--- a/samples/VSSample.Tests/VSSample.Tests.csproj
+++ b/samples/VSSample.Tests/VSSample.Tests.csproj
@@ -6,7 +6,7 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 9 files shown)


torosent:

Issue #1958, opened 2021-09-23
durableClient.CleanEntityStorageAsync() throws NRE, if durableClient was created with IDurableClientFactory.CreateClient()


scale-tone:
When using an IDurableClient instance injected via handler parameters in a normal way, everything works.

When obtaining that instance via IDurableClientFactory.CreateClient():
```
var durableClient = this._durableClientFactory.CreateClient(new DurableClientOptions
{
    TaskHub = taskHubName
});
```
, its .CleanEntityStorageAsync() method throws NullReferenceException with the following stack:
```
   at Microsoft.Azure.WebJobs.Extensions.DurableTask.DurableClient.Microsoft.Azure.WebJobs.Extensions.DurableTask.IDurableEntityClient.CleanEntityStorageAsync(Boolean removeEmptyEntities, Boolean releaseOrphanedLocks, CancellationToken cancellationToken) in D:\\a\\r1\\a\\azure-functions-durable-extension\\src\\WebJobs.Extensions.DurableTask\\ContextImplementations\\DurableClient.cs:line 561\r\n   
at DurableFunctionsMonitor.DotNetBackend.CleanEntityStorage.<>c__DisplayClass2_0.<<DfmCleanEntityStorageFunction>b__0>d.MoveNext() in C:\\projects\\github\\DurableFunctionsMonitor\\durablefunctionsmonitor.dotnetbackend\\Functions\\CleanEntityStorage.cs:line 43\r\n--- End of stack trace from previous location where exception was thrown ---\r\n   
at DurableFunctionsMonitor.DotNetBackend.HttpHandlerBase.<>c__DisplayClass2_0.<<HandleAuthAndErrors>b__0>d.MoveNext() in C:\\projects\\github\\DurableFunctionsMonitor\\durablefunctionsmonitor.dotnetbackend\\Common\\HttpHandlerBase.cs:line 36\r\n--- End of stack trace from previous location where exception was thrown ---\r\n   
at DurableFunctionsMonitor.DotNetBackend.Globals.HandleErrors(HttpRequest req, ILogger log, Func`1 todo) in C:\\projects\\github\\DurableFunctionsMonitor\\durablefunctionsmonitor.dotnetbackend\\Common\\Globals.cs:line 64
```

All other methods seem to work fine.

IDurableClientFactory instance being used is the default one, injected via ctor.

Changing DurableClientOptions.IsExternalClient doesn't make any difference.

Running under normal func host.

### App Details

[...]

PR #1971, opened 2021-10-08
Fix NullReferenceExceptions in DurableClient

sebastianburckhardt:
Fixes a bug reported by @scale-tone (#1958).

The problem is that `DurableClient` objects are sometimes constructed without a reference to a DurabeExtension. This can cause NullReferenceExceptions when methods are accessing `this.config`. There are a number of those situations (beyond just the one found in #1958). 

This PR fixes all of them:
- `CleanEntityStorageAsync` now gets the necessary information from the durableTaskOptions and from the durabilityProvider
- Added a constructor that can pass in a client factory, when constructing DurableClient without a DurableTaskExtension 
- This client factory is used (in place of `this.config.GetClient()`) when constructing another client from within DurableClient 
- Added null checks in a few more places
 
### Pull request checklist

* [ x] My changes **do not** require documentation changes
* [ x] I've added my notes to `release_notes.md`
* [ x] My changes **do not** need to be backported to a previous version
* [ x] My changes **do not** require any extra work to be leveraged by OutOfProc SDKs

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -7,5 +7,6 @@
 * * Fix fetching of large inputs for pending orchestrations on Azure Storage
 * * Updated TableQuery filter condition string generation to resolve invalid character issues
 * * Fixed stuck orchestration with duplicate message warning issue
+* * Fixed null reference exceptions thrown in DurableClient
 
 ## Breaking Changes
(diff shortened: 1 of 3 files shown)


cgillum (approved):
This change looks good to me, but we should really think about enabling nullable reference types in this codebase so the compiler can find these issues before users do. :)

PR #2015, opened 2021-11-17
Fix spurious validation errors for non-Azure-Storage backends

### LINKED ISSUE #2013: Cannot set PartitionCount larger than 16 for Netherite
### Description
Function host fails to start because an exception is thrown by the constructor of `AzureStorageDurabilityProviderFactory`, with message "PartitionCount must be an integer value between 1 and 16".
 
### Expected behavior
The constructor for `AzureStorageDurabilityProviderFactory` should not be called since the type of the provider is `Netherite`.

### Actual behavior
The constructor appears to be called due to dependency injection somewhere. In the constructor, the validation checks the number of partitions and throws an exception

### Relevant data
The call stack is not helpful, as it does not reveal why the constructor is called, supposedly because it is some kind of dependency injection.  

### Repro
Easy, just use any Netherite project and edit host.json to set PartitionCount to 32.

sebastianburckhardt:
The constructor of AzureStorageDurabilityProviderFactory performs validation on the options, even if this provider is not the provider that is actually selected for use. 

This causes problems with spurious validation exceptions, e.g. with the Netherite provider if the latter uses more than 16 partitions.
 
This fix performs a check in the constructor of AzureStorageDurabilityProviderFactory  to see if it is actually the selected provider, and skips the constructor if not. This avoids spurious errors caused by the configuration not being appropriate for this provider.

resolves #2013

### Pull request checklist


cgillum (approved):
LGTM!

sebastianburckhardt:
Some tests are consistently failing. Investigating now.

It looks like the problem is that some tests are using custom provider types, such as AzureStorageShortenedTimerDurabilityProviderFactory, so my code is overly simplistic (I cannot simply check whether the storage provider is the default provider or not).

PR #2033, opened 2021-12-09
Add SBOM generation task to our pipeline and prepare other compliance requirements

davidmrdavid:
Modeled after: https://github.com/Azure/azure-functions-eventgrid-extension/pull/98/files

This PR modifies our build pipeline in 2 ways:
(1) to include an SBOM manifest, which essentially documents our package dependencies.
(2) to publish built Nuget artifacts to our Azure DevOps feed (see [here](https://durabletaskframework.visualstudio.com/Durable%20Task%20Framework%20CI/_build/results?buildId=2682&view=artifacts&pathAsName=false&type=publishedArtifacts) for an examplev). For security and government compliance, we need to make sure our **release** pipeline simply _pushes_ these ADO artifacts forward to Nuget. We *cannot* rebuild the Nuget package at release time. As a result, the build pipeline now takes on some of the tasks of the release pipeline, such as code-signing.

So now, at every CI build, we'll be publishing an artifact to ADO for both the analyzer and the extension. At release time, we just need to make sure we reference the right artifact to push to Nuget. That change will come in later.

resolves N/A

### Pull request checklist

    * [x] Otherwise: Backport tracked by issue/PR #2034

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -119,6 +119,14 @@ jobs:
     - task: NuGetToolInstaller@1
       displayName: 'Use NuGet '
 
+    - task: DotNetCoreCLI@2
+      displayName: 'dotnet restore'
+      inputs:
+        command: restore
+        projects: '**/**/*.csproj'
+        feedsToUse: config
+        nugetConfigPath: 'azure-functions-durable-extension/.nuget/nuget.config'
+
     - task: VSBuild@1
       displayName: 'Build Durable Extension'
       inputs:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -165,6 +165,13 @@ jobs:
                               "ToolVersion": "1.0"
                             }
                         ]
+    - task: CopyFiles@2
+      displayName: 'Copy Files to: $(Build.ArtifactStagingDirectory)'
+      inputs:
+        SourceFolder: '$(Build.BinariesDirectory)'
+        Contents: '*.nupkg'
+        TargetFolder: '$(Build.ArtifactStagingDirectory)'
+        CleanTargetFolder: true
     - task: AzureArtifacts.drop-validator-task.manifest-generator-task.ManifestGeneratorTask@0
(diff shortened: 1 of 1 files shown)


davidmrdavid:
Surprised the CI failed. I'll investigate. My latest commit[should just have added comments so something's up. Still, the PR should still illustrate the basic change 🤔

amdeel:
Can you explain why we would want to add this to our CI pipeline that runs on every PR instead of our release pipeline?

davidmrdavid:
@amdeel: I'm doing this during our feature-merge CI runs because that's what other products within App Service seem to be doing to comply with this new requirements. This is the case for both the [python-worker](https://github.com/Azure/azure-functions-python-worker/blob/fd6f8c10ea62da7b733815d588dc5e0a62056042/azure-pipelines.yml#L204) and the [nodejs-worker](https://github.com/Azure/azure-functions-nodejs-worker/blob/v3.x/azure-pipelines.yml). I'm not certain if all products within App Service are doing that, I can ask, but my reading is that the wording of these security requirements is a bit vague and so, at the moment and due to the quickly-approaching deadline at the end of this year, folks are opting to apply them broadly and quickly.

davidmrdavid:
Maybe we can file a ticket to revisit these requirements at a later date? I'm sure other products would be interested in producing these artifacts with lesser frequency, but getting clarity on the minimal requirements will take time.

cgillum:
The EventGrid extension you modeled this after only seems to be doing it as part of a publish step though. That pipeline doesn't appear to be running any tests.

amdeel (changes requested):
I don't think we want to add this step to our CI pipeline. Publishing a signed package on every pull request would add additional slowdown to our existing pull request bottleneck right before a release.

As discussed offline, we should get a better understanding of what these requirements are and make changes to our release pipeline as necessary.

davidmrdavid:
@amdeel, I just got confirmation that we can opt to only do this during release proper. Thank you for the suggestion, I'll close this PR.

PR #2039, opened 2021-12-20
Patch binding expressions for OOProc

davidmrdavid:
Related to: https://github.com/Azure/azure-functions-durable-python/issues/282, https://github.com/Azure/azure-functions-durable-js/issues/231

OOProc activity inputs are doubly-serialized to prevent the data from being mutated within .NET; a recurring problem we've had in the past. However, this prevents expression bindings from receiving the right payload. For example, consider a Python activity containing a blob input binding which sets the blob to be read as the input to the activity.

This would look as follows

```JSON
{
  "scriptFile": "__init__.py",
  "bindings": [
    {
      "name": "name",
      "type": "activityTrigger",
      "direction": "in"
    },
    {
      "type": "blob",
      "direction": "in",
      "name": "inputBlob",
      "path": "incontainer/{name}",
      "connection": "AzureWebJobsStorage"
    }
  ]
}
```
Assume the given Activity input is "Tokyo".
In the input binding for the blob, we would expect the `path` to resolve to `incontainer/Tokyo`. However, due to our double serialization strategy, it currently receives `incontainer/"Tokyo"`. Notice the inserted quotes.

This PR removes those newly-inserted quotes by explicitly de-serializing the input payload right before assigning the data to the expression binding.

**TODO:** By fixing this bug, we're also potentially introducing a breaking change. If we're able to detect the end-user's JS/Python SDK **in this code-path**, then we should be able to properly version this logic and maintain backwards compatibility. However, I'm not sure that we have access to that information in this code-path. We should discuss this _asap_.

resolves N/A
### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
@@ -161,11 +161,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
                 // Non-dotnet payloads are doubly-serialized for protection against being interpreted/mutated within dotNET.
                 // Therefore, they also need to be doubly-deserialized when binding to expression-bindings.
+                // We avoid double-deserialized when the worker is unknown, for safety.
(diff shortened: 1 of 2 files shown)


lilyjma:
Some customers might be doing workarounds. So we need to be careful not to break them.

torosent:

PR #2049, opened 2022-01-11
Fix polling loop in HttpApi for Netherite

sebastianburckhardt:
Netherite is currently not following the standard semantics of WaitForOrchestrationAsync, in that it returns a TimeoutException instead of a null value. This is a bug, see issue microsoft/durabletask-netherite#104.

To fix this, we need to first fix the HttpAPI polling loop as it depends on the previous exception behavior. This PR does that, and also restructures the code a bit to make it easier to read.

This change should not affect any other storage providers.
 
### Pull request checklist

* [ x] My changes **do not** require documentation changes
* [ x] My changes **should not** be added to the release notes for the next release
* [ x] My changes **do not** need to be backported to a previous version
* [x ] My changes **do not** require any extra work to be leveraged by OutOfProc SDKs


davidmrdavid (approved):
If I understand correctly, functionality-wise this PR "just" removes the `try-catch` around `WaitForOrchestrationAsync`. This would make sense because we don't expect that API to throw a `TimeoutException`.

If this understanding it's correct, then LGTM!

PR #2054, opened 2022-01-12
Add azure-pipelines-release.yml

bachuv:
[skip ci]

Adding azure-pipelines-release.yml to add a manifest generator task and publish artifact task. We are creating this new pipeline to adhere to security requirements and add a manifest file as part of our artifacts (our current pipeline doesn't allow publishing artifacts because that step is not supported in a "Release pipeline").

### Pull request checklist

### DIFF
--- a/azure-pipelines-release.yml
+++ b/azure-pipelines-release.yml
@@ -23,14 +23,14 @@ steps:
 - task: VSBuild@1
   displayName: 'Build Durable Extension'
   inputs:
-    solution: '/WebJobs.Extensions.DurableTask.sln'
+    solution: '**/WebJobs.Extensions.DurableTask.sln'
     vsVersion: "16.0"
     configuration: Release
 
 - task: EsrpCodeSigning@1
   inputs:
     ConnectedServiceName: 'ESRP Service'
-    FolderPath: '/src/WebJobs.Extensions.DurableTask/bin/Release'
+    FolderPath: '**/src/WebJobs.Extensions.DurableTask/bin/Release'
     Pattern: '*DurableTask.dll'
     signConfigType: 'inlineSignParams'
     inlineOperation: |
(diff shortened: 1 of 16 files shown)


azure-pipelines-release.yml
```diff
@@ -0,0 +1,133 @@
+trigger: none
+pr: none
+
+pool:
+  vmImage: 'windows-latest'
```
bachuv, line 5: I'll update this once the 1ES VMs are fully configured to use.

azure-pipelines-release.yml
```diff
@@ -0,0 +1,133 @@
...
+      ]
+    SessionTimeout: '60'
+    MaxConcurrency: '50'
+    MaxRetryAttempts: '5'
+
+# Manifest Generator Task
+- task: ManifestGeneratorTask@0
+  displayName: 'Manifest Generator '
+  inputs:
+    BuildDropPath: '$(System.DefaultWorkingDirectory)/azure-functions-durable-extension'
```
davidmrdavid, line 120: I believe there's 3 possible values here:
1. The one you have right now
2. '$(Build.ArtifactStagingDirectory)'
3. $(System.DefaultWorkingDirectory)/azure-functions-durable-extension/src/WebJobs.Extensions.DurableTask/'

Just documenting these here in case we find that the current value doesn't produce the expected SBOM

davidmrdavid (commented):
I'm unsure if we've already associated this yml file with a pipeline action in Azure DevOps? If not, we'll need to test this by creating a new pipeline whose definition is sourced from this new file

amdeel (changes requested):
This looks good to me. The only change is that this should target main, not dev. We might end up deleting the NuGet push step, but we can probably just update it to push directly to NuGet later once the NuGet package ownership gets changed.

amdeel (approved):

bachuv:
Closing as these changes were made in #2056 with a cleaner commit history.

PR #2075, opened 2022-02-04
Increment version to 2.6.2

bachuv:
Incrementing version to 2.6.2

### Pull request checklist


amdeel:
> We have PRs ready to merge already that are increasing the version to 2.7.0. I don't think we need this PR.

Why are we incrementing versions in PRs? I think we should stick to the normal release version incrementing process.

In this case you have your PR ready to merge and that's probably the PR you're talking about with the version increase so we can ignore this 2.6.2 this release and close this PR. In the future we should get out of the habit of incrementing versions in feature PRs.

cgillum:
Including a minor version update in the PR ensures that we don't accidentally release new features or breaking changes in a release that isn't a major or minor update. We have two pending PRs that contain major new features and, had we merged them early accidentally, they would have gone out as 2.7.0 (which would be correct) instead of going out as 2.6.1 (which would have been incorrect). Given that we don't plan our release versions in advance, this felt like the safer option to me.

PR #2121, opened 2022-03-23
Allow Internal Code to Throw Exceptions in OOProc Shim

hossam-nasr:

Previously, in `ProcessAsyncActionsV1` in the `OutOfProcOrchestrationShim`, we didn't have an explicit API for `WhenAny` and `WhenAll` tasks. Instead, we scheduled all tasks wrapped in an `await Task.WhenAny()`, as seen [here](https://github.com/Azure/azure-functions-durable-extension/blob/56348c050a24a3f7f5fd94733b78489b863aff40/src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs#L243). Since the result of a `Task.WhenAny()` is another `Task`, this also had the nice side effect of wrapping any exceptions thrown by the sheduled `Task`s in a `Task` so that it can be passed to the SDKs, giving user code the ability to try-catch their own exceptions. However, since this approach didn't explicitly schedule an `WhenAll` task when the user asked for one, this led to some concurrency and race condition errors, where the orchestrator could end up in a bad state and get stuck. 

These errors were fixed in [`ProcessAsyncActionV2`](https://github.com/Azure/azure-functions-durable-extension/blob/56348c050a24a3f7f5fd94733b78489b863aff40/src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs#L196), where we introduced explicit durable APIs for `WhenAll` and `WhenAny` tasks. As a result, we removed the `WhenAny` wrapper. At first, this meant that any exception thrown by the `Task` itself would get thrown by the extension when the `Task` is awaited, crashing the process, and not allowing orchestrators to try-catch their own exceptions. To mitigate this, a try-catch statement was introduced in this PR #1957 which captures all exceptions and silently ignores them.

The problem is this try-catch statement would capture (and ignore) not just the exceptions thrown by the `Task`s, but also exceptions by our own logic, including, for example, [this check](https://github.com/Azure/azure-functions-durable-extension/blob/56348c050a24a3f7f5fd94733b78489b863aff40/src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs#L122) for long timers, which are currently unsupported in out of proc.

This PR addresses this problem by removing the try-catch block, and instead again wrapping the call to `InvokeAPIFromAction` in a `Task.WhenAny()` 

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
@@ -197,17 +197,20 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             foreach (AsyncAction action in actions)
             {
-                // Before awaiting the task, we wrap it in a `WhenAny` call
-                // to protect against possible exceptions. Exception handling
-                // and exception throwing is managed in the OOProc SDKs and
+                // This line may throw exceptions (usually to validate that some OOProc operation is valid), which we still want to surface
+                Task durableTask = this.InvokeAPIFromAction(action);
+
+                // Before awaiting the task, we wrap it in a try/catch block
+                // to protect against possible exceptions thrown by user code.
+                // Exception handling and exception throwing is managed in the OOProc SDKs and
                 // should not affect / interrupt this replay loop
                 try
                 {
-                    await this.InvokeAPIFromAction(action);
+                    await durableTask;
                 }
                 catch (Exception)
                 {
-                    // ignore exceptions, they should be handled by user-code instead.
+                    // Silently ignore exceptions thrown by user code
                 }
             }
         }


src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -201,14 +201,7 @@ private async Task ProcessAsyncActionsV2(AsyncAction[] actions)
...
                 // should not affect / interrupt this replay loop
-                try
-                {
-                    await this.InvokeAPIFromAction(action);
-                }
-                catch (Exception)
-                {
-                    // ignore exceptions, they should be handled by user-code instead.
-                }
+                await Task.WhenAny(this.InvokeAPIFromAction(action));
```
davidmrdavid, line 204: Thank you for this PR, this looks good to me; but I want to make sure @cgillum approves of the change as well.
In the meantime, could you please expand the comments above to justify why we use WhenAny instead of a try-catch? 

Otherwise, I could see a future maintainer removing the WhenAny for a try-catch again P . Let's break the cycle!

davidmrdavid (commented):
LGTM, just waiting for @cgillum's thoughts.

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
hossam-nasr: Yeah that makes sense, I added the comment. Another idea I had was to wrap `this.InvokeAPIFromAction()` in a separate custom class/Task that we create, rather than a `WhenAny` to make the intentions clearer

cgillum: I don't have any problem with this change but I'll admit I still don't understand how or why it works. For example, we say that WhenAny swallows exceptions, but since that's not the documented purpose of the `Task.WhenAny` API, it's not clear to me why this is the appropriate thing to do. It's also not clear to me why "swallowing exceptions" is a good thing; we mention that it's the SDKs responsibility to handle exceptions, but how is that related to the tasks in the C# layer?

Long story short, it feels like there's a lot of magic going on here and some context and it's not clear to me how easy it will be for future maintainers (or our futures selves) to understand what the intent of this code is if we find another problem with it later. More comments explaining more of the mechanics may help. However, I don't necessarily want to block on this issue - it's more of a suggestion.

cgillum (approved):

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
davidmrdavid: Let's make sure we remove the magic here, I think that's crucial for these OOProc bits. I'll coordinate with Hossam to add a few more comments, and I'll loop you back at that point. Thanks!

hossam-nasr: @cgillum After talking with @davidmrdavid I think we found a solution that achieves the same affect while using less "magic." I've updated the PR, let me know how this solution works now :)

davidmrdavid:
I'm unsure why the .NET Isolated test failed, but that seems unrelated to your change."
I think the latest commit removes the magic of "WhenAny", and comments the logic quite well. @cgillum, would you mind giving this just another look? Thanks!

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -197,17 +197,20 @@ private async Task ProcessAsyncActionsV2(AsyncAction[] actions)
         {
             foreach (AsyncAction action in actions)
             {
-                // Before awaiting the task, we wrap it in a `WhenAny` call
-                // to protect against possible exceptions. Exception handling
-                // and exception throwing is managed in the OOProc SDKs and
+                // This line may throw exceptions, which we still want to allow to surface
```
davidmrdavid, line 200:
```suggestion
                // This line may throw exceptions (usually to validate that some OOProc operation is valid), which we still want to surface
```

davidmrdavid (approved):
Left a small nit suggestion, but otherwise it looks good to me! Let's make sure the CI is all green before merging though

cgillum:
The .NET Isolated error is this one:

> /root/src/Worker.Extensions.DurableTask/DurableOrchestrator.cs(105,51): error CS0246: The type or namespace name 'IWorkerContext' could not be found (are you missing a using directive or an assembly reference?) [/root/src/Worker.Extensions.DurableTask/Worker.Extensions.DurableTask.csproj]

This feels like a legit error, but probably not caused by this PR. The appveyor run seems to have the same error. I may have broken this unintentionally as part of an unrelated myget.org update. I can investigate further to see if I can get this green again.

cgillum:
@hossam-nasr sorry about the interruption to your PR. Please merge the latest from `dev` to unblock this PR.

PR #2125, opened 2022-03-30
Change new OOProc orchestration trigger protocol from JSON to protobuf

cgillum:
## Overview

This PR changes the orchestration trigger payload protocol from JSON strings to protobuf strings. This includes both the request that gets sent to the language worker and the response that comes back. This only applies to the newer OOProc languages, .NET Isolated and Java.

The other big change in this PR is in the way we interpret errors from the OOProc worker. Previously, we made no attempt to interpret error messages. With this PR, we make a point of parsing out the error message and the stack trace details from the `RpcException`s that we get when an activity function fails.

### Why are we making this JSON --> protobuf change?
The old JSON schema for the orchestration trigger payload was problematic from the perspective of SDK code reuse. The newer OOProc SDKs are designed to work with a sidecar over gRPC. Having JSON as the protocol for Functions would then require use to either translate the JSON into protobuf or to translate the JSON into the SDK objects directly. This translation would be a lot of one-off code in each language SDK just to support functions. I deemed it much more economical to instead change the Durable extension to use the protobuf format instead so that language SDKs wouldn't need to do any protocol translation.

This change has the potential of also increasing efficiency since protobuf is much more compact and faster to serialize/deserialize compared to JSON. The downside is that it's much more difficult to debug if unexpected data is received at either end. This tradeoff seemed acceptable, however.

### Why are we making this error parsing change?
One of the features I'm working on for .NET Isolated and Java is the ability to create custom retry handlers. In order to create a reasonable experience, the error information from an activity needs to be in a format that a developer can reason about. This was not previously possible since the `RpcException` given to the extension contained unstructured error information. As part of https://github.com/Azure/durabletask/pull/681, we updated DT.Core to support structured failure details. In this PR, we parse the exception to construct a new `TaskFailureDetails` payload, which is a structure that defines an error in a language neutral way. Having this payload available to orchestrations enables developers to write much simpler error handling logic.

## Pull request checklist

* [x] My changes **do not** require documentation changes (this is for unreleased .NET Isolated and Java support)
* [x] My changes **should not** be added to the release notes for the next release (no impact for existing workloads)
    * [x] Otherwise: That work is being tracked here:
        * https://github.com/microsoft/durabletask-dotnet/pull/6
        * https://github.com/microsoft/durabletask-java/pull/19


cgillum:
Not sure why the commit history for this PR looks so weird. I may try to recreate my clone of this repo after this PR to see if that resolves it. I don't think it should cause any problems as long as I'm doing a squash-merge.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -92,6 +92,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.DurableTask.AzureStorage" Version="1.10.*" />
+    <PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="2.8.0" />
```
cgillum, line 95: At the time of writing, this package version is currently only on myget.org. We'll need to publish it to nuget.org prior to releasing this extension version.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -5557,8 +5557,8 @@ public async Task CustomSerializerSettings_TypeNameHandlingAll()
                 var status = await client.WaitForCompletionAsync(this.output);
 
                 Assert.Equal(OrchestrationRuntimeStatus.Completed, status?.RuntimeStatus);
-                Assert.Equal("World", status?.Input);
-                Assert.Equal("Hello, World!", status?.Output);
+                Assert.Equal("World", (string)status?.Input);
+                Assert.Equal("Hello, World!", (string)status?.Output);
```
cgillum, line 5561: I ran into an issue in an earlier iteration that caused this test to fail. Unfortunately, xunit doesn't know how to display `JValue` objects in the error messages (it always just shows "[]") so the test output was unfortunately useless. Casting the `JValue` properties to `string`, however, fixes the display problem.

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/HelloCitiesUntyped.cs
```diff
@@ -38,18 +39,19 @@ internal static class HelloSequenceUntyped
...
     /// Orchestrator function that calls the <see cref="SayHelloUntyped"/> activity function several times consecutively.
     /// </summary>
-    /// <param name="state">The opaque orchestration state that gets passed to the function. Code should never attempt to read this state.</param>
+    /// <param name="requestState">The serialized orchestration state that gets passed to the function.</param>
     /// <returns>Returns an opaque output string with instructions about what actions to persist into the orchestration history.</returns>
     [Function(nameof(HelloCitiesUntyped))]
-    public static string HelloCitiesUntyped([OrchestrationTrigger] string state) =>
-        DurableOrchestrator.LoadAndRun<string, string>(state, async (context, _) =>
+    public static string HelloCitiesUntyped([OrchestrationTrigger] string requestState, FunctionContext functionContext) =>
+        OrchestrationRunner.LoadAndRun<string, string>(requestState, async (context, _) =>
```
cgillum, line 46: Note that `OrchestrationRunner` is now defined in the Microsoft.DurableTask nuget package. This was possible now that we've moved away from the Functions-specific JSON protocol to the Durable Task-specific protobuf protocol.

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/HelloCitiesUntyped.cs
```diff
@@ -38,18 +39,19 @@ internal static class HelloSequenceUntyped
...
         {
             string result = "";
             result += await context.CallActivityAsync<string>(nameof(SayHelloUntyped), "Tokyo") + " ";
             result += await context.CallActivityAsync<string>(nameof(SayHelloUntyped), "London") + " ";
             result += await context.CallActivityAsync<string>(nameof(SayHelloUntyped), "Seattle");
             return result;
-        });
+        },
+        functionContext.InstanceServices);
```
cgillum, line 54: Passing `InstanceServices` allows us to get access to injected dependency services, which we didn't have before. In the final version of .NET Isolated support, developers hopefully won't have to worry about this or `OrchestrationRunner` at all.

cgillum (commented):
Adding some context comments.

davidmrdavid:
I noticed some of the tests here are failing. should I start reviewing this or would you prefer I wait until the tests are "green"?

cgillum:
@davidmrdavid feel free to start reviewing. The previous iteration was green and my latest commit was pretty small. I'll investigate what the failures are in the meantime.

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/HelloCitiesUntyped.cs
davidmrdavid: How does changing protocols enable this? I'm not entirely clear on that

cgillum:
Looks like the CI failed because of an Azure Storage SDK deadlock timeout expiration 😬. Re-running it got things green.

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/HelloCitiesUntyped.cs
davidmrdavid: Can you say a bit more about this? What services are we accessing in this case? This doesn't mean that customers are now allowed to use DI in their orchestrations, right?

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -347,14 +357,125 @@ public async Task CallActivityAsync(DispatchMiddlewareContext dispatchContext, F
...
+            {
+                Exception rpcException = e.InnerException;
+                if (TryGetRpcExceptionFields(rpcException.Message, out string? exception, out string? stackTrace))
+                {
+                    if (TrySplitExceptionTypeFromMessage(exception, out string? exceptionType, out string? exceptionMessage))
+                    {
+                        return new FailureDetails(exceptionType, exceptionMessage, stackTrace, innerFailure: null, isNonRetriable: false);
+                    }
+
+                    return new FailureDetails("(unknown)", exception, stackTrace, innerFailure: null, isNonRetriable: false);
```
davidmrdavid, line 384: I noticed `innerFailure` is always `null` in here. Why is that? Do we not except inner failures in OOProc PLs? I guess it might be worthwhile to define what an innerFailure would be in this case. I've been thinking of them as exceptions that were re-thrown, and therefore became wrapped in an outer exception.

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -347,14 +357,125 @@ public async Task CallActivityAsync(DispatchMiddlewareContext dispatchContext, F
...
+            out string? stackTrace)
+        {
+            exception = null;
+            stackTrace = null;
+
+            // The message string is formatted as follows:
+            // Result: {result}\nException: {message}\nStack: {stack}
+
+            const string ExceptionDelimeter = "\nException: ";
+            const string StackDelimeter = "\nStack: ";
```
davidmrdavid, line 409: What happens if a Java user throws an Exception whose message contains the substring `\nException`. Could that throw off our parsing here?

davidmrdavid (commented):
Made a first pass through this PR, left some questions. Thanks!

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/HelloCitiesUntyped.cs
cgillum: It wouldn't have made sense to move this code to the generic SDK because it relied on Functions-specific serialization logic (I consider the JSON serialization to be Functions-specific because the Durable Functions extension is the only place where it's used). By ditching the JSON serialization protocol and adopting the generic SDK's protobuf protocol, we no longer had any Functions-specific JSON protocols being used, even in the Functions case. Thus, I could easily consolidate more logic into the generic SDK.

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/HelloCitiesUntyped.cs
cgillum: Correct, customers are still not allowed to use DI in orchestrations. The primary use-cases are getting the Azure Function's `ILoggerFactory` for logging and for supporting custom `IDataConverter` implementations. Without this, customers wouldn't be able to use `ILogger` in their orchestrator code, nor could they override our serialization logic for inputs and outputs.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActivityContext.cs
```diff
@@ -61,14 +64,27 @@ internal JToken GetInputAsJson()
...
             {
-                var objectArray = this.messageDataConverter.Deserialize<object[]>(this.serializedInput);
-
-                if (objectArray?.Length != 1)
+                string actualInput;
+                if (this.inputsAreArrays)
+                {
+                    // The legacy behavior where the DTFx orchestration context schedules the activity results in an input that's
+                    // wrapped in an array. This is unfortunately quite inefficient because it requires us to do an extra serialization
+                    // round-trip in order to correctly interpret the input data.
```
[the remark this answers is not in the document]
cgillum: This is unfortunately embedded deep into DurableTask.Core, which makes changing it more difficult since it impacts all languages. If we remove the DurableTask.Core dependency from the .NET Isolated SDK, then we can remove this unnecessary overhead (similar to how Java can bypass this stuff).

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
cgillum: The problem is that Azure Functions doesn't preserve any of the structure of an exception or error from OOProc workers. Rather, we just get one flat `RpcException` that encodes the error information as a single string value.  We could look into updating the Functions OOProc gRPC protocol to support nested exceptions (like I recently did for DTFx) but that would be a lot of work. Unless Functions could better support nested exceptions, nested activity exceptions can't really be supported.

Note that we __do__ support nested exceptions in orchestrator functions. This is because we control the full payload for errors. The limitation is only for activity functions.

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
cgillum: For that specific case, no, because we only look for the first instance of `\nException`, which is always going to be the one we want. However, this logic will be thrown off if a Java (or a .NET Isolated) exception message were to contain `\nStack`. I might be able to fix that by using `LastIndexOf(StackDelimeter)` instead of `IndexOf(StackDelimeter)`, in which case we'll only look for the _last_ instance and bypass any that might be embedded in the exception message... Let me post an update for that.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActivityContext.cs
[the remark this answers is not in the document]
davidmrdavid: Thanks for the explanation. My 2 cents is that it would be great if we could document this blocker in the code, as I don't think the impact is immediately obvious :)

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
davidmrdavid: Sounds good, I'll await that update, do ping me :)

davidmrdavid (commented):
Thanks for the answers so far @cgillum.  My only blockers at this point are:
(1) documenting why `inputsAsArray` is hard to deprecate in code
(2) looking into fixing the `\nStack`-parsing edge case

Thanks!

cgillum:
I somehow messed up my local branch history and had to do a force-push of the latest iteration. The only notable change was related to this comment:

> looking into fixing the \nStack-parsing edge case

I changed `.IndexOf` to `.LastIndexOf` to handle the case where the exception message has `\nStack` in it for some reasons, which would otherwise mess up the parsing.

Regarding this:

> documenting why inputsAsArray is hard to deprecate in code

I don't recall exactly what the state of the code was when you requested this documentation. I feel the current set of comments are good enough, though. I don't think it's appropriate to go as far as talking about why it is/isn't hard to deprecate something in DTFx, though. That kind of comment belongs in DTFx itself.

I hope you're good with this latest update, assuming all tests pass!

davidmrdavid (approved):
:shipit:

PR #2134, opened 2022-04-09
Support Long Timers for Out of Proc Languages ⏳📈🚀

hossam-nasr:
This PR:

- Adds support for long timers for out-of-proc SDKs
- Adds a new schema version (to signify SDK support for long timers)
- Only allows OOProc to use the long timer API if using the new schema version

Extension-side of resolving https://github.com/Azure/azure-functions-durable-js/issues/325. JS SDK PR: https://github.com/Azure/azure-functions-durable-js/pull/340

### Pull request checklist

    * [x] Otherwise: That work is being tracked here: https://github.com/Azure/azure-functions-durable-js/issues/325

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -54,6 +54,23 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private MessageSorter messageSorter;
 
+        internal TimeSpan LongRunningTimerIntervalDuration
+        {
+            get
+            {
+                return this.durabilityProvider.LongRunningTimerIntervalLength;
+            }
+        }
+
+
+        internal TimeSpan MaximumShortTimerDuration
+        {
+            get
+            {
+                return this.durabilityProvider.MaximumDelayTime;
+            }
+        }
+
         internal DurableOrchestrationContext(DurableTaskExtension config, DurabilityProvider durabilityProvider, string functionName)
             : base(config, functionName)
         {
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -54,23 +54,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private MessageSorter messageSorter;
 
-        internal TimeSpan LongRunningTimerIntervalDuration
-        {
-            get
-            {
-                return this.durabilityProvider.LongRunningTimerIntervalLength;
-            }
-        }
-
-
-        internal TimeSpan MaximumShortTimerDuration
-        {
-            get
-            {
-                return this.durabilityProvider.MaximumDelayTime;
-            }
-        }
-
         internal DurableOrchestrationContext(DurableTaskExtension config, DurabilityProvider durabilityProvider, string functionName)
             : base(config, functionName)
         {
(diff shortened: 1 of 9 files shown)


src/Worker.Extensions.DurableTask/AssemblyInfo.cs
```diff
@@ -4,4 +4,4 @@
 using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
 
 // TODO: Find a way to generate this dynamically at build-time
-[assembly: ExtensionInformation("Microsoft.Azure.WebJobs.Extensions.DurableTask", "2.7.*")]
+[assembly: ExtensionInformation("Microsoft.Azure.WebJobs.Extensions.DurableTask", "2.8.*")]
```
davidmrdavid, line 7: did you modify this on purpose? Seems pretty low-level

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -81,6 +81,22 @@ internal bool IsReplaying
...
+        }
+
+        internal TimeSpan MaximumShortTimerDuration
+        {
+            get
+            {
+                return this.durabilityProvider.MaximumDelayTime;
+            }
+        }
+
```
davidmrdavid, line 99: I realize these names were originally my suggestion, but now I'm thinking it would be best to call them the same as the property they're accessing inside `durablityProvider`.

What do you think of renaming these methods to be `LongRunningTimerIntervalLength` and `MaximumDelayTime`?

davidmrdavid (commented):
It mostly looks good to me, but I left a few nits.
The only reason I'm not approving at this point is that `AssemblyInfo` change. What that change on purpose? If so, can you please explain it? Otherwise, I'd try removing it and making sure it all still works. Thanks!

src/Worker.Extensions.DurableTask/AssemblyInfo.cs
hossam-nasr: Yes, I changed this based on the last item on the PR checklist. I figured since we're moving to a new replay schema, it warranted a new minor version, but if I'm wrong I can revert this back to what it was

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
hossam-nasr: I don't mind changing `LongRunningTimerIntervalDuration` to `LongRunningTimerIntervalLength`, since both of those names are confusing to me anyway 😛 But I would vote to keep not change `MaximumShortTimerDuration` to `MaximumDelayTime` because the latter feels more ambiguous to me, but I don't have strong opinions either way.

Also, should those naming changes be reflected in the SDK payload (and the SDK variable names themselves) too?

davidmrdavid: I'm ok with that naming compromise - feel free to change the `LongRuning...` variable but keep the other one the same.

I don't think the SDK payload needs to have the same names, and the SDK can have whichever variable names it prefers so long as it parses the payload correctly :)

src/Worker.Extensions.DurableTask/AssemblyInfo.cs
davidmrdavid: I see. That new checklist item is fairly new, so I'm personally unsure how exactly to interpret it, but I'm almost certain we're not supposed to be updating it here :-)

@cgillum: can you please clarify that final bullet-point in the PR checklist?

cgillum: The idea is that if you are incrementing the extension version number then you must also increment the version number in this file to ensure they always match each other. For example, this update makes sense IFF we're also updating the version number in WebJobs.Extensions.DurableTask.csproj.

davidmrdavid: Thanks for clarifying @cgillum.

@hossam-nasr: I don't think we should update the extension version number in this PR, we usually do that during our release process, so let's please revert this change for now. During release ,we'll make sure this file is properly updated if necessary :) . Thanks!

hossam-nasr: Yeah that makes sense, sorry for the confusion and thanks for the clarification. I reverted this line

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
hossam-nasr: Done :)

davidmrdavid (approved):
:shipit: 
Thank you for this work, our users will really appreciate this change

Issue #2135, opened 2022-04-10
Not returning to suborchestrator after calling activity


kirk-marple:
### Description
> A clear and concise description of what the bug is. Please make an effort to fill in all the sections below; the information will help us investigate your issue.

We've successfully used Durable Functions for a year or more, and just recently getting an issue where the suborchestrator isn't returning from the activity.

This was working fine when we just had one set of Durable Functions, and we recently added two more, and I believe this is when the problem started.  

(UPDATE: just re-ran our original Durable Function, and realized it was seeing same behavior, but there's no functional code that's run after the suborchestrator's activity, so we never noticed it.  I'm not seeing our log messages that say the activity was successful, and the orchestration is left in Running state.)

I tried separating them out into separate function apps, and separate storage, thinking there was a problem using same storage account and task hub.

I'm totally stumped and blocked at this point.

I've trimmed my code down to the simple 'ping' case below.

### Expected behavior
> A clear and concise description of what you expected to happen.

Orchestration, suborchestration and activity to complete as expected.

### Actual behavior
> A clear and concise description of what actually happened.

I'm seeing these log messages in App Insights:
Attempting to create job [{job.Id}] entity workflow.
Sending ping [{ping}].
Received ping [{ping}].

But that's all I get; I never see:

Sent ping [{ping}].
Created job [{job.Id}] processing workflow.

It appears that the Ping method completes properly, and I can see it marked TaskCompleted.

After that, I see the events:
ExecutionStarted
OrchestrationStarted
OrchestrationCompleted

This reproduces exactly the same every time.  The serialized 'Job' is only about 5KB, so I don't think it's a memory issue.

Here's the key part I see in the logs (attached below).

Ping activity finishes, and after processing TaskCompleted, it executes CreatePingWorkflow but leaves itself in running state.
```
2022-04-10T07:33:41.422 [Information] fab972cedda44432a6441240e5a29dee:0: Task activity [Ping] completed successfully
2022-04-10T07:33:41.431 [Information] fab972cedda44432a6441240e5a29dee:0: Sending [TaskCompleted] message to knowledge-control-00 for instance 'fab972cedda44432a6441240e5a29dee:0'
2022-04-10T07:33:41.437 [Information] fab972cedda44432a6441240e5a29dee:0: Deleting [TaskScheduled] message from knowledge-workitems
[...]

Issue #2133, opened 2022-04-07
Instances are getting stuck Running even though execution has been completed. ExecutionCompleted is no longer happening.


philwhere:
### Description
My Durable Functions app has been running fine in Azure for while but has suddenly started to misbehave.
Instances are completing their entire execution for the orchestrator function, however they are no longer completing their instance status. The statuses stay as "Running" and the "ExecutionCompleted" event is not added to the instance history table.

### Expected behavior
At the end of an orchestration function, the instance should complete, fail or restart (ContinueAsNew).

### Actual behavior
The orchestrator runs all the way through logically but the instance does not end correctly.
I have found a single instance that has been marked as Completed since the issue appeared.

### Relevant source code snippets
(Chopped up and simplified example to make sense of the screenshots)
```csharp
        [FunctionName("Orchestrator")]
        public async Task RunOrchestrator(
            [OrchestrationTrigger] IDurableOrchestrationContext context,
            ILogger log)
        {
            var data = context.GetInput<OrchestratorInputModel>();
            try
            {
                if (data.Something)
                {
                    await context.CallActivityAsync("Calculate", data);
                    context.ContinueAsNew(data);
                    return;
                }
                await context.CallActivityAsync("Publish", data);
            }
            catch (SomeException exception)
            {
                await context.CallActivityAsync("UnrecoverableErrorHandler", exception);
            }
        }
```

### Known workarounds
Redeploying to point to a new task hub worked for a few weeks before the problem returned.

### App Details
- **Microsoft.Azure.WebJobs.Extensions.DurableTask**: 2.6.0
- **Azure Functions runtime version**: 3
- **Programming language used**: C#
- **Framework**: .NET Core 3.1

### Screenshots
Columns
![bandicam 2022-04-07 12-24-07-299](https://user-images.githubusercontent.com/21966883/162096455-69945ac1-debd-47be-af0b-7ddba2dc719a.png)

Correctly ending instance.
![bandicam 2022-04-07 12-21-42-707](https://user-images.githubusercontent.com/21966883/162096524-4819bde0-ed95-4087-8fca-36fa170d1bbf.png)

Instances that have run all the way through but are still running and missing the ExecutionCompleted line.
![bandicam 2022-04-07 12-21-54-390](https://user-images.githubusercontent.com/21966883/162096536-a69da605-4cd5-405e-96c6-e57bc95f9f1d.png)
[...]

PR #2139, opened 2022-04-18
[Do not merge] Add instrumentation to debug "stuck" sub-orchestrations

davidmrdavid:
_**Do not merge, for discussion only**_
Related: https://github.com/Azure/azure-functions-durable-extension/issues/2133, https://github.com/Azure/azure-functions-durable-extension/issues/2135

We recently received a few incident reports (internal and external) describing orchestrations that randomly started getting "stuck" in the past few months. They all share a few similarities:
* These orchestrations technically complete all work that they scheduled, but in their last replay they fail to actually change their status from "Running" to "Completed"
* These orchestrations all seem to log "FunctionAwaited" in their last replay, but schedule no new tasks
* As far as I can tell, these orchestrations all log the following sequence of telemetry events in their last replay: "OrchestrationExecuting, FunctionAwaited, OrchestrationExecuted, FunctionCompleted, ...". As far as I'm aware, we should not see `FunctionAwaited` and `FunctionCompleted` in the same replay, so something appears to have gone wrong.

From discussion with @cgillum, it seems that behavior like this would normally occur if an "illegal await" took place in the orchestration code, i.e that a non-durable API was awaited. So far, from the user code we've seen, we have gathered no evidence of this being the case.

To help us gather more evidence, I'm hoping to create an instrumented _private_ release to help us track what path the extension (and maybe DTFx) is taking. I added two new logs in areas I thought were relevant, but I could use more help in finding more.

Can folks point me to areas of the code where we could detect / gather more context around illegal awaits? Thanks!

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -820,6 +820,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
             if (!context.IsCompleted && !context.IsLongRunningTimer)
             {
+                this.TraceHelper.ExtensionInformationalEvent(
+                   context.HubName,
+                   context.InstanceId,
+                   context.Name,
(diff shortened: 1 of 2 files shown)


torosent:

PR #2141, opened 2022-04-19
Fixed a regression in Functions v3

cgillum:
Manual testing on Functions v3 revealed a runtime issue related to the managed identity PR.  The issue is that a particular API (`GetWebJobsConnectionSection`) in the WebJobs SDK can be compiled against but doesn't exist at runtime. This seems to be a problem with the WebJobs SDK or perhaps the packaging of the Functions core tools. Regardless, this PR works around the problem by copying the WebJobs SDK code that we rely on.

The issue only impacts Functions v3. Functions v4 (running on .NET 6) works fine as-is. The conditional compilation flags attempt to isolate the workaround to just Functions v3.

FYI @wsugarman, who helped me identify this workaround.


davidmrdavid (approved):
the change LGTM! Have we already filed a ticket about the API owners?

cgillum:
> Have we already filed a ticket about this to the API owners?

Nope - v3 is on a deprecation path and it's not customer impacting so I don't feel it's worth the energy to chase this down.

Issue #2138, opened 2022-04-18
2.7.0 release checklist


cgillum:
**Prep DTFx Release: (assigned to: @cgillum)**

- [x] Check DTFx package versions (either DT-AzureStorage only or if there were Core changes DT-AzureStorage, DT-Core, DT-Emulator, and DT-Redis)
- [x] Delete DTFx test packages from staging myget.
- [x] Publish DTFx packages to staging myget for testing. (either DT-AzureStorage only or if there were Core changes DT-AzureStorage and DT-Core)

**Prep Release (assigned to: @cgillum)**
_Due: <2-business-days-before-release>_
- [x] Update Durable Functions references (Analyzer? DTFx?) and check current version.
- [x] Locally, run `dotnet list package --vulnerable` to ensure the release is not affected by any vulnerable dependencies.
- [x] Add Durable Functions package to myget staging feed.
- [x] Check for package size, make sure it's not surprisingly heavier than a previous release.
  - The size increased from ~500 KB to about ~830 KB, likely because of the new .NET 6 target, in which case this is expected.
- [x] Merge dev into main. Person performing validation must approve PR. Important: Merge NOT Squash merge.

**Validation**
_Due: <1-business-days-before-release>_
- [x] Run private performance tests and ensure no regressions. **(assigned to: @cgillum)**
- [x] Smoke test Functions V1, Functions V2, and Functions V3 .NET apps. **(assigned to: @cgillum)**
  - [x] Functions V1
  - [x] Functions V2
  - [x] Functions V3 (this is automated by the CI)
  - [x] Functions V4
  - [x] Functions V4 - .NET Isolated
- [x] Smoke test JavaScript and Python apps. **(assigned to: @cgillum)**
  - [x] JavaScript (this is automated by the CI)
  - [x] Python
      - https://github.com/Azure/azure-functions-durable-python/issues/368)

**DTFx Release Completion (assigned to: @cgillum)**
_Due: <release-deadline>_
- [x] Publish DTFx packages to nuget (directly to nuget.org).
- [x] Publish release notes for DTFx.
- [ ] Patch increment DTFx packages that were released

**Release Completion (assigned to: )**
_Due: <release-deadline>_
- [x] Delete Durable Functions packages from myget.
- [x] Run Durable Functions release pipeline.
- [x] Push myget package to nuget (nuget.org extensions package option).
[...]

PR #2146, opened 2022-04-21
Ensure backwards compat in upperSchemaVersion handling

davidmrdavid:
Currently, Python SDK versions 1.1.0 to 1.1.3 only handle `upperSchemaVersion` values of 0 (V1) and 1 (V2). This is a problem, because `upperSchemaVersion` was supposed to support arbitrary integers to denote replay protocol versions and we just implemented `upperSchemaVersion.V3` in the Extension.

Technically speaking, the error is due to a bug in the Python SDK. However, due to the auto-upgrading nature of Bundles, I think it's important to consider how to avoid impacting existing SDKs when we're making a minor Extension release. This PR proposes one way to do that.

Essentially, we're "deprecating" the `upperSchemaVersion` field sent to the SDKs and instead utilizing a new field `newUpperSchemaVersion` to include protocol versions larger than V2. This will give the Python SDK the possibility of supporting new replay schemas without prevent older SDK releases from working with Extension Bundles.

Since this change is so minimal, I think we could include this change without needing to re-run all the validation work for Extension release `2.7.0`. The only affected path should be the OOProc smoke tests, which can be executed easily.  Alternatively, we can always make a `2.7.1` hotpatch release with this change after `2.7.0` is out. 

Related to # https://github.com/Azure/azure-functions-durable-python/issues/368 and https://github.com/Azure/azure-functions-durable-extension/issues/2138

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Bindings/OrchestrationTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/OrchestrationTriggerAttributeBindingProvider.cs
@@ -207,7 +207,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     new JProperty("instanceId", arg.InstanceId),
                     new JProperty("isReplaying", arg.IsReplaying),
                     new JProperty("parentInstanceId", arg.ParentInstanceId),
-                    new JProperty("upperSchemaVersion", SchemaVersion.V3),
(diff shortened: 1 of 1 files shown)


davidmrdavid:
I have validated that the JS and Python smoke tests pass with these changes.

cgillum:
What’s the impact to Durable JS when we reduce upperSchemaVersion back to 2? Do any newly introduced features stop working? Do we need a follow-up work item to restore any functionality that looks for upperSchemaVersion 3?

PR #2151, opened 2022-04-25
Replace .NET 6 with .NET Standard 2.0

cgillum:
In v2.7.0, we added a .NET 6.0 target to support gRPC-based out-of-proc. However, the .NET 6 target was incompatible with extension bundles and was therefore mostly useless. This PR removes the net60 TFM (target framework moniker) and goes back to only supporting .NET Standard 2.0 and .NET Framework 4.6.x.

The gRPC dependency was updated in the Microsoft.DurableTask.Sidecar project, which has been updated to 0.3.1 and no longer requires .NET 6.

One nice benefit of this change is that it reduces the size of the nuget package since there are two targets instead of one.

The previous version of the .NET 6 target code used various C# features that weren't supported in .NET Standard 2.0. To avoid needing to completely rewrite that code, some snippets of .NET code was copied into the `_CSharpLanguageHelpers.cs` file, following a standard pattern for using newer C# features in older .NET framework versions.

### Pull request checklist

* [x] ~~I have added all required tests (Unit tests, E2E tests)~~ (n/a)


cgillum:
Running into an issue with the V3 smoke test b/c the docker build is failing with the following errors:

> CSC : error CS8032: An instance of analyzer Microsoft.Extensions.Logging.Generators.LoggerMessageGenerator cannot be created from /root/.nuget/packages/microsoft.extensions.logging.abstractions/6.0.0/analyzers/dotnet/roslyn3.11/cs/Microsoft.Extensions.Logging.Generators.dll : Could not load file or assembly 'Microsoft.CodeAnalysis, Version=3.11.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. The system cannot find the file specified.. [/root/src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj]

> CSC : error CS8032: An instance of analyzer System.Text.Json.SourceGeneration.JsonSourceGenerator cannot be created from /root/.nuget/packages/system.text.json/6.0.0/analyzers/dotnet/roslyn3.11/cs/System.Text.Json.SourceGeneration.dll : Could not load file or assembly 'Microsoft.CodeAnalysis, Version=3.11.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. The system cannot find the file specified.. [/root/src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj]

I suspect this has something to do with some of the nuget package changes in the sidecar project. Still investigating.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -57,66 +57,55 @@ internal class LocalGrpcListener : IHostedService, IDisposable
 
         public void Dispose()
         {
-            this.host?.Dispose();
+            this.lifetime?.SignalShutdown();
+            this.lifetime?.Dispose();
         }
 
-        public async Task StartAsync(CancellationToken cancelToken = default)
+        public Task StartAsync(CancellationToken cancelToken = default)
```
davidmrdavid, line 64: Why did we end up dropping the `async` keyword?

davidmrdavid: I see in the StopAsync method we actually add the `async` keyword, which also throws me off :P . Is this all somehow related to async support in .net6?

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -43,6 +43,10 @@ public OutOfProcMiddleware(DurableTaskExtension extension)
         /// <exception cref="SessionAbortedException">Thrown if there is a recoverable error in the Functions runtime that's expected to be handled gracefully.</exception>
         public async Task CallOrchestratorAsync(DispatchMiddlewareContext dispatchContext, Func<Task> next)
         {
+#if FUNCTIONS_V1
+            await Task.Yield(); // Used to supress CS1998: This async method lacks "await" operators (by design)
```
davidmrdavid, line 47: Is there a reason why we wouldn't just suppress this warning?

src/WebJobs.Extensions.DurableTask/_CSharpLanguageHelpers.cs
```diff
@@ -0,0 +1,265 @@
...
+{
+    /// <summary>Represent a type can be used to index a collection either from the start or the end.</summary>
+    /// <remarks>
+    /// Index is used by the C# compiler to support the new index syntax.
+    /// <code>
+    /// int[] someArray = new int[5] { 1, 2, 3, 4, 5 } ;
+    /// int lastElement = someArray[^1]; // lastElement = 5
+    /// </code>
+    /// </remarks>
+    internal readonly struct Index : IEquatable<Index>
```
davidmrdavid, line 45: Just out of curiosity, how did you know to copy this specific class? Was there a compiler error saying this was missing? I just don't see anywhere in the code where you call this directly, so I'm unsure how this specific implementation ended up in our _CSharpLanguageHelpers. Thanks!

davidmrdavid (commented):
Left a few questions

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
cgillum: > Why did we end up dropping the async keyword?

Because we no longer use `await` in this method (there's no async methods being called anymore).

> I see in the StopAsync method we actually add the async keyword, which also throws me off :P . Is this all somehow related to async support in .net6?

No, it's simply an accidental artifact of refactoring the code to use different APIs. StopAsync can be implemented with or without `async`.

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
cgillum: It requires suppressing it at the method declaration level, which means you can only apply the suppression to the whole method or not at all. I only wanted to suppress it for `FUNCTIONS_V1`, so I suppressed in in this way.

src/WebJobs.Extensions.DurableTask/_CSharpLanguageHelpers.cs
cgillum: The C# compiler uses this type to implement the language features mentioned above in the comments. We're not using them directly. And yes, this struct is mentioned in the compiler error message but my knowledge of knowing to use it primarily came from researching how to use these newer language features when targeting .NET Standard 2.0.

cgillum:
The CI error uncovered a fundamental issue, unfortunately (or fortunately!), which required me to change my strategy a bit. The sidecar library now has separate targets for .NET 6 and .NET Standard. I did this to remove some dependencies from the .NET Standard 2.0 target which was breaking the .NET V3 smoke test.

src/WebJobs.Extensions.DurableTask/_CSharpLanguageHelpers.cs
```diff
@@ -0,0 +1,265 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
wsugarman, line 1: You're welcome to borrow some files that I've been using for `init`-only properties and all of the nullable attributes:
- https://github.com/wsugarman/Sweetener/blob/main/src/Shared/NullableAttributes.cs
- https://github.com/wsugarman/Sweetener/blob/main/src/Shared/IsExternalInit.cs

cgillum:
This PR has been replaced by https://github.com/Azure/azure-functions-durable-extension/pull/2157.

PR #2157, opened 2022-05-01
Remove sidecar dependency and add netcoreapp3.1 TFM

cgillum:
The purpose of this PR is to get Java support working with extension bundles. Java support already works with extensions.json, but it doesn't work with extension bundles because extension bundlers are not able to target the net60 TFM (which was required in the v2.7.0 release).

### Background

In https://github.com/Azure/azure-functions-durable-extension/pull/2151, I attempted to enable bundles support by removing the net60 TFM and having everything target netstandard2.0. This almost worked, except that the gRPC libraries made it impossible to support .NET Core 2.2, which meant we couldn't support Azure Functions v2.0. While it's true that both .NET Core 2.2 and Functions 2.0 are deprecated, there are still many Durable apps running on Functions V2 (we estimate 1000+ such apps), meaning we could accidentally break many of them if a) they depend on bundles or b) they try to upgrade to newer extension versions for bug fixes. This PR is a non-breaking alternative. The tradeoff is that we're not able to support V2 or V3 bundles. **With this PR, Java will be supported only on Functions V3 or higher and with bundles V4 or higher**. Durable will still be functional on Functions V2.

### Summary of changes

* Updates extension version to v2.7.1
* Removes the dependency on Microsoft.DurableTask.Sidecar, which required .NET 6. Everything we need is now directly included in the extension.
* Adds a netcoreapp3.1 target, which we're told can be supported in bundles V4.
* Adds a Functions V2 smoke test to ensure we don't accidentally break V2 in this change or any future changes.
* Removes the extra API documentation XML files. These were unnecessarily bloating the nuget package file since they aren't used for generating API documentation.
* Refactored WebJobs.Extensions.DurableTask.csproj to make it easier to understand which target has which dependencies (this file was getting a bit unruly)
* Added new compile-time constants, `FUNCTIONS_V2_OR_GREATER` and `FUNCTIONS_V3_OR_GREATER` to help manage which code gets compiled for which platform.

### Pull request checklist

* [x] My changes **do not** change the major or minor version of the WebJobs.Extensions.DurableTask package


test/FunctionsV2/WebJobs.Extensions.DurableTask.Tests.V2.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net60</TargetFramework>
```
davidmrdavid, line 4: What is the difference between these two?

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -16,53 +16,59 @@
...
   </ItemGroup>
-
-  <ItemGroup Condition="'$(TargetFramework)' != 'net462'">
+  
+  
+  <!-- 
+    .NET Standard 2.0 settings: These are for Functions V2 and above but may not contain some V3+ settings.
+    This is also the target for extension bundles V2 and V3.
+  -->
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
```
davidmrdavid, line 66: Should we add a comment noting Functions V2 is deprecated and that therefore we're less inclined to update these dependencies?

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -408,8 +401,8 @@ private static FailureDetails GetFailureDetails(Exception e)
             const string ExceptionDelimeter = "\nException: ";
             const string StackDelimeter = "\nStack: ";
 
-            ReadOnlySpan<char> messageSpan = rpcExceptionMessage;
-            int startException = messageSpan.IndexOf(ExceptionDelimeter);
+            ReadOnlySpan<char> messageSpan = rpcExceptionMessage.AsSpan();
+            int startException = messageSpan.IndexOf(ExceptionDelimeter.AsSpan());
```
davidmrdavid, line 405: I notice a handful of these `.AsSpan()` and `ToArray()` casts have been introduced. What happened that leads to these being _now_ necessary?

davidmrdavid (commented):
Left a few questions.
Also, given that you added a new V2 dotnet smoke test: can we also add a JS or Python smoke test using Function V2? Since we're dealing with the OOProc code-paths, it seems particularly relevant that we test an OOProc PL in V2.

test/FunctionsV2/WebJobs.Extensions.DurableTask.Tests.V2.csproj
cgillum: Sorry, this is an artifact of me going back and forth. They both seem to work, but the original `net6.0` seems to be what's in the official documentation. I'll revert this change.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
cgillum: I don't know what comment I'd write here. There are still a lot of customers using Durable on Functions V2 and I think the decision about whether to make updates to these dependencies may need to be made on a case-by-case basis.

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
cgillum: The change is that we've downgraded from .NET 6 to .NET Core 3.1 and have lost access to some of these convenient and/or high-performance enhancements.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
davidmrdavid: Fair enough. I was just concerned about the duplication of dependency declarations across TFMs, but I guess there's nothing we can really do about that. We can leave this as is

davidmrdavid (approved):
New changes make sense to me, thanks!
LGTM so long as CI goes green

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
cgillum: I originally considered removing the duplication, but I figured it's very possible that we may want to change the common dependencies across these two targets.  Having them be declared independently makes it easier for us to change one without accidentally changing the other (or otherwise adding complicated conditional logic).

cgillum:
Ugh...copy/paste error on the GitHub actions workflow name....will update this.

cgillum:
I'm going to test this against a modified V4 extension bundle that targets `netcoreapp3.1` (or `net6.0`) to ensure everything works properly end-to-end before merging.

davidmrdavid:
Sounds good. Do reach out if you need assistance with that!

cgillum:
Verified end-to-end using local preview bundles!

PR #2168, opened 2022-05-06
Throw on illegal await

davidmrdavid:
This adds a check for "illegal awaits" at the end of orchestrator code execution. If an illegal await is detected, we throw an exception.

This is necessary because currently we only check for illegal awaits each time a DF API is invoked, so we're missing the case where the illegal await occurs **without** a subsequent DF API being used.

This exception will not transition the DTFx-level orchestrator state to `Failed`, but it should result in Functions-level telemetry indicating that an illegal await took place.

Related to: https://github.com/Azure/durabletask/pull/725

### Pull request checklist

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -1,6 +1,6 @@
 ## New Features
 
-None
+- Improve telemetry to detect "illegal"/non-DF Tasks being awaited (https://github.com/Azure/azure-functions-durable-extension/pull/2168)
 
 ## Bug fixes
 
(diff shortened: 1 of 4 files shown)


davidmrdavid:
@cgillum

My testing focused on the scenario this was trying to catch: the illegal await at the **end** of the orchestrator code.

So I mostly just added `await Task.Delay(<someNum>);` at the end of simple orchestrator implementations in C# and checked that the exception was thrown, and it showed up in the `func cli` logs.

Frankly, I did not test other existing scenarios: I was hoping those would get caught in existing E2E tests, smoke tests, and perf tests. I did test that a  simple, legal, C# orchestrator is not affected by this change, but I haven't tried anything too complex.

If you'd like, I can grab the C# DF Patterns samples and triple check they continue to work with this change.

cgillum (approved):
Okay thanks - I think I feel reasonably good about this change.

PR #2181, opened 2022-05-18
Added smoke test for java

shreyas-gopalakrishna:
- Added simple java app with orchestration and activity for smoke test
- Added workflow for java

### Pull request checklist


cgillum (changes requested):
Instead of relying on extension bundles, we should use extension.csproj and reference the locally built extension. Otherwise, we won't be able to catch any regressions with this smoke test. Take a look at the other smoke tests we have (like for node.js or Python) as an example for how to do this.

test/SmokeTests/OOProcSmokeTests/durableJava/extensions.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
cgillum, line 3: Check to see if changing the `TargetFramework` value to `net6.0` will resolve the CI problem. The exception you see in the failure message indicates that the wrong flavor of the extension was loaded - i.e., the version which isn't compatible with the changes requried for Java to work.

cgillum (commented):
Glad to see this is working, but I worry about the fact that we're checking in a bunch of .jar files, which will need to be manually changed over time. Could we instead include the maven build tools as part of the Docker image, clone the Java SDK repo's main branch, and build it as part of the container build process? I feel like that would be more sustainable in the long run.

cgillum (approved):
I changed my mind about cloning the Java SDK repo (which currently isn't public anyways). I think it makes more sense to keep what you've done for now and move to using the public SDK once it's publicly available. @shreyas-gopalakrishna can you open an issue in this repo to track removing the jar files after the SDK dependency has been published to a publicly available maven repository?

PR #2194, opened 2022-05-27
Support instanceIdPrefix query parameter for HTTP instance queries

sebastianburckhardt:
It appears that the instancePrefix parameter is not currently available in the HTTP API when querying instances.

This is probably a simple oversight. This PR adds an optional query parameter 'instanceIdStartsWith' so that instances can be filtered by the instance prefix, just like already supported for the internal query SDKs.

[ ] Documentation PR is ready to merge and referenced in `pending_docs.md`
[x] I've added my notes to `release_notes.md`
[x] My changes **do not** need to be backported to a previous version
[x] I have added all required tests
[x] My changes **do not** require any extra work to be leveraged by OutOfProc SDKs
[x] My changes **do not** change the version of the WebJobs.Extensions.DurableTask package

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
+++ b/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
@@ -47,7 +47,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         private const string ShowHistoryOutputParameter = "showHistoryOutput";
         private const string ShowInputParameter = "showInput";
         private const string FetchStateParameter = "fetchState";
-        private const string InstanceIdStartsWithParameter = "instanceIdStartsWith";
+        private const string instanceIdPrefixParameter = "instanceIdPrefix";
         private const string CreatedTimeFromParameter = "createdTimeFrom";
         private const string CreatedTimeToParameter = "createdTimeTo";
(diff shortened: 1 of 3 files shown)


src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -47,6 +47,7 @@ internal class HttpApiHandler : IDisposable
         private const string ShowHistoryOutputParameter = "showHistoryOutput";
         private const string ShowInputParameter = "showInput";
         private const string FetchStateParameter = "fetchState";
+        private const string InstanceIdStartsWithParameter = "instanceIdStartsWith";
```
cgillum, line 50: In the SDKs, we call this "instance ID *prefix*". Any reason not to be consistent with this query string name?

cgillum (commented):
Thanks for putting this together so quickly! I'm just wondering about the name...

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
sebastianburckhardt: The "startswith" is a better name IMO, as it is more self-explanatory and is commonly used in other circumstances (e.g. Kusto and C#). 

Then again, consistency inside the codebase is also worth something.

I am o.k. with either.

PR #2196, opened 2022-05-27
Add explicit DTFx.Core dependency

davidmrdavid:
Currently, the Durable Extension takes a dependency on DTFx.AzureStorage, which transitively takes on a dependency on DTFx.Core.

This means that the extension cannot benefit from DTFx.Core changes without an accompanying DTFx.AzureStorage release, irrespective of whether any DTFx.AzureStorage changes warrant that release.

This PR adds an explicit DTFx.Core reference to the project, to circumvent this issue.


davidmrdavid:
FYI @amdeel, @bachuv

cgillum (approved):
:shipit:

PR #2215, opened 2022-06-22
Analyzer: Added Dependency Injection Analyzer

amdeel:
Added Dependency Injection Analyzer to our Deterministic Method Analyzers to flag variables used in deterministic methods that have been injected in orchestrator classes.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask.Analyzers/SyntaxNodeUtils.cs
+++ b/src/WebJobs.Extensions.DurableTask.Analyzers/SyntaxNodeUtils.cs
@@ -120,17 +120,17 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers
 
         public static bool IsInsideOrchestrationTrigger(SyntaxNode node)
         {
-            if (TryGetMethodDeclaration(node, out SyntaxNode methodDeclaration))
+            if (TryGetMethodDeclaration(node, out MethodDeclarationSyntax methodDeclaration))
             {
-                var parameters = ((MethodDeclarationSyntax)methodDeclaration).ParameterList.ChildNodes();
+                var parameters = methodDeclaration.ParameterList.ChildNodes();
 
                 foreach (SyntaxNode parameter in parameters)
                 {
                     var attributeLists = parameter.ChildNodes().Where(x => x.IsKind(SyntaxKind.AttributeList));
-                    foreach (SyntaxNode attribute in attributeLists)
+                    foreach (SyntaxNode attributeList in attributeLists)
                     {
                         //An AttributeList will always have a child node Attribute
-                        if (attribute.ChildNodes().First().ToString().Equals("OrchestrationTrigger"))
+                        if (attributeList.ChildNodes().First().ToString().Equals("OrchestrationTrigger"))
                         {
                             return true;
                         }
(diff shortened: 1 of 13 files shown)


src/WebJobs.Extensions.DurableTask.Analyzers/SyntaxNodeUtils.cs
```diff
@@ -196,6 +196,8 @@ private static bool TryGetAttribute(SyntaxNode node, string attributeName, out S
 
         internal static bool TryGetInvocationExpression(SyntaxNode node, out SyntaxNode invocationExpression) => TryGetContainingSyntaxNode(node, SyntaxKind.InvocationExpression, out invocationExpression);
 
+        internal static bool TryGetClassDeclaration(SyntaxNode node, out SyntaxNode classDeclaration) => TryGetContainingSyntaxNode(node, SyntaxKind.ClassDeclaration, out classDeclaration);
```
cgillum, line 199: Would it make sense to change this `out` parameter to be `ClassDeclarationSyntax` instead of `SyntaxNode`?

test/WebJobs.Extensions.DurableTask.Analyzers.Test/Orchestrator/DependencyInjectionAnalyzerTests.cs
```diff
@@ -0,0 +1,161 @@
...
+            };
+
+            VerifyCSharpDiagnostic(test, expectedDiagnostics);
+        }
+
+        protected override DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer()
+        {
+            return new DeterministicMethodAnalyzer();
+        }
+    }
```
cgillum, line 160: Can we add one more test with both an orchestrator and an activity function in the same class, but with the injected variable consumed only by the activity function (which should be legal)?  Something like this:

```csharp
public class HelloSequence
{
    private string injectedVariable;
    public HelloSequence(string injectedVariable)
    {
        this.injectedVariable = injectedVariable;
    }

    [FunctionName(""DIAnalyzerTestCases"")]
    public async Task Run([OrchestrationTrigger] IDurableOrchestrationContext context)
    {
        return context.CallActivityAsync<object>("DIAnalyzerTestActivity", null);
    }

    [FunctionName("DIAnalyzerTestActivity")]
    public async Task Run([ActivityTrigger] Object anything)
    {
        var usingDIVar = this.injectedVariable; // this should be fine
    }
}
```

cgillum (commented):
Some initial thoughts below.

test/WebJobs.Extensions.DurableTask.Analyzers.Test/Orchestrator/DependencyInjectionAnalyzerTests.cs
amdeel: Great point, I'll add that test.

yell0wfl4sh:
Hi @cgillum , @amdeel !

We are using DTF in production and have written all of our orchestrators where we use dependency injection for several configurations, for ex. ECS flighting. We are noticing several errors when we tried to upgrade "Microsoft.Azure.WebJobs.Extensions.DurableTask" to Version="2.8.0", where all dependency injected paths are now failing. This is big blocker for us for upgrading the version, and massive developer effort to rewrite all orchestrators where we move these units to activities. Also versioning these changes into our code in production also seems to challenge. 

We are trying to understand how such a breaking change was released without extensive description in release note. We would also like to know if there are any more these kind of gotcha's that we should keep in mind while coding and some good practices to properly roll out code change. We would also like to understand the importance of this change, as the existing model has been in production for couple of years now.


Thanks,
Yash

PR #2234, opened 2022-07-22
handle internal errors like OOM in entity processing by aborting the …

### LINKED ISSUE #2166: Azure Durable Entity cannot be reached anymore , seems like signals are being queued and not read
### Description
I am using Azure Durable Functions and have this running in production. All goes well, but for one specific Durable Entity, it seems I can no longer reach that entity anymore.  Signals, Calls... Everything seems to be ignored. And when using the DFM (Durable Functions Monitor), I remember seeing something like queue depth (or queue length) that was rather high.

### Expected behavior
When sending a signal to the entity, the logic to handle that signal should be performed.

### Actual behavior
No logic is being performed, as it seems all messages are queued / locked and not being picked up.

I don't believe the code snippets will be that relevant, as it's always been working for that entity and others as well.  It seems to be more of an infrastructure issue.

```csharp
await _entityClient.SignalEntityAsync(CotacolUserEntity.GetUserEntityId(userId), nameof(CotacolUserEntity.AddBookmarksAsync), bookmarks);
```

### Known workarounds
- I would be happy to clear that queue and lose any existing signals. How could I do this?
- I would even be happy to clear/remove that entity, but even that does not seem possible, as I have only found to clear state of an entity from inside an entity itself.
- Is there any troubleshooting tips someone can give?

### App Details

- **Durable Functions extension version**: 2.6.1 
- **Azure Functions runtime version**: 3.5.2.0
- **Programming language used**: C# 8.0

- **Timeframe issue observed**: since few weeks
- **Function App name**: cotacol-hunting-workers
- **Function name(s)**: CotacolUserEntity
- **Azure region**: West-Europe
- **Orchestration instance ID(s)**: @cotacoluserentity@4842724
- **Azure storage account name**: cotacolhuntingfx

sebastianburckhardt:
Customer observed a case (#2166) where an entity goes into a permanently failed state after encountering an OOM error.

This PR addresses this issue by
- catching OOM (and other) exceptions inside the `TaskEntityShim.Execute` method
- throwing a SessionAbortedException at some later point to ensure the workitem gets aborted

resolves #2166

### Pull request checklist

* [x] I've added my notes to `release_notes.md`

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -1128,11 +1128,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 }
             }
 
-            await entityContext.RunDeferredTasks();
-
-            // If there were internal errors, do not commit the batch, but instead rethrow
+            // If there were internal errors, throw a SessionAbortedException
             // here so DTFx can abort the batch and back off the work item
-            entityContext.ThrowInternalExceptionIfAny();
+            entityContext.AbortOnInternalError();
+
+            await entityContext.RunDeferredTasks();
         }
 
         internal string GetDefaultConnectionName()
--- a/release_notes.md
+++ b/release_notes.md
@@ -1,6 +1,7 @@
 ## Bug Fixes
 - Correctly Serialize HostStoppingEvent in ActivityShim (https://github.com/Azure/azure-functions-durable-extension/pull/2178)
 - Fix NotImplementedException for management API calls from Java client (https://github.com/Azure/azure-functions-durable-extension/pull/2193)
+- Handle OOM and other exceptions in entity shim by aborting the session (https://github.com/Azure/azure-functions-durable-extension/pull/2234)
 
 ## Enhancements
 - add optional 'instanceIdPrefix' query parameter to the HTTP API for instance queries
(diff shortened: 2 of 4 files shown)


src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -1128,11 +1128,11 @@ private async Task EntityMiddleware(DispatchMiddlewareContext dispatchContext, F
                 }
             }
 
-            await entityContext.RunDeferredTasks();
-
-            // If there were internal errors, do not commit the batch, but instead rethrow
+            // If there were internal errors, throw a SessionAbortedException
             // here so DTFx can abort the batch and back off the work item
-            entityContext.ThrowInternalExceptionIfAny();
```
cgillum, line 1135: I noticed that there were two places where we previously called `ThrowInternalExceptionIfAny`. Does the other location not require any changes?

cgillum (approved):
Just one question which may or may not require changes, but otherwise I think this change looks good.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
sebastianburckhardt: I think the other location is alright: it throws *after* DTFx has done its thing (persist history etc). At that point, throwing an exception is simply a way to tell the Functions telemetry that something went wrong inside this function. We do this for user code errors also, so they show up in the portal.

sebastianburckhardt: I should probably test that.

sebastianburckhardt: Ok, I tested it: it is fine. Those exceptions are being caught and wrapped, and then reported. My explanation above was not right though: it does not happen after DTFx finishes, but before.

cgillum (approved):

PR #2244, opened 2022-08-10
Analyzer: Excluded null-state attributes from BindingAnalyzer

### LINKED ISSUE #2222: Warning DF0112: Using binding 'NotNullWhen' with an OrchestrationTrigger method violates deterministic code constraints.
### Description
Getting `Warning DF0112: Using binding 'NotNullWhen' with an OrchestrationTrigger method violates deterministic code constraints.`

### Expected behavior
To my knowledge, `NotNullWhen` is perfectly deterministic, and is incorrectly being misinterpreted (because it's an attribute?).

### Actual behavior
A warning is logged during build.

```csharp
[FunctionName(nameof(DemoOrchestrator))]
public async Task DemoOrchestrator(
    [OrchestrationTrigger] IDurableOrchestrationContext context
){
  await DemoInlineFunction(context);
}

private static async Task DemoInlineFunction(
    IDurableOrchestrationContext context
){
   if(new DemoObject().TryGetValue(5, out var demoResult){
      context.SetCustomStatus("Doesn't matter");
   }
}

public class DemoObject{
  private static Dictionary<int, int> _map = new();
  public bool TryGetValue(int input, [NotNullWhen(true)] out int? demoOut)
  {
      if (_map.TryGetValue(input, out var c))
      {
          demoOut = c;
          return true;
      }
      demoOut = null;
      return false;
  }
}
```

### Known workarounds
No "good" workarounds.
The warning can be ignored or the `NotNullWhen` attribute can be removed

### App Details

- **Durable Functions extension version (e.g. v1.8.3)**: 2.7.2
- **Durable Functions Analyzer version**: 0.4.2

### Screenshots
N/A

### If deployed to Azure
N/A

amdeel:
This PR explicitly excludes null-state attributes from analysis when using the BindingAnalyzer.

resolves #2222 

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/Orchestrator/BindingAnalyzer.cs
+++ b/src/WebJobs.Extensions.DurableTask.Analyzers/Analyzers/Orchestrator/BindingAnalyzer.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System.Collections.Generic;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
@@ -32,7 +33,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers
                 if (descendant is AttributeSyntax attribute)
                 {
                     var attributeName = attribute.Name.ToString();
-                    if (attributeName != "OrchestrationTrigger")
+                    if (attributeName != "OrchestrationTrigger" && !IsExcludedAttributeName(attributeName))
                     {
                         var diagnostic = Diagnostic.Create(Rule, attribute.GetLocation(), attributeName);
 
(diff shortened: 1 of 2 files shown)


cgillum:
Looking at the original issue and looking at the test code for this PR, I was surprised that the analyzer was flagging parameters for methods that aren't functions. My expectation was that the analyzer would only consider function parameters -i.e., parameters in methods that have the `[FunctionName]` attribute.

Would that be another way to solve this problem?

amdeel:
@cgillum It's flagging the attributes because it's a method called in the orchetrator, so it's considered deterministic code. Do you think I should be checking code on orchestrator functions only for this analyzer instead?

cgillum:
I was just wondering why the analyzer would check method parameters on arbitrary methods. It wasn't clear to me how method parameters relate to orchestrator code constraints. I was assuming that this particular check was looking for illegal binding parameters, which are only relevant in function signatures.

amdeel:
@cgillum The deterministic analyzers currently analyze all orchestrator methods and methods called in orchestrators. I can just add a check on this analyzer specifically to prevent it from flagging code in methods without the FunctionName parameter on it, if you agree with that solution.

cgillum:
> I can just add a check on this analyzer specifically to prevent it from flagging code in methods without the FunctionName parameter on it

This sounds great to me. This approach also helps ensure that we don't accidentally flag other parameter attributes, including custom parameter attributes (those written by the user and not provided by the framework) that we couldn't possibly know about ahead of time.

cgillum (approved):

Issue #1418, opened 2020-07-29
Durable entity with null state persistence


emreertugrul:
Hi, 
I've created a durable entity function that does some action, and when a certain condition is met it deletes itself with `Entity.Current.DeleteState()`

When this is done the state is cleared (becomes null) but the entity still persists and is sent within the results when ListEntitiesAsync is called:
` var allRunningEntities = await client.ListEntitiesAsync(new EntityQuery
            {
                EntityName = nameof(PlayingMatchEntity),
                FetchState = true,
            }, new System.Threading.CancellationToken());`
Result:
```
  {
    "entityId": {
      "name": "playingmatchentity",
      "key": "4ilku4ut596biciph693j3pey"
    },
    "lastOperationTime": "2020-07-29T11:31:31.1441338Z",
    "state": null
  },
```
I cant find in any documentation on how to get rid of these null-state entities..

ConnorMcMahon:
This has been discussed in the past in [this issue](https://github.com/Azure/azure-functions-durable-extension/issues/931). I believe the conclusion reached in that issue was that you should use the PurgeInstanceHistoryAsync() API to remove these from the table storage that the `ListEntitiesAsync()` calls from.

In general, this is something that we should probably document better, or at least make more discoverable. @cgillum, @anthonychu, @sebastianburckhardt thoughts?

emreertugrul:
Yeah, the `PurgeInstanceHistoryAsync` seems to be available on IDurableOrchestrationClient which is not used (according to documentation at least) when you're dealing with only durable entities, instead IDurableEntityClient is used and the there arent any helpful methods there to purge..

ConnorMcMahon:
@emreertugrul to clarify, if you use IDurableClient (which is the union of the IDurableOrchestrationClient and the IDurableEntityClient), you can use that method. The entity's "orchestration instance id" is `@<entity-name>@<entity-key>`.

Perhaps IDurableEntityClient deserves a method like `PurgeEntityHistoryAsync()` that allows the usage of entity ids instead of just strings that represent orchestration instance ids.

sebastianburckhardt:
I think there are several issues here.
- the entities that don't have a state should be suppressed in the `ListEntitiesAsync()` results. This is currently being addressed in #1421. 
- Residual table entries (belonging to deleted entities) should be removed automatically when safe. This is tracked in #1065. 
[...]

PR #2251, opened 2022-08-17
Add support for implicit entity deletion in the durability provider

sebastianburckhardt:
In the Azure Storage provider, some metadata state has to be kept in storage even after entities are deleted, which has led to some confusion in the past (#1364, #1418) and may require users to calls to `CleanEntityStorage` to remove garbage.

With the new backends, there is no longer any need to keep such metadata. However, by default that data still remains there because there is not currently a mechanism in place to remove it.

This PR fixes this by allowing durability providers to remove entity state from storage immediately when an entity is deleted. The trick behind this mechanism is to use the conceptual "inverse" of the autostart feature: any auto-started orchestration that ends with a continueAsNew(null) can be considered to be implicitly deleted. If it should be needed again, such as if the entity receives more signals, it will just be autostarted again with input null, as if it had not been deleted.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
@@ -76,6 +76,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         public virtual bool GuaranteesOrderedDelivery => false;
 
         /// <summary>
+        /// Specifies whether this backend supports implicit deletion of entities.
+        /// </summary>
+        public virtual bool SupportsImplicitEntityDeletion => false;
+
+        /// <summary>
         /// JSON representation of configuration to emit in telemetry.
         /// </summary>
         public virtual JObject ConfigurationJson => EmptyConfig;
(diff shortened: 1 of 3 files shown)


cgillum (commented):
Would it be possible to add a test that verifies this behavior?

Are there any concerns about receiving new events for an entity after it has been implicitly deleted?

sebastianburckhardt:
> Would it be possible to add a test that verifies this behavior?

Sure, that makes sense.

> Are there any concerns about receiving new events for an entity after it has been implicitly deleted?

No, it works just as before, and just as designed: the entity will just be autostarted again. The rationale for the behavior is this: entities are never permanently deleted. The only thing that is created/deleted is the user-defined entity state. Whether or not an entity currently has any user-defined state or not, it can always receive and execute operations.

sebastianburckhardt:
Actually, about adding a test: I do not know how to write a unit test for this behavior in this repo. Our testing environment does not have any ability to run any storage provider other than the default Azure Storage provider. And the latter does not activate this behavior.

For what it's worth, I have tested this behavior in Netherite using the implementation here: https://github.com/microsoft/durabletask-netherite/pull/192

cgillum (approved):
I see. If this behavior is not active in the default Azure Storage implementation, then I think it's fine if we skip testing here since you already have tests in the Netherite repo.

PR #2256, opened 2022-08-22
Add support for cascading operations (terminate, suspend, resume).

Julio-Arroyo:
This is intended to be a proof-of-concept demo of how a simple, functional, and performant design of a cascading operations feature. The best way to read this draft PR is to focus on the big elements of the design (explained below), rather than the more minor details of code style (as much of the feedback in the suspend-resume PR has not been incorporated to this branch). If/when it is determined that this general design is a reasonable approach, I can clean this up to put it in release shape.

It exposes a new optional parameter so that operations (which at the moment include terminate, suspend, and resume), can "cascade down" or "fan-out" from the targeted orchestration to all suborchestrations.

The most important elements of this design include using history events for creating suborchestrations to maintain a collection of strings consisting of the instance IDs of suborchestrations. Therefore, when an event for terminate, suspend, or resume comes along, we can use the client operations to send the corresponding message to the suborchestrations.


davidmrdavid:
Wohoo! Thanks @Julio-Arroyo,

If you have the bandwidth, it would be great if you could pull in your changes that are currently merged in the extension repo (just pull from `dev`). I think that would help us review this work, and more easily get to a merge decision.

torosent:

PR #2258, opened 2022-08-23
Set extension version to 2.8.0

davidmrdavid:
As in title, routine PR.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask.xml
+++ b/src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask.xml
@@ -1016,12 +1016,16 @@
             <summary>
             Suspends a running orchestration instance.
             </summary>
+            <param name="instanceId">The ID of the orchestration instance to suspend.</param>
+            <param name="reason">The reason for suspending the orchestration instance.</param>
             <returns>A task that completes when the suspend message is enqueued if necessary.</returns>
         </member>
         <member name="M:Microsoft.Azure.WebJobs.Extensions.DurableTask.IDurableOrchestrationClient.ResumeAsync(System.String,System.String)">
             <summary>
             Resumes a suspended orchestration instance.
             </summary>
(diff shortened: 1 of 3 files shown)


cgillum:
As per the PR checklist, we need to also update [this line of code](https://github.com/Azure/azure-functions-durable-extension/blob/41f2d77a679add35a08ebc9ea41cc799f454cf7f/src/Worker.Extensions.DurableTask/AssemblyInfo.cs#L7) when making minor version changes to the extension.

FYI @jviau

davidmrdavid:
ah whoops, I had missed that by eagerly deleting the PR template. Let me fix that now

cgillum (approved):

jviau:
@cgillum it is possible to emit assembly attributes at build time via an MSBuild item group. The trick though, is how to determine what the version value should be at build time. How is it determined? Do we have that information available anywhere?

``` xml
<ItemGroup>
  <AssemblyAttribute Include="Microsoft.Azure.Functions.Worker.Extensions.Abstraction.ExtensionInformation" />
    <_Parameter1>Microsoft.Azure.WebJobs.Extensions.DurableTask</_Parameter1>
    <_Parameter2>2.8.*</_Parameter2>
</ItemGroup>
```

davidmrdavid:
@jviau : I think this is known at compile time if you grab them from this file: https://github.com/Azure/azure-functions-durable-extension/blob/ec4c49bc4808da4a95fac674ab3a34e6ca48079b/src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj#L7-L10

Is that a source of data you can access in this MSBuild snippet?

jviau:
Yes we _could_ access that in a MSBuild fashion, but is a dependency on that csproj something we want to take on? @cgillum was talking about moving this extension to the microsoft/durabletask-dotnet repo.

cgillum:
Yeah, my hope was that this could be a build step that doesn't require us to do anything manual. But you're right, we won't be able to do this automatically if we split things out to a new repo. We may just need to stick with manually updating this attribute, perhaps as part of the checklist for doing an extension release.

PR #2260, opened 2022-08-29
Make Linux File-Logger thread-safe

davidmrdavid:
**Motivation:** In some rare cases, we are seeing corruption in our linux telemetry. This appears to be due to concurrent writes to our log file. The root cause of these concurrent writes isn't entirely known, though some theories are being investigated.

For completeness, a leading theory is that these concurrent writes are occurring due to frequently recycling JobHosts (can occur due to config changes, failed health-checks, etc) which cause the Durable Extension to re-initialize several times, creating new and concurrent `LinuxAppServiceFileLogger`s that are, for a period of time, utilizing the same logfile.

**This PR:** This PR aims to make our logging infrastructure thread-safe.
To do this, 3 changes are introduced:
(1) `LinuxAppServiceFileLogger` now utilizes a semaphore to block competing threads from writing to the log-file at the same time.
(2) `LinuxAppServiceFileLogger` and `LinuxAppServiceLogger` now both implement the `IDisposable` method, to communicate that these classes ought to be "free'd" when a JobHost restart occurs.
(3) `LinuxAppServiceLogger` is not a local reference of the `DurableTaskExtension` class instead of a static reference. This ensure we can properly dispose of it, and its inner file-logger resource, in the case of a JobHost restart.

resolves: N/A

**Other comments:**  This PR is missing a test, though this kind of multithreading behavior is hard to reliably simulate. I'm curious to hear from you all if you think a test is worthwhile to set up for this, and your recommendations the setup. _Thanks!_

We will probably also need to backport this to DF V1, which I believe also supports linux logging.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -88,6 +88,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         private bool isTaskHubWorkerStarted;
         private HttpClient durableHttpClient;
         private EventSourceListener eventSourceListener;
+        private LinuxAppServiceLogger linuxLogger;
 
 #if FUNCTIONS_V1
         private IConnectionInfoResolver connectionInfoResolver;
@@ -339,6 +340,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             {
                 this.InitializeLinuxLogging();
             }
+
 #endif
 
             ConfigureLoaderHooks();
@@ -497,16 +499,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             string stampName = this.PlatformInformationService.GetLinuxStampName();
             string containerName = this.PlatformInformationService.GetContainerName();
 
-            // If running in linux, initialize the EventSource listener with the appropiate logger.
-            LinuxAppServiceLogger linuxLogger;
-            if (!inConsumption)
-            {
-                linuxLogger = new LinuxAppServiceLogger(writeToConsole: false, containerName, tenant, stampName);
-            }
-            else
-            {
-                linuxLogger = new LinuxAppServiceLogger(writeToConsole: true, containerName, tenant, stampName);
-            }
+            // in linux consumption, logs are emitted to the console.
+            // In other linux plans, they are emitted to a logfile.
+            this.linuxLogger = new LinuxAppServiceLogger(writeToConsole: inConsumption, containerName, tenant, stampName);
 
             // The logging service for linux works by capturing EventSource messages,
             // which our linux platform does not recognize, and logging them via a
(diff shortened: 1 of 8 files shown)


src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -515,16 +509,13 @@ private void InitializeLinuxLogging()
...
         }
 
         /// <inheritdoc />
         public void Dispose()
         {
-            // Not flushing the linux logger may lead to lost logs
-            // 40 seconds timeout because we write normally every 30 seconds, so we're just
-            // adding an extra 10 seconds to flush.
-            LinuxAppServiceLogger.Logger?.Stop(TimeSpan.FromSeconds(40));
+            this.linuxLogger?.Dispose();
```
NickCraver, line 518: Slight change here - need to put the linuxLogger as the last dispose, because you want to ditch the eventSourceListener that depends on it first.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceFileLogger.cs
```diff
@@ -19,8 +19,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
     /// We have modified their implementation to utilize syscall.rename instead of File.Move during file rolling.
     /// This change is necessary for older versions of fluent-bit, our logging infrastructure in linux dedicated, to properly deal with logfile archiving.
     /// </summary>
-    public class LinuxAppServiceFileLogger
+    internal class LinuxAppServiceFileLogger : IDisposable
```
NickCraver, line 22: Remove this from public is a breaking API change - any concern? I'm not sure of the rules with this library.

NickCraver (changes requested):
1 question and 1 quick gotcha on the ordering - looking good, and nice cleanup in a few spots :)

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
davidmrdavid: Ah, good catch, thanks! I have moved this down as the last step.
However, I figured this would ideally be order-agnostic, and the way to make this happen is, as I understand, to override the Dispose method of the eventSourceListener so that it would dispose of the logger as well. My last PR does this.

Let me know if that seems right to you!

src/WebJobs.Extensions.DurableTask/LinuxAppServiceFileLogger.cs
davidmrdavid: I believe this class was never meant for public usage, and in fact it's public-ness may be a problem for us. @cgillum: any thoughts?

davidmrdavid:
Hmm, seems some tests failed on this last run. Re-running them now in case it was just flakiness. Re-requesting another review to keep the momentum and because the diff changed.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
NickCraver: That change is absolutely fine..._but_ if you do it this way, I recommend removing the `linuxLogger` variable all together from this class - it no longer needs a direct reference at all :) It just needs to be a local passed to the event listener creation.

src/WebJobs.Extensions.DurableTask/LinuxAppServiceFileLogger.cs
cgillum: Agreed this should never have been public. I agree with changing it to be `internal`.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
davidmrdavid: Ah, of course! Removed

NickCraver (approved):
👍

davidmrdavid:
@cgillum: could I get your thumbs-up as well?

cgillum (approved):
:shipit:

PR #2267, opened 2022-09-07
Fix issue#2032 : Display input of activity function

### LINKED ISSUE #2032: Add new configuration setting that activates storage of input data to activities if you use blob storage backend
Hello!

Today our project use [DurableFunctionsMonitor](https://github.com/microsoft/DurableFunctionsMonitor) to an high extend and its awesome! One of its features is it can display the result of an activity. What we wish is it to display the input data as well to help us understand why we did get a specific result. If I understand this correctly in this [issue from DurableFunctionsMonitors github](https://github.com/microsoft/DurableFunctionsMonitor/issues/1) it’s not possible today due to Durable Functions Framework don’t store the input data because of billing purposes if you use blob storage as the backend. What I wish is to make this configurable, if we activate a setting, we will start to store the input data to activities.  

Thanks in advance!

nytiannn:
Fixes #2032
- Display the input of Activity function. 
- Add parameter "StoreInputsInOrchestrationHistory" at file src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs.
- When "StoreInputsInOrchestrationHistory" is set as true, the input of activity will be displayed at Taskhub.

Update:
- Add "ShowHistoryInput" parameter for GetStatusAsync method at file src\WebJobs.Extensions.DurableTask\ContextImplementations\DurableClient.cs.  Because the current method couldn't test the new added parameter "StoreInputsInOrchestrationHistory". Also, this parameter could allow customer query the history input on their own. 
- When "StoreInputsInOrchestrationHistory" is true:
       if "ShowHistoryInput" is true, the input will be included in the status. 
       if "ShowHistoryInput" is false, the input will not be included in the status. 
- When "StoreInputsInOrchestrationHistory" is false:
       if "ShowHistoryInput" is true, the input will be "null". 
       if "ShowHistoryInput" is false, the input will not be included in the status.


cgillum:
Hi @NaiyuanTIANNN, I realize this is only a draft PR, but it would be helpful if you could provide more information about the purpose of this PR in the PR description above. Similarly, it would be great if you could update the title of the PR to something more descriptive. It's not clear to me if this is intended to be a real PR or if you're just testing your permissions by publishing a test PR.

NaiyuanTIANNN:
@cgillum  Thank you, Chris! David also told me that and I already changed it. Thanks for your advice!

src/WebJobs.Extensions.DurableTask/Options/AzureStorageOptions.cs
```diff
@@ -175,6 +175,7 @@ public string TrackingStoreConnectionStringName
         /// <summary>
         /// Throws an exception if the provided hub name violates any naming conventions for the storage provider.
         /// </summary>
+
```
cgillum, line 178: Please remove this extra newline.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -206,6 +206,9 @@ public string HubName
         /// <summary>
         /// If UseAppLease is true, gets or sets the AppLeaaseOptions used for acquiring the lease to start the application.
         /// </summary>
+
+        public bool IncludeParameters { get; set; } = false;
```
cgillum, line 210: This change has inadvertently removed the documentation comments for the `AppLeaseOptions` property since you inserted your new method in between the existing method and its documentation comments. Please revert this change and instead declare your new property somewhere else, preferably below `AppLeaseOptions`.

Note that I can see the change in the generated documentation by looking at the changes to Microsoft.Azure.WebJobs.Extensions.DurableTask.xml, which is also included in this PR.

Also, please take some time to review this documentation which describes XML doc comments in C#: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/. Since your new `IncludeParameters` property is public, it will also require XML comments, just like all the other public properties declared in this file.

cgillum (changes requested):

cgillum:
One more thing, when creating a PR that resolves an issue in GitHub, please add a line in your PR description like this:

```
Fixes #2032
```

GitHub will automatically link your PR to the issue mentioned and will also automatically close the issue when your PR is merged. This is a best practice when using GitHub. You can learn more about this GitHub feature [here](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
NaiyuanTIANNN: Thanks for your reminder and help! I noticed it and have already changed it. Thanks for your help and @davidmrdavid. I already pushed the changed one here.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -203,6 +203,11 @@ public string HubName
         /// </summary>
         public bool UseAppLease { get; set; } = true;
 
+        /// <summary>
+        /// If IncludeParameters is true, the input of the activity function will be displayed.
+        /// </summary>
+        public bool IncludeParameters { get; set; } = false;
```
cgillum, line 209: Let's rename this to `StoreInputsInOrchestrationHistory`. This is a better description of what the feature does compared to "IncludeParameters", which is vague.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -203,6 +203,11 @@ public string HubName
         /// </summary>
         public bool UseAppLease { get; set; } = true;
 
+        /// <summary>
+        /// If IncludeParameters is true, the input of the activity function will be displayed.
```
cgillum, line 207: I suggest we update this comment. This setting works for both activities and sub-orchestrations. It's not clear what is meant when we say it "will be displayed", so I suggest wording that suggests that any tool can display this information.
```suggestion
        /// If <c>true</c>, the inputs of activity and sub-orchestrator functions will be saved to the orchestration history, making them visible to debugging tools. The default value is <c>false</c>.
```

cgillum (commented):
Thanks for the updates. A couple more suggestions.

@davidmrdavid should we add an end-to-end test that verifies the correct behavior?

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
NaiyuanTIANNN: Thanks, Chris! I have already updated it.

davidmrdavid:
Yep, we should @cgillum. @nytiannn, as we discussed - let's look to add that end-to-end test  :-)

amdeel:
One thing that's left to discuss is how we want to allow customers to query for this input data. Currently we filter out inputs from our history events in our GetStatusAsync API. We could add an additional optional parameter to this API similar to showHistoryOutput, but name it showHistoryInput instead. When this setting is true we wouldn't filter out any input data.

There is a slight cost to modifying existing APIs. While optional parameters don't count as breaking changes, changing the API can break customers test code, so we try to make changes to this APIs sparingly and only when necessary. But the change in this case does seem necessary to allow customers to actually use this input data.

ltouro:
I would suggest to consider the filtering as another feature and treat it separately. In my case, just seeing the input is way more important than to filter on it. Filtering the orchestration input is already enough for many cases

amdeel:
@ltouro Let me clarify my above comment, you will not be able to see activity input data when using GetStatusAsync at all unless we make a change.

ltouro:
I meant about the difference of showing a value and allowing to filter by it. I think the showing is simpler and more important. The filtering could follow up in a different PR

ltouro:
I meant about the difference of showing a value and allowing to filter by it. I think the showing is simpler and more important. The filtering could follow up in a different PR. 

Maybe I'm missing something about how input data is stored that requires filtering capability to show along with activity result?

amdeel:
@ltouro There's no filtering capability being added here. This is about showing the customer the activity input.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -953,6 +983,7 @@ private static void AddScheduledEventDataAndAggregate(ref Dictionary<string, Eve
             {
                 historyItem["ScheduledTime"] = taskScheduledData.Date;
                 historyItem["FunctionName"] = taskScheduledData.Name;
+                historyItem["Input"] = taskScheduledData.Input;
```
amdeel, line 986: With this change, you're always adding this input value but we should only display this when the customer is using this new showHistoryInput setting.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -203,6 +203,11 @@ public string HubName
         /// </summary>
         public bool UseAppLease { get; set; } = true;
 
+        /// <summary>
+        /// If <c>true</c>, the inputs of activity and sub-orchestrator functions will be saved to the orchestration history, making them visible to debugging tools. The default value is <c>false</c>.
```
amdeel, line 207: We should probably keep it general to cover all functions like entities. We could say something like "functions called by the orchestrator" instead of activity and sub-orchestrator functions.

amdeel (commented):
This looks great! Just a few small changes before we merge.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
nytiannn: Thanks!! I will make it clear.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -942,17 +972,29 @@ private static async Task<DurableOrchestrationStatus> GetDurableOrchestrationSta
             return this.httpApiHandler.CreateCheckStatusResponse(request, instanceId, attribute, returnInternalServerErrorOnFailure);
         }
 
-        private static void TrackNameAndScheduledTime(JObject historyItem, EventType eventType, int index, Dictionary<string, EventIndexDateMapping> eventMapper)
+        private static void TrackNameAndScheduledTime(JObject historyItem, EventType eventType, int index, Dictionary<string, EventIndexDateMapping> eventMapper, bool showHistoryInput)
         {
-            eventMapper.Add($"{eventType}_{historyItem["EventId"]}", new EventIndexDateMapping { Index = index, Name = (string)historyItem["Name"], Date = (DateTime)historyItem["Timestamp"] });
+            if (showHistoryInput)
```
amdeel, line 977: Correct me if I'm wrong but I think we can always add Input to this mapping and it wont actually affect what is displayed for the customer. So we shouldn't need to change behavior for showHistoryInput and can leave this method with input always being mapped. It should just be null if there isn't a value.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -5746,6 +5746,29 @@ public async Task ValidateExtensionLifecycleLogs()
...
+                   this.loggerProvider,
+                   options))
+            {
+                await host.StartAsync();
+                var client = await host.StartOrchestratorAsync(nameof(TestOrchestrations.SayHelloWithActivity), "foo", this.output);
+                await client.WaitForCompletionAsync(this.output);
+                var status = await client.InnerClient.GetStatusAsync(client.InstanceId, showHistory: true, showHistoryInput: true);
+
+                var input = status.History[1].Value<string>("Input");
+                Assert.Equal("[\"foo\"]", input);
```
amdeel, line 5766: Just to be a little more thorough we should add another activity to this test that doesn't take an input and make sure Input is still returned in the GetStatusAsync result as null. You might need to create a new orchestrator if you can't find one with two activities like this.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -5758,12 +5758,17 @@ public async Task TestStoreInputsInOrchestrationHistory()
                    options))
             {
                 await host.StartAsync();
-                var client = await host.StartOrchestratorAsync(nameof(TestOrchestrations.SayHelloWithActivity), "foo", this.output);
-                await client.WaitForCompletionAsync(this.output);
-                var status = await client.InnerClient.GetStatusAsync(client.InstanceId, showHistory: true, showHistoryInput: true);
+                var client1 = await host.StartOrchestratorAsync(nameof(TestOrchestrations.SayHelloWithActivity), "foo", this.output);
+                var client2 = await host.StartOrchestratorAsync(nameof(TestOrchestrations.SayHelloWithActivity), null, this.output);
```
amdeel, line 5762: You should make another orchestrator with two different activity functions. One function that takes a string input and one that doesn't take an input. In the scenario you currently have, null is being cast to a string and passed in the SayHello function. The reason I suggested this test was to make sure we return something for input when there isn't an input used at all.

amdeel:
We're making a lot of changes to the GetStatusAsync API to make it easier for customers to customize their results. This is great for the user but is exposing a big flaw in the current signature for GetStatusAsync. It's not ideal to keep adding these options and changing the API every time we want to add more customization. I suggest we add another API with a type to represent the filter options that would make this current API Obsolete. The new API can look like this:
`GetStatusAsync(string instanceId, bool showHistory = false, HistoryFilterOptions options = HistoryFilterOptions.DefaultOptions)`

davidmrdavid:
Thanks @nytiannn for your efforts here.
I personally don't love option 1 (new API name) because, to me, creating a new API name seems more confusing than having a new overload of the existing method.

Personally, I think either options 2 or 3 are _fine_ and would not block on either of them. While I agree that growing the number of parameters in a method without care can be an issue, I don't think we're there yet so I'm not concerned about option 2 at this moment. If I had to give a preference, I would suggest to option 3 since it gives you the most flexibility in the future. 

FYI @cgillum would be good to get your thoughts here so we can unblock this PR

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -462,7 +462,7 @@ async Task IDurableOrchestrationClient.RewindAsync(string instanceId, string rea
         }
 
         /// <inheritdoc />
-        async Task<DurableOrchestrationStatus> IDurableOrchestrationClient.GetStatusAsync(string instanceId, bool showHistory, bool showHistoryOutput, bool showInput)
+        async Task<DurableOrchestrationStatus> IDurableOrchestrationClient.GetStatusAsync(string instanceId, bool showHistory, bool showHistoryOutput, bool showInput, bool showHistoryInput)
```
cgillum, line 465: Rather than introducing a new parameter, let's simply rely on the existing `showInput` parameter. Developers probably won't understand the difference between `showInput` and `showHistoryInput` anyways. I think `showInput` can safely apply to both scenarios. This is preferable so that we don't need to introduce any new surface area. The downside is that it changes the behavior of an existing API, but I think it's okay because it enables a behavior that is much more obvious.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -481,7 +481,35 @@ async Task<DurableOrchestrationStatus> IDurableOrchestrationClient.GetStatusAsyn
                 return null;
             }
 
-            return await GetDurableOrchestrationStatusAsync(state, this.client, showHistory, showHistoryOutput);
+            return await GetDurableOrchestrationStatusAsync(state, this.client, showHistory, showHistoryOutput, showHistoryInput);
+        }
+
+        /// <inheritdoc/>
+        async Task<DurableOrchestrationStatus> IDurableOrchestrationClient.QueryStatusAsync(string instanceId, InstanceFilterOption instanceFilterOption)
```
cgillum, line 488: Let's not introduce a new method. I think we can accomplish our goals as I mentioned earlier more cheaply if we simply rely on the broader interpretation of the existing `showInput` parameter.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -895,13 +945,21 @@ private static async Task<DurableOrchestrationStatus> GetDurableOrchestrationSta
...
+                                        historyItem.Remove("Input");
+                                    }
+
                                     break;
                                 case EventType.TimerFired:
                                     historyItem.Remove("TimerId");
                                     break;
                                 case EventType.EventRaised:
-                                    historyItem.Remove("Input");
+                                    if (!showHistoryInput)
```
cgillum, line 958: Instead of repeating this same code in multiple places, let's just move it down below, outside of the `switch` statement. It's okay if we call `history.Remove("Input")` on a history event that doesn't have an `"Input"` field. It will just be a no-op. This design is safer because we never have to worry about accidentally forgetting any of the `case` statements.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
```diff
@@ -256,10 +256,19 @@ public interface IDurableOrchestrationClient
         /// </summary>
         /// <param name="instanceId">The ID of the orchestration instance to query.</param>
         /// <param name="showHistory">Boolean marker for including execution history in the response.</param>
-        /// <param name="showHistoryOutput">Boolean marker for including input and output in the execution history response.</param>
+        /// <param name="showHistoryOutput">Boolean marker for including output in the execution history response.</param>
         /// <param name="showInput">If set, fetch and return the input for the orchestration instance.</param>
```
cgillum, line 260: If we follow my suggestion and reinterpret `showInput` to apply to both the orchestration status and the history items, then we can revert all the changes to this file. However, we will need to update the `showInput` description here to say: `If set, fetch and return the input for the orchestration instance. If both <c>showHistory</c> and <see cref="DurableTaskOptions.StoreInputsInOrchestrationHistory"/> are also set to <c>true</c>, then the inputs for activity and sub-orchestration events in the orchestration history will also be returned.`

src/WebJobs.Extensions.DurableTask/DurableOrchestrationStatus.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
     /// </summary>
     /// <remarks>
     /// An external client can fetch the status of an orchestration instance using
-    /// <see cref="IDurableOrchestrationClient.GetStatusAsync(string, bool, bool, bool)"/>.
```
cgillum, line 14: If we follow my suggestion to reuse `showInput`, then we can revert all changes to this file.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -46,6 +46,8 @@ internal class HttpApiHandler : IDisposable
         private const string ShowHistoryParameter = "showHistory";
         private const string ShowHistoryOutputParameter = "showHistoryOutput";
         private const string ShowInputParameter = "showInput";
+        private const string ShowHistoryInputParameter = "showHistoryInput";
+        private const string InstanceFilterOption = "instanceFilterOption";
```
cgillum, line 50: With my suggestion to reuse `showInput`, I think we can revert all changes to this file.

cgillum (commented):
Hi @nytiannn. I would like to suggest we simplify the design and scope of this PR. Please see my suggestions below. I've considered your three options above, but I'd actually prefer a simpler alternative that doesn't involve any changes to the existing client API surface area.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -895,20 +895,25 @@ private static async Task<DurableOrchestrationStatus> GetDurableOrchestrationSta
...
                                 case EventType.ExecutionTerminated:
-                                    historyItem.Remove("Input");
+
                                     break;
                                 case EventType.TimerFired:
                                     historyItem.Remove("TimerId");
                                     break;
                                 case EventType.EventRaised:
-                                    historyItem.Remove("Input");
+
```
cgillum, line 904: nit: please remove the extra newlines before the `break` statement here and in the above cases. That way your change doesn't impact the original formatting pattern of the code.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
```diff
@@ -256,8 +256,8 @@ public interface IDurableOrchestrationClient
         /// </summary>
         /// <param name="instanceId">The ID of the orchestration instance to query.</param>
         /// <param name="showHistory">Boolean marker for including execution history in the response.</param>
-        /// <param name="showHistoryOutput">Boolean marker for including input and output in the execution history response.</param>
```
cgillum, line 259: Oh, interesting. I didn't realize that `showHistoryOutput` was also documented to show the input...

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -982,15 +982,15 @@ private async Task HelloWorldOrchestration_Activity_Main_Logic(string taskHubNam
 
                     if (showHistoryOutput)
                     {
-                        Assert.Null(status.History[0]["Input"]);
```
cgillum, line 985: Is this a breaking change? Were we previously returning the orchestration history when `showHistoryOutput` was true? If so, we should not change this behavior or else existing customers may be impacted.

Tip: we always have to be extremely careful about accidentally introducing breaking changes to the code. If a change requires us to modify a unit test like this, that's a good hint that we may be accidentally making a breaking change.

cgillum (commented):
Thanks @nytiannn. This PR is now much simpler, which I'm happy about. However, I am now noticing that it appears we are introducing a breaking change. Please see my comments.

test/Common/DurableTaskEndToEndTests.cs
cgillum: Oh, sorry, I think I was wrong. This is actually *not* a breaking change because the input, which was previously null, is now non-null. I looked at the diff wrong. I thought it was the other way around.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
cgillum: So I guess it was a bug that we never displayed the input even though we said we would. Since we never showed the output before, it should be safe for us to change the documentation here, and instead rely on `showInput`.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -1396,6 +1396,11 @@ internal async Task<bool> StartTaskHubWorkerIfNotStartedAsync()
                         await this.defaultDurabilityProvider.CreateIfNotExistsAsync();
                         await this.taskHubWorker.StartAsync();
 
+                        if (this.Options.StoreInputsInOrchestrationHistory == true)
```
cgillum, line 1399: In C#, when testing `bool` values in an `if` statement, it's expected that you don't explicitly compare it to `true` or `false`. Instead, you do either `if (value) {...}` or `if (!value) {...}`. See line 1386 above for an example. For this line, let's remove the `== true` so that you are being consistent with normal C# code style.

```csharp
if (this.Options.StoreInputsInOrchestrationHistory)
```

cgillum (approved):
After further review, I'm good with these changes. There are two stylistic issues I found, but nothing blocking.

davidmrdavid:
checking...

davidmrdavid (approved):
LGTM but let's wait to hear from @amdeel as well as she was involved throughout this PR

nytiannn:
I might need @amdeel 's review about the E2E test for this new parameter. I am not sure if this updated one satisfy her commits about covering the null case.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -5746,6 +5746,34 @@ public async Task ValidateExtensionLifecycleLogs()
...
+            DurableTaskOptions options = new DurableTaskOptions();
+            options.StoreInputsInOrchestrationHistory = true;
+
+            using (var host = TestHelpers.GetJobHostWithOptions(
+                   this.loggerProvider,
+                   options))
+            {
+                await host.StartAsync();
+                var client1 = await host.StartOrchestratorAsync(nameof(TestOrchestrations.SayHelloWithActivity), "foo", this.output);
+                var client2 = await host.StartOrchestratorAsync(nameof(TestOrchestrations.CallActivityWithNoInput), null, this.output);
```
amdeel, line 5762: One way to do this is just to put both activities in the same orchestrator and check input for both history results. But I wont block on this.

amdeel (approved):
This looks good! I'm glad that we're going to have this released for the customer soon.

amdeel (approved):

PR #2269, opened 2022-09-08
fix issue#1097 : Allow array inputs for method GetStatusAsync and PurgeInstanceHistoryAsync

### LINKED ISSUE #1097: IDurableOrchestrationClient.GetStatusAsync(IEnumerable<string> instanceIds)
I am trying to compare the results of a handfull of orchestrations.

Therefor I'd like to call the GetStatusAsync Method on the DurableOrchestrationClient and pass an array of instance Ids.

there is already a overloaded method that takes datetimes and runtime states to filter the result. It should be easy to integrate the ids here.

Similarly I think that should be included in the OrchestrationStatusQueryCondition class as well as the PurgeInstanceHistoryAsync method

nytiannn:
Fixes #1097

- Add overloaded method for **IDurableOrchestrationClient.GetStatusAsync** & **IDurableOrchestrationClient.PurgeInstanceHistoryAsync** to support array inputs.
- Return for new added **IDurableOrchestrationClient.GetStatusAsync** is a list of DurableOrchestrationStatus status. 
- Return for new added **IDurableOrchestrationClient.PurgeInstanceHistoryAsync** is an PurgeHistoryResult object which contains the number of instances deleted.


src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
```diff
@@ -261,6 +261,16 @@ public interface IDurableOrchestrationClient
...
 
+        /// <summary>
+        /// Get the status of multiple instances.
+        /// </summary>
+        /// <param name="instanceID"> List of multiple instance ID that needed to check that status.</param>
+        /// <param name="showHistory">Boolean marker for including execution history in the response.</param>
+        /// <param name="showHistoryOutput">Boolean marker for including input and output in the execution history response.</param>
+        /// <param name="showInput">If set, fetch and return the input for the orchestration instance.</param>
+        /// <returns>return a list of instances of orchestration status.</returns>
+        Task<IList<DurableOrchestrationStatus>> GetStatusAsync(IEnumerable<string> instanceID, bool showHistory = false, bool showHistoryOutput = false, bool showInput = true);
```
cgillum, line 272: Let's make the default value for `showInput` false. This is a better default for us to use for new APIs since fetching input data can be very expensive. 
```suggestion
        Task<IList<DurableOrchestrationStatus>> GetStatusAsync(IEnumerable<string> instanceID, bool showHistory = false, bool showHistoryOutput = false, bool showInput = false);
```

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -563,6 +597,19 @@ async Task<PurgeHistoryResult> IDurableOrchestrationClient.PurgeInstanceHistoryA
             return await this.DurabilityProvider.PurgeInstanceHistoryByInstanceId(instanceId);
         }
 
+        /// <inheritdoc />
+        async Task<IList<PurgeHistoryResult>> IDurableOrchestrationClient.PurgeInstanceHistoryAsync(IEnumerable<string> instanceId)
+        {
+            IList<PurgeHistoryResult> purgehistoryresultList = new List<PurgeHistoryResult>();
+            foreach (string id in instanceId)
+            {
+                PurgeHistoryResult result = await this.DurabilityProvider.PurgeInstanceHistoryByInstanceId(id);
```
cgillum, line 606: Should we do these purge operations in parallel instead of one-at-a-time? Otherwise, I worry that we're not adding much value with this new API.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -484,6 +484,40 @@ async Task<DurableOrchestrationStatus> IDurableOrchestrationClient.GetStatusAsyn
             return await GetDurableOrchestrationStatusAsync(state, this.client, showHistory, showHistoryOutput);
         }
 
+        /// <inheritdoc />
+        async Task<IList<DurableOrchestrationStatus>> IDurableOrchestrationClient.GetStatusAsync(IEnumerable<string> instanceID, bool showHistory, bool showHistoryOutput, bool showInput)
+        {
+            IList<DurableOrchestrationStatus> durableorchestrationstateList = new List<DurableOrchestrationStatus>();
+            foreach (string id in instanceID)
+            {
+                IList<OrchestrationState> stateList = null;
```
cgillum, line 493: Remove the initialization to `null`. This is a safer coding practice because then we're less likely to accidentally forget to assign this variable if this code ever needs to be refactored.
```suggestion
                IList<OrchestrationState> stateList;
```

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -484,6 +484,40 @@ async Task<DurableOrchestrationStatus> IDurableOrchestrationClient.GetStatusAsyn
...
+                    stateList = await this.DurabilityProvider.GetOrchestrationStateWithInputsAsync(id, showInput);
+                }
+                catch (NotImplementedException)
+                {
+                    // TODO: Ignore the show input flag. Should consider logging a warning.
+                    stateList = await this.client.ServiceClient.GetOrchestrationStateAsync(id, allExecutions: false);
+                }
+
+                OrchestrationState state = stateList?.FirstOrDefault();
+                DurableOrchestrationStatus currentidstate = new DurableOrchestrationStatus();
```
cgillum, line 505: This is a confusing variable name and it doesn't follow C# naming guidelines. Here's a good reference for C# naming guidelines: [link](https://github.com/ktaranov/naming-convention/blob/master/C%23%20Coding%20Standards%20and%20Naming%20Conventions.md#c-coding-standards-and-naming-conventions).

Here's my suggestion:
```suggestion
                var currentStatus = new DurableOrchestrationStatus();
```

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -484,6 +484,40 @@ async Task<DurableOrchestrationStatus> IDurableOrchestrationClient.GetStatusAsyn
...
+                if (state == null || state.OrchestrationInstance == null)
+                {
+                    currentidstate.InstanceId = id;
```
cgillum, line 508: I'm not sure it's a good idea to return results for orchestration instances that don't exist. Callers may not be able to distinguish between existing and non-existing instances. It may be preferable to instead add a `null` result to the output rather than adding an incomplete result to the output.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -484,6 +484,40 @@ async Task<DurableOrchestrationStatus> IDurableOrchestrationClient.GetStatusAsyn
             return await GetDurableOrchestrationStatusAsync(state, this.client, showHistory, showHistoryOutput);
         }
 
+        /// <inheritdoc />
+        async Task<IList<DurableOrchestrationStatus>> IDurableOrchestrationClient.GetStatusAsync(IEnumerable<string> instanceID, bool showHistory, bool showHistoryOutput, bool showInput)
+        {
+            IList<DurableOrchestrationStatus> durableorchestrationstateList = new List<DurableOrchestrationStatus>();
```
cgillum, line 490: Let's simplify the variable name. Also note that the current name doesn't follow the correct camelCase convention for C# local variables. Please be mindful of these coding conventions in future PRs. When in doubt, look at the surrounding code and try to copy those conventions.

```suggestion
            var results = new List<DurableOrchestrationStatus>();
```

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
```diff
@@ -261,6 +261,16 @@ public interface IDurableOrchestrationClient
...
         Task<DurableOrchestrationStatus> GetStatusAsync(string instanceId, bool showHistory = false, bool showHistoryOutput = false, bool showInput = true);
 
+        /// <summary>
+        /// Get the status of multiple instances.
+        /// </summary>
+        /// <param name="instanceID"> List of multiple instance ID that needed to check that status.</param>
+        /// <param name="showHistory">Boolean marker for including execution history in the response.</param>
+        /// <param name="showHistoryOutput">Boolean marker for including input and output in the execution history response.</param>
+        /// <param name="showInput">If set, fetch and return the input for the orchestration instance.</param>
+        /// <returns>return a list of instances of orchestration status.</returns>
```
cgillum, line 271:
```suggestion
        /// <returns>Returns a list of orchestration status values.</returns>
```

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -4431,6 +4431,70 @@ public async Task Purge_Single_Instance_History(bool extendedSessions, string st
...
+                await host.StartAsync();
+
+                var client1 = await host.StartOrchestratorAsync(nameof(TestOrchestrations.SayHelloWithActivity), "foo", this.output);
+                var client2 = await host.StartOrchestratorAsync(nameof(TestOrchestrations.SayHelloWithActivity), "bar", this.output);
+                var client3 = await host.StartOrchestratorAsync(nameof(TestOrchestrations.SayHelloWithActivity), "baz", this.output);
+
+                string firstInstanceId = client1.InstanceId;
+                string secondInstanceId = client2.InstanceId;
+                string thirdInstanceId = client3.InstanceId;
+                IList<string> instanceID = new List<string> { firstInstanceId, secondInstanceId, thirdInstanceId };
```
cgillum, line 4454: For variables that contain a list of values, it's best to make the variable name plural or add a suffix to make it clear that the variable contains multiple values. For example, `instanceIds` or `instanceIdList` would be better names.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -4431,6 +4431,70 @@ public async Task Purge_Single_Instance_History(bool extendedSessions, string st
             }
         }
 
+        [Theory]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        [MemberData(nameof(TestDataGenerator.GetBooleanAndFullFeaturedStorageProviderOptions), MemberType = typeof(TestDataGenerator))]
+        public async Task Purgemultipleinstancehistory(bool extendedSessions, string storageProvider)
```
cgillum, line 4437: Please follow C# method naming guidelines, which are to use PascalCase. This will make the code more readable to other developers.
```suggestion
        public async Task PurgeMultipleInstanceHistories(bool extendedSessions, string storageProvider)

```

cgillum (changes requested):

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
nytiannn: Hi, Chris. Thanks for your review. I have updated already and had a small question about the output. 

The Current output for this function is a list of PurgeHistoryResult. This class only have one attribute, InctanceDeleted, meaning the number of deleted instances. So, the return for this overloaded method looks like this: ["1", "1,","1,"........]. In this way, the advantage is that customer will directly know which instance is deleted and which one is not. 

Another option for the output is just return an instance of PurgeHistoryResult. This will look more concise. The number reflects the number of deleted instances, resulting to the unknown of which one not deleted. But when reading the code, I think as long as the instance ID is valid, it will be deleted successfully. Considering this, I am unsure if the "Purge History unsuccessfully" should be concerned. 

I am unsure which one is better. I would appreciate it if you could give me some advice. :)

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -563,6 +597,23 @@ async Task<PurgeHistoryResult> IDurableOrchestrationClient.PurgeInstanceHistoryA
             return await this.DurabilityProvider.PurgeInstanceHistoryByInstanceId(instanceId);
         }
 
+        async Task<List<PurgeHistoryResult>> IDurableOrchestrationClient.PurgeInstanceHistoryAsync(IEnumerable<string> instanceId)
```
cgillum, line 600: I agree with you that it would be better to return a single `PurgeHistoryResult` object instead of returning a list. This is also consistent with the other `PurgeInstanceHistoryAsync` method that uses time and status filters to purge multiple instances.
```suggestion
        async Task<PurgeHistoryResult> IDurableOrchestrationClient.PurgeInstanceHistoryAsync(IEnumerable<string> instanceId)
```

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -484,6 +484,40 @@ async Task<DurableOrchestrationStatus> IDurableOrchestrationClient.GetStatusAsyn
             return await GetDurableOrchestrationStatusAsync(state, this.client, showHistory, showHistoryOutput);
         }
 
+        /// <inheritdoc />
+        async Task<List<DurableOrchestrationStatus>> IDurableOrchestrationClient.GetStatusAsync(IEnumerable<string> instanceID, bool showHistory, bool showHistoryOutput, bool showInput)
```
davidmrdavid, line 488: Since the input here is an array, I think we should rename `IEnumerable<string> instanceID` to `IEnumerable<string> instanceIDs` instead, to denote its a collection of values :)

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -484,6 +484,40 @@ async Task<DurableOrchestrationStatus> IDurableOrchestrationClient.GetStatusAsyn
...
+                {
+                    currenStatus = null;
+                }
+                else
+                {
+                    currenStatus = await GetDurableOrchestrationStatusAsync(state, this.client, showHistory, showHistoryOutput);
+                }
+
+                results.Add(currenStatus);
+            }
```
davidmrdavid, line 516: Are we able to achieve this same result by just calling `GetStatusAsync` for each instanceID in the array? If so, I think we should do that instead.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -563,6 +597,27 @@ async Task<PurgeHistoryResult> IDurableOrchestrationClient.PurgeInstanceHistoryA
             return await this.DurabilityProvider.PurgeInstanceHistoryByInstanceId(instanceId);
         }
 
+        async Task<PurgeHistoryResult> IDurableOrchestrationClient.PurgeInstanceHistoryAsync(IEnumerable<string> instanceId)
```
davidmrdavid, line 600: Same feedback as above: can we change `instanceId` to `InstanceIDs`?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -563,6 +597,27 @@ async Task<PurgeHistoryResult> IDurableOrchestrationClient.PurgeInstanceHistoryA
...
+                {
+                    var purgeHistoryResult = await this.DurabilityProvider.PurgeInstanceHistoryByInstanceId(id);
+                    if (purgeHistoryResult.InstancesDeleted == 1)
+                    {
+                        instanceDeleted += 1;
+                    }
+                }));
+            }
+
+            await Task.WhenAll(tasks);
```
davidmrdavid, line 616: Hmm, the combination of `Task.Run` and `Task.WhenAll` in this block of code doesn't quite seem right to me.
My understanding is that, when you call `Task.Run`, you're asking for that `Task` to execute immediately.

However, that is also achieved in the `Task.WhenAll`: that also schedules all input tasks.

So that's all to say that I think this code could be simpler :)

Can you try the following:

When you create the list of `tasks`, create it as such:

```C#
tasks.add(this.DurabilityProvider.PurgeInstanceHistoryByInstanceId(id);
```

Then you'll have a list of Tasks that each returns a `PurgeHistoryResult`. From that list, you should be able to calculate the number of instances deleted :) . Let me know if you need with this.

davidmrdavid (commented):
Left some thoughts! I think it's almost there

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
amdeel, line 616: There's a much simpler way to do this. You can use a parallel for each loop instead of manually using tasks.

Also, make sure the int for instancesDeleted is incremented in a way that is safe for multiple threads to write to it.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -484,6 +484,19 @@ async Task<DurableOrchestrationStatus> IDurableOrchestrationClient.GetStatusAsyn
...
+        {
+            var results = new List<DurableOrchestrationStatus>();
+            var tasks = instanceIds.Select(async instanceId =>
+            {
+                var status = await ((IDurableOrchestrationClient)this).GetStatusAsync(instanceId);
+                results.Add(status);
+            });
+            await Task.WhenAll(tasks);
+            return results;
+            }
```
davidmrdavid, line 498: I think this code can be simplified with some C# magic.
I suggest rewriting it like this:

```suggestion
        async Task<IList<DurableOrchestrationStatus>> IDurableOrchestrationClient.GetStatusAsync(IEnumerable<string> instanceIds, bool showHistory, bool showHistoryOutput, bool showInput)
        {
            var tasks = instanceIds.Select(instanceId => ((IDurableOrchestrationClient)this).GetStatusAsync(instanceId));
            var results = await Task.WhenAll(tasks);
            return results;
        }
```

Note that, for this to work, you'll need to change the definition of `GetStatusAsync` in `IDurableOrchestrationClient.cs` to return an `IList` instead of a `List`.

Let me know what you think!

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -563,6 +576,22 @@ async Task<PurgeHistoryResult> IDurableOrchestrationClient.PurgeInstanceHistoryA
...
+            {
+                var purgeHistoryResult = await this.DurabilityProvider.PurgeInstanceHistoryByInstanceId(instanceId);
+                if (purgeHistoryResult.InstancesDeleted == 1)
+                {
+                    instanceDeleted += 1;
+                }
+            });
+            await Task.WhenAll(tasks);
+            PurgeHistoryResult result = new PurgeHistoryResult(instancesDeleted: instanceDeleted);
+            return result;
```
davidmrdavid, line 592: As with my previous comment, I recommend this rewrite:

```suggestion
            var tasks = instanceIds.Select(instanceId => this.DurabilityProvider.PurgeInstanceHistoryByInstanceId(instanceId));
            var results = await Task.WhenAll(tasks);

            var instancesDeleted = 0;
            foreach (var purgeHistoryResult in results)
            {
                if (purgeHistoryResult.InstancesDeleted == 1)
                {
                    instancesDeleted += 1;
                }
            }

            PurgeHistoryResult result = new PurgeHistoryResult(instancesDeleted: instancesDeleted);
            return result;
```

There's 2 main changes here:
(1) We're constructing the list of tasks directly, without the need to declare local async functions
(2) We're not modifying the shared `instancesDeleted` variable inside the Tasks, which is preferrable for data integrity in multi-threaded code :)

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
```diff
@@ -261,6 +261,16 @@ public interface IDurableOrchestrationClient
...
 
+        /// <summary>
+        /// Get the status of multiple instances.
+        /// </summary>
+        /// <param name="instanceIds"> List of multiple instance IDs that needed to check that status.</param>
+        /// <param name="showHistory">Boolean marker for including execution history in the response.</param>
+        /// <param name="showHistoryOutput">Boolean marker for including input and output in the execution history response.</param>
+        /// <param name="showInput">If set, fetch and return the input for the orchestration instance.</param>
+        /// <returns>Returns a list of orchestration status values.</returns>
+        Task<List<DurableOrchestrationStatus>> GetStatusAsync(IEnumerable<string> instanceIds, bool showHistory = false, bool showHistoryOutput = false, bool showInput = false);
```
davidmrdavid, line 272: In case you decide to take my previous suggestion on `DurableClient.cs`, then you'll need to change the return value here to be `IList` instead of `List`.

```suggestion
        Task<IList<DurableOrchestrationStatus>> GetStatusAsync(IEnumerable<string> instanceIds, bool showHistory = false, bool showHistoryOutput = false, bool showInput = false);
```

davidmrdavid (commented):
I think it's almost there!

In general, I think the construction of local `Task`s can be simplified. I've left some code change suggestions that exemplify how to do this.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
```diff
@@ -261,6 +261,16 @@ public interface IDurableOrchestrationClient
...
         Task<DurableOrchestrationStatus> GetStatusAsync(string instanceId, bool showHistory = false, bool showHistoryOutput = false, bool showInput = true);
 
+        /// <summary>
+        /// Get the status of multiple instances.
+        /// </summary>
+        /// <param name="instanceIds"> List of multiple instance IDs that needed to check that status.</param>
+        /// <param name="showHistory">Boolean marker for including execution history in the response.</param>
+        /// <param name="showHistoryOutput">Boolean marker for including input and output in the execution history response.</param>
+        /// <param name="showInput">If set, fetch and return the input for the orchestration instance.</param>
+        /// <returns>Returns a list of orchestration status values.</returns>
```
davidmrdavid, line 271: spelling suggestions:
```suggestion
        /// <summary>
        /// Get the status of multiple instances.
        /// </summary>
        /// <param name="instanceIds">The instanceIDs to query.</param>
        /// <param name="showHistory">Whether to include the instance history in the response.</param>
        /// <param name="showHistoryOutput">Whether to include the history output in the response.</param>
        /// <param name="showInput">Whether to include the input of the orchestration instance.</param>
        /// <returns>Returns a list of orchestration statuses.</returns>
```

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
```diff
@@ -279,6 +289,13 @@ public interface IDurableOrchestrationClient
         /// <returns>Returns an instance of <see cref="PurgeHistoryResult"/>.</returns>
         Task<PurgeHistoryResult> PurgeInstanceHistoryAsync(string instanceId);
 
+        /// <summary>
+        /// Purge the history for multiple instances.
+        /// </summary>
+        /// <param name="instanceIds">The IDs of the orchestration instance to purge.</param>
+        /// <returns>>Returns an instance of <see cref="PurgeHistoryResult"/>.</returns>
+        Task<PurgeHistoryResult> PurgeInstanceHistoryAsync(IEnumerable<string> instanceIds);
+
```
davidmrdavid, line 298: I suggest removing the extra `>` in the `returns` tag. Also a small wording suggestion.
```suggestion
        /// <summary>
        /// Purge the histories of multiple instances.
        /// </summary>
        /// <param name="instanceIds">The instanceIDs whose histories to purge.</param>
        /// <returns>A <see cref="PurgeHistoryResult"/> for the aggregate purge operation.</returns>
        Task<PurgeHistoryResult> PurgeInstanceHistoryAsync(IEnumerable<string> instanceIds);

```

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -4431,6 +4431,69 @@ public async Task Purge_Single_Instance_History(bool extendedSessions, string st
...
+                var client1 = await host.StartOrchestratorAsync(nameof(TestOrchestrations.SayHelloWithActivity), "foo", this.output);
+                var client2 = await host.StartOrchestratorAsync(nameof(TestOrchestrations.SayHelloWithActivity), "bar", this.output);
+                var client3 = await host.StartOrchestratorAsync(nameof(TestOrchestrations.SayHelloWithActivity), "baz", this.output);
+
+                string firstInstanceId = client1.InstanceId;
+                string secondInstanceId = client2.InstanceId;
+                string thirdInstanceId = client3.InstanceId;
+                var instanceIdList = new List<string> { firstInstanceId, secondInstanceId, thirdInstanceId };
+
+                await client1.InnerClient.PurgeInstanceHistoryAsync(instanceIdList);
```
davidmrdavid, line 4482: Should we add a call to `WaitForCompletionAsync` to ensure the orchestrations actually terminate prior to purging their history? I worry that otherwise we risk introducing a flaky test. This applies to both tests.

davidmrdavid (commented):
Almost there! Just some minor questions and suggestions.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -484,6 +484,40 @@ async Task<DurableOrchestrationStatus> IDurableOrchestrationClient.GetStatusAsyn
...
+                {
+                    currenStatus = null;
+                }
+                else
+                {
+                    currenStatus = await GetDurableOrchestrationStatusAsync(state, this.client, showHistory, showHistoryOutput, showInput);
+                }
+
+                results.Add(currenStatus);
+            }
```
davidmrdavid, line 516: The code here seems to be a duplicate of the code within `GetStatusAsync(string instanceId, bool showHistory, bool showHistoryOutput, bool showInput)`. Could we replace most of this with just a call to that other method?

Put differently, I think this code would be a bit simpler if we just had a for-loop that calls onto the `GetStatusAsync` that takes 1 instanceId at a time. Is there any reason not to do that?

davidmrdavid (commented):
Left some new feedback. I think we're close though, I'd just like to see a bit more code re-use.

Also, please see [this](https://github.com/Azure/azure-functions-durable-extension/pull/2269/files#r978895366) comment which I don't think has been addressed yet.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
```diff
@@ -279,6 +289,13 @@ public interface IDurableOrchestrationClient
         /// <returns>Returns an instance of <see cref="PurgeHistoryResult"/>.</returns>
         Task<PurgeHistoryResult> PurgeInstanceHistoryAsync(string instanceId);
 
+        /// <summary>
+        /// Purge the history for multiple instances.
+        /// </summary>
+        /// <param name="instanceIds">The IDs of the orchestration instance to purge.</param>
```
davidmrdavid, line 295: Minor typo: change `instance` to `instances`
```suggestion
        /// <param name="instanceIds">The IDs of the orchestration instances to purge.</param>
```

davidmrdavid (commented):
One final comment before my approval.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -563,6 +571,20 @@ async Task<PurgeHistoryResult> IDurableOrchestrationClient.PurgeInstanceHistoryA
...
+        async Task<PurgeHistoryResult> IDurableOrchestrationClient.PurgeInstanceHistoryAsync(IEnumerable<string> instanceIds)
+        {
+            var tasks = instanceIds.Select(instanceId => this.DurabilityProvider.PurgeInstanceHistoryByInstanceId(instanceId));
+            var results = await Task.WhenAll(tasks);
+            var instancesDeleted = 0;
+            foreach (var purgeHistoryResult in results)
+            {
+                    instancesDeleted += purgeHistoryResult.InstancesDeleted;
+            }
+
```
davidmrdavid, line 583: nitpick: seems this line was overly-indented.

```suggestion
            var instancesDeleted = 0;
            foreach (var purgeHistoryResult in results)
            {
                instancesDeleted += purgeHistoryResult.InstancesDeleted;
            }

```

nytiannn: Thanks for reminding, I didint notice that. I will update it!

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
```diff
@@ -261,6 +261,16 @@ public interface IDurableOrchestrationClient
...
         Task<DurableOrchestrationStatus> GetStatusAsync(string instanceId, bool showHistory = false, bool showHistoryOutput = false, bool showInput = true);
 
+        /// <summary>
+        /// Get the status of multiple instances.
+        /// </summary>
+        /// <param name="instanceIds"> The instanceIDs to query.</param>
+        /// <param name="showHistory">Boolean marker for including execution history in the response.</param>
+        /// <param name="showHistoryOutput">Boolean marker for including input and output in the execution history response.</param>
+        /// <param name="showInput">If set, fetch and return the input for the orchestration instance. If both <c>showHistory</c> and <see cref = "DurableTaskOptions.StoreInputsInOrchestrationHistory" /> are also set to<c>true</c>, then the inputs for activity and sub-orchestration events in the orchestration history will also be returned.</param>
+        /// <returns>Returns a list of orchestration statuses.</returns>
```
cgillum, line 271: We should be more descriptive about the order and contents of this list, because there are many different ways the implementation could choose to return the results and we want to make sure customers don't make wrong assumptions. Here is my suggestion:
```suggestion
        /// <returns>
        /// Returns a list of orchestration statuses. The length and order of the returned list will match the length and 
        /// order of <paramref name="instanceIds"/>. If any instance ID doesn't exist, a <c>null</c> value will be set in
        /// the corresponding list element.
        /// </returns>
```

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -4431,6 +4431,73 @@ public async Task Purge_Single_Instance_History(bool extendedSessions, string st
...
+                await client1.WaitForCompletionAsync(this.output);
+                var client2 = await host.StartOrchestratorAsync(nameof(TestOrchestrations.SayHelloWithActivity), "bar", this.output);
+                await client2.WaitForCompletionAsync(this.output);
+                var client3 = await host.StartOrchestratorAsync(nameof(TestOrchestrations.SayHelloWithActivity), "baz", this.output);
+                await client3.WaitForCompletionAsync(this.output);
+
+                string firstInstanceId = client1.InstanceId;
+                string secondInstanceId = client2.InstanceId;
+                string thirdInstanceId = client3.InstanceId;
+                var instanceIdList = new List<string> { firstInstanceId, secondInstanceId, thirdInstanceId };
```
cgillum, line 4484: To improve this test further, I suggest including an instance ID that doesn't exist, just like you did for the `GetStatusAsync_MultipleInstances` test. Alternatively, you could call `client1.InnerClient.PurgeInstanceHistoryAsync(instanceIdList);` a second time and validate that it succeeds and returns a result with zero deleted instances.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -4431,6 +4431,73 @@ public async Task Purge_Single_Instance_History(bool extendedSessions, string st
...
+
+                await client1.InnerClient.PurgeInstanceHistoryAsync(instanceIdList);
```
cgillum, line 4486: We should test the return value of `PurgeInstanceHistoryAsync` and confirm that the `PurgeHistoryResult.InstancesDeleted` property returns `3`.

cgillum (commented):
Just a few small things to fix but otherwise I think this PR looks good.

FYI @jviau, we'll want to think about parity for the .NET Isolated SDK for instance query and purging. Rather than having so many overloads, however, we should probably have a simple overload that takes an options parameter that we can add different conditions on.

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
```diff
@@ -261,6 +261,16 @@ public interface IDurableOrchestrationClient
...
+        Task<IList<DurableOrchestrationStatus>> GetStatusAsync(IEnumerable<string> instanceIds, bool showHistory = false, bool showHistoryOutput = false, bool showInput = false);
```
jviau, line 272: Can we use `IAsyncEnumerable` for the return value?

jviau, line 272: For methods that have a large list of individual "options" parameters, I like to wrap them into a record class. This is beneficial for a couple reasons:

1. You can add more options without making a signature change (which btw, is a breaking change on interfaces!)
2. Having a class makes it easy to store/re-use their query.
3. `record`'s with their immutability and `with` parameter are perfect for this scenario, as you can transform a shared query without impacting other uses of it.

To summarize, consider something like:

``` CSharp
IAsyncEnumerable<DurableOrchestrationStatus> GetStatusAsync(IEnumerable<string> instanceIds, GetStatusOptions? options = null);

public record GetStatusOptions(bool ShowHistory = false, bool ShowHistoryOutput = false, ShowInput = false);
```

jviau (commented):
@cgillum I agree on the options class instead of many overloads. I added that as a review comment in case we want to do that change here.

cgillum (approved):
LGTM!

@jviau for the purposes of this PR, I suggest we accept it as-is. It's already consistent with the existing APIs. I suggest we instead focus on the API improvements (options, IAsyncEnumerable, etc.) in the new .NET SDK that you're working on since that will eventually replace this in-proc SDK anyways. Let me know if that works for you.

jviau:
Yes, that works for me! I think I am already doing the options pattern there I suggested.

PR #2272, opened 2022-09-15
Increase version for 2.8.1 release

davidmrdavid:
As in title, routine PR


cgillum:
Good to see all the smoke tests are passing. I think that means we can safely check off those validations in the release checklist.

cgillum (approved):

PR #2280, opened 2022-09-28
Fix issue #2261: add parameter "eventTypes" for GetStatusAsync

### LINKED ISSUE #2261: Add API option to filter instance status history by eventType
On the [Get Instance Status API](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-http-api#get-instance-status) there's a `showHistory` parameter which returns a `historyEvents` array for the instance.  It would be good to be able to filter this to only specific event types.

N/A

Add a `$filter` query parameter or an explicit `includeEventType` type of parameter to enable filtering of the `historyEvents` result to only specific types of events.

Client-side filtering (not optimal).

**Additional context**
API Documentation: https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-http-api#get-instance-status

Current example response:

```json
{
  "createdTime": "2018-02-28T05:18:49Z",
  "historyEvents": [
      {
          "EventType": "ExecutionStarted",
          "FunctionName": "E1_HelloSequence",
          "Timestamp": "2018-02-28T05:18:49.3452372Z"
      },
      {
          "EventType": "TaskCompleted",
          "FunctionName": "E1_SayHello",
          "Result": "Hello Tokyo!",
          "ScheduledTime": "2018-02-28T05:18:51.3939873Z",
          "Timestamp": "2018-02-28T05:18:52.2895622Z"
      },
      {
          "EventType": "TaskCompleted",
          "FunctionName": "E1_SayHello",
          "Result": "Hello Seattle!",
          "ScheduledTime": "2018-02-28T05:18:52.8755705Z",
          "Timestamp": "2018-02-28T05:18:53.1765771Z"
      },
      {
          "EventType": "TaskCompleted",
          "FunctionName": "E1_SayHello",
          "Result": "Hello London!",
          "ScheduledTime": "2018-02-28T05:18:53.5170791Z",
          "Timestamp": "2018-02-28T05:18:53.891081Z"
      },
      {
          "EventType": "ExecutionCompleted",
          "OrchestrationStatus": "Completed",
          "Result": [
              "Hello Tokyo!",
              "Hello Seattle!",
              "Hello London!"
          ],
          "Timestamp": "2018-02-28T05:18:54.3660895Z"
      }
  ],
  "input": null,
  "customStatus": { "nextActions": ["A", "B", "C"], "foo": 2 },
  "lastUpdatedTime": "2018-02-28T05:18:54Z",
  "output": [
      "Hello Tokyo!",
      "Hello Seattle!",
      "Hello London!"
  ],
  "runtimeStatus": "Completed"
}
```

nytiannn:
Fixes #2261

- Add parameter **eventTypes** for API GetStatusAsync at file DurableClient.cs

- If eventTypes is set, the history returned will only include the event types that eventTypes provided. 

- If eventTypes is not set or null, the history returned is still the same as before, including ExecutionStarted,  TaskCompleted, ExecutionCompleted.


src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
```diff
@@ -258,8 +258,9 @@ public interface IDurableOrchestrationClient
         /// <param name="showHistory">Boolean marker for including execution history in the response.</param>
         /// <param name="showHistoryOutput">Boolean marker for including input and output in the execution history response.</param>
         /// <param name="showInput">If set, fetch and return the input for the orchestration instance.</param>
+        /// <param name="filter">If set, return the history with the event types filter included.</param>
         /// <returns>Returns a task which completes when the status has been fetched.</returns>
-        Task<DurableOrchestrationStatus> GetStatusAsync(string instanceId, bool showHistory = false, bool showHistoryOutput = false, bool showInput = true);
+        Task<DurableOrchestrationStatus> GetStatusAsync(string instanceId, bool showHistory = false, bool showHistoryOutput = false, bool showInput = true, IEnumerable<string> filter = null);
```
amdeel, line 263: Let's name this new field something a little more specific, like eventTypes. Also, instead of string we should make it an EventType collection.

amdeel (commented):
This is a good start. You'll need to edit HttpApiHandler as well to take this new input.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -910,12 +910,27 @@ private static async Task<DurableOrchestrationStatus> GetDurableOrchestrationSta
...
+                        for (int i = historyArray.Count - 1; i >= 0; i--)
+                        {
+                            var eventTypeName = historyArray[i].Value<string>("EventType");
+                            var eventType = (EventType)Enum.Parse(typeof(EventType), eventTypeName, true);
+                            if (!eventTypes.Contains(eventType))
+                            {
+                                historyArray.Remove(historyArray[i]);
+                            }
+                        }
+                    }
```
davidmrdavid, line 924: Just to be sure I follow: why are we iterating the array backwards? Is it for performance reasons?

davidmrdavid: Whatever the case, I think it would be good to document this in a comment :-)

test/Common/HttpApiHandlerTests.cs
```diff
@@ -474,7 +474,7 @@ public async Task HandleGetStatusRequestAsync_Failed_Orchestration_Config_Respon
             var instanceId = Guid.NewGuid().ToString();
             var clientMock = new Mock<IDurableClient>();
             clientMock
-                .Setup(x => x.GetStatusAsync(instanceId, false, false, true))
+                .Setup(x => x.GetStatusAsync(instanceId, false, false, true, null))
                 .Returns(Task.FromResult(list.First()));
             var httpApiHandler = new ExtendedHttpApiHandler(clientMock.Object);
```
davidmrdavid, line 480: Can we get a test that sets the `eventTypes` argument to a non-null value? It would be good to test it :)

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -910,12 +910,27 @@ private static async Task<DurableOrchestrationStatus> GetDurableOrchestrationSta
...
-                    indexList.Sort();
-                    foreach (var indexValue in indexList)
+                    if (eventTypes != null)
                     {
-                        historyArray.RemoveAt(indexValue - counter);
-                        counter++;
+                        for (int i = historyArray.Count - 1; i >= 0; i--)
+                        {
+                            var eventTypeName = historyArray[i].Value<string>("EventType");
+                            var eventType = (EventType)Enum.Parse(typeof(EventType), eventTypeName, true);
```
davidmrdavid, line 918: The `.Parse` operation may fail, and I think it would be good to handle that failure. As a result, instead of calling `.Parse` directly, I recommend calling [`.TryParse`](https://learn.microsoft.com/en-us/dotnet/api/system.enum.tryparse?view=net-7.0) which is a bit safer and returns `true` if the parsing succeeded, and `false` if it failed. You can use that result to log a warning and abort the operation if the parsing fails.

davidmrdavid (commented):
Almost there! Some thoughts and questions

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
nytiannn: Thanks, David! The historyArray includes all kind of eventTypes. And here we only want to only leave the ones that callers ask for. So, this for loop is for this edit operation: to remove the ones that callers don't need. Since the remove and traverse operation happens at the same time, I iterate the array backwards so that the item number won't be changed due to these operations. :)

davidmrdavid: That makes sense, thank you! Feel free to disregard my comment here then.

test/Common/HttpApiHandlerTests.cs
davidmrdavid: nevermind, I see this overlaps with the E2E test. no need for a unit test then

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -3008,6 +3009,32 @@ public async Task GetStatus_ShowInputFalse(string storageProvider)
             }
         }
 
+        [Fact]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        public async Task GetStatus_testeventTypes()
```
bachuv, line 3014: Can you update the test name `GetStatus_TestEventTypes` to match the existing naming convention?

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -3008,6 +3009,32 @@ public async Task GetStatus_ShowInputFalse(string storageProvider)
...
+                await client.WaitForCompletionAsync(this.output);
+                var eventList = new List<EventType> { EventType.TaskScheduled, EventType.TaskCompleted };
+                var status = await client.InnerClient.GetStatusAsync(client.InstanceId, showHistory: true, showHistoryOutput: false, showInput: false, eventList);
+
+                var eventTypeName1 = status.History[0].Value<string>("EventType");
+                var eventType1 = (EventType)Enum.Parse(typeof(EventType), eventTypeName1, true);
+                var eventTypeName2 = status.History[1].Value<string>("EventType");
+                var eventType2 = (EventType)Enum.Parse(typeof(EventType), eventTypeName2, true);
+                Assert.Equal(EventType.TaskScheduled, eventType1);
+                Assert.Equal(EventType.TaskCompleted, eventType2);
```
bachuv, line 3032: Should we add another Assert statement to check that we only received `TaskScheduled` and `TaskCompleted` events? Maybe check the length of `status.History`?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -910,12 +910,29 @@ private static async Task<DurableOrchestrationStatus> GetDurableOrchestrationSta
...
+                    }
+                    else
+                    {
+                        var counter = 0;
+                        indexList.Sort();
+                        foreach (var indexValue in indexList)
+                        {
+                            historyArray.RemoveAt(indexValue - counter);
+                            counter++;
+                        }
```
davidmrdavid, line 935: Can you please help me understand what this code does? it seems this would always execute before, so I'm trying to understand why we're not always running this logic anymore

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
```diff
@@ -258,8 +259,9 @@ public interface IDurableOrchestrationClient
         /// <param name="showHistory">Boolean marker for including execution history in the response.</param>
         /// <param name="showHistoryOutput">Boolean marker for including output in the execution history response.</param>
         /// <param name="showInput">If set, fetch and return the input for the orchestration instance. If both <c>showHistory</c> and <see cref = "DurableTaskOptions.StoreInputsInOrchestrationHistory" /> are also set to<c>true</c>, then the inputs for activity and sub-orchestration events in the orchestration history will also be returned.</param>
+        /// <param name="eventTypes">If set, return the history with the event types filter included.</param>
```
davidmrdavid, line 262: I would like try a more "functional" description. What do you think of the following description for `eventTypes`?:

If provided, and `showHistory` is set, the history is filtered to contain only the types of events specified in this parameter.

davidmrdavid (commented):
Two more questions before approval!

cgillum:
I recognize that I'm chiming in late, but if I'm being completely honest, I would prefer we not take this change. It adds new API surface area that I feel doesn't solve a problem that many customers will have. New APIs have to be documented, supported, and maintained, so there's always a cost to adding new things. I want to make sure the value we get from APIs is worth the cost. Querying instance history is primarily intended as a primitive, low-cost debugging tool and not something I think we should invest in more than we already have. Rather, we should push customers to using more official diagnostic tools for debugging.

I have a few technical gripes as well. For example, the `EventType` enum is defined in DurableTask.Core and we're trying hard to avoid leaking DurableTask.Core types and abstractions into the Durable Functions API layer. We also don't document these types and there are several types there which don't map to history events returned by the query APIs. Not only that, but we don't even document the schema of the JSON results, nor the behavior of how we combine/remove history events for this like TaskScheduled/TaskCompleted pairs, which means that customers don't have any guidance about how to use this new API effectively. All of this creates new confusion for customers that try to use it, and I don't it's worth spending time trying to fix that confusion.

I won't push back if the rest of the team insists we add this filtering feature, but I personally don't feel good about adding it.

davidmrdavid:
@cgillum: I think the reasons for your concern are well justified. I personally did not consider that some of the event types in DTFx did not map 1:1 to history event types in DF, and all that this implies in terms of gaps in docs and servicing. There was also an implicit assumption that, since the ticket remained open, that it had the team's implicit agreement as a worthwhile issue to implement

To prevent this kind of misunderstanding in the future, I think we should consider making more timely decisions about closing feature requests that we have concerns about. Perhaps we can use our triage meeting to make a pass over our current feature requests in case other such tickets exist.

Taking a step back, I know that there have been some discussion around moving some DF features into the DTFx layer, at which point I think the History events may be well positioned to match 1:1 across both layers, wouldn't they? I think that would go a long way to help here, but we aren't there now.

Also: I was somehow under the impression that this ticket was related to a requirement for DF Java? Was that not the case? I see the ticket was originally opened in the Java repo, so perhaps that's why I was confused. Can you please confirm that this was not a customer deliverable, @cgillum?

In any case, I don't feel strongly enough not to archive this PR if others agree with Chris' concerns. However, as I'm sure we all agree, because this PR has been opened for several weeks, I would like for us to try and prevent this misunderstanding in the future :-) .

@bachuv/ @amdeel

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
nytiannn: This part is also to edit the historyAarray list. In the original code, this part will delete some eventTypes and only leave several importane ones left to make the historyArray list more readable. Because we have more than 10 eventTypes and the non-edited list is quite long. So, if the caller wants to query other eventTypes, it will go to if condition. Otherwise, it will go to this else part, which is the original one, return only certain eventTypes.

cgillum:
@davidmrdavid I don't think this PR thread is a great place to have meta discussions about planning. It's probably best to do that elsewhere and focus the conversation here on the PR itself.

> I was somehow under the impression that this ticket was related to a requirement for DF Java? Was that not the case? I see the ticket was originally opened in the Java repo, so perhaps that's why I was confused.

There was an ask, but not a requirement, nor a design discussion. I personally would still push back on this feature, or at least significantly reduce the scope, e.g., to just the HTTP API layer. The customer that asked for this does not use .NET, so this PR is doing more than was asked for anyways.

davidmrdavid:
Thanks for clarifying, and you're right on the meta discussion :), thanks.

Just to summarize: I'm convinced by your points. I would just like to get just one more member of the team to chime in before making a decision.

nytiannn:
Thanks for the message. I just want to say I am ok with closing this PR. This new feature can help customers to query the event types they want, since our original code will only return certain types. But we don't know how many customers need this. And Chris' opinion makes many senses. This requires a lot of efforts to maintain. Also, at least to me, it's a little bit strange if this is only implemented in the Http API layer. So, I think it's fine to archive it. I learnt a lot from this PR and that's the most important thing.  And I agree with David's opinion that we should use our triage meeting to avoid this situation. It will be helpful. Thank you all! :)

bachuv:
I'm fine with closing this PR as well based on the discussion. I would like to get @amdeel's confirmation before closing it though just to keep everyone in the loop.

amdeel:
I'm fine with closing this PR and not moving forward with it. I don't feel strongly about this feature. I originally suggested it because a few customers reacted to the issue, we return the EventType in the JSON, and it seemed like an easy thing to implement for some quick value. Chris, you brought up some good points. I didn't initially consider that the type EventType is a DTFx concept. I can also see this leading to additional filter related asks and maintenance, which like you said would be a downside. And because we want customers to use other diagnostic tools I agree that this feature isn't necessary.

PR #2302, opened 2022-11-01
Adding function state change traces to improve orchestration monitoring query

bachuv:
[Creating this as a draft PR for now to make sure all tests pass]

This PR adds new traces to track function state changes. This helps us improve our monitoring query in the portal (Functions -> OrchestratorFunction -> Monitor tab -> App Insights query).

New query to list the status of orchestration instances:
```
traces
| where customDimensions.Category == "Host.Triggers.DurableTask"
| extend functionName = tostring(customDimensions["prop__functionName"])
| extend instanceId = tostring(customDimensions["prop__instanceId"])
| extend state = tostring(customDimensions["prop__functionState"])
| extend isReplay = tobool(tolower(customDimensions["prop__isReplay"]))
| where isReplay != true
| where functionName == <functionName>
| where state != ""
| summarize arg_max(timestamp, *) by instanceId
| project timestamp, instanceId, state
```

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -393,6 +393,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 await this.client.TerminateInstanceAsync(state.OrchestrationInstance, reason);
 
                 this.traceHelper.FunctionTerminated(this.TaskHubName, state.Name, instanceId, reason);
+
+                this.traceHelper.OrchestratorStateChange(
+                    this.TaskHubName,
+                    state.Name,
+                    instanceId,
+                    FunctionType.Orchestrator,
+                    FunctionState.Terminated,
+                    isReplay: false);
             }
             else
             {
--- a/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
@@ -962,6 +962,23 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 resource, attempt, delayMs, hubName, LocalAppName, LocalSlotName, ExtensionVersion);
         }
 
+        public void OrchestratorStateChange(
+            string hubName,
+            string functionName,
+            string instanceId,
+            FunctionType functionType,
+            FunctionState functionState,
+            bool isReplay)
+        {
+            if (this.ShouldLogEvent(isReplay))
+            {
+                this.logger.LogInformation(
+                    "{instanceId}: Function '{functionName} ({functionType})' state changed to {functionState}. IsReplay: {isReplay}. State: {state}. FunctionState: {functionState}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
+                    instanceId, functionName, functionType, functionState, isReplay, functionState, functionState, hubName,
+                    LocalAppName, LocalSlotName, ExtensionVersion, this.sequenceNumber++);
+            }
+        }
+
         private bool ShouldLogEvent(bool isReplay)
         {
             return this.traceReplayEvents || !isReplay;
--- a/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
@@ -170,6 +170,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     this.context.Name,
                     this.context.InstanceId,
                     isReplay: false));
+
+                this.Config.TraceHelper.OrchestratorStateChange(
+                    this.context.HubName,
+                    this.context.Name,
+                    this.context.InstanceId,
+                    FunctionType.Orchestrator,
+                    FunctionState.Failed,
+                    isReplay: false);
             }
         }
 
--- a/test/Common/TestHelpers.cs
+++ b/test/Common/TestHelpers.cs
@@ -598,7 +598,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
             {
                 $"{instanceId}: Function '{functionNames[0]} ({FunctionType.Orchestrator})' scheduled. Reason: NewInstance. IsReplay: False.",
                 $"{instanceId}: Function '{functionNames[0]} ({FunctionType.Orchestrator})' started. IsReplay: False. Input: \"World\"",
+                $"{instanceId}: Function '{functionNames[0]} ({FunctionType.Orchestrator})' state changed to Started. IsReplay: False. FunctionState: Started.",
                 $"{instanceId}: Function '{functionNames[0]} ({FunctionType.Orchestrator})' completed. ContinuedAsNew: False. IsReplay: False. Output: \"Hello, World!\"",
+                $"{instanceId}: Function '{functionNames[0]} ({FunctionType.Orchestrator})' state changed to Completed. IsReplay: False. FunctionState: Completed.",
             };
 
             return list;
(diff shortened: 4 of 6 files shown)


src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -170,6 +170,14 @@ internal void Rehydrate(string serializedInput)
                     this.context.Name,
                     this.context.InstanceId,
                     isReplay: false));
+
+                this.Config.TraceHelper.OrchestratorStateChange(
+                    this.context.HubName,
+                    this.context.Name,
+                    this.context.InstanceId,
+                    FunctionType.Orchestrator,
+                    FunctionState.Failed,
```
cgillum, line 179: Why is this considered a failure? I'm not super familiar with the entity code, but line 168 gives the impression that the entity is actually being *created*.

davidmrdavid: Relatedly - should this have `FunctionType.Entity` as a parameter instead?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -393,6 +393,14 @@ async Task IDurableOrchestrationClient.TerminateAsync(string instanceId, string
...
                 this.traceHelper.FunctionTerminated(this.TaskHubName, state.Name, instanceId, reason);
+
+                this.traceHelper.OrchestratorStateChange(
+                    this.TaskHubName,
+                    state.Name,
+                    instanceId,
+                    FunctionType.Orchestrator,
+                    FunctionState.Terminated,
+                    isReplay: false);
             }
```
davidmrdavid, line 404: Have we tested, on Azure, that terminating an orchestrator mid-execution won't lead it to accidentally emit a "Terminated" state-change log **and then** a "Completed" state-change log?

This was happening to us when when the following events took place:
1. orchestrator starts and schedules a 1 minute timer
2. before the 1 minute expires, the orchestrator is terminated. Therefore, the terminated state-change log is emitted
3. the time fires, and the orchestrator is resumed
4. the orchestrator performs a replay as a result of the timer firing. Therefore, it emits a completed state-change log.

I would appreciate if we could perform this test. It would help me gain confidence in the approach :)

davidmrdavid (commented):
Thanks for doing this work! I added a request for a manual test, please see my comment below

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -962,6 +962,23 @@ public void ExtensionWarningEvent(string hubName, string functionName, string in
...
+            FunctionType functionType,
+            FunctionState functionState,
+            bool isReplay)
+        {
+            if (this.ShouldLogEvent(isReplay))
+            {
+                this.logger.LogInformation(
+                    "{instanceId}: Function '{functionName} ({functionType})' state changed to {functionState}. IsReplay: {isReplay}. State: {state}. FunctionState: {functionState}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
+                    instanceId, functionName, functionType, functionState, isReplay, functionState, functionState, hubName,
+                    LocalAppName, LocalSlotName, ExtensionVersion, this.sequenceNumber++);
```
davidmrdavid, line 978: It seems we're logging `functionState` twice, aren't we? Once for `State: ` and another time for `FunctionState: `.

Should we just log it in the `State: ` field?

bachuv:
As an update, I removed the `functionState` property to make it simpler and updated the query in the description based on that change. I'm now looking into removing the additional "orchestrator state changed" log.

bachuv:
I looked into removing the additional state trace log and updating the existing `Activity.Tags` with the correct runtime status but realized that doesn't work for client APIs like Terminate and Suspend. For those APIs, I tried updating the `Tags` in the relevant methods in `DurableClient` but was seeing different values for `Activity.Current` which meant the runtime status value wouldn't get updated correctly. Let me know if there's another way of approaching this that I haven't considered.

bachuv:
I was able to remove the additional state trace log by adding a `RuntimeStatus` field to the existing state trace log instead. I also updated the runtime status query.

```
traces
| where customDimensions.Category == "Host.Triggers.DurableTask"
| extend functionName = tostring(customDimensions["prop__functionName"])
| extend instanceId = tostring(customDimensions["prop__instanceId"])
| extend state = tostring(customDimensions["prop__state"])
| extend isReplay = tobool(tolower(customDimensions["prop__isReplay"]))
| extend hubName = tostring(tolower(customDimensions["prop__hubName"]))
| extend runtimeStatus = tostring((customDimensions["prop__runtimeStatus"]))
| where isReplay != true
| where state != "Awaited"
| where runtimeStatus != ""
| summarize arg_max(timestamp, *) by instanceId
| order by timestamp asc
| project timestamp, instanceId, runtimeStatus
```

image

davidmrdavid:
Seems the FunctionsV2 tests are failing, @bachuv?

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -109,6 +109,7 @@
     <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers" Version="0.5.*" />
     <!-- Build-time dependencies -->
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.*" PrivateAssets="All" />
+    <PackageReference Include="Nunit3TestAdapter" Version="4.5.0" />
```
davidmrdavid, line 112: is this needed? Seems like an accidental addition

test/Common/TestHelpers.cs
```diff
@@ -483,8 +483,8 @@ public static Task DeleteTaskHubResources(string testName, bool enableExtendedSe
             if (testName.Equals("TimerCancellation", StringComparison.OrdinalIgnoreCase) ||
                 testName.Equals("TimerExpiration", StringComparison.OrdinalIgnoreCase))
             {
-                // It is assumed that the 4th log message is a timer message.
-                timeStamp = GetTimerTimestamp(logMessages[3].FormattedMessage);
+                // It is assumed that the 5th log message is a timer message.
+                timeStamp = GetTimerTimestamp(logMessages[4].FormattedMessage);
```
davidmrdavid, line 487: I didn't catch why this suddenly became the 5th log (whereas before in was the 4th). Can you please clarify?

davidmrdavid (commented):
Left 2 small questions

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
bachuv: Nope, that was added on accident. I'll remove it!

test/Common/TestHelpers.cs
bachuv: That was left over from the previous change when I was adding a new log for the runtime status change. For now, please ignore the changes in the test directory. I want to get feedback on the new approach (i.e. adding a new RuntimeStatus field to the existing state change log) before changing all of the tests since it's time consuming. Let me know if it's easier if I revert the changes in the test directory for now.

davidmrdavid: Thanks for the clarification. Now that a reviews have been requested again, do we have some context on why the position of this message changed?

davidmrdavid (commented):
With the exception of the `ResumingOrchestration` log, the code changes look good to me! It is certainly confusing that we log both a `State` and `RuntimeStatus` field. We'll need to tackle that _eventually_. Other than that, the overall change strategy makes sense to me. I think you should be good to update the test code

test/Common/TestHelpers.cs
bachuv, line 1: The test changes are out of date and I'll change them once I get feedback on the new field and query. Please ignore the changes in this file. Thanks!

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -329,9 +329,9 @@ public void ExtensionWarningEvent(string hubName, string functionName, string in
                 IsReplay: false);
 
             this.logger.LogInformation(
-                "{instanceId}: Resuming function '{functionName} ({functionType})'. Reason: {reason}. State: {state}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
-                instanceId, functionName, functionType, reason, FunctionState.Scheduled, hubName, LocalAppName,
-                LocalSlotName, ExtensionVersion, this.sequenceNumber++);
+                "{instanceId}: Resuming function '{functionName} ({functionType})'. Reason: {reason}. State: {state}. RuntimeStatus: {runtimeStatus}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}.",
+                instanceId, functionName, functionType, reason, FunctionState.Scheduled, OrchestrationRuntimeStatus.Pending, hubName,
```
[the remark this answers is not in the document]
cgillum: +1, this should be `Running`

cgillum (changes requested):

bachuv:
@davidmrdavid, I know the Netherite smoke test is expected to fail right now, but is the Python smoke test expected to fail?

davidmrdavid:
@bachuv: it isn't, but the error suggests the problem was a failure in downloading the docker image. I'm re-running the test now to see if it's transient.

cgillum (approved):
LGTM!

Issue #2324, opened 2022-11-23
Version 2.9.0 depends on Microsoft.DurableTask.Sidecar.Protobuf version that doesn't exist


jackpoz:
### Description
Trying to install version 2.9.0 from Visual Studio results in an error about Microsoft.DurableTask.Sidecar.Protobuf 1.0.0 not being found:
![image](https://user-images.githubusercontent.com/1153754/203512363-c8423dc5-eef8-4c45-955f-c6363b540002.png)

Checking https://www.nuget.org/packages/Microsoft.DurableTask.Sidecar.Protobuf , the latest version is 0.3.1
![image](https://user-images.githubusercontent.com/1153754/203512524-7a0d7f3b-877f-41bd-82cb-13c654443ffe.png)

### Expected behavior
It should be possible to install version 2.9.0 from Visual Studio without errors.

### Actual behavior
Version 2.9.0 cannot be installed due to dependency on a version of Microsoft.DurableTask.Sidecar.Protobuf that doesn't exist.

### Relevant source code snippets
No code

### Known workarounds
Fallback to version 2.8.1 .

### App Details

- Durable Functions extension version: 2.9.0
- Azure Functions runtime version: 3.0
- Programming language used: C#

### Screenshots
See screenshots above in the description

cgillum:
Hi @jackpoz @mmajcica. Sorry about that! Please try again now: https://www.nuget.org/packages/Microsoft.DurableTask.Sidecar.Protobuf/1.0.0

mmajcica:
It works now, thanks.

jackpoz:
that was fast, thanks 🚀

PR #2325, opened 2022-11-23
Add note about publishing protobuf package in release template

davidmrdavid:
Added a note to publish the Protobuf package during Extension releases, which is now a dependency. This should reports such as this one: https://github.com/Azure/azure-functions-durable-extension/issues/2324


davidmrdavid:
FYI @cgillum

bachuv (approved):

PR #2327, opened 2022-11-28
Configure dotnet out-of-proc worker options

jviau:
Introduces an `ObjectConverterShim` which bridges the `WorkerOptions.Serializer` (from Azure Functions SDK) with the durable `DataConverter`, aligning serialization between the two.

resolves #issue_for_this_pr

### Pull request checklist

### DIFF
--- a/src/Worker.Extensions.DurableTask/DurableTaskExtensionStartup.cs
+++ b/src/Worker.Extensions.DurableTask/DurableTaskExtensionStartup.cs
@@ -7,8 +7,13 @@ using Microsoft.Azure.Functions.Worker.Extensions.DurableTask;
 using Microsoft.DurableTask;
 using Microsoft.DurableTask.Client;
 using Microsoft.DurableTask.Converters;
+using Microsoft.DurableTask.Worker;
+using Microsoft.DurableTask.Worker.Shims;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
 [assembly: WorkerExtensionStartup(typeof(DurableTaskExtensionStartup))]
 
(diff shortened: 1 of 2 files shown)


jviau:
@cgillum I don't think we actually use `DataConverter` for `TaskActivity` serialization anywhere. Do you know what is used for that? Who handles binding for `TaskActivityAttribute`?

src/Worker.Extensions.DurableTask/DurableTaskExtensionStartup.cs
```diff
@@ -26,13 +32,34 @@ public override void Configure(IFunctionsWorkerApplicationBuilder applicationBui
...
+            return new DurableTaskShimFactory(options, factory); // For GrpcOrchestrationRunner
+        });
+
         applicationBuilder.UseMiddleware<DurableTaskFunctionsMiddleware>();
     }
+
+    private static DataConverter? GetConverter(IServiceProvider services)
+    {
+        WorkerOptions? worker = services.GetRequiredService<IOptions<WorkerOptions>>()?.Value;
+        return worker?.Serializer is ObjectSerializer serializer ? new ObjectConverterShim(serializer) : null;
```
cgillum, line 63: Is the expression `worker?.Serializer is ObjectSerializer serializer` simply a null check, or is there a possibility that some other type besides `ObjectSerializer` could be configured? This is one of the things I don't like about modern C# language features - it's unclear to me what the actual intent is.

cgillum, line 63: It looks like we only consider serializers that derive from `ObjectSerializer` in this code path. Can we/do we want to support users registering a `DataProvider` type directly?

cgillum (commented):
@jviau did you find the answer to your question already? For activities, we aren't doing any serialization/deserialization ourselves. We rely on the functions runtime/worker to figure out type serialization for us.

src/Worker.Extensions.DurableTask/DurableTaskExtensionStartup.cs
jviau: it's a null check here. Wound up like that after some refactors, but I see how a regular `is null` would be better. I'll change it.

jviau: I intentionally left that out for now as providing a `DataConverter` directly is only a partial solution (it does not affect activities in durable functions?). When we solve that gap I am open to supporting direct `DataConverter`'s

jviau: Added a comment explaining that.

cgillum (approved):
:shipit:

PR #2330, opened 2022-11-29
Fix GetInput issues due to object casting

jviau:
This is a sub-optimal solution to https://github.com/microsoft/durabletask-dotnet/issues/36. The problem lies in the fact that input is [already deserialized](https://github.com/microsoft/durabletask-dotnet/blob/main/src/Worker/Core/Shims/TaskOrchestrationShim.cs#L47) based on the input type provided to `GrpcOrchestrationRunner`. We use `object` because we do not know what their actual input type is, which ends up being the default type of whatever serializer we are using (`JsonElement` for example).

This PR introduces a workaround by re-serializing and then deserializing to the desired type. A better solution will be considered later. I think overall `GetInput<T>` should go away entirely, and we should use functions binding to let the input be a parameter.

### Pull request checklist

### DIFF
--- a/src/Worker.Extensions.DurableTask/DurableTaskFunctionsMiddleware.cs
+++ b/src/Worker.Extensions.DurableTask/DurableTaskFunctionsMiddleware.cs
@@ -3,25 +3,14 @@
 
 using System;
 using System.Diagnostics.CodeAnalysis;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.Functions.Worker.Middleware;
-using Microsoft.DurableTask;
-using Microsoft.DurableTask.Worker;
 using Microsoft.DurableTask.Worker.Grpc;
-using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.Functions.Worker.Extensions.DurableTask;
 
 internal class DurableTaskFunctionsMiddleware : IFunctionsWorkerMiddleware
 {
-    // TODO: Need to add call to action, i.e., pointer to some documentation.
-    private const string IllegalAwaitErrorMessage =
-        "An invalid asynchronous invocation was detected. This can be caused by awaiting non-durable tasks " +
-        "in an orchestrator function's implementation or by middleware that invokes asynchronous code.";
-
     public async Task Invoke(FunctionContext functionContext, FunctionExecutionDelegate next)
     {
         if (!IsOrchestrationTrigger(functionContext, out BindingMetadata? triggerMetadata))
@@ -87,7 +76,7 @@ internal class DurableTaskFunctionsMiddleware : IFunctionsWorkerMiddleware
             // If the middleware returns before the orchestrator function's context object was accessed and before
             // it completes its execution, then we know that either some middleware component went async or that the
             // orchestrator function did some illegal await as its very first action.
-            throw new InvalidOperationException(IllegalAwaitErrorMessage);
+            throw new InvalidOperationException(Constants.IllegalAwaitErrorMessage);
         }
 
         await orchestratorTask;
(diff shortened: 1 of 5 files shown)


src/Worker.Extensions.DurableTask/DurableTaskFunctionsMiddleware.cs
```diff
@@ -120,7 +128,24 @@ public FunctionsOrchestrationContext(TaskOrchestrationContext innerContext, Func
...
+            // deserialize again to convert to our desired type.
+            if (this.serializedInput is null)
+            {
+                object? input = this.innerContext.GetInput<object>();
+                if (input is null)
+                {
+                    return default!;
+                }
+
+                this.serializedInput = this.options.DataConverter.Serialize(input);
```
cgillum, line 145: Is there a less hacky way to do this, like converting the deserialized `JsonElement` into the target type?

jviau: Refactored this to include an optimal path for `JsonElement`. I retained the serialization logic because we are not guaranteed it will be `JsonElement`, the result of the `object` deserialization depends on what serializer they have configured.

src/Worker.Extensions.DurableTask/FunctionsOrchestrationContext.cs
```diff
@@ -0,0 +1,152 @@
...
+
+    [Obsolete("Not yet supported")]
+    public override Task<T> CallActivityAsync<T>(
+        Func<object?, T> activityLambda,
+        object? input = null,
+        TaskOptions? options = null)
+    {
+        this.EnsureLegalAccess();
+        return this.innerContext.CallActivityAsync(activityLambda, input, options);
+    }
```
cgillum, line 83: We should probably remove this. I'm not sure why I checked in it (might have been on accident).

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/GreetUser.cs
```diff
@@ -0,0 +1,46 @@
...
+            .ScheduleNewOrchestrationInstanceAsync(nameof(GreetUserOrchestration), payload);
+        logger.LogInformation("Created new orchestration with instance ID = {instanceId}", instanceId);
+
+        return durableContext.CreateCheckStatusResponse(req, instanceId);
+    }
+
+    [Function(nameof(GreetUserOrchestration))]
+    public static async Task<SayHelloPayload> GreetUserOrchestration([OrchestrationTrigger] TaskOrchestrationContext context)
+    {
+        SayHelloPayload? input = context.GetInput<SayHelloPayload>();
```
cgillum, line 32: I think we need to start looking into how to write unit tests for .NET Isolated DF sometime soon.

src/Worker.Extensions.DurableTask/FunctionsOrchestrationContext.InputConverter.cs
```diff
@@ -0,0 +1,67 @@
...
+
+    private class DefaultInputConverter : InputConverter
+    {
+        private readonly DataConverter converter;
+        private readonly string serializedInput;
+
+        public DefaultInputConverter(object input, DataConverter converter)
+        {
+            this.converter = converter;
+            this.serializedInput = converter.Serialize(input);
```
cgillum, line 34: Are we going to be doing this for base types, like `string`, `int`, etc.?  We didn't have to previously, right?

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/GreetUser.cs
jviau: Agreed.

src/Worker.Extensions.DurableTask/FunctionsOrchestrationContext.cs
jviau: This is an `abstract` method, so I need some implementation. Are you proposing this is removed from the base type? If so, that won't make next preview.

src/Worker.Extensions.DurableTask/FunctionsOrchestrationContext.InputConverter.cs
jviau: Previously base types did not work at all either. You would get an error like `error CS0030: Cannot convert type 'System.Text.Json.JsonElement' to 'string'`

However, I have a check in `GetInput<T>` which checks if the base type casts directly to the requested type and I return it directly.

cgillum (approved):

PR #2337, opened 2022-12-14
Allow for orchestration input to be a parameter

jviau:
Allows for functions-style orchestrations to express their input as a parameter instead of using `context.GetInput<T>();`. No attribute is needed, instead we will bind to the first parameter which has exactly 0 binding attributes (in or out). We do not care what the name of the parameter is or its type. This change also has the added benefit of letting us know what the input type is ahead of time for us to include in our `ITaskOrchestrator.InputType`, so that the inner `TaskOrchestrationContext` will deserialize directly to that type.

## Before

``` CSharp
[Function(nameof(MyOrchestration)]
public Task<string> MyOrchestration([OrchestrationTrigger] TaskOrchestrationContext context)
{
    MyInput input = context.GetInput<MyInput>();
}
```

## After

``` CSharp
[Function(nameof(MyOrchestration)]
public Task<string> MyOrchestration([OrchestrationTrigger] TaskOrchestrationContext context, MyInput input)
{
}
```

### Pull request checklist


cgillum:
Without having an attribute, how do we distinguish the orchestration input parameters from other binding parameters, like the `ILogger`, the `CancellationToken`, or other values that might get surfaced by either the worker or the trigger?

jviau:
Those are very good callouts. I will need to do some work for those scenarios. I think I have a lot of testing to do before I check this in.

jviau:
@cgillum I only found 2 types that also "loosely" bind (no attribute or "Source" paired with the context) like this input converter, which are `FunctionContext` and `CancellationToken`. I have added a manual check to exclude these. This is not a perfect solution as anyone can introduce another loose-binding converter like this, but I think it is _good enough_ for now.

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/DirectInput.cs
```diff
@@ -0,0 +1,69 @@
...
+    private const string ActivityName = "DirectInputActivity";
+
+    [Function(nameof(DirectInput))]
+    public static async Task<HttpResponseData> Start(
+        [HttpTrigger(AuthorizationLevel.Anonymous, "post")] HttpRequestData req,
+        [DurableClient] DurableClientContext durableContext,
+        FunctionContext executionContext)
+    {
+        ILogger logger = executionContext.GetLogger(nameof(DirectInput));
+        Input? input = await req.ReadFromJsonAsync<Input>();
```
davidmrdavid, line 30: I think this sample would benefit from only using `Input` to represent Orchestrator inputs.
Here, it also represents the input in the HTTP Trigger, which I think is a bit confusing. Can we consider separating these two notions?

Relatedly and even more nitpicky - I would prefer something like `MyInput` over `Input` as the typename here. I think the `My`- prefix helps communicate that there's nothing special about the type `Input` (i.e the user may choose) which I believe is what we're trying to convey here.

src/Worker.Extensions.DurableTask/OrchestrationInputConverter.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// </summary>
+        /// <param name="input">The input.</param>
+        public void PrepareInput(object? input)
+        {
+            if (this.type is null || input is null || this.context.Items.ContainsKey(OrchestrationInputKey))
+            {
+                return;
+            }
+
+            this.context.Items[OrchestrationInputKey] = input;
```
davidmrdavid, line 112: Can you please explain the purpose of checking  `this.context.Items.ContainsKey(OrchestrationInputKey))` in the if-statement above? Relatedly, I think this is worth documenting in a small comment as well

src/Worker.Extensions.DurableTask/OrchestrationInputConverter.cs
```diff
@@ -0,0 +1,115 @@
...
+        foreach (FunctionParameter parameter in definition.Parameters)
+        {
+            if (!IgnoredInputType(parameter.Type)
+                && !definition.InputBindings.ContainsKey(parameter.Name)
+                && !definition.OutputBindings.ContainsKey(parameter.Name))
+            {
+                // We take the first parameter we encounter which is not an explicitly ignored type
+                // and has exactly 0 bindings (in or out) as our input candidate.
+                return new(parameter.Type, context);
+            }
```
davidmrdavid, line 44: For safety - should we consider not short-circuiting this loop so that we raise an exception if more than 1 candidate is found? I would want us to avoid "silent" assumption violations if possible.

davidmrdavid (commented):
Some minor questions

davidmrdavid:
I'm really excited about this feature, would definitely like us to ship it.

@jviau: I'm not certain on what happens if the user writes down the wrong type in the orchestrator declaration. For example, what happens if their input is really of type `ObjA` but they write `ObjB` in the signature? Do we have a good error message here? If we don't, and considering that we're making a few assumptions in retrieving the input binding here, I think it would be good to have some actionable error messages. For example - we can always tell users that if something went wrong, they can go to the "legacy" model of retrieving the input from the DF context.

jviau:
@davidmrdavid there is no "correct" input type - at least between parameter and `.GetInput<T>()`. In both cases, input is a serialized blob and the type - be it the parameter type, or the generic in `GetInput<T>()` are what influences our deserialization. All we can do is attempt deserialization with whatever input type they have requested.

Additionally, the same issue you described exists with `GetInput<T>` - they could accidentally call `GetInput<ObjB>` instead of `GetInput<ObjA>`.

Due to the deserialization boundary, we do not know which is the "correct" type. Only the user does - it is entirely up to them to supply the correct type in both input modes.

src/Worker.Extensions.DurableTask/OrchestrationInputConverter.cs
jviau: The functions pipeline will handle throwing for us if there are unbound parameters at the end of input binding. I don't think this should throw in this case because sometimes it _could_ be valid as well - there might be another registered input converter that will handle the additional candidates.

Albeit this is all kind of fragile as the behavior here can change depending on what extensions the customer is using and in what order there are registered. Unfortunately, we cannot control for that, but I think this feature is worth it even with that potential flaw.

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/DirectInput.cs
jviau: I changed `Input` to `MyInput`. I did not change the HTTP trigger though, as I believe this is a common scenario to have a payload come in via HTTP and then also serve as the orchestration input. This is just a POCO after all, it can serve many purposes.

src/Worker.Extensions.DurableTask/OrchestrationInputConverter.cs
jviau: Added a comment to clarify that above.

src/Worker.Extensions.DurableTask/OrchestrationInputConverter.cs
cgillum: I agree this is a little fragile. My worry is that the Functions team could introduce a new type of input that we didn't account for and then we'll need to react by adding it to `IgnoredInputType`. Have you considered a design where we use a custom attribute to mark a particular parameter as being an orchestration/activity input?

jviau: A specific parameter is doable for sure, but it is slightly superfluous I find: `public async Task MyOrchestration([OrchestrationTrigger] TaskOrchestrationContext context, [OrchestrationInput] TInput input)`.

This fragility I mentioned is something already part of input binding as a whole and not unique to this scenario. Much like DI, the user can mess with the DI container and cause unexpected behavior. Here the user can also mess with input converters and cause all sorts of unexpected behavior. If users have an issue with this, the old way of `.GetInput<T>` will still be available to them.

cgillum (approved):
I'm good with this PR as it is now. Thanks for the discussion.

PR #2349, opened 2023-01-09
Add typename-preserving serializer for new PS external SDK

davidmrdavid:
The new external PowerShell SDK requires the provided History JSON array to have enough information to reconstruct its .NET types. This is because we will feed the History to initialize the replay logic of the C#-isolated logic, which manages the replay of this new PS SDK.

To that purpose, this PR modifies `OrchestrationTriggerBinding` so that, when the external PS SDK is enabled (controlled by envvar "ExternalDurablePowerShellSDK"), the provided History JSON array includes the necessary fields to de-serialize the events back into their original DurableTask types in the new PS SDK.

N/A

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Bindings/OrchestrationTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/OrchestrationTriggerAttributeBindingProvider.cs
@@ -219,8 +219,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             {
                 var input = arg.GetInputAsJson();
 
-                var usesExternalDurableSDK = this.platormInformation.UsesExternalPowerShellSDK();
-                JsonSerializer serializer = usesExternalDurableSDK ? TypePreservingSerializer : DefaultSerializer;
+                var usesExternalPowerShellSDK = this.platormInformation.UsesExternalPowerShellSDK();
+                JsonSerializer serializer = usesExternalPowerShellSDK ? TypePreservingSerializer : DefaultSerializer;
                 var history = JArray.FromObject(arg.History, serializer);
 
                 // due to Python only supporting up to SchemaVersion V2 from SDK versions 1.1.0 to 1.1.3,
--- a/src/WebJobs.Extensions.DurableTask/Bindings/OrchestrationTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/OrchestrationTriggerAttributeBindingProvider.cs
@@ -22,12 +22,20 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         private readonly string connectionName;
         private static readonly JsonSerializer DefaultSerializer = JsonSerializer.Create();
 
+#pragma warning disable CS0612 // Type or member is obsolete
+        private readonly IPlatformInformation platormInformation;
+#pragma warning restore CS0612 // Type or member is obsolete
+
         public OrchestrationTriggerAttributeBindingProvider(
             DurableTaskExtension config,
-            string connectionName)
+            string connectionName,
+#pragma warning disable CS0612 // Type or member is obsolete
+            IPlatformInformation platormInformationService)
+#pragma warning restore CS0612 // Type or member is obsolete
         {
             this.config = config;
             this.connectionName = connectionName;
+            this.platormInformation = platormInformationService;
         }
 
         public Task<ITriggerBinding?> TryCreateAsync(TriggerBindingProviderContext context)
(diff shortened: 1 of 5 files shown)


davidmrdavid:
FYI @michaelpeng36

cgillum:
I will need some time before I can get to this. One question I'm interested in discussing is whether we want/need to continue carrying the legacy JSON protocol forward for the new PowerShell SDK, or whether we can instead use the newer protobuf protocol.

davidmrdavid:
@cgillum: adding the new protobuf protocol to all the OOProc SDKs is certainly top of my list, but given that this is an implementation detail, **and** that PS users have been waiting a lot time for new features and perf improvements, I would prefer not make this a blocking requirement at this time. But yes, I expect to start onboarding the JS, Python, and PS SDKs onto the new protocol early this year.

src/WebJobs.Extensions.DurableTask/Bindings/OrchestrationTriggerAttributeBindingProvider.cs
```diff
@@ -197,11 +215,14 @@ public Task<IListener> CreateListenerAsync(ListenerFactoryContext context)
...
             }
 
-            private static string OrchestrationContextToString(DurableOrchestrationContext arg)
+            private string OrchestrationContextToString(DurableOrchestrationContext arg)
             {
-                var history = JArray.FromObject(arg.History, DefaultSerializer);
                 var input = arg.GetInputAsJson();
 
+                var usesExternalPowerShellSDK = this.platormInformation.UsesExternalPowerShellSDK();
+                JsonSerializer serializer = usesExternalPowerShellSDK ? TypePreservingSerializer : DefaultSerializer;
```
jviau, line 223: This change does slightly concern me with the longevity of it. The `$type` is Newtonsoft.Json specific (I think it has its own way of writing / reading the type name) - so how robust is this for if we want to change to STJ later on? Even if the change is only on the other end of this (whoever needs the type information).

Also, will this affect any other serialization with powershell? Will admin APIs be affected?

davidmrdavid: Good point on $type being Newtonsoft.JSON specific. I think we don't have to be too concerned about longetivity of this strategy since the JSON-based OOProc protocol implemented in this path should be "soon" deprecated in favor of the protobuf-based protocol that C#-isolated and Java utilize today.

Currently, I plan on GA'ing the new PS SDK with the protobuf implementation (not supported today) and so, as soon as that's enabled, we should be able to deprecate this JSON-based type-preserving serializer.

As for whether this affects any other serialization with PowerShell. I don't think so. This is only serializing the DF payload. All other data going into the PS worker is serialized independently.

So, to summarize - I think of this change as a temporary means of enabling the new PS SDK to be previewed by customers. Then I'll implement the protobuf protocol prior to GA, so we should't have to support this change in the long run. 

Does that make sense and/or seem reasonable?

src/WebJobs.Extensions.DurableTask/IPlatformInformation.cs
```diff
@@ -112,5 +112,11 @@ public interface IPlatformInformation
         /// </summary>
         /// <returns>The application container name.</returns>
         string GetContainerName();
+
+        /// <summary>
+        /// Determines whether the user has opted in to the external PowerShell SDK.
+        /// </summary>
+        /// <returns>True if the user has opted in to the external PowerShell SDK. False otherwise.</returns>
+        bool UsesExternalPowerShellSDK();
```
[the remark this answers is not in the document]
davidmrdavid: We don't have that notion unfortunately.

I also think I should give more context to the `UsesExternalPowerShellSDK` check, that'll clarify things.

Essentially, to opt into the new PS SDK, users need to set an environment variable. This method checks whether that variable was set. Therefore, it is unavoidably very feature-specific and doesn't correspond to a more general "durable engine" version change.

If we're concerned about this type growing too large, I can always read this environment variable directly in the code path where I need it, instead of adding it as a utility method in this class. I don't feel strongly one way or another.

src/WebJobs.Extensions.DurableTask/Bindings/OrchestrationTriggerAttributeBindingProvider.cs
jviau: Ah great, that plan sounds good to me!

src/WebJobs.Extensions.DurableTask/IPlatformInformation.cs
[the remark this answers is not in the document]
jviau: This is fine for now. Maybe down the line we can refactor if this continues to grow.

jviau (approved):
Looks good. If you have not already, can you do some spot testing with other Durable runtimes and validating they are not impacted by this?

davidmrdavid:
Yup, confirmed. Plus, this is an opt-in behavior (needs a special envvar) so users should be protected against it. Merging

PR #2368, opened 2023-01-26
Fix entity deserialization for class-based entities so it reflects custom serialization settings

### LINKED ISSUE #2361: SignalEntityAsync doesn't seem to leverage custom JsonSerializationSetings from custom IMessageSerializerSettingsFactory
### Description
I've got a durable entity defined in class scope that has properties with private setters.  When i call SingnalEntityAsync all these properties are set to default when the actual handle method is called.  If i add a [JsonProperty] to these properties with a private setter, their value is correct.

I tried setting a private resolver contract with
```
public class PrivateResolver : DefaultContractResolver
    {
        protected override JsonProperty CreateProperty(MemberInfo member, MemberSerialization memberSerialization)
        {
            var prop = base.CreateProperty(member, memberSerialization);
            if (!prop.Writable)
            {
                var property = member as PropertyInfo;
                var hasPrivateSetter = property?.GetSetMethod(true) != null;
                prop.Writable = hasPrivateSetter;
            }
            return prop;
        }
    }

    internal class CustomMessageSerializerSettingsFactory : IMessageSerializerSettingsFactory
    {
        public JsonSerializerSettings CreateJsonSerializerSettings()
        {
            return new JsonSerializerSettings()
            {
                ContractResolver = new PrivateResolver(),
                TypeNameHandling = TypeNameHandling.None,
                DateParseHandling = DateParseHandling.None,
            };
        }
    }

 builder.Services.AddSingleton<IMessageSerializerSettingsFactory, CustomMessageSerializerSettingsFactory>();
```

i'm seeing my PrivateResolver::CreateProperty being called with the expected MemberInfos so DI seems configured right, but all SignalEntityAsync's from the DurableClient don't seem to set values for properties with private setters still.

### Expected behavior
Expect properties with private setters to have appropriate value 

### Actual behavior
value is set to default (null)

see above

### Known workarounds
explicitly setting a [JsonProperty] attribute on each property and the value appears

### App Details

- **Durable Functions extension version (e.g. v1.8.3)**: 2.9.1
- **Azure Functions runtime version (1.0 or 2.0)**: 4
- **Programming language used**: c#

### Screenshots
N/A

### If deployed to Azure
N/A

sebastianburckhardt:

resolves #2361 

### Pull request checklist

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -1,1 +1,3 @@
+## Bug fixes
 
+- Fix deserialization of class-based entities to use custom serializer settings (resolves https://github.com/Azure/azure-functions-durable-extension/issues/2361)
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
@@ -302,7 +302,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             {
                 try
                 {
-                    JsonConvert.PopulateObject(this.State.EntityState, result);
+                    JsonConvert.PopulateObject(this.State.EntityState, result, this.messageDataConverter.JsonSettings);
                 }
                 catch (Exception e)
                 {
--- a/test/Common/DurableTaskEndToEndTests.cs
+++ b/test/Common/DurableTaskEndToEndTests.cs
@@ -4299,6 +4299,41 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
         }
 
         /// <summary>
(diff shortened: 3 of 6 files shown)


cgillum (approved):
Quick scan looks goot to me!

cgillum:
The test failure looks interesting. Does this change impact the output of logging?

sebastianburckhardt:
Since the only line that changed is in the code that deserializes class-based entities, I think it should affect only tests that contain such entities. From what I can see, none of the reported test failures contain any entities at all, so I think those test failures are not caused by this change. I am rerunning the tests now to see if the result changes.

cgillum:
Looks like we're green. Merging...

PR #2372, opened 2023-01-27
Remove DurableClientContext, allow for direct binding of DurableTaskClient

### LINKED ISSUE #2240: CreateCheckStatusResponse returns wrong status code
### Description
According to the documentation [here](https://github.com/Azure/azure-functions-durable-extension/blob/d9a254ee4cc8f8cfbfb15eebff3c3f4a38b1e7de/src/Worker.Extensions.DurableTask/DurableClientContext.cs#L42)
```csharp
/// <returns>An HTTP 202 response with a Location header and a payload containing instance control URLs.</returns>
```
and [here](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-http-features?tabs=csharp#async-operation-tracking)
> The target HTTP trigger returns an HTTP 202 response with a Location header that has the value "statusQueryGetUri".

`CreateCheckStatusResponse` should return a 202 (Accepted) status code.

However, the [`DefaultDurableClientContext`](https://github.com/Azure/azure-functions-durable-extension/blob/d9a254ee4cc8f8cfbfb15eebff3c3f4a38b1e7de/src/Worker.Extensions.DurableTask/DefaultDurableClientContext.cs#L52) creates a response with status code 201 (Created):
```csharp
HttpResponseData response = request.CreateResponse(HttpStatusCode.Created);
```

### Expected behavior
The response created by `CreateCheckStatusResponse` should have a status code of 202 (Accepted).

### Actual behavior
The response has a status code of 201 (Created).

### Known workarounds
You can overwrite the status code manually:
```csharp
response.StatusCode = HttpStatusCode.Accepted;
```

### App Details

- **Microsoft.Azure.Functions.Worker**: 1.8.0
- **Microsoft.Azure.Functions.Worker.Extensions.DurableTask**: 0.4.1-beta
- **Azure Functions runtime version**: 4.x

jviau:
This PR removes `DurableClientContext`, instead allowing for direct binding to `DurableTaskClient` via `[DurableClient]` attribute. The HTTP methods on `DurableClientContext` were moved to extension methods for `DurableTaskClient`. The HTTP extension was also modified to (1) use `WorkerOptions.Serializer` and (2) be `202Accepted` by default to match the description.

resolves #2240

### Pull request checklist


src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// <param name="instanceId">The ID of the orchestration instance to check.</param>
+    /// <param name="statusCode">The status code.</param>
+    /// <param name="cancellation">The cancellation token.</param>
+    /// <returns>An HTTP response with a Location header and a payload containing instance control URLs.</returns>
+    public static HttpResponseData CreateCheckStatusResponse(
+        this DurableTaskClient client,
+        HttpRequestData request,
+        string instanceId,
+        HttpStatusCode statusCode,
+        CancellationToken cancellation = default)
```
cgillum, line 49: Seems a little odd to pass a `CancellationToken` to a synchronous method that doesn't do any I/O, no?

src/Worker.Extensions.DurableTask/DurableTaskExtensionStartup.cs
```diff
@@ -54,6 +54,7 @@ public override void Configure(IFunctionsWorkerApplicationBuilder applicationBui
             return new DurableTaskShimFactory(options, factory); // For GrpcOrchestrationRunner
         });
 
+        applicationBuilder.Services.Configure<WorkerOptions>(o => o.InputConverters.RegisterAt<DurableTaskClientConverter>(0));
```
cgillum, line 57: Is this step necessary? I thought the registration of input converters was done via the attribute?

cgillum (commented):
I really like this change.

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
jviau: It does do I/O though! It writes to the request stream. I took it in because the `ObjectSerializer.Serialize` method I use accepts it.

src/Worker.Extensions.DurableTask/DurableTaskExtensionStartup.cs
jviau: Oh I need to remove that attribute haha. It doesn't work. Input binding attributes pointing to their input converter is not supported (yet).

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
cgillum: Hmm, yeah, I guess so. It's weird that it's in synchronous code instead of async code, but I suppose that's because HTTP triggers in Azure Functions buffer the request and response streams? Anyways, doesn't hurt to have this, I guess, and maybe makes it more future-proof.

src/Worker.Extensions.DurableTask/DurableTaskExtensionStartup.cs
cgillum: I see - yeah, I guess if it only works with the target class (which isn't doable in our case), then we should just remove that attribute.

cgillum (approved):
:shipit:

PR #2380, opened 2023-02-07
Document, in PR template, that new EventIds should lie within our supported range

davidmrdavid:
As in title. It also links to a PR example showing how to extend that range.


cgillum (approved):
:shipit:

PR #2390, opened 2023-02-15
Fix bug in entity batch processing that caused entity state size explosion

sebastianburckhardt:
The current implementation of the entity batching contained a bug which caused entities that experienced a high load to stay in a suspended state for a long time (because the resume message was starved).

As a result, we observed situations in which the internal queue in the entity state grows explosively.

This PR
- fixes the suspected reason causing the long delay of the resume message.
- adds a safety measure that un-suspends any entity if it has been suspended for 10 seconds.
- adds very detailed tracing to make it easier to diagnose issues with entity size and entity batch processing in the future.

### Pull request checklist

* [x] Otherwise: Ensure the EventIds are within the supported range in our existing Windows infrastructure.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -983,6 +983,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     entityContext.State.PutBack(lockHolderMessages);
                 }
 
+                // mitigation for ICM358210295 : un-suspend an entity if the last continue-as-new happened at least 10 seconds ago
+                if (entityContext.State.Suspended
+                    && runtimeState.ExecutionStartedEvent?.Timestamp < DateTime.UtcNow - TimeSpan.FromSeconds(10))
+                {
+                    entityContext.State.Suspended = false;
+                    entityShim.AddTraceFlag(EntityTraceFlags.MitigationResumed);
+                }
+
                 if (!entityContext.State.Suspended)
                 {
                     entityShim.AddTraceFlag('2');
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -876,6 +876,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
             try
             {
+                entityShim.AddTraceFlag('1');
+
                 // 1. First time through the history
                 // we count events, add any under-lock op to the batch, and process lock releases
                 foreach (HistoryEvent e in runtimeState.Events)
@@ -967,6 +969,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                                 // this is a continue message.
                                 // Resumes processing of previously queued operations, if any.
                                 entityContext.State.Suspended = false;
+                                entityShim.AddTraceFlag(EntityTraceFlags.Resumed);
                             }
 
                             break;
@@ -981,6 +984,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
                 if (!entityContext.State.Suspended)
                 {
+                    entityShim.AddTraceFlag('2');
+
                     // 2. We add as many requests from the queue to the batch as possible,
                     // stopping at lock requests or when the maximum batch size is reached
                     while (entityContext.State.MayDequeue())
@@ -989,6 +994,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                         {
                             // we have reached the maximum batch size already
                             // insert a delay after this batch to ensure write back
+                            entityShim.AddTraceFlag(EntityTraceFlags.BatchSizeLimit);
                             entityShim.ToBeContinuedWithDelay();
                             break;
                         }
@@ -1009,14 +1015,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             }
             catch (Exception e)
             {
-                entityContext.CaptureInternalError(e);
+                entityContext.CaptureInternalError(e, entityShim);
             }
 
             WrappedFunctionResult result;
 
             if (entityShim.OperationBatch.Count > 0 && !this.HostLifetimeService.OnStopping.IsCancellationRequested)
             {
-                // 3a. Start the functions invocation pipeline (billing, logging, bindings, and timeout tracking).
+                // 3a. (function execution) Start the functions invocation pipeline (billing, logging, bindings, and timeout tracking).
                 result = await FunctionExecutionHelper.ExecuteFunctionInOrchestrationMiddleware(
                     entityShim.GetFunctionInfo().Executor,
                     new TriggeredFunctionData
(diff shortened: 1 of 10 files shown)


src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -876,6 +876,8 @@ private async Task EntityMiddleware(DispatchMiddlewareContext dispatchContext, F
 
             try
             {
+                entityShim.AddTraceFlag('1');
+
```
davidmrdavid, line 880: what does `TraceFlag('1')` do? Can we add a small inline comment clarifying this?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -979,8 +983,18 @@ private async Task EntityMiddleware(DispatchMiddlewareContext dispatchContext, F
                     entityContext.State.PutBack(lockHolderMessages);
                 }
 
+                // mitigation for ICM358210295 : un-suspend an entity if the last continue-as-new happened at least 10 seconds ago
```
davidmrdavid, line 986: nitpick: can we say this more directly? For example: resume entity processing if it has been suspended for at least 10 seconds

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -768,19 +841,19 @@ public void ExtensionWarningEvent(string hubName, string functionName, string in
...
+               functionName,
+               functionState,
+               instanceId,
+               details,
+               (int)statusCode,
+               reason,
+               functionType,
+               ExtensionVersion,
+               IsReplay: false,
+               latencyMs);
```
davidmrdavid, line 856: can we un-do this small whitespace diff?

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -290,9 +331,32 @@ async Task ExecuteOperationsInBatch()
...
+                            {
+                                // we have spent a significant time on this batch already
+                                this.AddTraceFlag(EntityTraceFlags.SignificantTimeElapsed);
+                                return true;
+                            }
+                            else
+                            {
+                                return false;
+                            }
+                        }
```
davidmrdavid, line 357: stylistic nit: any reason for this is a local function? I agree it's helpful to have it abstracted out as a named method, but I think that this is _just_ long enough to be refactored into a class-level method.

src/WebJobs.Extensions.DurableTask/EntityScheduler/MessageSorter.cs
```diff
@@ -48,8 +57,13 @@ public void LabelOutgoingMessage(RequestMessage message, string destination, Dat
 
             DateTime timestamp = now;
 
+            bool countingEntries = false;
+            int destinationCount = 0;
+
             if (this.SendHorizon + reorderWindow + MinIntervalBetweenCollections < now)
             {
+                countingEntries = true;
+
```
davidmrdavid, line 66: I realize this is pre-existing code, so just asking this for my own understanding: can you please clarify what each of terms in the if-statement mean? As in - what is the `SendHorizon`, `reorderWindow`, etc. Just trying to reason about this check. Thanks!

davidmrdavid (commented):
Just leaving some questions as an entity-internal newie :) . Most of the PR makes sense, so I think I should be ready to approve shortly after this round.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
sebastianburckhardt: Sure. TraceFlags are essentially a "log inside the log". They create a string of log characters. This is because it can be tricky to figure out what path was taken through the code.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
sebastianburckhardt: Sure, I agree.

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
sebastianburckhardt: I suppose it is a matter of preference. I like local functions because code that belongs together logically stays in close proximity in the source file which helps me to read it. For example, I don't have to define and pass "stopwatch" or "onHostStopping" as arguments, and when you read the function you can directly see where those variables come from.

src/WebJobs.Extensions.DurableTask/EntityScheduler/MessageSorter.cs
sebastianburckhardt: the message sorter is designed to reorder/deduplicate messages. As a first approximation, it works like this:
- on sending side: for each destination, timestamp of last message sent. When sending a message, it includes that timestamp in the "Predecessor" field.
- on receiving side: for each source, timestamp of last message received. If it receives a message specifying "Predecessor" that matches what was recorded as last message, receive it and update the recorded timestamp. If it is smaller discard it as duplicate. If it is larger, put it into a buffer for out-of-order message until it matches.

now, this algorithm is modified so it stops making any reordering or deduplication efforts outside the specified tolerance window (default is 30 minutes). The `SendHorizon` is that window on the sender side, and the `ReceiveHorizon` is that window on the receiver side.

sebastianburckhardt: And in order to not spend too much time going through the data structures to clean out irrelevant entries, we use `MinIntervalBetweenCollections`

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
sebastianburckhardt: ok.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
davidmrdavid: I like, I can see it being really useful.

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
davidmrdavid: Fair enough, it is indeed just a matter of preference. I won't block on this.

davidmrdavid (approved):

src/WebJobs.Extensions.DurableTask/EntityScheduler/MessageSorter.cs
davidmrdavid: I see. That definitely allows me to understand the code better but I think I'm still missing a piece of the puzzle. I'd love to sync with you over the details of this later. For now, the changes make sense to me.

sebastianburckhardt:
I am still not seeing these new trace events (e.g. EventId 233) show up in our telemetry. @davidmrdavid , in your experience, is this long delay normal? I know we had to update the filters that control which EventIds get collected, but I would have expected this to take effect by now.

davidmrdavid:
@sebastianburckhardt: if I recall correctly, we had to make a change _somewhere_ in Antares to allow for this EventId, and antares releases can be rather slow (like how we had to wait for the scale controller for Netherite/MSSQL). Some Antares bits release more often than others, but assuming this is part of the slowest payload, then this delay would be "normal".

@cgillum: I recall you made this change (extending the trace eventIds supported by our platform). Do you know if that change should have released already?

PR #2411, opened 2023-03-02
Add host.json setting for new Distributed Tracing feature

bachuv:
This PR adds a host.json setting so customers can easily configure the new distributed tracing feature. 

A new `TelemetryModule` called `DurableFunctionsTelemetryModule` from DurableTask.ApplicationInsights is used if the new host.json setting, `DurableDistributedTracingVersion`, is set to `V2`.

```
"durableTask": {
      "tracing": {
        "DistributedTracingEnabled": true,
        "Version": "V2"
      }
    }
```

Related DurableTask PR - https://github.com/Azure/durabletask/pull/871

image

### Pull request checklist


src/WebJobs.Extensions.DurableTask/Options/TraceOptions.cs
```diff
@@ -64,6 +64,13 @@ public class TraceOptions
         /// </summary>
         public string DistributedTracingProtocol { get; set; } = "HttpCorrelationProtocol";
 
+        /// <summary>
+        /// Gets or sets a flag indicating whether to enable the new version of distributed tracing
+        /// that is built using OpenTelemetry.
+        /// The default value is false.
+        /// </summary>
+        public bool NewDistributedTracingEnabled { get; set; } = false;
+
```
bachuv, line 73: I left this name for now, but I want to change it before merging since it's not very self-explanatory. The current setting `DistributedTracingEnabled` is used in various different places for the previous implementation so I was hesitant to use that setting. I'm open to suggestions on how to handle this.

src/WebJobs.Extensions.DurableTask/Correlation/ITelemetryActivator.cs
```diff
@@ -10,6 +11,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Correlation
     /// </summary>
     public interface ITelemetryActivator
     {
+        /// <summary>
+        /// Configuration used for Application Insights.
+        /// </summary>
+        TelemetryConfiguration Configuration { get; }
+
```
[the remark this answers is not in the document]
davidmrdavid: Do we have alternative places in which to store this? Otherwise, I'd think this is fine enough :)

src/WebJobs.Extensions.DurableTask/Options/TraceOptions.cs
davidmrdavid: Are we looking to completely replace the old implementation of Distributed Tracing once this goes into public preview? If so - I'd see the benefit in re-using the name.

bachuv: The plan right now is to keep the old implementation until GA.

src/WebJobs.Extensions.DurableTask/Correlation/ITelemetryActivator.cs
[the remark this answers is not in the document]
bachuv: I ended up removing this property.

src/WebJobs.Extensions.DurableTask/Options/TraceOptions.cs
bachuv: I like this idea! Existing apps that enabled distributed tracing wouldn't break and the OTel compatible implementation is opt-in.

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
```diff
@@ -15,12 +17,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Correlation
...
-    public class TelemetryActivator : ITelemetryActivator
+    public class TelemetryActivator : ITelemetryActivator, IAsyncDisposable
     {
         private readonly DurableTaskOptions options;
-        private TelemetryClient telemetryClient;
+        private readonly INameResolver nameResolver;
         private EndToEndTraceHelper endToEndTraceHelper;
-        private INameResolver nameResolver;
+        private TelemetryClient telemetryClient;
+        private IAsyncDisposable disposable;
```
cgillum, line 26: Any reason to not just call this what it is?
```suggestion
        private IAsyncDisposable telemetryModule;
```

src/WebJobs.Extensions.DurableTask/Options/TraceOptions.cs
cgillum: > We might as well keep the old stuff around but give a way to control what mode the customer uses.

@jviau just to clarify, are you suggesting we have this extra setting just for preview and then remove it for GA?

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
```diff
@@ -39,17 +43,40 @@ public TelemetryActivator(IOptions<DurableTaskOptions> options, INameResolver na
...
+                this.SetupTelemetryConfiguration();
+                if (this.options.Tracing.NewDistributedTracingEnabled)
+                {
+                    DurableTelemetryModule module = new DurableTelemetryModule();
+                    module.Initialize(this.telemetryConfiguration);
+                    this.disposable = module;
+                }
+                else
+                {
+                    this.SetUpDistributedTracing();
```
cgillum, line 69: Should we rename this method to something like `SetUpDeprecatedDistributedTracing()` to make it clear that this applies only to the old implementation?

cgillum (commented):
A left a few comments/questions.

Also, I noticed that this code requires `APPINSIGHTS_INSTRUMENTATIONKEY`. However, shouldn't we also support [APPLICATIONINSIGHTS_CONNECTION_STRING](https://learn.microsoft.com/en-us/azure/azure-functions/functions-app-settings#applicationinsights_connection_string), which is recommended and also required for supporting sovereign clouds? I know this is technically outside the scope of this PR, but we should definitely do this work before releasing for public preview.

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
```diff
@@ -129,7 +158,7 @@ private void SetUpTelemetryClient()
                     message: "'APPINSIGHTS_INSTRUMENTATIONKEY' isn't defined in the current environment variables, but Distributed Tracing is enabled. Please set 'APPINSIGHTS_INSTRUMENTATIONKEY' to use Distributed Tracing.");
             }
 
-            this.telemetryClient = new TelemetryClient(config);
+            this.telemetryConfiguration = config;
```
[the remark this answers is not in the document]
bachuv: Thanks for the suggestion! That makes sense. It's also used in `SetupTelemetryClient()`, but I just pass it in as a parameter now.

src/WebJobs.Extensions.DurableTask/Options/TraceOptions.cs
jviau: @cgillum I don't think we should remove the existing distributed tracing even after GA. At least not right away. We should give a deprecation window so customers that already use it have a chance to migrate over. We may want to track this work, but we should probably give an in-portal alert for customers to switch over to new distributed tracing, and give a timeline that it will be removed at a certain date (we can decide on that date).

jviau:
> Also, I noticed that this code requires APPINSIGHTS_INSTRUMENTATIONKEY. However, shouldn't we also support [APPLICATIONINSIGHTS_CONNECTION_STRING](https://learn.microsoft.com/en-us/azure/azure-functions/functions-app-settings#applicationinsights_connection_string), which is recommended and also required for supporting sovereign clouds? I know this is technically outside the scope of this PR, but we should definitely do this work before releasing for public preview.

Better yet, we can just DI the existing `TelemetryClient` or `IOptions<TelemetryConfiguration>` and not worry about building it ourselves at all. Either of these types works for import, I tested it with the SmokeTest v3 sample.

src/WebJobs.Extensions.DurableTask/Options/TraceOptions.cs
cgillum: Okay, but we probably need @lilyjma to weigh in on this decision.

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
```diff
@@ -80,7 +79,7 @@ public void Initialize(ILogger logger)
             }
         }
 
-        private void SetUpDistributedTracing()
+        private void SetUpDeprecatedDistributedTracing()
```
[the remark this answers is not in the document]
cgillum: This is a preview feature that we never intend to GA so V1 feels misleading. Deprecated feels like the right term to me, even before we GA the new experience.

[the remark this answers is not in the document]
cgillum: But just to argue with myself - I suppose another way to look at it is that this has been our only option for so long that we might as well treat it as if it was GA. From that perspective, I suppose it's fine to treat these as V1 and V2.

cgillum:
> > Also, I noticed that this code requires APPINSIGHTS_INSTRUMENTATIONKEY. However, shouldn't we also support [APPLICATIONINSIGHTS_CONNECTION_STRING](https://learn.microsoft.com/en-us/azure/azure-functions/functions-app-settings#applicationinsights_connection_string), which is recommended and also required for supporting sovereign clouds? I know this is technically outside the scope of this PR, but we should definitely do this work before releasing for public preview.
> 
> Better yet, we can just DI the existing `TelemetryClient` or `IOptions<TelemetryConfiguration>` and not worry about building it ourselves at all. Either of these types works for import, I tested it with the SmokeTest v3 sample.

I'm fine with either approach. @bachuv can you look into this?

bachuv:
> > > Also, I noticed that this code requires APPINSIGHTS_INSTRUMENTATIONKEY. However, shouldn't we also support [APPLICATIONINSIGHTS_CONNECTION_STRING](https://learn.microsoft.com/en-us/azure/azure-functions/functions-app-settings#applicationinsights_connection_string), which is recommended and also required for supporting sovereign clouds? I know this is technically outside the scope of this PR, but we should definitely do this work before releasing for public preview.
> > 
> > 
> > Better yet, we can just DI the existing `TelemetryClient` or `IOptions<TelemetryConfiguration>` and not worry about building it ourselves at all. Either of these types works for import, I tested it with the SmokeTest v3 sample.
> 
> I'm fine with either approach. @bachuv can you look into this?

@cgillum Yes, of course! I'll look into it for public preview.

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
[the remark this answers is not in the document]
jviau: > I suppose another way to look at it is that this has been our only option for so long that we might as well treat it as if it was GA

This is how I feel about it. Unless we find that leaving it in causes actual issues, then we might as well leave it be for existing users.

src/WebJobs.Extensions.DurableTask/Options/TraceOptions.cs
lilyjma: Thanks for looping me in. 

I've discussed with Varshi. Since replacing the old implementation would break current users, Jacob's suggestion sounds reasonable. Also, because the old implementation has been in preview for >1 year, we'll need to treat it as GA when retiring it, _assuming_ that it has >1k customers. Per Azure Retirement process, we'll need to give 18 month notice to retire the old implementation when there's a GA replacement (we could also replace with preview but that would require CTO approval). If we have <1k customers on the old implementation, the process would be easier. So the first step is to get the number of customers using the old implementation.

src/WebJobs.Extensions.DurableTask/Options/DurableDistributedTracingVersion.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// <summary>
+    /// Represents configuration options for Distributed Tracing in Durable Functions.
+    /// </summary>
+    public enum DurableDistributedTracingVersion
+    {
+        /// <summary>
+        /// Original implementation of Distributed Tracing in Durable Functions
+        /// that can be configured to use HttpCorrelationProtocol or W3CTraceContext.
+        /// </summary>
+        V1 = 0,
```
jviau, line 15: We should probably add a `None` value, just to set us up to remove the explicit enabled property on the options (or at least tweak it to interact with this enum, and in a future PR we can do this).

Make sure to then also update the `TelemetryActivator` to no-op when `None` is set.

``` CSharp
None = 0,
V1 = 1,
V2 = 2,
```

src/WebJobs.Extensions.DurableTask/Options/TraceOptions.cs
```diff
@@ -61,15 +62,15 @@ public class TraceOptions
...
 
         /// <summary>
-        /// Gets or sets a flag indicating whether to enable the new version of distributed tracing
-        /// that is built using OpenTelemetry.
-        /// The default value is false.
+        /// Gets or sets a <see cref="DistributedTracingVersion" /> value indicating which version of distributed tracing to use.
+        /// The default value is 'DurableDistributedTracingVersion.V1'.
         /// </summary>
-        public bool NewDistributedTracingEnabled { get; set; } = false;
+        public DurableDistributedTracingVersion DurableDistributedTracingVersion { get; set; } = DurableDistributedTracingVersion.V1;
```
jviau, line 73: What do you think of naming the property `Version`?
```suggestion
        public DurableDistributedTracingVersion Version { get; set; } = DurableDistributedTracingVersion.V1;
```

jviau (approved):

bachuv:
Thanks for approving, @jviau! I'll merge these changes after https://github.com/Azure/durabletask/pull/648 is merged since this PR is dependent on https://github.com/Azure/durabletask/pull/648.

@cgillum, can you take a look at the recent changes?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -17,6 +17,7 @@
 using DurableTask.Core.Exceptions;
 using DurableTask.Core.History;
 using DurableTask.Core.Middleware;
+using Microsoft.ApplicationInsights.Extensibility;
```
cgillum, line 20: Is this change required?

cgillum:
> @cgillum Yes, of course! I'll look into it for public preview.

If you don't want to do Jacob's proposal for private preview, can we at least add support for `APPLICATIONINSIGHTS_CONNECTION_STRING` in the private preview? I feel like it would be a mistake to release any preview without support for both configuration options, especially `APPLICATIONINSIGHTS_CONNECTION_STRING` which is the only one we recommend.

bachuv:
@cgillum, that makes sense! I added support for `APPLICATIONINSIGHTS_CONNECTION_STRING`.

cgillum (approved):
Looks good - thanks!

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
```diff
@@ -48,6 +48,12 @@ public ValueTask DisposeAsync()
             return this.telemetryModule?.DisposeAsync() ?? default;
         }
 
+        /// <inheritdoc/>
+        public void Dispose()
+        {
+            this.telemetryModule?.DisposeAsync().AsTask().GetAwaiter().GetResult();
+        }
+
```
davidmrdavid, line 56: I'm a bit concerned about this pattern, as I understand that calling `GetAwaiter()` is generally anti-pattern. Do we use disposing pattern (`.GetAwaiter().GetResult`) anywhere else in the codebase?

davidmrdavid (commented):
Just a small question

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
bachuv: Yes, I see it used in DurableTaskExtension.cs ([example](https://github.com/Azure/azure-functions-durable-extension/blob/18d85bd8c0d41d5936cac46054e3839066ef65a7/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs#L559))  and DurableTaskListener.cs ([example](https://github.com/Azure/azure-functions-durable-extension/blob/18d85bd8c0d41d5936cac46054e3839066ef65a7/src/WebJobs.Extensions.DurableTask/Listener/DurableTaskListener.cs#L88)).

davidmrdavid (approved):

PR #2414, opened 2023-03-06
Add Extension-GUID field to telemetry and dependency updates

davidmrdavid:
~Same as: https://github.com/Azure/durabletask/pull/872~

~Creating this PR so the team can review and make suggestions for logs and mitigations to add in the private package being used to diagnose and resolve IcM 360157507~

Changes in this PR:
* Create and log a GUID per DF Extension instance, so that we can detect if there are multiple DF extensions co-existing in a Functions Host at the same time. This also serves as a proxy for a process ID.
* ExtensionGUID is also logged in the message field of some TaskHub shutdown logs. This will be helpful in identifying _which_ Extension's taskhub is shutting down.
* Also increased the DTFx dependencies and added references to our new ADO-based private feed


src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -88,6 +88,7 @@ public class DurableTaskExtension :
         private bool isTaskHubWorkerStarted;
         private HttpClient durableHttpClient;
         private EventSourceListener eventSourceListener;
+        private Guid extensionGUID;
```
cgillum, line 91: Can this be made `readyonly`? It looks like we only assign it in the constructor. Also, per standard C# naming conventions, it should be `extensionGuid`.
```suggestion
        private readonly Guid extensionGuid;
```

davidmrdavid: Sure, good idea. I'll incorporate this feedback without the GitHub UI so that I can easily propagate the naming change via VS

cgillum (approved):

PR #2441, opened 2023-04-07
Prepare v1.0.2 release of worker package

jviau:


nuget.config
```diff
@@ -2,8 +2,9 @@
 
 <configuration>
   <packageSources>
-    <add key="nuget.org" value="https://www.nuget.org/api/v2/" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
     <add key="azure-appservice-staging" value="https://www.myget.org/F/azure-appservice-staging/api/v3/index.json" />
     <add key="azure-appservice" value="https://www.myget.org/F/azure-appservice/api/v3/index.json" />
```
davidmrdavid, line 7: nit: should we remove these?

davidmrdavid (approved):

nuget.config
jviau: I think we can do that in a separate PR. As part of the formal work to move our nuget packages. I only added the new feed so I could coordinate the public NuGet.org push together.

PR #2452, opened 2023-04-21
Add target-based scaling support for Azure Storage

davidmrdavid:
## This PR

This PR adds target-based scaling support for the AzureStorage backend.

## Changes

The requirements were relatively simple: we needed to update our WebJobs SDK dependency and then implement and expose some new interfaces: `ITargetScaler` and `ITargetScalerProvider`. This PR provides such an implementation for the Azure Storage backend. 

## Backwards compatibility

These new interfaces are not available in older versions of WebJobs, meaning that we need to use the C# preprocessor to prevent older versions of the Functions runtime from trying to load those types. This has worked for Functions V1 and V2. However, our bundles-mandated TFMs cannot differentiate between Functions V3 and V4. 

Once we release this change, the Durable Extension will no longer be compatible with Functions V3. Therefore, prior to merge, we should update our documentation to warn users of this change.

## Merge pre-requisites

In order to prevent apps w/ extension bundles from auto-upgrading to an incompatible version of the DF Extension, there will be a patch release to Functions V3 to stop automatic bundle upgrades. We need to wait for that patch release to complete.

## Testing
This PR adds unit tests to validate that we request the expected number of workers. We use the same inputs as in our internal design document.

We have also done **local** manual testing validating that the Host recognizes these changes and considers them when issuing scale votes.

## Equivalent PRs for other backends
* Netherite: https://github.com/microsoft/durabletask-netherite/pull/265
* MSSQL: https://github.com/microsoft/durabletask-mssql/pull/169

## References
* Equivalent EventHubs PR - https://github.com/Azure/azure-sdk-for-net/pull/33675
* Equivalent ServiceBus PR - https://github.com/Azure/azure-sdk-for-net/pull/31367

_Work done in collaboration with_ @bachuv.

### Pull request checklist


src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTriggerMetrics.cs
```diff
@@ -16,7 +17,7 @@ internal class DurableTaskTriggerMetrics : ScaleMetrics
         /// The number of messages across control queues. This will
         /// be in the form of a serialized array of ints, e.g. "[1,2,3,4]".
         /// </summary>
-        public string ControlQueueLengths { get; set; }
+        public virtual string ControlQueueLengths { get; set; }
```
davidmrdavid, line 20: made this virtual to enable mocking. Same with `WorkItemQueueLength=

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
```diff
@@ -0,0 +1,61 @@
...
+        public async Task<TargetScalerResult> GetScaleResultAsync(TargetScalerContext context)
+        {
+            var metrics = await this.metricsProvider.GetMetricsAsync();
+
+            // compute activityWorkers: the number of workers we need to process all activity messages
+            var workItemQueueLength = metrics.WorkItemQueueLength;
+            double activityWorkers = Math.Ceiling(workItemQueueLength / (double)this.MaxConcurrentActivities);
+
+            var serializedControlQueueLengths = metrics.ControlQueueLengths;
+            var controlQueueLengths = JsonConvert.DeserializeObject<IReadOnlyList<int>>(serializedControlQueueLengths);
```
davidmrdavid, line 45: For reasons unknown to me - we get the control queue lengths already serialized. Technically speaking, I could add a new field to obtain the raw data, but I decided to first get feedback on this decision.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        private readonly DurableTaskMetricsProvider metricsProvider;
+        private readonly TargetScalerResult scaleResult;
+        private readonly DurabilityProvider durabilityProvider;
+
+        public DurableTaskTargetScaler(string functionId, DurableTaskMetricsProvider metricsProvider, DurabilityProvider durabilityProvider)
+        {
+            this.metricsProvider = metricsProvider;
+            this.scaleResult = new TargetScalerResult();
+            this.TargetScalerDescriptor = new TargetScalerDescriptor(functionId);
```
bachuv, line 26: I see that `functionId` is used to create the `TargetScalerDescriptor`. Are there any other uses for `functionId`?

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.Azure.WebJobs.Extensions.DurableTask.ContextImplementations;
 using Microsoft.Azure.WebJobs.Extensions.DurableTask.Correlation;
 using Microsoft.Azure.WebJobs.Extensions.DurableTask.Options;
+using Microsoft.Azure.WebJobs.Host.Scale;
```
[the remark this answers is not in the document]
davidmrdavid: yep, good catch. Removing it :)

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
davidmrdavid: I don't think so. I believe this is literally just the function name as a string. It's called `functionId` for convention, but it might as well be `functionName`.

cgillum:
Question on this point:

> Once we release this change, the Durable Extension will no longer be compatible with Functions V3. Therefore, prior to merge, we should update our documentation to warn users of this change.

I'm worried that updating our documentation won't be enough. Although Functions V3 has been deprecated, I worry that some users may still try to use it for one reason or another. Can we consider making some kind of proactive announcement that this new version of the DF extension no longer supports Functions V3, and perhaps include the expected error message for improved SEO when customers inevitably run into this?

cgillum:
Also, @davidmrdavid, please fix the merge conflicts so that we can be confident about the changes we're reviewing in WebJobs.Extensions.DurableTask.csproj.

davidmrdavid:
@cgillum: thanks for the heads up on merge conflicts, I missed those. They've been addressed.

> I'm worried that updating our documentation won't be enough. Although Functions V3 has been deprecated, I worry that some users may still try to use it for one reason or another. Can we consider making some kind of proactive announcement that this new version of the DF extension no longer supports Functions V3, and perhaps include the expected error message for improved SEO when customers inevitably run into this?

Yep, I agree with this. I'll obtain the exact error message tomorrow-ish and we can look to include that in some proactive announcement, as well as in this PR and maybe even our release notes.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskMetricsProvider.cs
```diff
@@ -0,0 +1,79 @@
...
+
+#if !FUNCTIONS_V1
+using System;
+using System.Threading.Tasks;
+using DurableTask.AzureStorage.Monitoring;
+using Microsoft.Extensions.Logging;
+using Microsoft.WindowsAzure.Storage;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Listener
```
cgillum, line 12: For consistency with existing code, please remove ".Listener" from the namespace of these two new files.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
```diff
@@ -39,7 +42,15 @@ internal sealed class DurableTaskScaleMonitor : IScaleMonitor<DurableTaskTrigger
             this.storageAccount = storageAccount;
             this.logger = logger;
             this.performanceMonitor = performanceMonitor;
+            this.durableTaskMetricsProvider = durableTaskMetricsProvider;
+
+#if FUNCTIONS_V3_OR_GREATER
+            this.scaleMonitorDescriptor = new ScaleMonitorDescriptor($"{this.functionId}-DurableTaskTrigger-{this.hubName}".ToLower(), this.functionId);
+#else
+#pragma warning disable CS0618 // Type or member is obsolete
```
cgillum, line 50: Can you improve the comment to explain why it's okay for us to disable the obsolete warning?

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskMetricsProvider.cs
```diff
@@ -0,0 +1,79 @@
...
+                    metrics.WorkItemQueueLatency = heartbeat.WorkItemQueueLatency.ToString();
+                }
+            }
+
+            return metrics;
+        }
+
+        private DisconnectedPerformanceMonitor GetPerformanceMonitor()
+        {
+            if (this.performanceMonitor == null)
```
cgillum, line 65: In what situations do we expect `this.performanceMonitor` to be null?

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTriggerMetrics.cs
cgillum: I suggest we just make all of them virtual for consistency, unless you think there's a reason not to.

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
```diff
@@ -200,6 +203,17 @@ internal static OrchestrationInstanceStatusQueryCondition ConvertWebjobsDurableC
         }
 
 #if !FUNCTIONS_V1
+
+        internal DurableTaskMetricsProvider GetMetricsProvider(
+            string functionName,
+            string hubName,
+            CloudStorageAccount storageAccount,
+            ILogger logger,
+            DisconnectedPerformanceMonitor performanceMonitor = null)
```
cgillum, line 212: I recommend not using default parameter values for internal methods. I'd rather have the caller explicitly pass null when making internal method calls.

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
```diff
@@ -208,12 +222,30 @@ internal static OrchestrationInstanceStatusQueryCondition ConvertWebjobsDurableC
             string connectionName,
             out IScaleMonitor scaleMonitor)
         {
+            var storageAccount = this.storageAccountProvider.GetStorageAccountDetails(connectionName).ToCloudStorageAccount();
+            var metricsProvider = this.GetMetricsProvider(functionName, hubName, storageAccount, this.logger);
```
cgillum, line 226: Please use explicit variable types instead of `var` since we can't infer the type by simply reading the code.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -1589,10 +1589,52 @@ internal IScaleMonitor GetScaleMonitor(string functionId, FunctionName functionN
...
+            /// <summary>
+            /// Construct a placeholder target scaler.
+            /// </summary>
+            /// <param name="functionId">The function ID.</param>
+            public NoOpTargetScaler(string functionId)
+            {
+                this.TargetScalerDescriptor = new TargetScalerDescriptor(functionId);
+            }
+
+            public TargetScalerDescriptor TargetScalerDescriptor { get; private set; }
```
cgillum, line 1628: Can the `private set` be removed?
```suggestion
            public TargetScalerDescriptor TargetScalerDescriptor { get; }
```

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
```diff
@@ -0,0 +1,79 @@
...
+            // When running on ScaleController V3, ILogger logs are forwarded to Kusto.
+            var scaleControllerLog = $"Target worker count for {this.functionId}: {numWorkersToRequest}. " +
+                $"Metrics used: workItemQueueLength={workItemQueueLength}. controlQueueLengths={serializedControlQueueLengths}. " +
+                $"maxConcurrentOrchestrators={this.MaxConcurrentOrchestrators}. maxConcurrentActivities={this.MaxConcurrentActivities}";
+
+            // target worker count should never be negative
+            if (numWorkersToRequest < 0)
+            {
+                scaleControllerLog = "Tried to request a negative worker count." + scaleControllerLog;
+                this.logger.LogError(scaleControllerLog);
```
cgillum, line 70: Shouldn't we be writing structured logs that can be included in our Kusto tables?

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            this.functionId = functionId;
+            this.metricsProvider = metricsProvider;
+            this.scaleResult = new TargetScalerResult();
+            this.TargetScalerDescriptor = new TargetScalerDescriptor(this.functionId);
+            this.durabilityProvider = durabilityProvider;
+            this.logger = logger;
+        }
+
+        public TargetScalerDescriptor TargetScalerDescriptor { get; private set; }
```
cgillum, line 33:
```suggestion
        public TargetScalerDescriptor TargetScalerDescriptor { get; }
```

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
```diff
@@ -0,0 +1,79 @@
...
+            var scaleControllerLog = $"Target worker count for {this.functionId}: {numWorkersToRequest}. " +
+                $"Metrics used: workItemQueueLength={workItemQueueLength}. controlQueueLengths={serializedControlQueueLengths}. " +
+                $"maxConcurrentOrchestrators={this.MaxConcurrentOrchestrators}. maxConcurrentActivities={this.MaxConcurrentActivities}";
+
+            // target worker count should never be negative
+            if (numWorkersToRequest < 0)
+            {
+                scaleControllerLog = "Tried to request a negative worker count." + scaleControllerLog;
+                this.logger.LogError(scaleControllerLog);
+                throw new Exception(scaleControllerLog);
```
cgillum, line 71: Is throwing an exception the behavior we want? Or should we fail more gracefully?

cgillum (commented):
Added some questions/comments.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskMetricsProvider.cs
davidmrdavid: I'm not sure, this logic predates me. You can find it already in our repo here: https://github.com/Azure/azure-functions-durable-extension/blob/5c6172ca353b5ce59a301170fb3daeb57ae4f4f3/src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs#L53-L65

In fact, I did not meant to duplicate this logic (it is now here and in the `DurableTaskScaleMonitor` class), I must have forgotten to put it on a shared location

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTriggerMetrics.cs
davidmrdavid: Fair enough, they're all virtual now :)

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskMetricsProvider.cs
davidmrdavid: It's been removed, and so have all imports of it

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskScaleMonitor.cs
davidmrdavid: I added a comment. Essentially, this new constructor is not compatible with Functions V1 -> V3. Unfortunately, we don't have a way to exclude that whole group, but we can exclude Functions V1 and V2 and have those use the old, obsolote, constructor. That's what we're doing here.

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
davidmrdavid: Makes sense. I realized after this comment that this parameter was always null, so I just removed it anyways.

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
davidmrdavid: Done!

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
davidmrdavid: Incorporated in my next commit

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
davidmrdavid: To be clear, this is not going to the Durable Functions Kusto table, but to the ScaleController table. It seems they want us to log using their ILogger implementation, and from there they have custom logic to forward that to Kusto. I'm just following their convention here.

If you'd like, I can _additionally_ log this data in our Kusto table via some structured log. I'd have to look whether or not we have the correct EventSource listener on the **ScaleController-host** to even do this though. After all, this code doesn't run in the context of the Functions runtime, but in the context of the Scale Controller process.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
davidmrdavid: I can't immediately think of a more graceful solution here. My hunch is that if the scaling logic is requesting a negative number, then something is terribly wrong and we should fail exceptionally so the ScaleController can take over and decide how to deal with it. Do you have any suggestions?

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskMetricsProvider.cs
davidmrdavid: Let me try to refactor this so that we're not duplicating this logic. Looking into that shortly.

davidmrdavid: duplication removed in this commit: https://github.com/Azure/azure-functions-durable-extension/pull/2452/commits/2e6a3175c4c6a7e229aad58b4aed3ff053f76700

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
cgillum: So you're saying this code is executed by the scale controller and not in the Functions host? Are there some doc comments we can add to make this clear? I think that also helps clarify my question about the direct use of ILogger.

cgillum (approved):
These changes look good to me. My only ask would be to consider adding more comments to make clear the cases where code is run in the Scale Controller host instead of running in the Functions host since that has big implications in terms of what context we're allowed to assume.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
davidmrdavid: My understanding is that ScaleController V3 calls this directly, although @bachuv just pointed me to a [comment](https://github.com/Azure/azure-webjobs-sdk/issues/2909) that suggests that this code may also be accessed in the runtime driven scale case **as well,** which would go through the Host. So perhaps there's two cases: one mediated by the Functions Host, and one where it's accessed directly by the Scale Controller.

I'll look to get a conclusive answer on this. For now, I've pushed a series of comments reflecting my current understanding. I won't merge until we have confirmation from the ScaleController team.

bachuv (approved):

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
davidmrdavid: I spoke with Alexey today to get clarity on this.

The caller of this code depends on whether the user is running with or without a VNET.

If running **without** a VNET, the ScaleController process calls this directly.
If running with a VNET, then the customer needs to enable **runtime-driven scaling** and in that case it is the Host which provides calls this code for us.

I think this may complicate things a little bit, especially around throwing exceptions. We should confirm what's the preferred error handling behavior in the VNET case.

davidmrdavid: @alrod: if runtime-driven scaling is enabled, and the DF Extension encounters an error in calculating it's target worker count (for a example, if it wrongly determines that it needs a negative target worker count), is safe to throw an exception for the Host to catch or should we handle the failure more gracefully?

test/FunctionsV2/WebJobs.Extensions.DurableTask.Tests.V2.csproj
```diff
@@ -16,7 +16,7 @@
 
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="4.19.4" />
-    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.31" />
+    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.36" />
```
alrod, line 19: Can we update the version to latest https://www.nuget.org/packages/Microsoft.Azure.WebJobs/3.0.37

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -89,7 +89,7 @@
   </PropertyGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
     <PackageReference Include="Azure.Identity" Version="1.5.0" />
-    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.31" />
+    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.36" />
```
alrod, line 92: Lets update to the latest 3.0.37

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
alrod: Is the code even make since to exists?

`int numWorkersToRequest = (int)Math.Max(activityWorkers, orchestratorWorkers);`
Is possible activityWorkers and orchestratorWorkers are < 0 ?

src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
```diff
@@ -10,6 +10,7 @@
 using DurableTask.AzureStorage.Tracking;
 using DurableTask.Core;
 using Microsoft.Extensions.Logging;
+using Microsoft.WindowsAzure.Storage;
```
nytian, line 13: Is the target-based scaling support still based on the old deprecated AS backend?

PR #2456, opened 2023-04-25
Ensure function abort on shutdown

### LINKED ISSUE #2454: Durable function fails with `TaskCanceledException` and never gets retried

jviau:
Fixes an issue with gRPC out-of-proc durable workers where all in-flight activities and orchestrations would fail during host shutdown. The root issue is that the function invocation actually _completes_, but in a failed state, during host shutdown. The fix is to check for a combination of failure and host shutting down post invocation, then ensure we abort the session in this case.

resolves #2454

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
+++ b/src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
@@ -135,8 +135,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     context.SetResult(
                         response.Actions.Select(ProtobufUtils.ToOrchestratorAction),
                         response.CustomStatus);
-#pragma warning restore CS0618 // Type or member is obsolete (not intended for general public use)
                 },
+#pragma warning restore CS0618 // Type or member is obsolete (not intended for general public use)
             };
 
             FunctionResult functionResult;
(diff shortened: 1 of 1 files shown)


src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -145,6 +145,12 @@ public async Task CallOrchestratorAsync(DispatchMiddlewareContext dispatchContex
                 functionResult = await function.Executor.TryExecuteAsync(
                     input,
                     cancellationToken: this.HostLifetimeService.OnStopping);
+                if (!functionResult.Succeeded)
+                {
+                    // Shutdown can surface as a completed invocation in a failed state.
+                    // Re-throw so we can abort this invocation.
+                    this.HostLifetimeService.OnStopping.ThrowIfCancellationRequested();
```
cgillum, line 152: I assume that this throws `OperationCanceledException`, and that our existing logic knows how to gracefully abort when this exception is encountered?

jviau: Correct - this is inside a `try/catch` block, which will catch that exception and rethrow a `SessionAbortedException`

cgillum (approved):

davidmrdavid (approved):
version rev looks good

PR #2462, opened 2023-05-02
Add ScaleController V3 integration

davidmrdavid:
## This PR
This PR allows Scale Controller V3 to receive TBS worker requests for DF. This PR builds upon the changes in https://github.com/Azure/azure-functions-durable-extension/pull/2452 so please review that PR first.

Since this PR builds upon the changes in https://github.com/Azure/azure-functions-durable-extension/pull/2452, that' is also the branch we're targeting.

## Main Changes

The main objective of this PR is the implementation of `DurableTaskTriggersScaleProvider`, which the Scale Controller Host will utilize to make scaling decisions. 

This class implements  `IScaleMonitorProvider` and `ITargetScalerProvider` by delegating to the `GetTargetScaler` and `GetScaleMonitor` implementations of https://github.com/Azure/azure-functions-durable-extension/pull/2452.

When constructing  `DurableTaskTriggersScaleProvider`, we receive the DI services provider from the Scale Controller Host and the Sync Triggers payload for the function app. We read the user's config (hubName, backend type, etc) from the Sync Triggers payload.

Scale Controller V3 will obtain an instance of this class by invoking `private static IWebJobsBuilder AddDurableScaleForTrigger(this IWebJobsBuilder builder, TriggerMetadata triggerMetadata)` during initialization.

Finally, note that `DurableTaskTriggersScaleProvider` will not be constructed in the Functions Host. It is constructed only by the  ScaleController Host through a call to `AddDurableScaleForTrigger`.

## Secondary Changes

For code re-use, this PR removes a several scaling utility methods defined in `DurableTaskExtension.cs` and puts them inside a new file: `Scale/ScaleUtils.cs`. That way, we're able to re-use these utilites in the
`DurableTaskTriggersScaleProvider` implementation.

We also add `#nullable enable` in all new files.

## References:
* Equivalent ServiceBus PR: https://github.com/Azure/azure-sdk-for-net/pull/34557
* Equivalent EventHubs PR: https://github.com/alrod/azure-sdk-for-net/commit/78e99c24a8923c30082d75b2342bfe93507ea114

resolves #issue_for_this_pr

### Pull request checklist


src/WebJobs.Extensions.DurableTask/Scale/ScaleUtils.cs
```diff
@@ -0,0 +1,118 @@
...
+                    out ITargetScaler targetScaler))
+            {
+                return targetScaler;
+            }
+            else
+            {
+                // the durability provider does not support target-based scaling.
+                // Create an empty target scaler to avoid exceptions (unless target-based scaling is actually turned on).
+                return new NoOpTargetScaler(functionId);
+            }
```
davidmrdavid, line 95: unknown: should we "fail fast" on the scalehost if TBS is not supported?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -150,13 +150,13 @@ public DurableTaskExtension()
             this.nameResolver = nameResolver ?? throw new ArgumentNullException(nameof(nameResolver));
             this.loggerFactory = loggerFactory ?? throw new ArgumentNullException(nameof(loggerFactory));
             this.PlatformInformationService = platformInformationService ?? throw new ArgumentNullException(nameof(platformInformationService));
-            this.ResolveAppSettingOptions();
+            DurableTaskOptions.ResolveAppSettingOptions(this.Options, this.nameResolver);
```
davidmrdavid, line 154: made this method static so it may be re-used in `DurableTaskTriggersScaleProvider`

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -150,13 +150,13 @@ public DurableTaskExtension()
...
             this.PlatformInformationService = platformInformationService ?? throw new ArgumentNullException(nameof(platformInformationService));
-            this.ResolveAppSettingOptions();
+            DurableTaskOptions.ResolveAppSettingOptions(this.Options, this.nameResolver);
 
             ILogger logger = loggerFactory.CreateLogger(LoggerCategoryName);
 
             this.TraceHelper = new EndToEndTraceHelper(logger, this.Options.Tracing.TraceReplayEvents);
             this.LifeCycleNotificationHelper = lifeCycleNotificationHelper ?? this.CreateLifeCycleNotificationHelper();
-            this.durabilityProviderFactory = this.GetDurabilityProviderFactory(this.Options, logger, orchestrationServiceFactories);
+            this.durabilityProviderFactory = GetDurabilityProviderFactory(this.Options, logger, orchestrationServiceFactories);
```
davidmrdavid, line 159: Also made this static so it may be re-used in `GetDurabilityProviderFactory`.

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
```diff
@@ -107,6 +109,33 @@ public static IWebJobsBuilder AddDurableTask(this IWebJobsBuilder builder)
...
+            {
+                serviceProvider = serviceProvider ?? resolvedServiceProvider;
+                return scalerProvider.Value;
+            });
+
+            builder.Services.AddSingleton<ITargetScalerProvider>(resolvedServiceProvider =>
+            {
+                serviceProvider = serviceProvider ?? resolvedServiceProvider;
+                return scalerProvider.Value;
+            });
```
davidmrdavid, line 133: The explanation for this lazy initialization pattern is here: https://github.com/Azure/azure-sdk-for-net/pull/34557/files#r1170772719

src/WebJobs.Extensions.DurableTask/Scale/DurableTaskTriggersScaleProvider.cs
```diff
@@ -0,0 +1,127 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Scale
+{
+    internal class DurableTaskTriggersScaleProvider : IScaleMonitorProvider, ITargetScalerProvider
+    {
+        private IScaleMonitor monitor;
+        private ITargetScaler targetScaler;
```
cgillum, line 20: Make these `readonly`

src/WebJobs.Extensions.DurableTask/Scale/DurableTaskTriggersScaleProvider.cs
```diff
@@ -0,0 +1,127 @@
...
+            IServiceProvider serviceProvider,
+            TriggerMetadata triggerMetadata)
+        {
+            string functionId = triggerMetadata.FunctionName;
+            FunctionName functionName = new FunctionName(functionId);
+
+            var options = this.GetOptions(serviceProvider, triggerMetadata);
+
+            var durabilityProviderFactory = this.GetDurabilityProviderFactory(serviceProvider, options);
+            var defaultDurabilityProvider = durabilityProviderFactory.GetDurabilityProvider();
```
cgillum, line 32: Please use explicit types instead of `var` when it's not obvious what the type is.

src/WebJobs.Extensions.DurableTask/Scale/DurableTaskTriggersScaleProvider.cs
```diff
@@ -0,0 +1,127 @@
...
+                options.HubName);
+        }
+
+        private DurableTaskOptions GetOptions(IServiceProvider serviceProvider, TriggerMetadata triggerMetadata)
+        {
+            var nameResolver = serviceProvider.GetService<INameResolver>();
+
+            // the metadata is the sync triggers payload
+            var metadataString = triggerMetadata.Metadata.ToString();
+            var metadata = JsonConvert.DeserializeObject<DurableTaskMetadata>(metadataString);
```
cgillum, line 61: Is `triggerMetadata.Metadata` a `JObject`? If so, the more direct way to turn this into an instance of a class is to call `ToObject<T>()`
```suggestion
            var metadata = triggerMetadata.Metadata.ToObject<DurableTaskMetadata>();
````

File: src/WebJobs.Extensions.DurableTask/Scale/DurableTaskTriggersScaleProvider.cs [hunk 8]
```diff
@@ -0,0 +1,127 @@
...
+            [DefaultValue(null)]
+            public int? MaxConcurrentOrchestratorFunctions { get; set; }
+
+            [JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate)]
+            [DefaultValue(null)]
+            public int? MaxConcurrentActivityFunctions { get; set; }
+
+            [JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate)]
+            [DefaultValue(null)]
+            public IDictionary<string, object>? StorageProvider { get; set; }
```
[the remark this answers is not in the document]
cgillum: I assumed that this was implied by the fact that these are nullable properties and have `[DefaultValue(null)]` attributes.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
```diff
@@ -38,46 +39,55 @@ public DurableTaskTargetScaler(string functionId, DurableTaskMetricsProvider met
...
+
+                // compute orchestratorWorkers: the number of workers we need to process all orchestrator messages.
+                // We bound this result to be no larger than the partition count
+                var upperBoundControlWorkers = Math.Ceiling(controlQueueMessages / (double)this.MaxConcurrentOrchestrators);
+                var orchestratorWorkers = Math.Min(activeControlQueues, upperBoundControlWorkers);
+
+                int numWorkersToRequest = (int)Math.Max(activityWorkers, orchestratorWorkers);
+                this.scaleResult.TargetWorkerCount = numWorkersToRequest;
+
+                // When running on ScaleController V3, ILogger logs are forwarded to Kusto.
```
cgillum, line 65: Did you mean to change these comments? I recalled you added this extra detail in the previous PR but now it seems you're reverting it back.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
```diff
@@ -38,46 +39,55 @@ public DurableTaskTargetScaler(string functionId, DurableTaskMetricsProvider met
...
+                var scaleControllerLog = $"Target worker count for '{this.functionId}' is '{numWorkersToRequest}'. " +
+                    metricsLog;
+
+                // target worker count should never be negative
+                if (numWorkersToRequest < 0)
+                {
+                    scaleControllerLog = $"Error: target worker count for '{this.functionId}' was negative: '{this.functionId}'." +
+                        "An exception was thrown." + metricsLog;
+                    this.logger.LogError(scaleControllerLog);
+                    throw new Exception(scaleControllerLog);
```
cgillum, line 77: I'm just now noticing this, but it's bad practice to throw `Exception`. Consider using `InvalidOperationException` instead.

cgillum: Actually, doesn't this result in double-exception and double logging? If I'm reading the code correctly, we catch this exception a few lines down and make another call to `this.logger.LogError`. It seems like we need to reconsider the design here, because using exceptions as control flow like this is also bad practice.

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
```diff
@@ -107,6 +109,33 @@ public static IWebJobsBuilder AddDurableTask(this IWebJobsBuilder builder)
             return builder;
         }
 
+#if FUNCTIONS_V3_OR_GREATER
+        /// <summary>
+        /// Adds the IScaleMonitor and ITargetScaler providers for the Durable Triggers.<see cref="IWebJobsBuilder"/>.
```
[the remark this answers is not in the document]
davidmrdavid: I really need to get into the habit of this. Please keep on reminding me, but I'll try to start remembering :)

src/WebJobs.Extensions.DurableTask/Scale/DurableTaskTriggersScaleProvider.cs
davidmrdavid: Thanks.  Yeah I see I've been doing this. Will be addressed in my next commit

src/WebJobs.Extensions.DurableTask/Scale/ScaleUtils.cs
```diff
@@ -0,0 +1,120 @@
...
+using System;
+using System.Threading.Tasks;
+
+#if !FUNCTIONS_V1
+using Microsoft.Azure.WebJobs.Host.Scale;
+#endif
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Scale
+{
+    internal class ScaleUtils
```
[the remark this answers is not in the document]
davidmrdavid: It used to hold state, until I refactored it a few days ago. Made static, thanks!

src/WebJobs.Extensions.DurableTask/Scale/DurableTaskTriggersScaleProvider.cs
davidmrdavid: Cool trick. Incorporated on my next commit

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
```diff
@@ -107,6 +109,33 @@ public static IWebJobsBuilder AddDurableTask(this IWebJobsBuilder builder)
...
+        /// </summary>
+        /// <param name="builder">The <see cref="IWebJobsBuilder"/> to configure.</param>
+        /// <returns>Returns the provided <see cref="IWebJobsBuilder"/>.</returns>
+        public static IWebJobsBuilder AddDurableScaleForTrigger(this IWebJobsBuilder builder, TriggerMetadata triggerMetadata)
```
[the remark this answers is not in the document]
davidmrdavid: I agree the name is weird (and there's no a "durable scale" operation, whatever that means :-) ), I'm just adhering to the naming conventions in the ScV3 repo.

Other builder methods include:
* AddTimerScaleForTrigger
* AddServiceBusScaleForTrigger
* AddCosmosDbScaleForTrigger
* AddAzureStorageQueuesScaleForTrigger
etc.

I could definitely rename this to `AddDurableFunctionsScaleForTrigger` if that helps. However, I think the convention of `Add<Extension>ScaleForTrigger` means that the naming will be strange no matter what.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -229,6 +230,25 @@ public string HubName
         // to mock the value from ExtensionConfigContext. It should not be used in production code paths.
         internal Func<Uri> WebhookUriProviderOverride { get; set; }
 
+        internal static void ResolveAppSettingOptions(DurableTaskOptions options, INameResolver nameResolver)
+        {
+            if (options == null)
+            {
+                throw new InvalidOperationException($"{nameof(options)} must be set before resolving app settings.");
```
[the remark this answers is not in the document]
davidmrdavid: I don't mind making this change but I want to clarify that this method isn't just copied for `DurableTaskExtension.cs`, it's been refactored out of `DurableTaskExtension.cs` so that it can be used _here_ (for TBS purposes) and in the `DurableTaskExtension.cs`. With that in mind, do you still want me to change `InvalidOperationException`, or keep it as is.

I could also just copy the code in case we want different exception types in different locations.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
davidmrdavid: Ah I didn't meant to revert these changes, it was just a bad merge. I'm adding these back.

src/WebJobs.Extensions.DurableTask/Scale/ScaleUtils.cs
```diff
@@ -0,0 +1,120 @@
...
+                    connectionName,
+                    out IScaleMonitor scaleMonitor))
+            {
+                return scaleMonitor;
+            }
+            else
+            {
+                // the durability provider does not support runtime scaling.
+                // Create an empty scale monitor to avoid exceptions (unless runtime scaling is actually turned on).
+                return new NoOpScaleMonitor($"{functionId}-DurableTaskTrigger-{hubName}".ToLower(), functionId);
```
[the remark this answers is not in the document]
davidmrdavid: Yes, this PR refactors the location of these classes from `DurableTaskExtension` to `ScaleUtils`, where I think it's more convenient to have them.

Your comment made me realize that I had forgotten to remove the old declarations of these methods and classes in `DurableTaskExtension.cs`. That's been addressed in my latest commit: https://github.com/Azure/azure-functions-durable-extension/pull/2462/commits/5e7fc18b9f3e29f9bc91d5c26d5800aa374debc5 .

src/WebJobs.Extensions.DurableTask/Scale/ScaleUtils.cs
davidmrdavid: Yeah I don't disagree, although I don't have a ScaleController-provided ILogger instance handy in this class. Let me try to get some clarity on this edge case from the Sc team

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
davidmrdavid: Yeah you're right, it's getting double logged; good call out.

A part of me wants a try-catch surrounding this whole operation so we can log any exceptions that arise, and to include in those logs the contextual information needed to debug this. That's what I do in the outer catch block.

Do you think it would suffice to remove the `LogError` when the target worker count is negative, and simply throw an exception? Alternatively, I could refactor the outer try-catch so that it ends right when the target worker count is calculated (i.e line 71)

src/WebJobs.Extensions.DurableTask/Scale/ScaleUtils.cs
```diff
@@ -0,0 +1,120 @@
...
+
+            /// <summary>
+            /// A descriptive name.
+            /// </summary>
+            public ScaleMonitorDescriptor Descriptor { get; private set; }
+
+            /// <inheritdoc/>
+            Task<ScaleMetrics> IScaleMonitor.GetMetricsAsync()
+            {
+                throw new InvalidOperationException("The current DurableTask backend configuration does not support runtime scaling");
```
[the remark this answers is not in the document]
davidmrdavid: This is actually the same behavior we have today (the class already exists, it's just been moved to `ScaleUtils`), see here in our `dev` branch: https://github.com/Azure/azure-functions-durable-extension/blob/5c6172ca353b5ce59a301170fb3daeb57ae4f4f3/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs#L1596-L1629

The reasoning for this implementation decision is here: https://github.com/Azure/azure-functions-durable-extension/pull/1009/files#r341763309

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
```diff
@@ -107,6 +109,33 @@ public static IWebJobsBuilder AddDurableTask(this IWebJobsBuilder builder)
             return builder;
         }
 
+#if FUNCTIONS_V3_OR_GREATER
+        /// <summary>
+        /// Adds the <see cref="IScaleMonitor"/> and <see cref="ITargetScaler"/> providers for the Durable Triggers.
+        /// </summary>
+        /// <param name="builder">The <see cref="IWebJobsBuilder"/> to configure.</param>
+        /// <returns>Returns the provided <see cref="IWebJobsBuilder"/>.</returns>
+        public static IWebJobsBuilder AddDurableScaleForTrigger(this IWebJobsBuilder builder, TriggerMetadata triggerMetadata)
```
alrod, line 118: Please make the method as internal. We decided to use reflection to call the extensions methods.

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
alrod: Please change the pater to:
https://[github.com/Azure/azure-sdk-for-net/pull/38673](https://github.com/Azure/azure-sdk-for-net/pull/38673)

test/FunctionsV2/WebJobs.Extensions.DurableTask.Tests.V2.csproj
```diff
@@ -16,7 +16,7 @@
 
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="4.19.4" />
-    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.36" />
+    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.37" />
```
alrod, line 19: Can we add e2e test, something like:
https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/tests/ScaleHostEndToEndTests.cs

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
davidmrdavid: Thanks, it's been addressed here: https://github.com/Azure/azure-functions-durable-extension/pull/2462/commits/3800635bc41957c9ec60880755dc5b4914d90868

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
davidmrdavid: Sounds good, it's been addressed here: https://github.com/Azure/azure-functions-durable-extension/pull/2462/commits/3800635bc41957c9ec60880755dc5b4914d90868

davidmrdavid: @alrod: does this mean that we need to change how ScV3 calls into our extension method? If so, can you please link @bachuv to an example of how to change this in the Sc codebase? thanks!

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
```diff
@@ -107,6 +109,22 @@ public static IWebJobsBuilder AddDurableTask(this IWebJobsBuilder builder)
...
 
+#if FUNCTIONS_V3_OR_GREATER
+        /// <summary>
+        /// Adds the <see cref="IScaleMonitor"/> and <see cref="ITargetScaler"/> providers for the Durable Triggers.
+        /// </summary>
+        /// <param name="builder">The <see cref="IWebJobsBuilder"/> to configure.</param>
+        /// <returns>Returns the provided <see cref="IWebJobsBuilder"/>.</returns>
+        internal static IWebJobsBuilder AddDurableScaleForTrigger(this IWebJobsBuilder builder, TriggerMetadata triggerMetadata)
+        {
+            // this segment adheres to the followings pattern: https://github.com/Azure/azure-sdk-for-net/pull/38673/files
```
cgillum, line 120: This is really weird at first glance, but I think I get it. 😅

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
cgillum: At a minimum we should remove the first `LogError` and rely on the one inside the catch-block. But I wonder if we even need that one since we're throwing. Wouldn't it be expected that some other code catches these exceptions and logs them as errors?

src/WebJobs.Extensions.DurableTask/Scale/DurableTaskTriggersScaleProvider.cs
```diff
@@ -0,0 +1,124 @@
...
+        {
+            var nameResolver = serviceProvider.GetService<INameResolver>();
+
+            // the metadata is the sync triggers payload
+            var metadata = triggerMetadata.Metadata.ToObject<DurableTaskMetadata>();
+
+            var options = serviceProvider.GetService<IOptions<DurableTaskOptions>>().Value;
+
+            // The property `taskHubName` is always expected in the SyncTriggers payload
+            options.HubName = metadata?.TaskHubName ?? throw new Exception($"Expected `taskHubName` property in SyncTriggers payload but found none. Payload: {triggerMetadata.Metadata}");
```
cgillum, line 63: Avoid throwing `Exception`. It's preferred to throw `InvalidOperationException` if you don't have any better option.

src/WebJobs.Extensions.DurableTask/Scale/DurableTaskTriggersScaleProvider.cs
```diff
@@ -0,0 +1,124 @@
...
+                defaultDurabilityProvider,
+                functionId,
+                functionName,
+                connectionName,
+                options.HubName);
+        }
+
+        private DurableTaskOptions GetOptions(IServiceProvider serviceProvider, TriggerMetadata triggerMetadata)
+        {
+            var nameResolver = serviceProvider.GetService<INameResolver>();
```
cgillum, line 55: Minor nit: but I think it's consider best practice to have dependency-injected services like `INameResolver` and `IOptions<DurableTaskOptions>` as explicit constructor parameters instead of using `IServiceProvider`.

src/WebJobs.Extensions.DurableTask/Scale/DurableTaskTriggersScaleProvider.cs
```diff
@@ -0,0 +1,124 @@
...
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Scale
+{
+    internal class DurableTaskTriggersScaleProvider : IScaleMonitorProvider, ITargetScalerProvider
+    {
+        private readonly IScaleMonitor monitor;
+        private readonly ITargetScaler targetScaler;
+
+        public DurableTaskTriggersScaleProvider(
+            IServiceProvider serviceProvider,
```
cgillum, line 23: nit: I think it's considered good practice to declare your dependent services explicitly in your constructor rather than taking `IServiceProvider` and fetching them manually. It certainly makes the code more unit testable.

For example, you could replace `IServiceProvider` here with parameters for `INameResolver`, `ILoggerFactory`, `IOptions<DurableTaskOptions>`, and `IEnumerable<IDurabilityProviderFactory>`.

cgillum (commented):
Added a few more comments. I think the only one I consider blocking is the one about double (or maybe even triple) logging errors.

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
alrod: I will change calling the method once we have the nuget package and update the reference in SC v3.x.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
```diff
@@ -38,46 +39,58 @@ public DurableTaskTargetScaler(string functionId, DurableTaskMetricsProvider met
...
-            // and the ScaleController is injecting it's own custom ILogger implementation that forwards logs to Kusto.
-            var scaleControllerLog = $"Target worker count for {this.functionId}: {numWorkersToRequest}. " +
-                $"Metrics used: workItemQueueLength={workItemQueueLength}. controlQueueLengths={serializedControlQueueLengths}. " +
-                $"maxConcurrentOrchestrators={this.MaxConcurrentOrchestrators}. maxConcurrentActivities={this.MaxConcurrentActivities}";
-
-            // target worker count should never be negative
-            if (numWorkersToRequest < 0)
+            DurableTaskTriggerMetrics? metrics = null;
+            try
             {
```
alrod, line 44: To fall back to regular scale for Netherite and MSSQL you need to throw `NonSupportedException`:

[azure-sdk-for-net/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs at main · Azure/azure-sdk-for-net (github.com)](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs#L61)

 
Do you have this implemented for durable?

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
```diff
@@ -38,46 +39,58 @@ public DurableTaskTargetScaler(string functionId, DurableTaskMetricsProvider met
...
-            // When running on ScaleController V3, ILogger logs are forwarded to the ScaleController's Kusto table.
-            // This works because this code does not execute in the Functions Host process, but in the ScaleController process,
-            // and the ScaleController is injecting it's own custom ILogger implementation that forwards logs to Kusto.
-            var scaleControllerLog = $"Target worker count for {this.functionId}: {numWorkersToRequest}. " +
-                $"Metrics used: workItemQueueLength={workItemQueueLength}. controlQueueLengths={serializedControlQueueLengths}. " +
-                $"maxConcurrentOrchestrators={this.MaxConcurrentOrchestrators}. maxConcurrentActivities={this.MaxConcurrentActivities}";
-
-            // target worker count should never be negative
-            if (numWorkersToRequest < 0)
+            DurableTaskTriggerMetrics? metrics = null;
```
alrod, line 42: To fall back to regular scale for Netherite and MSSQL you need to throw NonSupportedException:

[azure-sdk-for-net/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs at main · Azure/azure-sdk-for-net (github.com)](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs#L61)

Do you have this implemented for durable?

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
bachuv: @alrod, do you know what the expected behavior is for logging and throwing exceptions? Do we log the error or will it be handled in the scale controller?

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
bachuv: I updated the exception [here](https://github.com/Azure/azure-functions-durable-extension/blob/becd05ebc2e6956221fd2dcaf9a0f15c1de24b85/src/WebJobs.Extensions.DurableTask/Scale/ScaleUtils.cs#L115).

Yes, we can add a test.

test/FunctionsV2/WebJobs.Extensions.DurableTask.Tests.V2.csproj
bachuv: Yes, we'll work on this as we don't have any existing tests in this repo for TBS scaling. We've been testing these changes in the scale controller repo by running the integration test there.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
cgillum: Still looking for an update on this.

alrod: `numWorkersToRequest` is calcuated as:
`int numWorkersToRequest = (int)Math.Max(activityWorkers, orchestratorWorkers);`

How `numWorkersToRequest `can be negative?

If you want to throw an exception from `GetScaleResultAsync` I would just thorwing `InvalidOperationException/ArgumentOutOfRangeException` without writing the error log - Scale Controller will take care about this

Something like:
https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs#L94

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskListener.cs
```diff
@@ -51,21 +52,25 @@ internal sealed class DurableTaskListener : IListener
...
+                ScaleUtils.GetScaleMonitor(
+                    this.config.DefaultDurabilityProvider,
                     this.functionId,
                     this.functionName,
-                    this.connectionName));
+                    this.connectionName,
+                    this.config.Options.HubName));
 
 #endif
 #if FUNCTIONS_V3_OR_GREATER
```
alrod, line 66: In SC we have "FUNCTIONS_V3_OR_GREATER", right?

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
```diff
@@ -107,6 +111,22 @@ public static IWebJobsBuilder AddDurableTask(this IWebJobsBuilder builder)
...
+            builder.Services.AddSingleton(serviceProvider => new DurableTaskTriggersScaleProvider(serviceProvider.GetService<IOptions<DurableTaskOptions>>(), serviceProvider.GetService<INameResolver>(), serviceProvider.GetService<ILoggerFactory>(), serviceProvider.GetService<IEnumerable<IDurabilityProviderFactory>>(), triggerMetadata));
```
alrod, line 123: We can pass serviceProvider

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
```diff
@@ -107,6 +111,22 @@ public static IWebJobsBuilder AddDurableTask(this IWebJobsBuilder builder)
...
+            builder.Services.AddSingleton<IScaleMonitorProvider>(serviceProvider => serviceProvider.GetRequiredService<DurableTaskTriggersScaleProvider>());
```
alrod, line 124: There was a bug in this pattern, please change according to:
https://github.com/Azure/azure-sdk-for-net/pull/38756

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
davidmrdavid: Thanks @alrod. I agree that the current code cannot provide negative numbers, my check is just ensuring that remains true if the code is ever refactored. It may a bit too defensive, so I don't mind removing it if others think is overkill :).

@cgillum: I've removed the double logging on this commit: https://github.com/Azure/azure-functions-durable-extension/pull/2462/commits/05332642f6340b53cb6069c06eced3f7ed8e1273

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskListener.cs
davidmrdavid: I believe so. SC has TFM `.net6.0` and `FUNCTIONS_V3_OR_GREATER` targets `netcoreapp3.1`, which is "older" than `.net6.0`. So yes, I believe this code triggers in SC.

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
```diff
@@ -38,46 +39,54 @@ public DurableTaskTargetScaler(string functionId, DurableTaskMetricsProvider met
...
+                }
+
+                this.logger.LogInformation(scaleControllerLog);
+                return this.scaleResult;
+            }
+            catch (Exception ex)
+            {
+                var metricsLog = $"Metrics: workItemQueueLength={metrics?.WorkItemQueueLength}. controlQueueLengths={metrics?.ControlQueueLengths}. " +
+                    $"maxConcurrentOrchestrators={this.MaxConcurrentOrchestrators}. maxConcurrentActivities={this.MaxConcurrentActivities}";
+                var errorLog = $"Error: target worker count for '{this.functionId}' resulted in exception." + metricsLog + $"Exception: {ex}";
```
cgillum, line 87: Please add whitespace/punctuation around `metricsLog` to make the exception message more readable.
```suggestion
                var errorLog = $"Error: target worker count for '{this.functionId}' resulted in exception. " + metricsLog + $". Exception: {ex}";
```

Consider just having a single string template for this instead of two to make it easier to reason about what the actual error message will look like.

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
davidmrdavid: Thanks, addressed here: https://github.com/Azure/azure-functions-durable-extension/pull/2462/commits/1975b414394693f7464e743dccae0ca23814fb98

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
```diff
@@ -38,46 +39,54 @@ public DurableTaskTargetScaler(string functionId, DurableTaskMetricsProvider met
...
+                throw new Exception(errorLog, ex);
```
cgillum, line 88: nit: we're effectively double-logging the exception here my including the same exception information in both `errorLog` and as an inner exception `ex`. This can be confusing for someone trying to debug scale controller logs.

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
davidmrdavid: Thanks @alrod. I think the code did this originally, but as per this comment we decided not to resolve services in the constructor: https://github.com/Azure/azure-functions-durable-extension/pull/2462#discussion_r1336409281

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
davidmrdavid: You're right, I think I was being overly verbose here. I've updated the code to remove the string representation of the error and instead just rely on the "innerException" parameter: https://github.com/Azure/azure-functions-durable-extension/pull/2462/commits/f5c6d3f1c14622f3383ca7055f71a82bb79c2a9e

cgillum (approved):

src/WebJobs.Extensions.DurableTask/Scale/ScaleUtils.cs
```diff
@@ -0,0 +1,120 @@
...
+#if !FUNCTIONS_V1
+using Microsoft.Azure.WebJobs.Host.Scale;
+#endif
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Scale
+{
+    internal static class ScaleUtils
+    {
+#if !FUNCTIONS_V1
+        internal static IScaleMonitor GetScaleMonitor(DurabilityProvider durabilityProvider, string functionId, FunctionName functionName, string? connectionName, string hubName)
```
alrod, line 17: Can we also add unit tests for `ScaleUtils.GetTargetScaler` and  `ScaleUtils.GetScaleMonitor`

davidmrdavid: Just added them: https://github.com/Azure/azure-functions-durable-extension/pull/2462/commits/635aa753269dc859ceb27242fd45856117ce9582

davidmrdavid: and expanded them here: https://github.com/Azure/azure-functions-durable-extension/pull/2462/commits/3ad0f2897415867cfa9faaaea8cbeb505c3d0b92

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
davidmrdavid: added: https://github.com/Azure/azure-functions-durable-extension/pull/2462/commits/3ad0f2897415867cfa9faaaea8cbeb505c3d0b92

davidmrdavid: this was incorporated in my latest commit

src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
davidmrdavid: Changed the exception type to InvalidOperationException: https://github.com/Azure/azure-functions-durable-extension/pull/2462/commits/7f47657f53b5206168a4c92ad9181e8486f567b9

test/FunctionsV2/DurableTaskTargetScalerTests.cs
```diff
@@ -132,5 +137,43 @@ public void TestGetScaleMonitor(bool supportsScaleMonitor)
...
+                var expectedSubString = "scale monitors to sample";
+                if (isTbsEnabled)
+                {
+                    expectedSubString = "target scalers to sample";
+                }
+
+                bool containsExpectedLog = this.loggerProvider.GetAllLogMessages().Select(p => p.FormattedMessage ?? "").Any(p => p.Contains(expectedSubString));
+                Assert.True(containsExpectedLog);
+            }
+        }
```
davidmrdavid, line 177: @alrod: an E2E test, using the pre-existing DF testing infra, has been added here ^

davidmrdavid: I realize this is not exactly like the E2E tests you suggested (doing so would have taken more time due to how this codebase is structured and the DF initialization boilerplate), but I think this, in addition to the unit tests we already have, should have sufficient testing coverage.

alrod (approved):

PR #2489, opened 2023-06-07
Fix missing dedupe states for gRPC client server impl

### LINKED ISSUE #2466: Singleton out of process orchestrator two running at the same time
Hi,
We've used the singleton pattern for our orchestrator to make sure only one execution is performed at one time, because of a critical section regarding two sequential operations, see  https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-singletons?tabs=csharp#singleton-example

Now when we've tested it, sometimes the orchestration will have to instances running at the same time. We can see logs from one activity in the orchestrator is running at the same time and failing because of it.

Is this pattern broken in the out of process worker?

jviau:
Fixes missing dedupe states in the gRPC server client when scheduling a new orchestration via `IOrchestrationServiceClient`. Without the dedupe states, we would effectively restart already running orchestrations when it was scheduled again.

resolves #2466

### Pull request checklist

    * [ ] Otherwise: I've added my notes to `release_notes.md` -- TODO

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -195,7 +195,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 throw new ArgumentException($"Instance ID lengths must not exceed {MaxInstanceIdLength} characters.");
             }
 
-            var dedupeStatuses = this.GetStatusesNotToOverride();
+            OrchestrationStatus[] dedupeStatuses = this.GetStatusesNotToOverride();
             Task<OrchestrationInstance> createTask = this.client.CreateOrchestrationInstanceAsync(
                 orchestratorFunctionName, DefaultVersion, instanceId, input, null, dedupeStatuses);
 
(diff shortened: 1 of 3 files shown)


src/WebJobs.Extensions.DurableTask/OverridableStates.cs
```diff
@@ -21,4 +24,33 @@ public enum OverridableStates
...
+#pragma warning disable SA1649 // File name should match first type name Justification: pairing extension methods with enum.
+    internal static class OverridableStatesExtensions
+#pragma warning restore SA1649 // File name should match first type name
+    {
+        private static readonly OrchestrationStatus[] NonRunning = new OrchestrationStatus[]
+        {
+            OrchestrationStatus.Running,
+            OrchestrationStatus.ContinuedAsNew,
+            OrchestrationStatus.Pending,
+        };
```
cgillum, line 40: Does `Suspended` need to be in this list?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -213,18 +213,8 @@ async Task<string> IDurableOrchestrationClient.StartNewAsync<T>(string orchestra
...
-                {
-                    OrchestrationStatus.Running,
-                    OrchestrationStatus.ContinuedAsNew,
-                    OrchestrationStatus.Pending,
-                };
-            }
-
-            return new OrchestrationStatus[0];
+            OverridableStates overridableStates = this.durableTaskOptions.OverridableExistingInstanceStates;
+            return overridableStates.ToDedupeStatuses();
```
cgillum, line 217: I'm confused. The method name is `GetStatusesNotToOverride`, but the value we're returning seems to be the opposite of this: `overridableStates`. Why is this logic correct? Assuming it is correct, can we rename something to make it _obviously_ correct?

src/WebJobs.Extensions.DurableTask/OverridableStates.cs
jviau: Yeah, I think that is a good change. Might even be a bug fix for all other DF languages

cgillum:
I think this Java issue will get resolved by this PR as well: https://github.com/microsoft/durabletask-java/issues/135

cgillum (approved):

PR #2498, opened 2023-06-22
Remove pushing to MyGet step from release pipeline

davidmrdavid:
As in title, just removing an obsolete step.


bachuv (approved):
Thanks for removing this step!

PR #2509, opened 2023-07-07
Allow durability provider to disable status checks on RaiseEventAsync

sebastianburckhardt:
## Problem Description

The current implementation of RaiseEventAsync first reads the status of the orchestration instance from storage, before enqueueing the raised events. This has negative performance implications on Netherite where the status check is not a quick cheap operation as for Azure Storage (see microsoft/durabletask-netherite#276)

Interestingly, our documentation is ambiguous about whether this check should happen:

- our Azure docs [Handling external events in Durable Functions - Azure | Microsoft Learn](https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-external-events?tabs=csharp) do just say the "event is discarded" if there is no orchestration instance with that instance id. It does not say anything about an exception being thrown. Also, it does not mention what happens if the instance is not in a running state. 
- For the isolated client (FunctionsDurableTaskClient), the Intellisense on the RaiseEventAsync says  "Raised events for a completed or non-existent instance will be silently discarded". This matches the semantics of DurableTask. I would like this to be the behavior when using Netherite. This makes sense particularly because Netherite is meant for high-scale situations where performance is more important than helping beginning users to understand how to correctly use events.
- For the in-process client (DurableClient), the Intellisense on the RaiseEventAsync says "If the specified instance is not found or not running, this operation will throw an exception." This is indeed what that method does. I expect that seeing these exception can be helpful for users who are unsure about how to design their application in a way that avoids races.
 
## Proposed Solution

This PR adds a flag to the durability provider that allows the durability provider to disable the status check.

It also disables the status check in the case where RaiseEventAsyncInternal is called by CleanEntityStorageAsync since the check is unwanted in that case also.

### Pull request checklist

[ ]  Documentation PR is ready to merge and referenced in `pending_docs.md`
[ ]  I've added my notes to `release_notes.md`

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -231,7 +231,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 throw new ArgumentNullException(nameof(eventName));
             }
 
-            return this.RaiseEventInternalAsync(this.client, this.TaskHubName, instanceId, eventName, eventData);
+            return this.RaiseEventInternalAsync(this.client, this.TaskHubName, instanceId, eventName, eventData, this.durabilityProvider.CheckStatusBeforeRaiseEvent);
         }
 
         /// <inheritdoc />
@@ -256,7 +256,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             DurableClient durableClient = (DurableClient)this.GetDurableClient(taskHubName, connectionName);
             TaskHubClient taskHubClient = durableClient.client;
 
-            return this.RaiseEventInternalAsync(taskHubClient, taskHubName, instanceId, eventName, eventData);
+            return this.RaiseEventInternalAsync(taskHubClient, taskHubName, instanceId, eventName, eventData, this.durabilityProvider.CheckStatusBeforeRaiseEvent);
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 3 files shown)


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -231,7 +231,7 @@ Task IDurableOrchestrationClient.RaiseEventAsync(string instanceId, string event
                 throw new ArgumentNullException(nameof(eventName));
             }
 
-            return this.RaiseEventInternalAsync(this.client, this.TaskHubName, instanceId, eventName, eventData);
+            return this.RaiseEventInternalAsync(this.client, this.TaskHubName, instanceId, eventName, eventData, this.durabilityProvider.CheckStatusBeforeRaiseEvent);
```
cgillum, line 234: Rather than passing this flag as a method argument, you can reduce the size of this change by having `RaiseEventInternalAsync` query this new property directly.

cgillum (approved):
This change looks good to me. It would be great if we could also make this configurable in host.json. Some users of the Azure Storage backend have complained that the current status check behavior is problematic for them because of how it's non-deterministic.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
sebastianburckhardt: I think I will leave it as it is since it allows me to always disable the status check when the function is called from CleanEntityStorageAsync (in that context the check is always superfluous since the entity status was already checked earlier).

PR #2450, opened 2023-04-18
updated environment variable read to TryAdd

amdeel:
This PR addresses an issue where the customer's function can fail occasionally with and ArgumentException "An item with the same key has already been added." In the case of concurrent reads.

### Pull request checklist

* [ ] My changes **do not** require documentation changes
    * [ ] Otherwise: Documentation PR is ready to merge and referenced in `pending_docs.md`
* [ ] My changes **should not** be added to the release notes for the next release
    * [ ] Otherwise: I've added my notes to `release_notes.md`
* [ ] My changes **do not** need to be backported to a previous version
    * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [ ] I have added all required tests (Unit tests, E2E tests)
* [ ] My changes **do not** require any extra work to be leveraged by OutOfProc SDKs
    * [ ] Otherwise: That work is being tracked here: #issue_or_pr_in_each_sdk
* [ ] My changes **do not** change the version of the WebJobs.Extensions.DurableTask package
    * [ ] Otherwise: major or minor version updates are reflected in `/src/Worker.Extensions.DurableTask/AssemblyInfo.cs`
* [ ] My changes **do not** add EventIds to our EventSource logs
    * [ ] Otherwise: Ensure the EventIds are within the supported range in our existing Windows infrastructure. You may validate this with a deployed app's telemetry. You may also extend the range by completing a PR such as [this one](https://msazure.visualstudio.com/One/_git/AAPT-Antares-Websites/pullrequest/7463263?_a=files).

PR #2532, opened 2023-08-03
Make environment dictionary concurrency-safe

davidmrdavid:
Replaces: https://github.com/Azure/azure-functions-durable-extension/pull/2450

From the old description: This PR addresses an issue where the customer's function can fail occasionally with an ArgumentException "An item with the same key has already been added". This may occur on current accesses to the dictionary. To make these operations safe, we use a concurrent dictionary.

resolves #issue_for_this_pr

### Pull request checklist

    * [x] Otherwise: Documentation PR is ready to merge and referenced in `pending_docs.md`
    * [x] Otherwise: Backport tracked by issue/PR #issue_or_pr
    * [x] Otherwise: That work is being tracked here: #issue_or_pr_in_each_sdk
    * [x] Otherwise: Ensure the EventIds are within the supported range in our existing Windows infrastructure. You may validate this with a deployed app's telemetry. You may also extend the range by completing a PR such as [this one](https://msazure.visualstudio.com/One/_git/AAPT-Antares-Websites/pullrequest/7463263?_a=files).

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DefaultPlatformInformation.cs
+++ b/src/WebJobs.Extensions.DurableTask/DefaultPlatformInformation.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 #nullable enable
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using Microsoft.Extensions.Logging;
 
@@ -16,7 +17,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 #pragma warning restore CS0612 // Type or member is obsolete
     {
         private readonly INameResolver nameResolver;
-        private readonly Dictionary<string, string> cachedEnviromentVariables;
+        private readonly ConcurrentDictionary<string, string> cachedEnviromentVariables;
         private readonly EndToEndTraceHelper traceHelper;
 
         private WorkerRuntimeType? workerRuntimeType;
@@ -28,7 +29,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             ILogger logger = loggerFactory.CreateLogger(DurableTaskExtension.LoggerCategoryName);
             this.traceHelper = new EndToEndTraceHelper(logger, traceReplayEvents: false);
 
-            this.cachedEnviromentVariables = new Dictionary<string, string>();
+            this.cachedEnviromentVariables = new ConcurrentDictionary<string, string>();
         }
 
         private string? ReadEnviromentVariable(string variableName)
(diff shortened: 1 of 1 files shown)


src/WebJobs.Extensions.DurableTask/DefaultPlatformInformation.cs
bachuv, line 1: Should we add a unit test for this? Is there an existing one that covers this scenario?

src/WebJobs.Extensions.DurableTask/DefaultPlatformInformation.cs
```diff
@@ -37,7 +38,7 @@ public DefaultPlatformInformation(INameResolver nameResolver, ILoggerFactory log
             if (!this.cachedEnviromentVariables.TryGetValue(variableName, out value))
             {
                 value = this.nameResolver.Resolve(variableName);
-                this.cachedEnviromentVariables.Add(variableName, value);
+                this.cachedEnviromentVariables.TryAdd(variableName, value);
```
bachuv, line 41: Should we add a log if the variable already exists (i.e. `TryAdd()` returns `false`)? Would that be helpful for investigations?

davidmrdavid: I'm not sure this log would be very informative, let me explain why.

In this path, we're caching the values of environment variables (representing app settings) in a dictionary. This is simply so that, if we have read the variable in the past, we don't have to do another IO operation again.

Since we don't support dynamically changing app settings (at least not without a restart), if we already have a dictionary entry for a given app setting, then we can ignore subsequent attempts at caching that value because the entry we already have, and the one we would replace it with, are the same.

Finally, this codepath has always been concurrent, so it's possible that two threads might race to try to update this dictionary, but since they are both guaranteed to read the same value, I don't think it's too useful to log that a given thread lost the update race. I think it's safe to silently ignore because it's idempotent :) .

src/WebJobs.Extensions.DurableTask/DefaultPlatformInformation.cs
davidmrdavid: Since the bug fixed here is a race condition, it's rather difficult to test. We do have tests for this class in general, but none specifically for race conditions in it.

Since the change is so minimal: we're just making a regular dictionary into a concurrent dictionary so that it can tolerate races, I think the change is safe. If you feel strongly though, I can research into making an E2E test that tries to trigger the race condition across several parallel threads or something like that.

src/WebJobs.Extensions.DurableTask/DefaultPlatformInformation.cs
bachuv: That makes sense, thanks for the explanation!

bachuv: That makes sense, thanks for the explanation!

src/WebJobs.Extensions.DurableTask/DefaultPlatformInformation.cs
bachuv: I don't feel strongly about it.

bachuv (approved):

PR #2544, opened 2023-08-14
Use IDurableClient for gRPC OOP languages

jviau:
This PR changes `LocalGrpcListener` to use `IDurableClient` in select places as part of its implementation. Previously we used `IOrchestrationService` directly to implement the gRPC calls. However, this means logic that is part of `TaskHubClient` is not ran. This is not a problem today, but it will be an issue for our distributed tracing improvements. Our tracing work has logic in `TaskHubClient` that is essential for it to all work.

This PR also introduces more orchestrations to the smoke test project for easier manual testing of multiple orchestration patterns.

resolves an issue for distributed tracing work

### Pull request checklist


src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -346,10 +313,12 @@ private DurabilityProvider GetDurabilityProvider(ServerCallContext context)
...
 
-            private OrchestrationStatus[] GetStatusesNotToOverride()
+            private IDurableClient GetClient(ServerCallContext context)
             {
-                OverridableStates overridableStates = this.extension.Options.OverridableExistingInstanceStates;
-                return overridableStates.ToDedupeStatuses();
+                string? taskHub = context.RequestHeaders.GetValue("Durable-TaskHub");
+                string? connectionName = context.RequestHeaders.GetValue("Durable-ConnectionName");
+                var attribute = new DurableClientAttribute() { TaskHub = taskHub, ConnectionName = connectionName };
+                return this.extension.GetClient(attribute);
```
cgillum, line 321: The above three lines are equivalent to those in `GetDurabilityProvider`. Should we factor them out into a separate method to make the code more robust to potential future changes in the gRPC header contract?

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/Patterns.cs
```diff
@@ -0,0 +1,186 @@
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Azure.Functions.Worker.Http;
+using Microsoft.DurableTask;
+using Microsoft.DurableTask.Client;
+using Microsoft.Extensions.Logging;
+
+namespace DurableTest
+{
+    public class Patterns
```
cgillum, line 9: Thoughts on creating automation to run these new smoke tests?

bachuv: That would be great! Are these tests not run as part of [this smoke test](https://github.com/Azure/azure-functions-durable-extension/blob/8dd369afe54793f9d8c2702c6cc7a06f88cfdbf0/.github/workflows/smoketest-dotnet-isolated-v4.yml)? I thought it would be picked up just by looking at the file path that's used for the test, but maybe I'm missing something.

bachuv (commented):
Looks good to me! I would like to address the comments from Chris before approving though.

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/Patterns.cs
jviau: I am actually unfamiliar with the test. I can move these if we want. I added them because I was verifying how more complex orchestration calls behave with distributed tracing here.

jviau: I removed this file for now, we can add it in a separate PR if we want more advanced tests.

cgillum (approved):
High-level question: this seems like a pretty obvious thing to do, and I don't recall why we didn't consider it originally. Do you see any downsides with using `IDurableClient` instead of `IOrchestrationService`?

bachuv (approved):

jviau:
> High-level question: this seems like a pretty obvious thing to do, and I don't recall why we didn't consider it originally. Do you see any downsides with using `IDurableClient` instead of `IOrchestrationService`?

No, I don't see any. I cannot say exactly why it wasn't done this way to start (I didn't write it). I am guessing maybe it was because the `GrpcDurableTaskClient` was written with `IOrchestrationService`, so this was done the same way without thinking too deeply about it.

PR #2550, opened 2023-08-22
Patch GH automation

davidmrdavid:
This PR makes 2 changes to our GH automation:
(1) It makes our Python smoke test run on Functions V4, given that Functions V2 is EOL and is incompatible with some of the new "Python V2 programming model" changes
(2) It fixes our GH automation so that new issues have the "needs: triage" label added automatically again. This stopped working recently when FabricBot migrated to a new config language

### DIFF
--- a/.github/workflows/smoketest-python37-v4.yml
+++ b/.github/workflows/smoketest-python37-v4.yml
@@ -1,4 +1,4 @@
-name: Smoke Test - Python 3.7 on Functions V2
+name: Smoke Test - Python 3.7 on Functions V4
 
 on:
   push:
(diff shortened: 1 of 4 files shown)


cgillum:
> It fixes our GH automation so that new issues have the "needs: triage" label added automatically again. This stopped working recently when FabricBot migrated to a new config language

Should a similar fix be applied to our other repos? I noticed that the MSSQL repo rules weren't working anymore, so there were some issues being missed.

.github/workflows/smoketest-python37-v4.yml
```diff
@@ -17,6 +17,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - name: Run V2 Python 3.7 Smoke Test
+    - name: Run V4 Python 3.7 Smoke Test
```
cgillum, line 20: I was recently informed that [Python 3.7 has reached EOL](https://devguide.python.org/versions/). Should we update this to a newer Python version?

cgillum (approved):

davidmrdavid:
@cgillum: Yes, I'm already porting this to all other repos.

.github/workflows/smoketest-python37-v4.yml
davidmrdavid: I don't feel strongly, but I like testing with a minimal version of Python just so that we try (within reason) to prevent breaking customers where possible. It's still a useful signal knowing that our SDKs continue working in old versions of Python.

PR #2572, opened 2023-09-07
implement passthrough middleware for entities

sebastianburckhardt:
With new core entity support, storage providers can now dispatch entity work items separately, and DT.Core processes them through in a separate dispatcher.

This PR updates the middleware passthrough to reflect this. 

This PR depends on https://github.com/Azure/durabletask/pull/953. 

### Pull request checklist

is targeting feature branch `feature/core-entities`.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Bindings/EntityTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/EntityTriggerAttributeBindingProvider.cs
@@ -9,6 +9,7 @@ using Microsoft.Azure.WebJobs.Host.Bindings;
 using Microsoft.Azure.WebJobs.Host.Listeners;
 using Microsoft.Azure.WebJobs.Host.Protocols;
 using Microsoft.Azure.WebJobs.Host.Triggers;
+using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
@@ -57,6 +58,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private class EntityTriggerBinding : ITriggerBinding
         {
+            private static readonly IReadOnlyDictionary<string, object?> EmptyBindingData = new Dictionary<string, object?>(capacity: 0);
+
             private readonly DurableTaskExtension config;
             private readonly ParameterInfo parameterInfo;
             private readonly FunctionName entityName;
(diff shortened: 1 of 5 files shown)


src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -235,6 +237,159 @@ public async Task CallOrchestratorAsync(DispatchMiddlewareContext dispatchContex
...
+#pragma warning restore CS0618 // Type or member is obsolete (not intended for general public use)
+                },
+            };
+
+            FunctionResult functionResult;
+            try
+            {
+                functionResult = await functionInfo.Executor.TryExecuteAsync(
+                    input,
+                    cancellationToken: this.HostLifetimeService.OnStopping);
```
jviau, line 337: Add the following lines here:

``` CSHarp
if (!functionResult.Succeeded)
{
    // Shutdown can surface as a completed invocation in a failed state.
    // Re-throw so we can abort this invocation.
    this.HostLifetimeService.OnStopping.ThrowIfCancellationRequested();
}
```

jviau (approved):

src/WebJobs.Extensions.DurableTask/Bindings/EntityTriggerAttributeBindingProvider.cs
```diff
@@ -95,31 +98,52 @@ private class EntityTriggerBinding : ITriggerBinding
...
+#if FUNCTIONS_V3_OR_GREATER
+                else if (value is RemoteEntityContext remoteContext)
+                {
+                    // Generate a byte array which is the serialized protobuf payload
+                    // https://developers.google.com/protocol-buffers/docs/csharptutorial#parsing_and_serialization
+                    var entityBatchRequest = remoteContext.Request.ToEntityBatchRequest();
+
+                    // We convert the binary payload into a base64 string because that seems to be the most commonly supported
+                    // format for Azure Functions language workers. Attempts to send unencoded byte[] payloads were unsuccessful.
+                    string encodedRequest = ProtobufUtils.Base64Encode(entityBatchRequest);
```
jviau, line 137: one of the reasons we used proto encoding for orchestrations is that the `OrchestrationRuntimeState` wasn't serializable itself, and this was a quick way to reach serialization. If the built in DT.Core `EntityBatchRequest` and `EntityBatchResult` types are serializable, we can stick with those if you prefer.

sebastianburckhardt: It is indeed serializable. But I don't have much reason to replace the protobuf serialization as done right now with a different one.

PR #2576, opened 2023-09-11
Entities: Add worker side entity trigger and logic

jviau:
Adds the dotnet isolated worker side trigger and logic to allow for entity invocation. Many APIs are stubbed out to what we expect the final dependencies to look like.

resolves https://github.com/microsoft/durabletask-dotnet/issues/188

### Pull request checklist

### DIFF
--- a/src/Worker.Extensions.DurableTask/TaskEntityDispatcher.cs
+++ b/src/Worker.Extensions.DurableTask/TaskEntityDispatcher.cs
@@ -3,15 +3,17 @@
 
 using System;
 using System.Threading.Tasks;
+using Microsoft.DurableTask.Entities;
+using Microsoft.DurableTask.Worker.Grpc;
 using Microsoft.Extensions.DependencyInjection;
 
-namespace Microsoft.Azure.Functions.Worker.Extensions.DurableTask;
+namespace Microsoft.Azure.Functions.Worker;
 
 /// <summary>
 /// Represents a task entity dispatch invocation.
 /// </summary>
 /// <remarks>
-/// This type is used to aid in dispatching a <see cref="EntityTriggerAttribute"/> to the operation reciever object.
+/// This type is used to aid in dispatching a <see cref="EntityTriggerAttribute"/> to the operation receiver object.
 /// </remarks>
 public sealed class TaskEntityDispatcher
 {
@@ -35,7 +37,7 @@ public sealed class TaskEntityDispatcher
     {
         if (entity == null)
         {
-            throw ArgumentNullException(nameof(entity));
+            throw new ArgumentNullException(nameof(entity));
         }
 
         this.Result = await GrpcEntityRunner.LoadAndRunAsync(this.request, entity);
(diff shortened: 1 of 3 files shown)


src/Worker.Extensions.DurableTask/TaskEntityDispatcher.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// <returns></returns>
+    public Task DispatchAsync<T>()
+    {
+        if (typeof(ITaskEntity).IsAssignableFrom(typeof(T)))
+        {
+            ITaskEntity entity = (ITaskEntity)ActivatorUtilities.GetServiceOrCreateInstance<T>(this.services);
+            return this.DispatchAsync(entity);
+        }
+
+        return this.DispatchAsync(new StateEntity<T>());
```
davidmrdavid, line 62: I'm definitely a tad confused about the different cases for `T` here. 

It appears that in the `if` case `<T>` corresponds to a user-level Entity class declaration, like the standard `CounterEntity` class for example. Do I have that right?

I'm not sure I understand what the else-case corresponds to though. I'd appreciate an explanation there :) . Thanks!

davidmrdavid: Thanks, this makes sense.  think it would be good to add a version of this as inlined comment in the implementation.

src/Worker.Extensions.DurableTask/TaskEntityDispatcher.cs
```diff
@@ -0,0 +1,69 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.DurableTask;
```
jviau, line 8: Any thoughts on namespace here? Should this be `Microsoft.DurableTask.Entities` to go with all the other entity types?

src/Worker.Extensions.DurableTask/TaskEntityDispatcher.cs
jviau: For comments, I would prefer we write msdocs and then add a link to that doc in the comments.

src/Worker.Extensions.DurableTask/TaskEntityDispatcher.cs
```diff
@@ -0,0 +1,72 @@
...
+
+/// <summary>
+/// Represents a task entity dispatch invocation.
+/// </summary>
+/// <remarks>
+/// This type is used to aid in dispatching a <see cref="EntityTriggerAttribute"/> to the operation reciever object.
```
sebastianburckhardt, line 14:
```suggestion
/// This type is used to aid in dispatching a <see cref="EntityTriggerAttribute"/> to the operation receiver object.
```

src/Worker.Extensions.DurableTask/TaskEntityDispatcher.cs
sebastianburckhardt: In the user code this class always appear together with `EntityTriggerAttribute` so maybe `Microsoft.Azure.Functions.Worker` makes sense so they don't need an extra using statement.

src/Worker.Extensions.DurableTask/TaskEntityDispatcher.cs
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /// <summary>
+    /// <para>Dispatches the entity trigger to an instance of the provided <typeparamref name="T"/>.</para>
+    /// <para>
+    /// If <typeparamref name="T"/> is <see cref="ITaskEntity"/>, it will be activated from <see cref="IServiceProvider"/>
+    /// and then be dispatched to.
+    /// </para>
+    /// <para>
+    /// If <typeparamref name="T"/> is not <see cref="ITaskEntity"/>, it is assumed the type represents the entity state
```
sebastianburckhardt, line 51:
```suggestion
    /// If <typeparamref name="T"/> is not <see cref="ITaskEntity"/>, it is assumed <typeparamref name="T"/> represents the entity state
```

src/Worker.Extensions.DurableTask/TaskEntityDispatcher.cs
sebastianburckhardt: +1 for deferring further revisions of the XML comments until after we have updated the msdocs. I think this is something we probably want to do for the entire feature branch.

@davidmrdavid , there is an example for a simple counter  here: https://github.com/Azure/azure-functions-durable-extension/pull/2584

sebastianburckhardt (approved):
I marked up some small errors in my comments.

Otherwise, looks good to me!

src/Worker.Extensions.DurableTask/TaskEntityDispatcher.cs
jviau: I see most extensions do put their attributes in that namespace. Not a fan as it bloats the namespace, but yeah we can follow that example.

PR #2584, opened 2023-09-13
add an entity example to the DotNetIsolated smoke test project.

sebastianburckhardt:
Adds a simple entity example to the DotNetIsolated smoke test.

### Pull request checklist

goes into feature branch. Checklist items are tracked by the feature branch.

### DIFF
--- a/test/SmokeTests/OOProcSmokeTests/DotNetIsolated/Counter.cs
+++ b/test/SmokeTests/OOProcSmokeTests/DotNetIsolated/Counter.cs
@@ -24,7 +24,7 @@ public class Counter
     [DataMember]
     public int CurrentValue { get; set; }
 
-    public void Add(int amount, TaskEntityContext context)
+    public void Add(int amount)
     {
         this.CurrentValue += amount;
     }
--- a/test/SmokeTests/OOProcSmokeTests/DotNetIsolated/Counter.cs
+++ b/test/SmokeTests/OOProcSmokeTests/DotNetIsolated/Counter.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-using System.Runtime.Serialization;
 using System.Text;
 using Azure.Core;
 using Microsoft.Azure.Functions.Worker;
@@ -18,10 +17,8 @@ namespace DotNetIsolated;
 /// <summary>
 /// A simple counter, demonstrating entity use.
 /// </summary>
-[DataContract]
 public class Counter
 {
-    [DataMember]
     public int CurrentValue { get; set; }
 
     public void Add(int amount)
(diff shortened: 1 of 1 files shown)


test/SmokeTests/OOProcSmokeTests/DotNetIsolated/Counter.cs
```diff
@@ -0,0 +1,144 @@
...
+using Microsoft.DurableTask.Entities;
+using Microsoft.Extensions.Logging;
+
+namespace DotNetIsolated;
+
+
+/// <summary>
+/// A simple counter, demonstrating entity use.
+/// </summary>
+[DataContract]
```
jviau, line 21: Data attributes are not necessary. They might even be misleading; STJ does not support them by default. You have to implement a completely custom `STJ.DataContractResolver`. Would be better to remove and let customers figure out serialization through existing STJ examples/docs.

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/Counter.cs
```diff
@@ -0,0 +1,144 @@
...
+            int currentValue = response.ReadStateAs<Counter>()!.CurrentValue;
+            var httpResponse = request.CreateResponse(System.Net.HttpStatusCode.OK);
+            httpResponse.Headers.Add("Content-Type", "text/plain; charset=utf-8");
+            httpResponse.WriteString($"{currentValue}\n");
+            return httpResponse;
+        }
+    }
+
+    [Function(nameof(DeleteCounter))]
+    public static async Task<HttpResponseData> DeleteCounter(
```
jviau, line 126: Can you add a comment explaining how delete is an implicit operation.

``` CSharp
// All entities have a "delete" operation built into them.
```

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/Counter.cs
sebastianburckhardt: The goal was to clarify to the reader that this class is going to be serialized when saving the entity state.
Its generally something I like to make very clear, even if attributes are not technically required.

What is the "STJ way" to accomplish this? Also, what is STJ?

sebastianburckhardt: (For now I have just removed this)

davidmrdavid: I think STJ is `System.Text.JSON`

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/Counter.cs
davidmrdavid: does that mean 'delete' is a reserved operation name? May be worth calling that out if so

sebastianburckhardt: It's not exactly 'reserved' it just has a default implemntation already defined.

PR #2633, opened 2023-10-10
Call EnsureLegalAccess from EntityFeature in dotnet-isolated

jviau:
Removes a check that asserts the `TaskOrchestrationContext` was accessed. It does not add much value and instead causes issues:

1. As we add extensions / extra functionality to the `TaskOrchestrationContext`, we have to ensure everything is wrapped to make the `IsAccessed=true` call. This is cumbersome.
2. An orchestration could validly have 0 work off the context and be purely computational based on its inputs, we should not block this.
    1. e.g: an orchestration could follow a branch that based on the input it decides there is 0 work to be done and just returns.

resolves #issue_for_this_pr

### Pull request checklist

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -1,5 +1,7 @@
 # Release Notes
 
+## Microsoft.Azure.Functions.Worker.Extensions.DurableTask v1.1.0-preview.1
+
 ### New Features
 
 - Updates to take advantage of new core-entity support
@@ -8,6 +10,7 @@
 ### Bug Fixes
 
 - Address input issues when using .NET isolated (#2581)[https://github.com/Azure/azure-functions-durable-extension/issues/2581]
+- No longer fail orchestrations which return before accessing the `TaskOrchestrationContext`.
 
 ### Breaking Changes
 
--- a/src/Worker.Extensions.DurableTask/FunctionsOrchestrationContext.cs
+++ b/src/Worker.Extensions.DurableTask/FunctionsOrchestrationContext.cs
@@ -36,6 +36,8 @@ internal sealed partial class FunctionsOrchestrationContext : TaskOrchestrationC
         this.LoggerFactory = functionContext.InstanceServices.GetRequiredService<ILoggerFactory>();
     }
 
+    public bool IsAccessed { get; private set; }
+
     public override TaskName Name => this.innerContext.Name;
 
     public override string InstanceId => this.innerContext.InstanceId;
@@ -53,7 +55,7 @@ internal sealed partial class FunctionsOrchestrationContext : TaskOrchestrationC
 
     public override T GetInput<T>()
     {
-        this.ThrowIfIllegalAccess();
+        this.EnsureLegalAccess();
 
         object? input = this.innerContext.GetInput<object>();
         if (input is T typed)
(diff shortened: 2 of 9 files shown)


cgillum:
Just to make sure I understand, can you provide a concrete example of this?
> an orchestration could follow a branch that based on the input it decides there is 0 work to be done and just returns.

The closest thing I can think of is when an orchestration is waiting on an external event, in which case it _does_ access the context, but what's another scenario you had in mind where the code _doesn't_ access the context?

src/Worker.Extensions.DurableTask/FunctionsOrchestrationContext.cs
```diff
@@ -71,60 +69,51 @@ public override T GetInput<T>()
...
-    }
-
-    /// <summary>
-    /// Throws if accessed by a non-orchestrator thread or marks the current object as accessed successfully.
-    /// </summary>
-    private void EnsureLegalAccess()
     {
         this.ThrowIfIllegalAccess();
-        this.IsAccessed = true;
+        return this.innerContext.WaitForExternalEvent<T>(eventName, cancellationToken);
```
cgillum, line 116: Why wouldn't we throw in this case?

jviau: Throw where? `WaitForExternalEvent` still has the `this.ThrowIfIllegalAccess();` in it.

jviau:
> Just to make sure I understand, can you provide a concrete example of this?
> 
> > an orchestration could follow a branch that based on the input it decides there is 0 work to be done and just returns.
> 
> The closest thing I can think of is when an orchestration is waiting on an external event, in which case it _does_ access the context, but what's another scenario you had in mind where the code _doesn't_ access the context?

I believe there are cases where a customer could reasonably write an orchestration that never accesses a context. I don't have a concrete example, but I could see an orchestration being written where it examines the input and decides there is no work to be done, returning immediately. I think overall this check is unnecessarily restrictive without much benefit.

1. We will still throw based on the thread changing.
2. This error is not even descriptive for why it happens. When I saw the error, I was confused why it was happening.
3. The check does not even work that well, we saw customers with async middleware and it would fail when they accessed the context. This check did not even catch that.

release_notes.md
```diff
@@ -1,5 +1,7 @@
 # Release Notes
 
+## Microsoft.Azure.Functions.Worker.Extensions.DurableTask v1.1.0-preview.1
```
sebastianburckhardt, line 3: Should this be preview.2 or preview.3?

src/Worker.Extensions.DurableTask/FunctionsOrchestrator.cs
```diff
@@ -58,14 +58,6 @@ internal class FunctionsOrchestrator : ITaskOrchestrator
         FunctionsOrchestrationContext orchestrationContext)
     {
         Task orchestratorTask = next(functionContext);
-        if (!orchestratorTask.IsCompleted && !orchestrationContext.IsAccessed)
```
sebastianburckhardt, line 61: I am not entirely sure what problems this was supposed to catch, so I can't really tell if there is something that is no longer being checked. Based on your comments earlier I suppose this check did not have a well-defined function.

sebastianburckhardt (approved):
some minor comments, looks good to me.

release_notes.md
jviau: I have it as `preview.1` because this is most likely what we will release publicly. What we have now is `entities-preview.N`, which is just for internal testing.

jviau:
I added the async check back in, but made a wrapper to `TaskOrchestrationEntityFeature` to call `EnsureLegalAccess`. We can evaluate removing this check separately.

src/Worker.Extensions.DurableTask/FunctionsOrchestrator.cs
jviau: Yeah I want to remove it, but don't want to block this fix on that discussion. So I added it back in, we can discuss that later.

PR #2636, opened 2023-10-12
Remove extra nuget.config

jviau:
Unblock build by removing extra nuget.config and pointing ADO yml to the one at repo root.


test/SmokeTests/OOProcSmokeTests/DotNetIsolated/DotNetIsolated.csproj
```diff
@@ -16,7 +16,7 @@
       This preview package is to enable better NuGet restore behavior with custom feeds.
       TODO: upgrade to 1.16.0 when it's released.
     -->
-    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.16.0-preview1" OutputItemType="Analyzer" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.16.0-preview2" OutputItemType="Analyzer" />
```
jviau, line 19: Secondary change which fixes this project in VisualStudio.

.nuget/NuGet.Config
bachuv, line 1: Can we update this nuget.config file with the correct package sources instead of deleting it? I noticed that the tests are failing since the CI can't find the correct nuget.config file and I believe another .yml file needs to be updated. I think it would be easier to maintain if we update the sources instead of the nuget.config file location.

jviau: Found and removed the additional references to the removed nuget.config

bachuv (approved):

PR #2648, opened 2023-10-18
fix translation of legacy query to new entity query support

sebastianburckhardt:
Fix bug in how entity queries are translated to the new entity query support in `IEntityOrchestrationService`.

### Pull request checklist

tracked with feature branch.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -641,7 +641,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 var result = await entityBackendQueries.QueryEntitiesAsync(
                     new EntityBackendQueries.EntityQuery()
                     {
-                        InstanceIdStartsWith = query.EntityName != null ? $"${query.EntityName}" : null,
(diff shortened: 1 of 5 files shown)


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -641,7 +641,7 @@ async Task<EntityQueryResult> IDurableEntityClient.ListEntitiesAsync(EntityQuery
                 var result = await entityBackendQueries.QueryEntitiesAsync(
                     new EntityBackendQueries.EntityQuery()
                     {
-                        InstanceIdStartsWith = query.EntityName != null ? $"${query.EntityName}" : null,
+                        InstanceIdStartsWith = query.EntityName != null ? $"@{query.EntityName.ToLowerInvariant()}@" : null,
```
jviau, line 644: Will this cause any issues with new durable functions?

sebastianburckhardt: not in this case... `EntityName` is already the entity name only, it was defined a bit differently for that query.

sebastianburckhardt:
This PR now also pulls in the fixes from preview.2 that were merged in https://github.com/Azure/durabletask/pull/999

davidmrdavid (approved):

PR #2653, opened 2023-11-01
HTTP APIs for .NET Isolated Parity

bachuv:
This PR adds HTTP API parity (`CallHttpAsync()`) for .NET Isolated.
 
Changes include:
- New `DurableHttpRequest` and `DurableHttpResponse` types in the Worker project
- New `TaskOrchestrationContext` extension method for `CallHttpAsync()`
- Additional changes in `OutOfProcMiddleware` to execute the `BuiltIn::HttpActivity` activity function by using the existing `TaskHttpActivityShim` (`BuiltIn::HttpActivity` is the reserved name to know when a TaskActivity should be an HTTP activity).

### Pull request checklist


src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -540,6 +581,23 @@ public async Task CallActivityAsync(DispatchMiddlewareContext dispatchContext, F
             dispatchContext.SetProperty(activityResult);
         }
 
+        private static DurableHttpRequest? ConvertDurableHttpRequest(string? inputString)
+        {
+            IList<dynamic>? input = JsonConvert.DeserializeObject<IList<dynamic>>(inputString);
```
bachuv, line 586: I'm wondering if there's a better way to deserialize the `DurableHttpRequest` here since using `dynamic` might not be reliable. I ended up using this since `DurableHttpRequest` is serialized using the System.Text.Json serializer in the worker extension, but deserialized using the Newtonsoft.Json serializer in the webjobs extension, which leads to different json payloads. Let me know if there's a recommended way to go about this.

src/Worker.Extensions.DurableTask/DurableHttpRequest.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.Net.Http;
+using System.Text.Json.Serialization;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Azure.Functions.Worker;
+
+/// <summary>
+/// Request used to make an HTTP call through Durable Functions.
+/// </summary>
+public class DurableHttpRequest
```
bachuv, line 15: Leaving a note to add support for the following properties as well:

- ITokenSource tokenSource
- bool asynchronousPatternEnabled
- TimeSpan? timeout
- HttpRetryOptions httpRetryOptions

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -416,6 +418,45 @@ public async Task CallActivityAsync(DispatchMiddlewareContext dispatchContext, F
                 throw new InvalidOperationException($"An activity was scheduled but no {nameof(TaskScheduledEvent)} was found!");
             }
 
+            if (!string.IsNullOrEmpty(scheduledEvent.Name) && scheduledEvent.Name.StartsWith("BuiltIn::HttpActivity"))
```
jviau, line 421: Instead of manually invoking the activity here, can we test if it starts with `BuiltIn::`, and if so, we can just continue to the next call in the chain:

``` CSharp
if (scheduledEvent.Name?.StartsWith("BuiltIn::", StringComparison.OrdinalIgnoreCase) ?? false)
{
    await next();
    return;
}
```

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
jviau: I recommend writing custom converter for STJ so that it will produce identical serialization payload to what we expect for newtonsoft, then you can serialize from STJ and deserialize directly into Newtonsoft.

src/Worker.Extensions.DurableTask/DurableHttpRequest.cs
jviau: `ITokenSource` is the one I don't know if we will be able to support fully - or if we should. It is a lot more complex because these are two different assemblies and processes, so the types won't line up. We can at least add support for managed identities. But not custom `ITokenSource`.

jviau (changes requested):

src/Worker.Extensions.DurableTask/HttpHeadersConverter.cs
```diff
@@ -0,0 +1,103 @@
...
+using System.Collections.Generic;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Azure.Functions.Worker;
+
+// StringValues does not deserialize as you would expect, so we need a custom mechanism
+// for serializing HTTP header collections
+internal class HttpHeadersConverter : JsonConverter<object>
```
jviau, line 14: Does directly declaring the dictionary here not work?

```suggestion
internal class HttpHeadersConverter : JsonConverter<IDictionary<string, StringValues>>
```

src/Worker.Extensions.DurableTask/DurableHttpResponse.cs
```diff
@@ -0,0 +1,83 @@
...
+    /// </summary>
+    [JsonPropertyName("content")]
+    public string Content { get; }
+
+    /// <summary>
+    /// Creates a DurableHttpResponse from an HttpResponseMessage.
+    /// </summary>
+    /// <param name="httpResponseMessage">HttpResponseMessage returned from the HTTP call.</param>
+    /// <returns>A <see cref="Task{TResult}"/> representing the result of the asynchronous operation.</returns>
+    public static async Task<DurableHttpResponse> CreateDurableHttpResponseWithHttpResponseMessage(HttpResponseMessage httpResponseMessage)
```
jviau, line 60: nit: I do one of two naming methods for these static creation methods: `From{Type*}` (where Type* isn't exact, but a good enough name of what the type is). Or just a `Create`. I tend to prefer the former.

```suggestion
    public static async Task<DurableHttpResponse> FromHttpResponse(HttpResponseMessage httpResponseMessage)
```

src/Worker.Extensions.DurableTask/DurableHttpRequest.cs
bachuv: I see, so you're saying to only add a `ManagedIdentityTokenSource` for .NET Isolated? This would mean that the `CallHttpAsync` API would have different arguments for .NET In proc and Isolated though. Is that not confusing for customers?

In proc:
`CallHttpAsync(.. ITokenSource tokenSource, ..)`

Isolated:
`CallHttpAsync(.. ManagedIdentityTokenSource tokenSource ..)`

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
bachuv: I tested it, but wasn't able to get that working.

src/Worker.Extensions.DurableTask/DurableHttpRequest.cs
jviau: The arguments would need to be different yes, but not exactly how you mentioned. We wouldn't be exposing the direct types in the out of proc API as it is a different process. We would have to design and implement some semantic way to indicate what auth behavior they want to happen in the host process.

bachuv: Ok, I think that's out of scope for this PR then, but we can add it in the future.

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
jviau: Do you know why it didn't work? Were you able to debug into the rest of the pipeline?

src/Worker.Extensions.DurableTask/DurableHttpResponseConverter.cs
```diff
@@ -0,0 +1,138 @@
...
+using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Text;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Azure.Functions.Worker;
+internal class DurableHttpResponseConverter : JsonConverter<DurableHttpResponse>
```
jviau, line 13: Can you explain why a full converter was necessary here? Particularly, why did `JsonPropertyName` attributes not work? I know I mentioned writing your own converter, which may be what needs to happen, but I misspoke and really meant to just customize JSON conversion (through attributes or converter).

I'd recommend using attributes if possible, as they are less code to maintain. You can start with an attribute approach, serialize both the in-proc and out-of-proc types using the respective JSON libraries, and compare them for what you need to change.

src/Worker.Extensions.DurableTask/DurableHttpResponseConverter.cs
```diff
@@ -0,0 +1,138 @@
...
+
+    public override void Write(
+        Utf8JsonWriter writer,
+        DurableHttpResponse response,
+        JsonSerializerOptions options)
+    {
+        writer.WriteStartObject();
+
+        // status code
+        writer.WriteString("status code", response.StatusCode.ToString());
```
jviau, line 106: Note sure if `status code` is intentional. 
```suggestion
        writer.WriteString("statusCode", response.StatusCode.ToString());
```

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
bachuv: When I manually invoked the activity, I was serializing the output from `shim.RunAsync()` before setting it as the result. If I use the code provided, the output is throwing an error when it deserializes - `"One or more errors occurred. (The JSON value could not be converted to System.String. Path: $ | LineNumber: 0 | BytePositionInLine: 1.)",` because I think it's not serialized. Do you know if there's a way to serialize the output before it is returned?

[This line](https://github.com/Azure/azure-functions-durable-extension/blob/9f750ebc77fda1f52c5042dcdb41e6eb52ee7fdb/src/WebJobs.Extensions.DurableTask/ProtobufUtils.cs#L121) is the last place a breakpoint is hit before an exception is thrown.

src/Worker.Extensions.DurableTask/DurableHttpResponseConverter.cs
bachuv: I wrote the full converter based off of your previous comment, but without it I kept running into serialization errors like `Unexpected token while deserializing object: PropertyName. Path '[0].uri', line 1, position 36.` when deserializing the JSON.

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
jviau: I tested this change locally and did not encounter issues. Maybe it had to happen along with JSON changes at the same time? We can sync tomorrow to see what is going on.

src/Worker.Extensions.DurableTask/DurableHttpResponseConverter.cs
jviau: Lets sync tomorrow on this - I was able to get conversion working with just attributes.

src/Worker.Extensions.DurableTask/DurableHttpRequest.cs
```diff
@@ -0,0 +1,83 @@
...
+        Uri uri,
+        IDictionary<string, StringValues>? headers = null,
+        string? content = null,
+        bool asynchronousPatternEnabled = true,
+        TimeSpan? timeout = null,
+        HttpRetryOptions? httpRetryOptions = null)
+    {
+        this.Method = method;
+        this.Uri = uri;
+        this.Headers = HttpHeadersHelper.CreateCopy(headers);
```
jviau, line 32: I wonder what value creating a copy of the headers offers here? We never mutate them ourselves. We can probably forgo this copy and leave it to users to know not to mutate the dictionary they supply if they don't want it to change.

src/Worker.Extensions.DurableTask/HttpRetryOptions.cs
```diff
@@ -0,0 +1,113 @@
...
+using DurableTaskCore = DurableTask.Core;
+
+namespace Microsoft.Azure.Functions.Worker;
+
+/// <summary>
+/// Defines retry policies that can be passed as parameters to various operations.
+/// </summary>
+public class HttpRetryOptions
+{
+    private readonly DurableTaskCore.RetryOptions coreRetryOptions;
```
jviau, line 16: This is not necessary/used in the worker, so it can be removed. This type can be simplified to just have the properties/defaults set directly with the expected defaults.

src/Worker.Extensions.DurableTask/HttpRetryOptions.cs
```diff
@@ -0,0 +1,113 @@
...
+        set { this.coreRetryOptions.MaxNumberOfAttempts = value; }
+    }
+
+    /// <summary>
+    /// Gets or sets the list of status codes upon which the
+    /// retry logic specified by this object shall be triggered.
+    /// If none are provided, all 4xx and 5xx status codes
+    /// will be retried.
+    /// </summary>
+    public IList<HttpStatusCode> StatusCodesToRetry { get; set; } = new List<HttpStatusCode>();
```
jviau, line 102: When adding a mutable collection to a type, I prefer to leave the setter off, so they would append to this during initialization.

src/Worker.Extensions.DurableTask/RetryOptions.cs
jviau, line 1: Don't believe this type is used?

src/Worker.Extensions.DurableTask/HttpRetryOptions.cs
```diff
@@ -0,0 +1,113 @@
...
+
+    internal RetryOptions GetRetryOptions()
```
jviau, line 104: This is not called anywhere, we can remove it.

src/Worker.Extensions.DurableTask/TaskOrchestrationContextExtensionMethods.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    /// <summary>
+    /// Makes an HTTP call using the information in the DurableHttpRequest.
+    /// </summary>
+    /// <param name="context">The task orchestration context.</param>
+    /// <param name="req">The DurableHttpRequest used to make the HTTP call.</param>
+    /// <returns>DurableHttpResponse</returns>
+    public static async Task<DurableHttpResponse> CallHttpAsync(this TaskOrchestrationContext context, DurableHttpRequest req)
+    {
+        string responseString = await context.CallActivityAsync<string>(Constants.HttpTaskActivityReservedName, req);
```
jviau, line 24: Pass the `DurableHttpResponse` directly into the type param instead of `string`.

jviau (changes requested):

src/Worker.Extensions.DurableTask/HTTP/DurableHttpRequest.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// </summary>
+    [JsonPropertyName("headers")]
+    [JsonConverter(typeof(HttpHeadersConverter))]
+    public IDictionary<string, StringValues>? Headers { get; }
+
+    /// <summary>
+    /// Content passed with the HTTP request made by the Durable Function.
+    /// </summary>
+    [JsonPropertyName("content")]
+    public string? Content { get; }
```
jviau, line 62: nit: I typically only put the required properties in the constructor. Something like this can just be a `{ get; set; }`. The only properties that need to be in the constructor are `Method` and `Uri`.

src/Worker.Extensions.DurableTask/HTTP/DurableHttpResponse.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// </summary>
+    [JsonPropertyName("headers")]
+    [JsonConverter(typeof(HttpHeadersConverter))]
+    public IDictionary<string, StringValues>? Headers { get; }
+
+    /// <summary>
+    /// Content returned from an HTTP request.
+    /// </summary>
+    [JsonPropertyName("content")]
+    public string? Content { get; }
```
jviau, line 49: For this one, only `StatusCode` is needed in the ctor. `Headers` and `Content` can be `{ get; init; }`. (Using `init` because want this to be immutable)

src/Worker.Extensions.DurableTask/HTTP/HttpHeadersConverter.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Text.Json.Serialization;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.DurableTask.Http;
+
+// StringValues does not deserialize as you would expect, so we need a custom mechanism
+// for serializing HTTP header collections
+internal class HttpHeadersConverter : JsonConverter<IDictionary<string, StringValues>>
+{
+    public override bool CanConvert(Type objectType)
```
jviau, line 16: No need to override this - the base class will handle this.

src/Worker.Extensions.DurableTask/HTTP/HttpMethodConverter.cs
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public override HttpMethod Read(
+        ref Utf8JsonReader reader,
+        Type objectType,
+        JsonSerializerOptions options)
+    {
+        HttpMethod method = HttpMethod.Get;
+
+        if (reader.TokenType != JsonTokenType.StartObject)
```
jviau, line 25: This doesn't look right. I wouldn't expect a start-object here at all, it is going to be a literal string token. Implementation should be a pure `return new HttpMethod(reader.GetString());`. No need to perform checks ahead. If it is `null`, this won't be called. If this isn't a string, `.GetString()` will throw (I think), which works for us.

src/Worker.Extensions.DurableTask/HTTP/HttpRetryOptions.cs
```diff
@@ -0,0 +1,103 @@
...
+    /// <exception cref="ArgumentException">
+    /// The <paramref name="firstRetryInterval"/> value must be greater than <see cref="TimeSpan.Zero"/>.
+    /// </exception>
+    public HttpRetryOptions(TimeSpan firstRetryInterval, int maxNumberOfAttempts, IList<HttpStatusCode>? statusCodesToRetry = null)
+    {
+        this.MaxRetryInterval = DefaultMaxRetryinterval;
+
+        this.firstRetryInterval = firstRetryInterval;
+        this.maxNumberOfAttempts = maxNumberOfAttempts;
+        this.statusCodesToRetry = statusCodesToRetry ?? new List<HttpStatusCode>();
```
jviau, line 39:
```suggestion
        this.StatusCodesToRetry = statusCodesToRetry ?? new List<HttpStatusCode>();
```

and delete the field `statusCodesToRetry`

src/Worker.Extensions.DurableTask/HTTP/HttpRetryOptions.cs
```diff
@@ -0,0 +1,103 @@
...
+    /// <summary>
+    /// Gets or sets the first retry interval.
+    /// </summary>
+    /// <value>
+    /// The TimeSpan to wait for the first retries.
+    /// </value>
+    [JsonPropertyName("FirstRetryInterval")]
+    public TimeSpan FirstRetryInterval
+    {
+        get { return this.firstRetryInterval; }
```
jviau, line 51: backing field is not necessary. Can be `public TimeSpan FirstRetryInterval { get; set; }`

src/Worker.Extensions.DurableTask/HTTP/HttpRetryOptions.cs
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /// <summary>
+    /// Gets or sets the max retry interval.
+    /// </summary>
+    /// <value>
+    /// The TimeSpan of the max retry interval, defaults to 6 days.
+    /// </value>
+    [JsonPropertyName("MaxRetryInterval")]
+    public TimeSpan MaxRetryInterval { get; set; }
```
jviau, line 62: nit: instead of setting in the ctor, you can inline the default like so:

```suggestion
    public TimeSpan MaxRetryInterval { get; set; } = DefaultMaxRetryinterval;
```

src/Worker.Extensions.DurableTask/HTTP/HttpRetryOptions.cs
```diff
@@ -0,0 +1,103 @@
...
+
+    /// <summary>
+    /// Gets or sets the backoff coefficient.
+    /// </summary>
+    /// <value>
+    /// The backoff coefficient used to determine rate of increase of backoff. Defaults to 1.
+    /// </value>
+    [JsonPropertyName("BackoffCoefficient")]
+    public double BackoffCoefficient { get; set; }
```
jviau, line 71: Need to default this to 1.

```suggestion
    public double BackoffCoefficient { get; set; } = 1;
```

src/Worker.Extensions.DurableTask/HTTP/HttpRetryOptions.cs
```diff
@@ -0,0 +1,103 @@
...
+
+    /// <summary>
+    /// Gets or sets the timeout for retries.
+    /// </summary>
+    /// <value>
+    /// The TimeSpan timeout for retries, defaults to <see cref="TimeSpan.MaxValue"/>.
+    /// </value>
+    [JsonPropertyName("RetryTimeout")]
+    public TimeSpan RetryTimeout { get; set; }
```
jviau, line 80: Also need to set the default:
```suggestion
    public TimeSpan RetryTimeout { get; set; } = TimeSpan.MaxValue;
```

src/Worker.Extensions.DurableTask/HTTP/HttpRetryOptions.cs
```diff
@@ -0,0 +1,103 @@
...
+
+    /// <summary>
+    /// Gets or sets the max number of attempts.
+    /// </summary>
+    /// <value>
+    /// The maximum number of retry attempts.
+    /// </value>
+    [JsonPropertyName("MaxNumberOfAttempts")]
+    public int MaxNumberOfAttempts
```
jviau, line 89: backing field also not needed:

```suggestion
    public int MaxNumberOfAttempts { get; set; }
```

src/Worker.Extensions.DurableTask/TaskOrchestrationContextExtensionMethods.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    /// <summary>
+    /// Makes an HTTP call using the information in the DurableHttpRequest.
+    /// </summary>
+    /// <param name="context">The task orchestration context.</param>
+    /// <param name="request">The DurableHttpRequest used to make the HTTP call.</param>
+    /// <returns>DurableHttpResponse</returns>
+    public static async Task<DurableHttpResponse> CallHttpAsync(this TaskOrchestrationContext context, DurableHttpRequest request)
+    {
+        DurableHttpResponse response = await context.CallActivityAsync<DurableHttpResponse>(Constants.HttpTaskActivityReservedName, request);
```
jviau, line 25: nit: arg validation, can remove `async`.

``` CSharp
public static Task<DurableHttpResponse> CallHttpAsync(this TaskOrchestrationContext context, DurableHttpRequest request)
{
    if (context is null)
    {
    }

    return context.CallActivityAsync<DurableHttpResponse>(Constants.HttpTaskActivityReservedName, request);
}
```

src/Worker.Extensions.DurableTask/TaskOrchestrationContextExtensionMethods.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// <summary>
+    /// Makes an HTTP call to the specified uri.
+    /// </summary>
+    /// <param name="context">The task orchestration context.</param>
+    /// <param name="method">HttpMethod used for api call.</param>
+    /// <param name="uri">uri used to make the HTTP call.</param>
+    /// <param name="content">Content passed in the HTTP request.</param>
+    /// <param name="retryOptions">The retry option for the HTTP task.</param>
+    /// <returns>A <see cref="Task{DurableHttpResponse}"/>Result of the HTTP call.</returns>
+    public static async Task<DurableHttpResponse?> CallHttpAsync(this TaskOrchestrationContext context, HttpMethod method, Uri uri, string? content = null, HttpRetryOptions? retryOptions = null)
```
jviau, line 39: nit: call the other overload for consistency:

``` CSharp
public static Task<DurableHttpResponse?> CallHttpAsync(this TaskOrchestrationContext context, HttpMethod method, Uri uri, string? content = null, HttpRetryOptions? retryOptions = null)
{
    DurableHttpRequest request = new DurableHttpRequest(
            method: method,
            uri: uri,
            content: content,
            httpRetryOptions: retryOptions);
   return context.CallHttpAsync(request);
}
```

src/Worker.Extensions.DurableTask/Worker.Extensions.DurableTask.csproj
```diff
@@ -54,5 +54,8 @@
       <PackagePath>content/SBOM</PackagePath>
     </Content>
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="HTTP\" />
```
jviau, line 58: this can be removed.

jviau (changes requested):

src/Worker.Extensions.DurableTask/HTTP/DurableHttpRequest.cs
```diff
@@ -19,20 +19,10 @@ public class DurableHttpRequest
     /// </summary>
     public DurableHttpRequest(
```
jviau, line 20: nit: can inline params now
```suggestion
    public DurableHttpRequest(HttpMethod method, Uri uri)
```

src/Worker.Extensions.DurableTask/HTTP/DurableHttpResponse.cs
```diff
@@ -17,16 +17,10 @@ public class DurableHttpResponse
     /// Initializes a new instance of the <see cref="DurableHttpResponse"/> class.
     /// </summary>
     /// <param name="statusCode">HTTP Status code returned from the HTTP call.</param>
-    /// <param name="headers">Headers returned from the HTTP call.</param>
-    /// <param name="content">Content returned from the HTTP call.</param>
     public DurableHttpResponse(
```
jviau, line 20: nit: can inline param now:
```suggestion
    public DurableHttpResponse(HtttpStatusCode statusCode)
```

src/Worker.Extensions.DurableTask/TaskOrchestrationContextExtensionMethods.cs
```diff
@@ -36,16 +39,13 @@ public static async Task<DurableHttpResponse> CallHttpAsync(this TaskOrchestrati
...
     /// <returns>A <see cref="Task{DurableHttpResponse}"/>Result of the HTTP call.</returns>
-    public static async Task<DurableHttpResponse?> CallHttpAsync(this TaskOrchestrationContext context, HttpMethod method, Uri uri, string? content = null, HttpRetryOptions? retryOptions = null)
+    public static Task<DurableHttpResponse> CallHttpAsync(this TaskOrchestrationContext context, HttpMethod method, Uri uri, string? content = null, HttpRetryOptions? retryOptions = null)
     {
-        DurableHttpRequest request = new DurableHttpRequest(
-            method: method,
-            uri: uri,
-            content: content,
-            httpRetryOptions: retryOptions);
+        DurableHttpRequest request = new DurableHttpRequest(method, uri);
```
jviau, line 44: style nit: use object initializer

```suggestion
        DurableHttpRequest request = new DurableHttpRequest(method, uri)
        {
            Content = content,
            HttpRetryOptions = retryOptions,
        };
```

jviau (approved):

PR #2694, opened 2023-12-11
Proposed: Zip .NET in-proc perf tests to allow for automated perf tests

davidmrdavid:
**Please see the internal thread on this PR for more context**

This PR adds two steps to our release pipeline:
(1) It creates a zip-deployable payload of our .NET in-proc perf tests
(2) It moves that .zip to the directory where our ADO publishing payload is located.

With these two changes, ADO should be able to pick up our .NET in-proc perf tests and consume them in an automated performance testing runner. Doing this should reduce the friction in our release process.

PR #2701, opened 2023-12-21
Add a nightly release pipeline

davidmrdavid:
This is a follow up to: https://github.com/Azure/azure-functions-durable-extension/pull/2694

In this PR, I make a copy of the "Release Extension 2.x" pipeline that triggers **nightly** **iff** there's any new code merged in `dev`. If we incorporate this pipeline, we should be able to create automation that runs e2e perf tests on the same night as new code is merged into the `dev` branch. That should help us catch performance regressions faster.

Additionally, this also paves the way for potentially created a nightly package on ADO, which may have its own servicing wins. That said, this PR is primarily for E2E testing automation.

### DIFF
--- a/azure-pipelines-release-nightly.yml
+++ b/azure-pipelines-release-nightly.yml
@@ -7,7 +7,7 @@ schedules:
   displayName: Daily midnight build
   branches:
     include:
-    - dev
+    - dajusto/add-nightly-release
 
 trigger: none
 pr: none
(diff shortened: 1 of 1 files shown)


cgillum (approved):
I wonder if maybe we don't need the package signing steps since I assume these aren't "official" releases, but either way this looks good to me!

davidmrdavid:
@cgillum: I could see arguments going both ways. I'd say let's sign it just since it's automated anyways. Thanks for the quick review :)

PR #2718, opened 2024-01-10
Add CreateCheckStatusResponseAsync APIs

### LINKED ISSUE #2717: Isolated Process with DurableClient and ASP.NET Core integration will throw exception on CreateCheckStatusResponse
**Description**
This issue is re-raised here as result of discussion in [azure-functions-durable-extension issue: 2683](https://github.com/Azure/azure-functions-durable-extension/issues/2683)

Packages used:
- [Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore v1.1.1](https://www.nuget.org/packages/Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore/1.1.1)
- [Microsoft.Azure.Functions.Worker.Extensions.DurableTask v1.1.0](https://www.nuget.org/packages/Microsoft.Azure.Functions.Worker.Extensions.DurableTask/1.1.0)

When using these versions of the functions packages that provides ASP.NET Core integration and DurableTaskClient, or newer, the method [CreateCheckStatusResponse](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs) will throw an InvalidOperationException because the implementation writes to the `HttpResponseData` through serialization synchronous.

This point has been raised in the (azure-functions-durable-extension repo)[https://github.com/Azure/azure-functions-durable-extension/issues/2683], and there the following points came to light:
- It is unclear when using Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore package if you should still use a `HttpTrigger` function that accepts and returns `HttpRequestData` \ `HttpResponseData`. This as of v1.1.1 the package seem to provide support through [AspNetCoreHttpRequestData.cs](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/extensions/Worker.Extensions.Http.AspNetCore/src/HttpDataModel/AspNetCoreHttpRequestData.cs) and does seem to play nice for now, but it is not clear if this is recommended
  - If it is not advised can this be added to [asp.net core integration documentation](https://learn.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide#aspnet-core-integration)
- From Durable's teams perspective they are using the type as intended so the issue is re-raised here. So no async version will be introduced.
- Not clear whom/where a async version of CreateCheckStatusResponse has to be made for AspNetCore integration

All that I know is that to solve the exception is that an extension method version would need to be introduced that performs the serialization asynchronous, that extension method version would than be taken changes should the non-asynchronous version be updated and to avoid the work around.

Reference:
https://github.com/Azure/azure-functions-durable-extension/issues/2683

**Expected behavior**
When using ASP.NET Core integration and the durable task extension method there should be no exception thrown, however as synchronous writing is not allowed since ASP.NET Core 3+ a Async version will be needed to be exposed.

**Actual behavior**
System.InvalidOperationException: Synchronous operations are disallowed. Call WriteAsync or set AllowSynchronousIO to true instead
```
Result: An unhandled exception occurred.
Exception: System.AggregateException: One or more errors occurred. (Synchronous operations are disallowed. Call WriteAsync or set AllowSynchronousIO to true instead.)
  ---> System.InvalidOperationException: Synchronous operations are disallowed. Call WriteAsync or set AllowSynchronousIO to true instead.
    at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpResponseStream.Write(Byte[] buffer, Int32 offset, Int32 count)
    at Azure.Core.Serialization.JsonObjectSerializer.Serialize(Stream stream, Object value, Type inputType, CancellationToken cancellationToken)
    at Microsoft.Azure.Functions.Worker.DurableTaskClientExtensions.CreateCheckStatusResponse(DurableTaskClient client, HttpRequestData request, String instanceId, HttpStatusCode statusCode, CancellationToken cancellation) in /_/src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs:line 97
    at Microsoft.Azure.Functions.Worker.DurableTaskClientExtensions.CreateCheckStatusResponse(DurableTaskClient client, HttpRequestData request, String instanceId, CancellationToken cancellation) in /_/src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs:line 34
```

```csharp
[Function("SimpleHttpTriggerHttpData")]
public async Task<HttpResponseData> Run(
[HttpTrigger(AuthorizationLevel.Anonymous, "get")] HttpRequestData req,
[DurableClient] DurableTaskClient durableClient)
{
       return durableClient.CreateCheckStatusResponse(req, "FakeId"); //Will result in exception
}
```

**Known Workarounds**
One option is to copy the `CreateCheckStatusResponse` into the consumers project and make a asynchronous version, however this does introduce having to keep an eye on the original code every time the package is updated in case of changes.
It does need modification to use reflection as the `FunctionsDurableTaskClient` needs to be accessed to get query parameters, thus not nice.

Another option from that ASP.NET Core provided for those updating in the past was to apply the following configuration which will allow calling synchronous writes to the HttpRequest. However this would be rather avoided:
```
services.Configure<KestrelServerOptions>(options =>
{
    options.AllowSynchronousIO = true;
});
```

**Addition requested**
When using ASP.NET Core integration it would be nice to also have a `CreateCheckStatusResponse` version that accepts `HttpRequest` and returns an `ActionResult` to keep it more consistent with the ASP.NET Core integration HttpTriggers. However that would be just a nice to have.

jviau:
With AspNetCore support via extension in dotnet isolated, customers are unable to use our `CreateCheckStatusResponse` APIs due to synchronous IO not being allowed by default. While we don't want to add direct AspNetCore support at this time (would require bringing in that extension package, which is a no-go for us without our own separate package). We can in the short term add `CreateCheckStatusResponseAsync` which will fix the no-sync IO issue.

resolves #2717

### Pull request checklist

    * [x] Otherwise: This change only applies to Durable Functions v2.x and **will not** be merged to branch v3.x.

### DIFF
--- a/src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
+++ b/src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Net;
 using System.Threading;
+using System.Threading.Tasks;
 using Azure.Core.Serialization;
 using Microsoft.Azure.Functions.Worker.Http;
 using Microsoft.DurableTask.Client;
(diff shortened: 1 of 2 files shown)


jviau:
Will update release notes before checking in (just waiting on another PR to avoid a merge conflict)

PR #2745, opened 2024-02-16
Patch deserialization bug in 2.13.1

davidmrdavid:
As in title. Also adds de-/serialization tests in the smoke tests thanks to @bachuv

### DIFF
--- a/test/SmokeTests/OOProcSmokeTests/durableJS/DurableFunctionsOrchestratorJS/index.js
+++ b/test/SmokeTests/OOProcSmokeTests/durableJS/DurableFunctionsOrchestratorJS/index.js
@@ -19,7 +19,7 @@ module.exports = df.orchestrator(function* (context) {
     outputs.push(yield context.df.callActivity("Hello", "Tokyo"));
     outputs.push(yield context.df.callActivity("Hello", "Seattle"));
     outputs.push(yield context.df.callActivity("Hello", "London"));
-    outputs.push(yield context.df.callActivity("Hello", "123"));
+    outputs.push(yield context.df.callActivity("Hello", 123));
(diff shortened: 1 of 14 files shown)


cgillum (approved):
Assuming we've done the manual validation of the fix and things are working as expected, LGTM.

test/SmokeTests/OOProcSmokeTests/durableJS/DurableFunctionsOrchestratorJS/index.js
```diff
@@ -13,12 +13,15 @@ const df = require("durable-functions");
 
 module.exports = df.orchestrator(function* (context) {
     const outputs = [];
+    const city = {city:"Paris", country:"France"};
 
     // Replace "Hello" with the name of your Durable Activity Function.
     outputs.push(yield context.df.callActivity("Hello", "Tokyo"));
     outputs.push(yield context.df.callActivity("Hello", "Seattle"));
     outputs.push(yield context.df.callActivity("Hello", "London"));
+    outputs.push(yield context.df.callActivity("Hello", "123"));
```
davidmrdavid, line 22: nit: can we add a case with numbers as input?

davidmrdavid (commented):
LGTM as well. Just left a tiny nitpick for the JS case

test/SmokeTests/OOProcSmokeTests/durableJS/DurableFunctionsOrchestratorJS/index.js
davidmrdavid: This is testing a string with numbers, which is a good edgfe case, but one with "real" ints would be good

bachuv: Yes, I didn't mean to make it a string. I updated it to `123`.

Issue #2743, opened 2024-02-13
Durable Entities stay locked when critical section is left via exception


ranasch:
### Description
Running .net8 isolated durable function which has a critical section guarded by a durable entity. When the orchestration fails, the entity keeps locked and no further orchestrations can proceed.

### Expected behavior
The the critical section (using) is left via exception, the build in dispose should release the lock of the orchestration on the entity, otherwise it stays locked forever and no other orchestration can obtain a lock.

### Actual behavior
When the critical section is left via exception, the lock is not released on the entity, because the release is not persisted on the Azure Table.

### Relevant source code snippets
```csharp
public static class SayHelloOrchestration
{
    [Function(nameof(SayHelloOrchestration))]
    public static async Task<List<string>> RunOrchestrator([OrchestrationTrigger] TaskOrchestrationContext context)
    {
        var outputs = new List<string>();
        ILogger logger = context.CreateReplaySafeLogger(nameof(SayHelloOrchestration));

        logger.LogInformation($"Saying hello ({context.InstanceId})");

        await using (var handle = await context.AcquireResourceLockAsync(Guid.Empty.ToString()))
        {
            outputs.Add(await context.CallActivityAsync<string>(nameof(SayHelloActivity), $"Brisbane ({context.InstanceId})"));
            throw new Exception("Test");
            outputs.Add(await context.CallActivityAsync<string>(nameof(SayHelloActivity), $"Sydney ({context.InstanceId})"));
        }
        logger.LogInformation($"Saying goodbye ({context.InstanceId})");
        return outputs;
    }
}

```
The created entity will stay be locked by the critical section, unless you also implement a custom Dispose, which by calling the Entity a final time ensures the unlock state is persisted in the table storage:

```csharp
    public static class ResourceLockExtensions
    {
        public static async Task<IAsyncDisposable> AcquireResourceLockAsync(this TaskOrchestrationContext context, string instanceKey)
        {
            var entityId = new EntityInstanceId(nameof(ResourceLockEntity), instanceKey);
            var entityHandle = await context.Entities.LockEntitiesAsync(entityId);

            return new ResourceLockHandle(context, entityId, entityHandle);
        }
    }

    public sealed class ResourceLockHandle : IAsyncDisposable
    {
        private readonly TaskOrchestrationContext _context;
        private readonly EntityInstanceId _entityId;
        private readonly IAsyncDisposable _entityHandle;

[...]

PR #2748, opened 2024-02-20
Correctly propagate orchestration results of a failed orchestration

sebastianburckhardt:
As diagnosed in #2743, the current code in `OutOfProcMiddleware.CallOrchestratorAsync` does not correctly propagate the results of a failed orchestration to the backend.  

The problem is that when an orchestrator function fails, the OrchestratorExecutionResult that was constructed by the executor is ignored, and instead an alternate OrchestratorExecutionResult.ForFailure is constructed from the exception returned by the function result. This is a problem because 

1.  the original failure details are lost instead of being persisted in the history.
2.  any orchestrator actions that took place prior to the failure (e.g. sending lock release messages) are ignored.

This PR fixes the problem by propagating the original `OrchestratorExecutionResult` that describes the failure (if there is one).

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteOrchestratorContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteOrchestratorContext.cs
@@ -64,6 +64,20 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             return this.executionResult ?? throw new InvalidOperationException($"The execution result has not yet been set using {nameof(this.SetResult)}.");
         }
 
+        internal bool TryGetOrchestrationErrorDetails(out string details)
+        {
+            if (this.failure != null)
+            {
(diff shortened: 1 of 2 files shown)


lilyjma:
@cgillum / @jviau - requested for your review because there are some customers asking about this. There's a DF release starting next week, so it'd be great if this fix could go out with that.

cgillum (approved):
LGTM.

I'm working on a process for creating end-to-end tests. Is there an orchestration that you used for testing this fix which we can borrow and include as part of an end-to-end test? (Creating such a test can be done separately from this PR - I'm just asking now in case I forget later)

sebastianburckhardt:
> Is there an orchestration that you used for testing this fix which we can borrow and include as part of an end-to-end test?

Yes. All the tests are in here: https://github.com/Azure/azure-functions-durable-extension/pull/2612
The specific test that would have detected this problem is called `FaultyCriticalSection`.

PR #2753, opened 2024-03-05
Add GitHub Actions improvements: allow actions to be triggered manually, implement (best effort) fail fast, and get MSSQL working again

davidmrdavid:
This PR incorporates a series of GitHub Actions improvements:
(1) It adds a `workflow_dispatch:` trigger for our actions, allowing them to be triggered through the UI, without needing a PR. See: https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow
(2) There's a known error in the `pwsh` GitHub Actions runners that prevents scripts from failing fast on an error. This makes debugging difficult because it's hard to find the first error on a pipeline. I added a manual `Exit-OnError` step in our scripts that exists manually whenever a step fails.
(3) The MSSQL smoke tests stopped working, apparently due to the "additionalRunFlags" set of dynamic parameters. We weren't really making use of this as far as I can tell, so I removed it from the script and now it is working again.

### DIFF
--- a/.github/workflows/smoketest-dotnet-isolated-v4.yml
+++ b/.github/workflows/smoketest-dotnet-isolated-v4.yml
@@ -1,6 +1,7 @@
 name: Smoke Test - .NET Isolated on Functions V4
 
 on:
+  workflow_dispatch:
   push:
     branches: [ main, dev ]
     paths:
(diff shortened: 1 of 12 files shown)


bachuv (approved):
LGTM!

PR #2779, opened 2024-04-01
Log caught exception in LocalGrpcListener's CreateInstance method

davidmrdavid:
In the LocalGrpcListener's CreateInstance method, we catch an `InvalidException` and disregard it's contents. This PR ensures we log it's contents so we can better debug exceptions in this code path.

resolves N/A

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
@@ -167,7 +167,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                         functionName: request.Name,
                         instanceId: request.InstanceId,
                         message: $"Failed to start instanceId {request.InstanceId} due to internal exception.\n Exception trace: {ex}.");
-                    throw new RpcException(new Status(StatusCode.Internal, $"Failed to start instance with ID {request.InstanceId}. Does this instance ID already exist?.\n"));
+                    throw new RpcException(new Status(StatusCode.Internal, $"Failed to start instance with ID {request.InstanceId}.\nInner Exception message: {ex.Message}."));
                 }
             }
 
--- a/src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
@@ -155,7 +155,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                         InstanceId = instanceId,
                     };
                 }
-                catch (OrchestrationAlreadyExistsException ex)
+                catch (OrchestrationAlreadyExistsException)
                 {
                     throw new RpcException(new Status(StatusCode.AlreadyExists, $"An Orchestration instance with the ID {request.InstanceId} already exists."));
                 }
--- a/src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
@@ -10,6 +10,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using DurableTask.Core;
 using DurableTask.Core.Entities;
+using DurableTask.Core.Exceptions;
 using DurableTask.Core.History;
 using DurableTask.Core.Query;
 using DurableTask.Core.Serializing.Internal;
--- a/src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
@@ -155,11 +155,18 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                         InstanceId = instanceId,
                     };
                 }
-                catch (InvalidOperationException ex)
+                catch (OrchestrationAlreadyExistsException ex)
                 {
-                    string errorMessage = $"Failed to start instance with ID {request.InstanceId}. Does this instance ID already exist?.\n" +
-                        $"Inner exception: {ex}";
-                    throw new RpcException(new Status(StatusCode.AlreadyExists, errorMessage));
+                    throw new RpcException(new Status(StatusCode.AlreadyExists, $"An Orchestration instance with the ID {request.InstanceId} already exists."));
+                }
+                catch (Exception ex)
+                {
+                    this.extension.TraceHelper.ExtensionWarningEvent(
+                        this.extension.Options.HubName,
+                        functionName: request.Name,
+                        instanceId: request.InstanceId,
+                        message: $"Failed to start instanceId {request.InstanceId} due to internal exception.\n Exception trace: {ex}.");
+                    throw new RpcException(new Status(StatusCode.Internal, $"Failed to start instance with ID {request.InstanceId}. Does this instance ID already exist?.\n"));
                 }
             }
 
--- a/src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
@@ -155,9 +155,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                         InstanceId = instanceId,
                     };
                 }
-                catch (InvalidOperationException)
+                catch (InvalidOperationException ex)
                 {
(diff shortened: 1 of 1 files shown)


src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -155,9 +155,11 @@ public async override Task<P.CreateInstanceResponse> StartInstance(P.CreateInsta
                         InstanceId = instanceId,
                     };
                 }
-                catch (InvalidOperationException)
+                catch (InvalidOperationException ex)
                 {
-                    throw new RpcException(new Status(StatusCode.AlreadyExists, $"An Orchestration instance with the ID {request.InstanceId} already exists."));
+                    string errorMessage = $"Failed to start instance with ID {request.InstanceId}. Does this instance ID already exist?.\n" +
+                        $"Inner exception: {ex}";
+                    throw new RpcException(new Status(StatusCode.AlreadyExists, errorMessage));
```
davidmrdavid, line 162: I couldn't do the usual way of creating an innerException (adding it as a constructor param) because `RpcException` does not support it. So I had to add it in the message itself.

jviau: We shouldn't be including an inner exception here. `RpcException` becomes the response to the gRPC call. Including the entire inner exception stack trace here does not sound wise.

Instead, I recommend adding an explicit `ILogger.LogDebug` if this is really important.

cgillum: Shouldn't it be a warning?

davidmrdavid:
> I imagine this is due to customer complaints of seeing RpcException when making client calls. 

@jviau - this came from IcM 487112392, where a customer reported thair their application fails with "An Orchestration instance with the ID (GUID) already exists" even though the ID is brand new.  Debugging this is turning out to be tricky, and we concluded that difficulty would have been eased had we preserved the `InvalidOperationException` info, instead of just discarding it and throwing this RpcException. So this PR is trying to do just that, but I see your point about proper `RpcException` usage. Let me think about that.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
davidmrdavid: taking it one step further - shouldn't we use `this.extension.TraceHelper.ExtensionWarningEvent` so it propagates to Kusto?

cgillum: Yes, we should be using `TraceHelper` instead of `ILogger` to ensure that this gets into Kusto.

cgillum:
> > I imagine this is due to customer complaints of seeing RpcException when making client calls.
> 
> @jviau - this came from IcM 487112392, where a customer reported thair their application fails with "An Orchestration instance with the ID (GUID) already exists" even though the ID is brand new. Debugging this is turning out to be tricky, and we concluded that difficulty would have been eased had we preserved the `InvalidOperationException` info, instead of just discarding it and throwing this RpcException. So this PR is trying to do just that, but I see your point about proper `RpcException` usage. Let me think about that.

@davidmrdavid I think what @jviau is suggesting doesn't need to change anything you're doing in this PR. It's a small user experience improvement we can make directly in `microsoft/durabletask-dotnet` repo, independent of this PR.

jviau:
Do you think there is some other cause that throws the invalid exception? We could look through our backends and see if we can determine any other cases it could throw.

cgillum:
Yes, @sebastianburckhardt suggested that there might be some other `InvalidOperationException`s that are being wrongly translated. This PR will help us confirm this since we'll be logging the exception details for the first time.

sebastianburckhardt:
yeah, for example this one:
```
throw new InvalidOperationException($"Netherite backend failed to start: {this.startupException.Message}", this.startupException);
```

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
jviau: Be careful about logging warnings here, as some of the `InvalidOperationExceptions` are expected behavior, so it will throw customers off to see those as warnings.

Can we take a look into the backends and see exactly what exception they throw on conflict and try to narrow down our `ALREADY_EXISTS` check to that?

jviau:
My thoughts on this:

1. Lets spend some time examining the 3 backends and see (1) do they support dedupe? (2) if so, what is the exact exception they throw? (3) narrow down the catch for that exact exception if possible (maybe even check the message).
2. If after step 1 we still want the log, we should (1) **not** include the exception in the `RpcException`, this is too large of a message to send across. (2) add it to a log statement but be careful about verbosity. If we start emitting warnings for by-design dedupe exceptions and detectors go off, this will cause customers to think something is broken when it isn't (this has happened before in other non-durable scenarios).

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -155,10 +156,19 @@ public async override Task<P.CreateInstanceResponse> StartInstance(P.CreateInsta
                         InstanceId = instanceId,
                     };
                 }
-                catch (InvalidOperationException)
+                catch (OrchestrationAlreadyExistsException)
                 {
                     throw new RpcException(new Status(StatusCode.AlreadyExists, $"An Orchestration instance with the ID {request.InstanceId} already exists."));
                 }
+                catch (Exception ex)
```
davidmrdavid, line 163: I feel like the following would be ideal
```suggestion
                catch (OrchestrationAlreadyExistsException)
                {
                    throw new RpcException(new Status(StatusCode.AlreadyExists, $"An Orchestration instance with the ID {request.InstanceId} already exists."));
                }
                ```suggestion
                catch (InvalidOperationException ex) when (ex.Message.EndsWith("already exists.")) // for older versions of DTF.AS and DTFx.Netherite
                {
                    
                    throw new RpcException(new Status(StatusCode.AlreadyExists, $"An Orchestration instance with the ID {request.InstanceId} already exists."));
                }
                catch (Exception ex)
```

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -155,10 +156,19 @@ public async override Task<P.CreateInstanceResponse> StartInstance(P.CreateInsta
...
+                {
+                    this.extension.TraceHelper.ExtensionWarningEvent(
+                        this.extension.Options.HubName,
+                        functionName: request.Name,
+                        instanceId: request.InstanceId,
+                        message: $"Failed to start instanceId {request.InstanceId} due to internal exception.\n Exception trace: {ex}.");
+                    throw new RpcException(new Status(StatusCode.Internal, $"Failed to start instance with ID {request.InstanceId}. Does this instance ID already exist?.\n"));
```
cgillum, line 170: I think "Does this instance ID exist?" might be problematic because it could lead readers into making wrong assumptions when investigating the issue. I suggest we leave that out. However, it might be worth including `ex.Message` in this exception so that users can more easily get a hint about what's wrong without needing to search through Functions logs.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
cgillum: Maybe, though it won't ever be used for DT.AzureStorage if we release these changes together. It would really only be applicable for old versions of Netherite.

davidmrdavid: Oh, right - we control the DF.AS version. That mitgates this somewhat.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
davidmrdavid: Good point. Incorporated.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
davidmrdavid: @cgillum: do you have any strong reason _not_ to incorporate this? Even if it just affects Netherite, that seems sufficient reason to me.

davidmrdavid: @cgillum: In that case I added it :-). Thanks!

cgillum (approved):
LGTM!

jviau (approved):

Issue #2771, opened 2024-03-25
Potential memory leak when using DefaultAzureCredential for Storage Account


jeromekol:
### Description
We use Durable Entities to manage state for one of our applications. The flow looks like this: IoT Hub => Azure Functions => Update Durable entity. We consume around 200 messages per second. Recently we moved from in-process Azure Functions with .NET 6 to out-of-process Azure Functions running .NET 8. Managed Identity is used to connect to the Storage Account for Durable Entities, by configuring the appsetting according to [documentation](https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference?tabs=blob&pivots=programming-language-csharp#connecting-to-host-storage-with-an-identity). We did not have any issues with .NET 6 and in-process model. However, when we moved to .NET 8 we saw a steady rise of memory (up to 5gb) until the function was eventually restarted by Azure and freeing the memory. We were able to solve that issue by changing from identity-based connection to connection string (incl. storage access key).

### Expected behavior
No steady rise of memory usage.

### Actual behavior
Once the function starts it takes up to one day until the majority of the memory is used up and until the function gets unresponsive. That eventually triggers a self-heal action by the Azure infrastructure and restarts the function.

### Relevant source code snippets
no snippets, just configuration on appsettings.

### Known workarounds
We use Storage Connection strings for the time being.

### App Details

- **Durable Functions extension version (e.g. v1.8.3)**: 1.1.1
- **Azure Functions runtime version (1.0 or 2.0)**: 4
- **Programming language used**: c# (dotnet-isolated)

### Screenshots
> If applicable, add screenshots to help explain your problem.
Memory before and after change (only configuration changes)
![memory_leak_durable_func](https://github.com/Azure/azure-functions-durable-extension/assets/35995674/389aedeb-fbe3-4cde-a3c1-8b80c1ad3014)

Screenshot from WinDbg after analyzing a memory dump:
![windbg_mem_durable_func](https://github.com/Azure/azure-functions-durable-extension/assets/35995674/f235450c-e57c-48e9-b629-6c596795b6dc)

### If deployed to Azure
> We have access to a lot of telemetry that can help with investigations. Please provide as much of the following information as you can to help us investigate!

- **Timeframe issue observed**:
- **Function App name**:
- **Function name(s)**:
- **Azure region**: 
- **Orchestration instance ID(s)**: d16b2edbfc7f42f28b93d17bc64bdca2 (that is an executionId, not sure if it is the one needed)
[...]

PR #2782, opened 2024-04-05
Suppress execution context from token renewal

jviau:
Suppress execution context from the background token renewal operations. The renewal uses the logger, and so we will capture whatever current log scope is present, leaking that scope into this background thread. This is both a small memory leak and potential object disposed failure (the log scope or its contents may be disposed at any point).

resolves #issue_for_this_pr

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Auth/AzureCredentialFactory.cs
+++ b/src/WebJobs.Extensions.DurableTask/Auth/AzureCredentialFactory.cs
@@ -72,6 +72,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Auth
                 RefreshOffset = tokenRefreshOffset,
             };
 
+            // The token credential will make background callbacks to renew the token.
+            // We suppress async flow to avoid logging scope from being captured as we do not know
+            // where this will be called from first.
+            using AsyncFlowControl flowControl = System.Threading.ExecutionContext.SuppressFlow();
             return new AppAuthTokenCredential(
                 value.Token,
                 (o, t) => this.RenewTokenAsync((TokenRenewalState)o, t),
--- a/release_notes.md
+++ b/release_notes.md
@@ -23,6 +23,8 @@
 
 ### Bug Fixes
 
+- Fix execution context / log scope leak in token renewal (https://github.com/Azure/azure-functions-durable-extension/pull/2782)
+
 ### Breaking Changes
 
 ### Dependency Updates


src/WebJobs.Extensions.DurableTask/Auth/AzureCredentialFactory.cs
```diff
@@ -72,6 +72,10 @@ public AppAuthTokenCredential Create(IConfiguration configuration, TimeSpan toke
                 RefreshOffset = tokenRefreshOffset,
             };
 
+            // The token credential will make background callbacks to renew the token.
+            // We suppress async flow to avoid logging scope from being captured as we do not know
+            // where this will be called from first.
+            using AsyncFlowControl flowControl = System.Threading.ExecutionContext.SuppressFlow();
```
cgillum, line 78: What logging scope is currently being captured, and why does it need to be suppressed? I'm missing a bit of background context on this. Just wondering how confident we feel that this fix will be effective.

jviau: We use `ILogger` in the trace helper here. The log scope of whatever was present before calling `.Create` is captured.

I did manual tests with an `AsyncLocal` and verified before the value set outside is present during renew, and after the change it is no longer present.

cgillum: Makes sense at a high level. The thing I can't wrap my head around is that the thing we captured appears to have a lifetime that's different from this AzureCredentialFactory (a singleton). How is that possible? Shouldn't singletons only capture context that also has a singleton lifecycle? For this particular crash, it appears that a singleton object captured a per-request object, which seems very concerning to me.

jviau:
I wonder if this will address https://github.com/Azure/azure-functions-durable-extension/issues/2771?

src/WebJobs.Extensions.DurableTask/Auth/AzureCredentialFactory.cs
jviau: Yes, a singleton did capture a per-request object! But it didn't capture it through DI (nor at time of construction). It captured it through execution context and at the time/where `.Create` was called. Async locals are very powerful, but also lead to subtle issues like this.

So yes, it is a singleton, but the 'scope' I talk about in the PR is not related to service scope directly.

cgillum: Thanks for the clarification. I think I understand it now. Makes me much more wary of async locals as silent app killers. 😅

cgillum (approved):

PR #2783, opened 2024-04-05
Add MSBuild step to auto-generate the .NET worker's AssemblyInfo based on the webjobs extension's version.

davidmrdavid:
Another step in the effort to minimize our release burden. This adds a "before build" step to the webjobs extension that re-generates the AssemblyInfo file of the `Worker.Extensions.DurableTask` project, so that it is kept up to date with the "Version" tag of the WebJobs extension.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
+++ b/src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
@@ -138,6 +138,9 @@
   </ItemGroup>
 
 
+	<!-- The "AssemblyInfo.cs" file for the .NET isolated worker extension needs to be kept up to date with the
+	WebJobs extension version. This pre-build task re-generates that AssemblyInfo file based on this csproj's
+	Version tag.-->
 	<Target Name="GenerateTextFileOnBuild" BeforeTargets="BeforeBuild">
 		<ItemGroup>
 			<FileContent 
--- a/src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
+++ b/src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
@@ -137,4 +137,22 @@
     </Content>
   </ItemGroup>
 
+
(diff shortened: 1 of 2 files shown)


cgillum:
I think there's a simpler way to do this, which I added to the MSSQL project just the other day. Take a look and see if it applies: https://github.com/microsoft/durabletask-mssql/pull/213

davidmrdavid:
Yep, Jacob just linked it to me - funny timing

davidmrdavid:
@jviau yeah those are fair concerns. Let me provide my 2 cents on them:

> Accidentally release Worker extension pointing to an unreleased WebJobs extension.

In my mind, this is a solvable problem - there should be quality gates in ADO preventing us from doing this. I'm happy to park this particular investment if you think that having those checks in place should be a pre-requisite here. In theory, this should be "as easy as" creating smoke tests with a limited `nuget.config` file that points only to NuGet (no ADO).

> Unable to build dotnet isolated smoke tests project as it will try to nuget-restore an unpublished WebJobs extension.

You got me with this one :-) . 

My take is that - when running smoke tests for PR validation, we should run against with the version of the WebJobs extension currently in the branch (because we _usually_ release the WebJobs extension and the .NET isolated extension + SDK at the same time). But that would require some hackery in the build (maybe we can pack the extension and load it locally, not from any remote package index?).

I can think of various solutions here but first I'd like to check the premise - do you agree that it would be better to have the .NET isolated smoke test use the WebJobs extension built from the repo (so it can test unreleased code), or do you think it should always testing against the latest NuGet payload? I guess an easy answer is - it should do **both**.

jviau:
> I can think of various solutions here but first I'd like to check the premise - do you agree that it would be better to have the .NET isolated smoke test use the WebJobs extension built from the repo (so it can test unreleased code), or do you think it should always testing against the latest NuGet payload? I guess an easy answer is - it should do both.

I think smoke tests should use the WebJobs extension built from the repo. But then have a separate step in release which will verify we have released the correct version to the appropriate feed already (so check ADO when publishing to ADO, check nuget when publishing to nuget).

PR #2792, opened 2024-04-12
Update MSSQL smoke test

bachuv:
This PR removes the `SA_PASSWORD` hard coded environment variable from the MSSQL smoke test. Instead, a password for the SQL server is set as a repository secret.

### Pull request checklist


.github/workflows/smoketest-mssql-inproc-v4.yml
```diff
@@ -17,12 +17,10 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
-    env:
-      SA_PASSWORD: NotASecret!12
```
cgillum, line 21: This environment variable is also needed in the DurableTask.SqlServer code, so I don't think it will be sufficient to only set it for the SQL container. If it's possible to set a `SA_PASSWORD` environment variable on the fly, then that would be ideal. Otherwise, we could potentially just save a random string that we generate once as a `SA_PASSWORD` environment variable for GitHub actions.

bachuv: > This environment variable is also needed in the DurableTask.SqlServer code

Just curious, how is this value (in the extension repo) connected to DurableTask.SqlServer code?

> If it's possible to set a SA_PASSWORD environment variable on the fly, then that would be ideal

Could this be done in the e2e-test.ps1 script or does it need to be done before that?

> Otherwise, we could potentially just save a random string that we generate once as a SA_PASSWORD environment variable for GitHub actions.

Where would we save this?

cgillum: > how is this value (in the extension repo) used in the DurableTask.SqlServer code?

Sorry, I misspoke, it's actually `DurableTask.SqlServer.Tests` where it's used. This is how we're able to build a connection string to talk to the database as part of our integration tests. You can see where it's used in [SharedTestHelpers.cs](https://github.com/microsoft/durabletask-mssql/blob/d6e0361572bf60e72ede663a4d46404add6cbffe/test/DurableTask.SqlServer.Tests/Utils/SharedTestHelpers.cs#L38).

> Could this be done in the e2e-test.ps1 script or does it need to be set before that?

Not sure. It depends on whether environment variable changes from one task/step apply to subsequent tasks/step. I have some doubts about this, but it depends on how GitHub actions internally schedules these tasks/steps.

> Where would we save this?

It looks like we would need to do this in the GitHub Actions configuration, as described here: https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#creating-secrets-for-a-repository. I can make you a repo owner so that you have permissions to do this.

bachuv: > It looks like we would need to do this in the GitHub Actions configuration, as described here: https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#creating-secrets-for-a-repository. I can make you a repo owner so that you have permissions to do this.

I think this would be more straightforward and easier to maintain. Can you make me a repo owner so I can try adding the environment variable? Thanks!

cgillum: You should already be an owner.

bachuv: I added a `SA_PASSWORD` repository secret and that worked.

cgillum (approved):
LGTM!

PR #2799, opened 2024-04-22
Add versioning support(phase1)

skyao:
TBD

resolves #issue_for_this_pr

#### webjobs updates

- IDurableOrchestrationClient

  Add override StartNewAsync() method to set instanceVersion.

- DurableClient

  Update StartNewAsync() method to set instanceVersion correctly to call TaskHubClient.CreateOrchestrationInstanceAsync() method

- LocalGrpcListener

  Update to call IDurableOrchestrationClient.StartNewAsync() method with version from CreateInstanceRequest.

- src\WebJobs.Extensions.DurableTask\Microsoft.Azure.WebJobs.Extensions.DurableTask.xml

  Auto update for new added override StartNewAsync() method with instanceVersion parameter.

#### worker updates

- FunctionsOrchestrationContext

  Implemented new added abstract method to get instance version.

    So that in the orchestration code, we get the instance version by this context:
  
    ```c#
        public static async Task<List<string>> RunOrchestrator(
            [OrchestrationTrigger] TaskOrchestrationContext context)
        {
    			string instanceVersion = context.InstanceVersion;
          ......
        }
    ```

    The implementation of TaskOrchestrationContext above is FunctionsOrchestrationContext.

### dependencies

This PR is depended on following PRs

- https://github.com/microsoft/durabletask-dotnet/pull/295
- https://github.com/Azure/azure-functions-durable-extension/pull/2799

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
@@ -163,6 +163,25 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         Task<string> StartNewAsync<T>(string orchestratorFunctionName, string instanceId, T input);
 
         /// <summary>
+        /// Starts a new instance of the specified orchestrator function.
+        /// </summary>
+        /// <remarks>
+        /// If an orchestration instance with the specified ID already exists, the existing instance
+        /// will be silently replaced by this new instance.
+        /// </remarks>
+        /// <param name="orchestratorFunctionName">The name of the orchestrator function to start.</param>
+        /// <param name="instanceId">The ID to use for the new orchestration instance.</param>
+        /// <param name="instanceVersion">The version to use for the new orchestration instance.</param>
+        /// <param name="input">JSON-serializable input value for the orchestrator function.</param>
+        /// <typeparam name="T">The type of the input value for the orchestrator function.</typeparam>
+        /// <returns>A task that completes when the orchestration is started. The task contains the instance id of the started
+        /// orchestratation instance.</returns>
+        /// <exception cref="ArgumentException">
+        /// The specified function does not exist, is disabled, or is not an orchestrator function.
+        /// </exception>
+        Task<string> StartNewAsync<T>(string orchestratorFunctionName, string instanceId, string instanceVersion, T input);
+
+        /// <summary>
         /// Sends an event notification message to a waiting orchestration instance.
         /// </summary>
         /// <remarks>
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -175,7 +175,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         }
 
         /// <inheritdoc />
-        async Task<string> IDurableOrchestrationClient.StartNewAsync<T>(string orchestratorFunctionName, string instanceId, T input)
+        async Task<string> IDurableOrchestrationClient.StartNewAsync<T>(string orchestratorFunctionName, string instanceId, string instanceVersion, T input)
         {
             if (this.ClientReferencesCurrentApp(this))
             {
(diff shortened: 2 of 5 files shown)


src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs
```diff
@@ -162,6 +162,25 @@ public interface IDurableOrchestrationClient
...
+        /// <param name="instanceId">The ID to use for the new orchestration instance.</param>
+        /// <param name="instanceVersion">The version to use for the new orchestration instance.</param>
+        /// <param name="input">JSON-serializable input value for the orchestrator function.</param>
+        /// <typeparam name="T">The type of the input value for the orchestrator function.</typeparam>
+        /// <returns>A task that completes when the orchestration is started. The task contains the instance id of the started
+        /// orchestratation instance.</returns>
+        /// <exception cref="ArgumentException">
+        /// The specified function does not exist, is disabled, or is not an orchestrator function.
+        /// </exception>
+        Task<string> StartNewAsync<T>(string orchestratorFunctionName, string instanceId, string instanceVersion, T input);
```
cgillum, line 182: Adding a new method without a default implementation to an existing interface is a breaking change, which we can't allow. We will need to find a way to add this overload without introducing a breaking change.

skyao: Another solution is trying to reuse orchestratorFunctionName, we can add version to this parameter with a predefined seperator(using `@` as example, we can make desicion later if we accept this solution) :

orchestratorFunctionName="HelloOrchestrator"
orchestratorFunctionName="HelloOrchestrator@1.2.0"

In the implementation, we can try to split orchestratorFunctionName to get real orchestratorFunctionName and optional version:

orchestratorFunctionName="HelloOrchestrator" --> orchestratorFunctionName="HelloOrchestrator", version=""
 orchestratorFunctionName="HelloOrchestrator@1.2.0" --> orchestratorFunctionName="HelloOrchestrator", version="1.2.0"

This solution should work fine without breaking change, but it's just a bit ugly.

jviau: `@` is already used for entity support and has some existing semantics to it. And if we do go with some special character separator, we would have to consider the impact to existing instance IDs that might already contain that character.

Adding support here is challenging without a breaking change. The simplest route would be a new interface `IDurableOrchestrationClient2` and we add it there.

skyao: we can change `@` to something else like `||`. 

This combine of orchestratorFunctionName and version will only be used to call this StartNewAsync() method:

- Before calling StartNewAsync, there are two separated variables: orchestratorFunctionName and version, we just combine them as "orchestratorFunctionNam||version" to pass them into StartNewAsync() method with one orchestratorFunctionName paremeter.

- In StartNewAsync() method, we will split orchestratorFunctionName paremeter by `||` at first and get back real orchestratorFunctionName and version, then "orchestratorFunctionName||version" won't use any more.

skyao: @cgillum @jviau would you please have a look at this PR and make a decision to make it work without breaking changes.

cgillum: Let's go with @jviau's suggestion, and NOT add support for versioning in the WebJobs APIs. These WebJobs APIs are already moving towards deprecation. Let's instead focus on just the APIs defined in `microsoft/durabletask-dotnet`. If you agree, then we can close this PR.

skyao: I agree. 

Yet This PR is used by me in the updated quickstart to show how versioning works in durable-functions, as well as:

https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-create-first-csharp?pivots=code-editor-vscode#add-functions-to-the-app

Without this PR, I can't how an end-to-end process to do versioning.

cgillum:
This work has been completed in other PRs with an updated design.

PR #2812, opened 2024-05-09
Initial changes for extension channel

jviau:
Draft PR to demonstrate how durable would switch to using the existing gRPC channel provided by the functions worker. This is a very old branch and will have many conflicts / out of date changes that will need to be updated.

This change does move and more and more towards specializing the WebJobs extension for out-of-proc workers. This makes me wonder if there is value in splitting the extension: one for in-proc, one for all others (and common assemblies as needed).

## Known Issues / Remaining Work

- Requires .NET6.0, which from what I understand is problematic for bundles
- Need to register the `IDurableTaskClientProvider`: https://github.com/microsoft/durabletask-dotnet/blob/c65608e41c6e8f841833ba48aaee85c4826e0fc9/src/Client/Core/DependencyInjection/ServiceCollectionExtensions.cs#L21


davidmrdavid:
I worry the .NET6 requirement might be a showstopper, I remember it being so when trying to make similar changes for Java, unless I'm misremembering.

jviau:
> I worry the .NET6 requirement might be a showstopper, I remember it being so when trying to make similar changes for Java, unless I'm misremembering.

If we do this as part of the major version rev then I think it is possible to make it work but will be challenging. Bundles is the biggest question to me: can they support .net6.0 packages?

cgillum:
Bundles v4 supports .NET Core 3.1, so we can't take a dependency on .NET 6+ unless there's also a new major bundles version release.

jviau:
Requesting bundles support .NET6 is a reasonable ask. .netcoreapp3.1 is a well out of SLA, so it is problematic for extensions to be stuck on it due to bundles.

PR #2829, opened 2024-05-23
Override new "base" Terminate API

davidmrdavid:
Follow up to: https://github.com/Azure/azure-functions-durable-extension/pull/2802

The PR above made the `FunctionsDurableTaskClient` override new overloads from `DurableTaskClient` which were causing issues as reported here: https://github.com/microsoft/durabletask-dotnet/issues/282#issuecomment-2114590648

Seems we missed the `TerminateInstanceAsync` API in the PR above, so this PR adds it.

This PR also adds what appears to be the first set of unit tests for the .NET isolated worker extension. In it, we're testing against this particular exception: that the `TerminateInstanceAsync` API can be invoked without triggering exceptions. I tested with and without the code change and confirmed that the test is fails without the change, and passes with it.

**Note:** the new test project I added won't run as part of the CI automatically. I think to do that, I need to create a new ADO pipeline. Given that we're migrating our CI to a different ADO organization, now may not be the right time to add it either, so I'd like to do that in a different PR.

resolves https://github.com/microsoft/durabletask-dotnet/issues/282#issuecomment-2114590648

### Pull request checklist

### DIFF
--- a/src/Worker.Extensions.DurableTask/AssemblyInfo.cs
+++ b/src/Worker.Extensions.DurableTask/AssemblyInfo.cs
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System.Runtime.CompilerServices;
 using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
 
 // TODO: Find a way to generate this dynamically at build-time
 [assembly: ExtensionInformation("Microsoft.Azure.WebJobs.Extensions.DurableTask", "2.13.3")]
+[assembly: InternalsVisibleTo("Worker.Extensions.DurableTask.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100cd1dabd5a893b40e75dc901fe7293db4a3caf9cd4d3e3ed6178d49cd476969abe74a9e0b7f4a0bb15edca48758155d35a4f05e6e852fff1b319d103b39ba04acbadd278c2753627c95e1f6f6582425374b92f51cca3deb0d2aab9de3ecda7753900a31f70a236f163006beefffe282888f85e3c76d1205ec7dfef7fa472a17b1")]
--- a/src/Worker.Extensions.DurableTask/FunctionsDurableTaskClient.cs
+++ b/src/Worker.Extensions.DurableTask/FunctionsDurableTaskClient.cs
@@ -1,7 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.DurableTask;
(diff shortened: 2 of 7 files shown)


src/Worker.Extensions.DurableTask/FunctionsDurableTaskClient.cs
```diff
@@ -85,9 +85,10 @@ public override AsyncPageable<OrchestrationMetadata> GetAllInstancesAsync(Orches
     }
 
     public override Task TerminateInstanceAsync(
-        string instanceId, object? output = null, CancellationToken cancellation = default)
+        string instanceId, TerminateInstanceOptions? options = null, CancellationToken cancellation = default)
     {
-        return this.inner.TerminateInstanceAsync(instanceId, output, cancellation);
+        object? output = options?.Output;
+        return this.inner.TerminateInstanceAsync(instanceId, output: output, cancellation);
```
jviau, line 91: We should be passing the options directly to the inner call:

```suggestion
        return this.inner.TerminateInstanceAsync(instanceId, options, cancellation);
```

jviau (changes requested):

src/Worker.Extensions.DurableTask/AssemblyInfo.cs
```diff
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System.Runtime.CompilerServices;
 using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
 
 // TODO: Find a way to generate this dynamically at build-time
 [assembly: ExtensionInformation("Microsoft.Azure.WebJobs.Extensions.DurableTask", "2.13.3")]
+[assembly: InternalsVisibleTo("Worker.Extensions.DurableTask.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100cd1dabd5a893b40e75dc901fe7293db4a3caf9cd4d3e3ed6178d49cd476969abe74a9e0b7f4a0bb15edca48758155d35a4f05e6e852fff1b319d103b39ba04acbadd278c2753627c95e1f6f6582425374b92f51cca3deb0d2aab9de3ecda7753900a31f70a236f163006beefffe282888f85e3c76d1205ec7dfef7fa472a17b1")]
```
davidmrdavid, line 9: So that internal classes can be tested in the tests project. This is a standard pattern, and we use it for the DF WebJobs Extension tests afaik.

src/Worker.Extensions.DurableTask/HTTP/HttpMethodConverter.cs
```diff
@@ -20,7 +20,8 @@ public override bool CanConvert(Type objectType)
         Type objectType,
         JsonSerializerOptions options)
     {
-        return new HttpMethod(reader.GetString());
+        string readerString = reader.GetString() ?? string.Empty;
+        return new HttpMethod(readerString);
```
davidmrdavid, line 24: There was a small warning that `reader.GetString()` might return `null`. So I added this to make sure the method was well typed.

jviau (approved):

PR #2844, opened 2024-06-04
Minimize telemetry surface area

davidmrdavid:
Under certain conditions (user opting in to tracing raw inputs and outputs), our telemetry may capture sensitive information. For example, through logs of exceptions, inputs, output, etc. T

This PR aims to remove the sensitive information by implementing these guidelines:

(1) User-provided inputs and outputs are **never** part of telemetry as-is. This includes the "reason" label for operations like terminate or rewind, as well as the simpler cases of inputs and outputs to orchestrators and entities.

(2) When application-level exceptions are logged, we only log a sanitized version of them that includes the exception type, and the exception stack trace. The exception **message** is absent, as it may contain sensitive information.

And that's it. To this end, I've refactored many methods in the `EndToEndTracer` class, as well as their callers. In most cases, the refactoring is simply to avoid logging sensitive **parameters** via ETW, but still logging them to the user's Application Insights.  

The more complicated refactorings are all around logging exceptions.  In the easy case - I simply refactored the string parameter containing the exception string into an actual `Exception` type, that I can use to create a sanitized exception string containing only the exception type and stack trace. Other cases where the exception object wasn't available in the immediate caller are handled on a case-by-case basis.

Finally - some diffs in this PR are minor improvements like adding nullable analysis and removing build warnings.

resolves N/A

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
@@ -5,6 +5,7 @@ using System;
 using System.Diagnostics;
 using System.Net;
 using DurableTask.Core.Common;
+using DurableTask.Core.Exceptions;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
@@ -338,12 +339,26 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             string hubName,
             string functionName,
             string instanceId,
-            string reason,
-            string sanitizedReason,
+            Exception exception,
             FunctionType functionType,
             bool isReplay,
             int taskEventId = -1)
         {
+
+            string reason = exception.Message;
+            if (exception is OrchestrationFailureException orchestrationFailureException)
+            {
+                reason = orchestrationFailureException.Details;
+            }
+
+            string sanitizedReason = $"{exception.GetType().FullName}\n{exception.StackTrace}";
+
+            if (isReplay)
+            {
+                reason = $"(replayed {exception.GetType().Name})";
+                sanitizedReason = reason;
+            }
+
             if (this.ShouldLogEvent(isReplay))
             {
                 EtwEventSource.Instance.FunctionFailed(
@@ -363,19 +378,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     instanceId, functionName, functionType, reason, isReplay, FunctionState.Failed, OrchestrationRuntimeStatus.Failed, hubName,
                     LocalAppName, LocalSlotName, ExtensionVersion, this.sequenceNumber++, taskEventId);
             }
-        }
-
-        public void FunctionFailed(
-            string hubName,
-            string functionName,
-            string instanceId,
-            Exception reason,
-            FunctionType functionType,
-            bool isReplay,
-            int taskEventId = -1)
-        {
-            string sanitizedException = $"{reason.GetType().FullName}\n{reason.StackTrace}";
-            this.FunctionFailed(hubName, functionName, instanceId, reason.ToString(), sanitizedException, functionType, isReplay, taskEventId);
         }
 
         public void FunctionAborted(
--- a/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
@@ -59,23 +59,23 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
 #pragma warning disable SA1117 // Parameters should be on same line or separate lines
 
-        private string SanitizeString(string? rawPayload, out string iloggerPayload, bool isReplay = false)
+        private void SanitizeString(string? rawPayload, out string iloggerString, out string durableKustoTableString, bool isReplay = false)
         {
             if (isReplay)
             {
-                iloggerPayload = rawPayload ?? "(replay)";
-                return "(replay)";
+                iloggerString = rawPayload ?? "(replay)";
+                durableKustoTableString = "(replay)";
             }
 
             string payload = rawPayload ?? "(null)";
             int numCharacters = payload.Length;
             string sanitizedPayload = $"(Redacted {numCharacters} characters)";
 
-            iloggerPayload = this.shouldTraceRawData ? payload : sanitizedPayload;
-            return sanitizedPayload;
+            iloggerString = this.shouldTraceRawData ? payload : sanitizedPayload;
+            durableKustoTableString = sanitizedPayload;
         }
 
-        private string SanitizeException(Exception? exception, out string iloggerExceptionString, bool isReplay = false)
+        private void SanitizeException(Exception? exception, out string iloggerExceptionString, out string durableKustoTableString, bool isReplay = false)
         {
             string exceptionString = exception != null ? exception.ToString() : string.Empty;
             if (exception is OrchestrationFailureException orchestrationFailureException)
@@ -88,10 +88,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
             if (isReplay)
             {
-                return "(replay)";
+                durableKustoTableString = "(replay)";
             }
 
-            return sanitizedString;
+            durableKustoTableString = sanitizedString;
         }
 
         public void ExtensionInformationalEvent(
@@ -171,7 +171,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             bool isReplay,
             int taskEventId = -1)
         {
-            string sanitizedInput = this.SanitizeString(input, out string loggerInput, isReplay: isReplay);
+            this.SanitizeString(input, out string loggerInput, out string sanitizedInput, isReplay: isReplay);
 
             if (this.ShouldLogEvent(isReplay))
             {
@@ -259,7 +259,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             bool isReplay,
             int taskEventId = -1)
         {
-            string sanitizedOutput = this.SanitizeString(output, out string loggerOutput, isReplay: isReplay);
+            this.SanitizeString(output, out string loggerOutput, out string sanitizedOutput, isReplay: isReplay);
 
             if (this.ShouldLogEvent(isReplay))
             {
@@ -289,7 +289,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             string instanceId,
             string reason)
         {
-            string sanitizedReason = this.SanitizeString(reason, out string loggerReason);
+            this.SanitizeString(reason, out string loggerReason, out string sanitizedReason);
 
             FunctionType functionType = FunctionType.Orchestrator;
 
@@ -316,7 +316,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             string instanceId,
             string reason)
         {
-            string sanitizedReason = this.SanitizeString(reason, out string loggerReason);
+            this.SanitizeString(reason, out string loggerReason, out string sanitizedReason);
 
             FunctionType functionType = FunctionType.Orchestrator;
 
@@ -343,7 +343,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             string instanceId,
             string reason)
         {
-            string sanitizedReason = this.SanitizeString(reason, out string loggerReason);
+            this.SanitizeString(reason, out string loggerReason, out string sanitizedReason);
 
             FunctionType functionType = FunctionType.Orchestrator;
 
@@ -370,7 +370,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             string instanceId,
             string reason)
         {
-            string sanitizedReason = this.SanitizeString(reason, out string loggerReason);
+            this.SanitizeString(reason, out string loggerReason, out string sanitizedReason);
 
             FunctionType functionType = FunctionType.Orchestrator;
 
@@ -400,7 +400,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             bool isReplay,
             int taskEventId = -1)
         {
-            string sanitizedReason = this.SanitizeException(exception, out string loggerReason, isReplay);
+            this.SanitizeException(exception, out string loggerReason, out string sanitizedReason, isReplay);
             this.FunctionFailed(hubName, functionName, instanceId, loggerReason, sanitizedReason, functionType, isReplay, taskEventId);
         }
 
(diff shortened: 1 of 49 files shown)


src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -239,26 +238,6 @@ public void ExtensionWarningEvent(string hubName, string functionName, string in
             }
         }
 
-        public void ProcessingOutOfProcPayload(
```
davidmrdavid, line 242: this wasn't used

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -237,7 +237,7 @@ public override async Task<string> Execute(OrchestrationContext innerContext, st
                         this.context.Name,
                         this.context.InstanceId,
                         this.entityTraceInfo.TraceFlags,
-                        this.context.InternalError.ToString());
+                        this.context.InternalError.SourceException);
```
davidmrdavid, line 240: not immediately sure if this is right

davidmrdavid:
Just realized a bunch of logging unit tests broke, which is good, but it means I need to update them. Still, I'd appreciate a quick pass on the proposed refactoring in the meantime. Thanks

davidmrdavid:
Tests are passing. I'd appreciate a review here, @cgillum. Thanks!

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -54,6 +59,41 @@ public static string LocalSlotName
...
+
+            string payload = rawPayload ?? "(null)";
+            int numCharacters = payload.Length;
+            string sanitizedPayload = $"(Redacted {numCharacters} characters)";
+
+            iloggerPayload = this.shouldTraceRawData ? payload : sanitizedPayload;
+            return sanitizedPayload;
+        }
+
+        private string SanitizeException(Exception? exception, out string iloggerExceptionString, bool isReplay = false)
```
cgillum, line 78: The design of this API is a bit confusing. It's not clear what the difference is supposed to be between the output (return value) and the `out` variable. Rather than having a named `out` parameter and an unnamed return value, it would be better from a code understandability perspective to have two named `out` parameters and just return `void`. Having this kind of usage clarity is especially important in this case because making a mistake in interpreting the outputs of this method could cause us to accidentally leak sensitive information.

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -348,22 +380,36 @@ public void ExtensionWarningEvent(string hubName, string functionName, string in
...
+            string sanitizedReason = this.SanitizeException(exception, out string loggerReason, isReplay);
+            this.FunctionFailed(hubName, functionName, instanceId, loggerReason, sanitizedReason, functionType, isReplay, taskEventId);
+        }
+
         public void FunctionFailed(
             string hubName,
             string functionName,
             string instanceId,
             string reason,
+            string sanitizedReason,
```
cgillum, line 412: I noticed that `sanitizedReason` is going to ETW while `reason` is going to `ILogger`, but don't we send both of these payloads to Kusto, ETW to DurableFunctionsEvents and ILogger to FunctionsLogs? I know that's not the case for the DTFx tracing code, but I thought the WebJobs `ILogger` logs went to FunctionsLogs.

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -95,7 +95,7 @@ public async Task CallOrchestratorAsync(DispatchMiddlewareContext dispatchContex
                 this.Options.HubName,
                 functionName.Name,
                 instance.InstanceId,
-                isReplaying ? "(replay)" : this.extension.GetIntputOutputTrace(startEvent.Input),
+                startEvent.Input,
```
cgillum, line 98: I can't tell by looking at the diffs, but are we using `GetInputOutputTrace` at all anymore in the code?

davidmrdavid: At the time this of writing this comment, there was a single leftover use, but that was an accident. Starting from this commit (https://github.com/Azure/azure-functions-durable-extension/pull/2844/commits/43f63fc330a5e19723e9273d2196f3e8a5564f45) you can see the method has been removed.

So yes - this PR intends to remove this method altogether. Instead of sanitizing the logs at each calling site of our EndToEndTraceHelper, I opted to centralizing the sanitization in the EndtoEndTraceHelper itself. I think that should help minimize the chance of sanitization errors.

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
davidmrdavid: Good point - I have refactored the implementation to use two `out` variables.

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
davidmrdavid: You're right. Throughout this PR, we only sanitize the logs to `DurableFunctionsEvents` and not the ILogger-powered logs to `FunctionsLogs`. This is because the ILogger logs are also sent to the user's Application Insights instance, and I was worried about changing the logging behavior to a user-facing component.

So I'm choosing to delay our decision to sanitize the FunctionsLogs for now so that we can unblock sanitizing the DF Kusto table. Does that seem reasonable?

cgillum: OK, thanks for clarifying. It would be good to confirm what the plan is for sanitizing `FunctionsLogs` just to make sure that the work done in this PR isn't redundant or need to be reversed. Happy to chat about this offline.

davidmrdavid: Let's discuss this offline. But in general I think the clean up done here is unlikely to need to be reversed. If anything, I think it's more likely that it will need to be expanded to include FunctionsLogs as well. I'm just opting to merge a minimal improvement for now.

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -54,6 +59,41 @@ public static string LocalSlotName
 
 #pragma warning disable SA1117 // Parameters should be on same line or separate lines
 
+        private void SanitizeString(string? rawPayload, out string iloggerString, out string durableKustoTableString, bool isReplay = false)
+        {
+            if (isReplay)
+            {
+                iloggerString = rawPayload ?? "(replay)";
```
bachuv, line 66: I see `iloggerString` assigned here, but it's reassigned on line 74 (if `isReplay` is true). Seems like this line is unnecessary? Let me know if I'm misunderstanding this.

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -54,6 +59,41 @@ public static string LocalSlotName
...
+            {
+                exceptionString = orchestrationFailureException.Details;
+            }
+
+            string sanitizedString = exception != null ? $"{exception.GetType().FullName}\n{exception.StackTrace}" : string.Empty;
+            iloggerExceptionString = this.shouldTraceRawData ? exceptionString : sanitizedString;
+
+            if (isReplay)
+            {
+                durableKustoTableString = "(replay)";
```
bachuv, line 91: Why do we assign `durableKustoTableString` here if we are reassigning it on line 94 (if `isReplay` is true)?

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
bachuv: Can we add a test or run an E2E test to validate this change?

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
```diff
@@ -54,6 +59,41 @@ public static string LocalSlotName
 
 #pragma warning disable SA1117 // Parameters should be on same line or separate lines
 
+        private void SanitizeString(string? rawPayload, out string iloggerString, out string durableKustoTableString, bool isReplay = false)
```
bachuv, line 62: Do we have unit or E2E tests to check if we're logging the expected information correctly?

davidmrdavid: Yes, added some unit tests for the sanitizers here: https://github.com/Azure/azure-functions-durable-extension/pull/2844/commits/e16b9126e6e20bb2ec24e96b65ce7fc5fdaeb17d

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
davidmrdavid: I think this behavior was buggy, thanks for calling this out.
To be honest - part of why this method reads to strange is that current own logging is pretty inconsistent. **Sometimes** when the orchestrator is emitting a replay log, it replaces the inputs/outputs/exceptions with "(replay)", but sometimes it replaces it with "(replayed"), and sometimes it doesn't replace it at all.

I'm _ok_ making a small breaking change here to simplify this behavior. In the latest version of this PR, I completely drop the replacing of data with "(replay)" when "isReplay" is true. That simplifies the sanitizer's behavior and makes the logging more consistent across the board.

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
davidmrdavid: Answered here: https://github.com/Azure/azure-functions-durable-extension/pull/2844#discussion_r1663332537 . Let's look to continue the discussion _there._

src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
bachuv: Thanks for the explanation! I agree that we should make the behavior consistent across replays, but I'm not sure that we should remove this information. Do you know what the best practices are to make breaking changes with telemetry? Should we just add `(replay)` and then add the payload in the same string so we're not removing information that customers might use? I'm open to thoughts about this. I just want to get a better understanding.

bachuv: Resolving this comment after syncing with David offline about this.

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
bachuv: Do the tests that you added cover this scenario?

test/FunctionsV2/EndToEndTraceHelperTests.cs
```diff
@@ -0,0 +1,109 @@
...
+                Assert.Equal(expectedString, iLoggerString);
+            }
+            else
+            {
+                // If raw data is not being traced,
+                // kusto and the ilogger should get the same data
+                Assert.Equal(iLoggerString, kustoTableString);
+            }
+        }
+
```
bachuv, line 57: Can we remove this blank line to get rid of the warnings?

davidmrdavid: Good catch, just did! -> https://github.com/Azure/azure-functions-durable-extension/pull/2844/commits/bac3b3376cc75c1b4d8a8f4fd10a5b4b84257b85

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
davidmrdavid: Not quite, but I think I can explain why this makes sense and why it may be odd to write a test for this. I'm gonna have to write a small wall of text, sorry about that!

First - some background.

So this `InternalError` variable is of type ExceptionDispatchInfo, which is a built-in .NET type described here: https://learn.microsoft.com/en-us/dotnet/api/system.runtime.exceptionservices.exceptiondispatchinfo?view=net-8.0. In short, and as described in the docs, you can use this class to _capture_ and exception and re-throw it later without losing the original stack trace.

Here's an example of that capture and rethrow behavior: https://github.com/Azure/azure-functions-durable-extension/blob/829777ae7efb3dbbbab396f4bd4d0f76bc3c0955/src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs#L444 . The benefit here is the **throw** will preserve the original stack trace, instead of generating a new one at the current throw location.

As the documentation also states, the **SourceException** property allows you to obtain the original exception captured within.

Second - some experiments

I wrote the following Console app to demonstrate the logging behavior of `ExceptionDispatchInfo`, see below:

```C#
// See https://aka.ms/new-console-template for more information
using System.Runtime.ExceptionServices;

var rawException = new Exception("to be captured");
var wrappedException = ExceptionDispatchInfo.Capture(rawException);

var rawException2 = wrappedException.SourceException;
Console.WriteLine($"rawException: {rawException.ToString()}");
Console.WriteLine($"wrappedException: {wrappedException.ToString()}");
Console.WriteLine($"rawException2: {rawException2.ToString()}");
```
 
The output is the following:

```
rawException: System.Exception: to be captured
wrappedException: System.Runtime.ExceptionServices.ExceptionDispatchInfo
rawException2: System.Exception: to be captured
```

As you can see, logging the original exception, and the **SourceException** yields the same result. But logging the `<ExceptionDispatchInfo>.toString()` yields simply " System.Runtime.ExceptionServices.ExceptionDispatchInfo".

_So - what does this mean?_ It means that, in the code **today**, we're not logging the exception at all in **EntityBatchFailed**, we're just logging the string "System.Runtime.ExceptionServices.ExceptionDispatchInfo".

You can confirm this on Kusto, by running the following query:

```kql
All("DurableFunctionsEvents")
| where PreciseTimeStamp between(ago(3d).. now()) // p >= datetime(2024-06-19 16:24:00) and PreciseTimeStamp <= datetime(2024-06-19 16:25:00)
| where TaskName =~ "EntityBatchFailed" and Details contains "System.Runtime.ExceptionServices.ExceptionDispatchInfo"
| project TaskName, Details
| take 5
``` 

and you'll see

TaskName | Details
-- | --
EntityBatchFailed | System.Runtime.ExceptionServices.ExceptionDispatchInfo
EntityBatchFailed | System.Runtime.ExceptionServices.ExceptionDispatchInfo
EntityBatchFailed | System.Runtime.ExceptionServices.ExceptionDispatchInfo
EntityBatchFailed | System.Runtime.ExceptionServices.ExceptionDispatchInfo
EntityBatchFailed | System.Runtime.ExceptionServices.ExceptionDispatchInfo
```

**So what is this change doing then?**

Well, I am **assuming** that we meant to log the exception, not this "ExceptionDispatchInfo" constant, so I believe there's a bug. The fix I'm implementing here is to log the actual exception, via **SourceException**, and to use the sanitization utilities I've implemented to sanitize it. So yes, this is a breaking change, but I'd argue it's more akin to a bug fix - we were previously logging a constant string, instead of exception data that can be used for debugging.

Hope this makes sense, let me know what you think!

bachuv: This makes sense. Thanks for the detailed explanation!

bachuv (approved):
LGTM!

PR #2845, opened 2024-06-07
Cache TokenCredential

bachuv:
This PR caches `TokenCredential` to resolve a memory issue that has been reported when managed identity is enabled. With these changes, we only create the `TokenCredential` object once instead of every time that `GetStorageAccountDetails()` is called. This also means that `TokenRenewalState` is also created once.

These changes were verified by enabling managed identity locally using [these instructions](https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-configure-durable-functions-with-credentials#configure-your-app-to-use-client-secret-credentials) and measuring the memory in VS with and without the changes in this PR.

### Pull request checklist

    * [x] Otherwise: This change only applies to Durable Functions v2.x and **will not** be merged to branch v3.x.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/AzureStorageAccountProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/AzureStorageAccountProvider.cs
@@ -49,6 +49,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             AzureStorageAccountOptions account = connectionInfo.Get<AzureStorageAccountOptions>();
             if (account != null)
             {
+                // To avoid memory leaks, we cache the TokenCredential, which can be retrieved using a `key`
+                // There are two ways of generating the key, corresponding to the two setting sets that can be used to set up System-assigned Identities (see: https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-configure-durable-functions-with-credentials#add-managed-identity-configuration-in-the-azure-portal)
                 string key = !string.IsNullOrEmpty(account.AccountName) ? account.AccountName : account.BlobServiceUri.ToString();
 
                 TokenCredential credential = this.cachedTokenCredentials.GetOrAdd(
--- a/src/WebJobs.Extensions.DurableTask/AzureStorageAccountProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/AzureStorageAccountProvider.cs
@@ -49,8 +49,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             AzureStorageAccountOptions account = connectionInfo.Get<AzureStorageAccountOptions>();
             if (account != null)
             {
+                string key = !string.IsNullOrEmpty(account.AccountName) ? account.AccountName : account.BlobServiceUri.ToString();
+
                 TokenCredential credential = this.cachedTokenCredentials.GetOrAdd(
-                    account.AccountName,
+                    key,
                     attr => this.credentialFactory.Create(connectionInfo));
 
                 return new StorageAccountDetails
--- a/src/WebJobs.Extensions.DurableTask/AzureStorageAccountProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/AzureStorageAccountProvider.cs
@@ -2,10 +2,10 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
 using DurableTask.AzureStorage;
 using Microsoft.Azure.WebJobs.Extensions.DurableTask.Options;
 using Microsoft.Extensions.Configuration;
-using System.Collections.Concurrent;
 
 #if !FUNCTIONS_V1
 using Microsoft.Azure.WebJobs.Extensions.DurableTask.Auth;
--- a/src/WebJobs.Extensions.DurableTask/AzureStorageAccountProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/AzureStorageAccountProvider.cs
@@ -5,6 +5,8 @@ using System;
 using DurableTask.AzureStorage;
 using Microsoft.Azure.WebJobs.Extensions.DurableTask.Options;
 using Microsoft.Extensions.Configuration;
+using System.Collections.Concurrent;
+
 #if !FUNCTIONS_V1
 using Microsoft.Azure.WebJobs.Extensions.DurableTask.Auth;
 using Microsoft.WindowsAzure.Storage.Auth;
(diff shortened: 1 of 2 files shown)


src/WebJobs.Extensions.DurableTask/AzureStorageAccountProvider.cs
```diff
@@ -44,7 +49,11 @@ public StorageAccountDetails GetStorageAccountDetails(string connectionName)
             AzureStorageAccountOptions account = connectionInfo.Get<AzureStorageAccountOptions>();
             if (account != null)
             {
-                TokenCredential credential = this.credentialFactory.Create(connectionInfo);
+                string key = !string.IsNullOrEmpty(account.AccountName) ? account.AccountName : account.BlobServiceUri.ToString();
```
bachuv, line 52: I added this since customers can enable managed identity using 2 different set of settings:

```
AzureWebJobsStorage__accountName: For example: mystorageaccount123
```

OR
```
AzureWebJobsStorage__blobServiceUri: Example: https://mystorageaccount123.blob.core.windows.net/

AzureWebJobsStorage__queueServiceUri: Example: https://mystorageaccount123.queue.core.windows.net/

AzureWebJobsStorage__tableServiceUri: Example: https://mystorageaccount123.table.core.windows.net/
```

src/WebJobs.Extensions.DurableTask/AzureStorageAccountProvider.cs
```diff
@@ -19,6 +21,9 @@ internal sealed class AzureStorageAccountProvider : IStorageAccountProvider
 #if !FUNCTIONS_V1
         private readonly ITokenCredentialFactory credentialFactory;
 
+        private readonly ConcurrentDictionary<string, TokenCredential> cachedTokenCredentials =
+            new ConcurrentDictionary<string, TokenCredential>();
+
```
davidmrdavid, line 26: Is there a situation where we should remove elements from this dictionary?

src/WebJobs.Extensions.DurableTask/AzureStorageAccountProvider.cs
davidmrdavid, line 52:
```suggestion
                // To avoid memory leaks, we cache the TokenCredential, which can be retrieved using a `key`
                // There are two ways of generating the key, corresponding to the two setting sets that can be used to set up System-assigned Identities (see: https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-configure-durable-functions-with-credentials#add-managed-identity-configuration-in-the-azure-portal)
                string key = !string.IsNullOrEmpty(account.AccountName) ? account.AccountName : account.BlobServiceUri.ToString();
```

davidmrdavid (commented):
Some thoughts. In particular, curious if there's a "disposing" hook where we're supposed to remove elements from `cachedTokenCredentials`.

src/WebJobs.Extensions.DurableTask/AzureStorageAccountProvider.cs
davidmrdavid: Just a suggestion to add a bit more context ^

src/WebJobs.Extensions.DurableTask/AzureStorageAccountProvider.cs
```diff
@@ -44,7 +49,11 @@ public StorageAccountDetails GetStorageAccountDetails(string connectionName)
             AzureStorageAccountOptions account = connectionInfo.Get<AzureStorageAccountOptions>();
             if (account != null)
             {
-                TokenCredential credential = this.credentialFactory.Create(connectionInfo);
+                string key = !string.IsNullOrEmpty(account.AccountName) ? account.AccountName : account.BlobServiceUri.ToString();
+
+                TokenCredential credential = this.cachedTokenCredentials.GetOrAdd(
+                    key,
+                    attr => this.credentialFactory.Create(connectionInfo));
```
AnatoliB, line 56: Doesn't this token eventually expire? We seem to be storing it forever, but doesn't it get invalidated on the server side after a long enough period? Shouldn't we refresh it sometimes?

bachuv: Tagging @jviau because from what I remember he addressed this in a weekly sync when we originally discussed this memory leak, but let me know if I am remembering incorrectly. @jviau, do you know if there's any additional work we have to do to handle the token refresh?

jviau: `TokenCredentials` are long lived objects which refresh their token internally

jviau (changes requested):

src/WebJobs.Extensions.DurableTask/AzureStorageAccountProvider.cs
AnatoliB: Thank you @jviau. Are you saying that _the background auto refresh task leaks memory_, am I reading this correctly? Or the leak happens only when a new instance of `TokenCredential` is created?

jviau: It is the number of background tasks that adds up with each creation of the credential. The tasks logic itself does not leak memory but the task is itself leaked.

AnatoliB: Ok, I understand now, thank you.

src/WebJobs.Extensions.DurableTask/AzureStorageAccountProvider.cs
bachuv: I thought it should be `connectionName` too, but then realized that it was always the default value "Storage". I see the argument to keep it `connectionName` though since `connectionInfo` is the input for `credential`.

davidmrdavid: If `connectionName` is the value from which `accountName` and `blobServiceUri` are derived, then I think it's reasonable to use it as the key. @bachuv - is there a particular concern that you have with it always having "Storage" as the default value?

bachuv: > If connectionName is the value from which accountName and blobServiceUri are derived

It looks like `accountName` and `blobServiceUri` come from `connectionInfo` actually. Let me know if I misunderstood your comment.

>  @bachuv - is there a particular concern that you have with it always having "Storage" as the default value?

I don't have any concerns about always having "Storage" as the default value. I mostly just wanted to point out that I don't see us using `connectionName` in the managed identity code path.

src/WebJobs.Extensions.DurableTask/AzureStorageAccountProvider.cs
jviau: I can't think of one. I believe we want all of these for the lifetime of the application.

src/WebJobs.Extensions.DurableTask/AzureStorageAccountProvider.cs
jviau: How connection info works is that it will retrieve the config section for `AzureWebJobs{ConnectionName}`. In this case, you are getting `AzureWebJobsStorage` section, which includes blobServiceUri and all that. Since this is an `IConfiguration` object, it _does_ include all the managed identity information (`AzureStorageAccountOptions` is just a subset of what is in that section) and that is what the `ITokenCredentialFactory` ends up using.

So, the `connectionName`, ultimately maps to the config section which contains all of the managed identity properties, is the correct key to use.

jviau (approved):

PR #2849, opened 2024-06-12
Provide actionable suggestions for when webhooks are not configured

davidmrdavid:
We periodically receive questions about apps failing with the following error "Webhooks are not configured". This often occurs when running in K8s environments.

This PR augments our error message to provide guidance on how to work around this error.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
+++ b/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
@@ -1242,7 +1242,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         internal Uri GetWebhookUri()
         {
             string errorMessage = "Webhooks are not configured. This may occur if the environment variable `WEBSITE_HOSTNAME` is not set (should be automatically set for Azure Functions). " +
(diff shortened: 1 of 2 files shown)


src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -1241,7 +1241,9 @@ internal HttpCreationPayload GetInstanceCreationLinks()
 
         internal Uri GetWebhookUri()
         {
-            return this.webhookUrlProvider?.Invoke() ?? throw new InvalidOperationException("Webhooks are not configured");
+            string errorMessage = "Webhooks are not configured. This may occur if the environment variable `WEBSITE_HOSTNAME` is not set (should be automatically set for Azure Functions). " +
```
jviau, line 1244: nit on `<ip-address>:<port>` - should we instead use something around "URI" wording? So it is more clear it is open to ip-address:port or a DNS name

davidmrdavid: Good idea - implemented in commit: https://github.com/Azure/azure-functions-durable-extension/pull/2849/commits/3621b121c9a106ceae985ab35dc3f96a631853c9

PR #2856, opened 2024-06-18
Enable External Durable Client Managed Identity Support

nytian:
As titled. 

When external durable client is used, it will use the `StandardConnectionInfoProvider` instead of `WebJobsConnectionInfoProvider`. This PR updates the `Resolve()` at `StandardConnectionInfoProvider` so that it can read the configuration required for identity based authentication.

Also, when using an external durable client within a Functions app (but not within a Durable Functions app),   `AzureWebJobsStorage `configuration is required by Functions Host. To achieve this, the resolution process includes logic to first check the connection name with the `AzureWebJobs` prefix and then verify the standard connection name.

End-to-end test have been made with asp net core app and functions app at samples dir. 

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/StandardConnectionInfoProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/StandardConnectionInfoProvider.cs
@@ -44,10 +44,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// <inheritdoc />
         public IConfigurationSection Resolve(string name)
         {
-            // This implementation is a replication of the WEbJobsConnectionInfo Provider used for the internal durable client.
-            // The original code can be found at: 
+            // This implementation is a replica of the WebJobsConnectionInfoProvider used for the internal durable client.
+            // The original code can be found at:
             // https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/WebJobsConnectionInfoProvider.cs#L37.
-            // We need to first check the configuration section with the AzureWebJobs prefix, as this is the default name within the Functions app.
+            // We need to first check the configuration section with the AzureWebJobs prefix, as this is the default name within the Functions app whether it's internal or external client.
             string prefixedConnectionStringName = "AzureWebJobs" + name;
             IConfigurationSection section = this.configuration?.GetSection(prefixedConnectionStringName);
 
--- a/src/WebJobs.Extensions.DurableTask/StandardConnectionInfoProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/StandardConnectionInfoProvider.cs
@@ -23,10 +23,36 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
         }
 
+        // This implementation is IConfigurationSection.Exists. Details: https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Configuration.Abstractions/src/ConfigurationExtensions.cs#L78
+        // Functions host v1 (.net462 framework) doesn't support this method so we implement a substitute one here. 
+        private bool IfExists(IConfigurationSection section)
+        {
+            if (section == null)
+            {
+                return false;
+            }
+
+            if (section.Value == null)
+            {
+                return section.GetChildren().Any();
+            }
+
+            return true;
+        }
+
         /// <inheritdoc />
         public IConfigurationSection Resolve(string name)
         {
-            return this.configuration.GetSection(name);
+            string prefixedConnectionStringName = "AzureWebJobs" + name;
+            IConfigurationSection section = this.configuration?.GetSection(prefixedConnectionStringName);
+
+            if (!this.IfExists(section))
+            {
+                // next try a direct unprefixed lookup
+                section = this.configuration?.GetSection(name);
+            }
+
+            return section;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/WebJobs.Extensions.DurableTask/StandardConnectionInfoProvider.cs
```diff
@@ -23,10 +24,36 @@ public StandardConnectionInfoProvider(IConfiguration configuration)
             this.configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
         }
 
+        // This implementation is IConfigurationSection.Exists. Details: https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Configuration.Abstractions/src/ConfigurationExtensions.cs#L78
```
davidmrdavid, line 27: nit -on phrasing
```suggestion
        // This implementation is a lone of `IConfigurationSection.Exists` found herehttps://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Configuration.Abstractions/src/ConfigurationExtensions.cs#L78
```

src/WebJobs.Extensions.DurableTask/StandardConnectionInfoProvider.cs
```diff
@@ -23,10 +24,36 @@ public StandardConnectionInfoProvider(IConfiguration configuration)
...
+
+            return true;
+        }
+
         /// <inheritdoc />
         public IConfigurationSection Resolve(string name)
         {
-            return this.configuration.GetSection(name);
+            string prefixedConnectionStringName = "AzureWebJobs" + name;
+            IConfigurationSection section = this.configuration?.GetSection(prefixedConnectionStringName);
```
davidmrdavid, line 48: I'm a bit confused about this. Some questions:
(1) Are there other connection strings with prefix "AzureWebJobs" that aren't "AzureWebJobsStorage"?
(2) Why is it that we only apply this logic to connections with the "AzureWebJobs" prefix?
(3) How do we know the "name" parameter doesn't already include the "AzureWebJobs" prefix :) ?

Thanks!

davidmrdavid (commented):
Left some questions

src/WebJobs.Extensions.DurableTask/StandardConnectionInfoProvider.cs
nytian: That's a good question, it's a little confusing here. This is a special case that is quite hard to explain  and let me know if anything is unclear to you. 

This is a scenario that specially for using a durable client in Functions App. When the Functions app started, the Functions Host will look for `AzureWebJobsStorage` (side note: this is the default value, the connection name could be others, eg: MyCustomeStorage, but functions host will alwasy look for one with AzureWebJobs prefix, in this case it will be AzureWebJobsMyCustomeStorage) especially . If  there is no such a value, the host will fail. So we have to check the connection name with `AzureWebJobs` first so that the host can start successfully. Once the Host successfully started, then the method creating the external durable client will be called. In here, it will only look for connection name as `xxx` especially with no prefix. 

```

_client = clientFactory.CreateClient(new DurableClientOptions
{
    ConnectionName = "xxx",
});

```

So to answer the second question, this is the only scenario that we need a prefix check, only at Functions the configuration always starts with AzureWebJobs so we don't need to check other prefix. 

For (3), if it already includes, the section will be null, and thus the Resolve will look for the name itself with no prefix.

nytian: Our `WebJobsConnectionInfoProvider` which is for the internal durable client also use similar strategy.  You can check it [here ](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/WebJobsConnectionInfoProvider.cs#L38). And it said that this logic is basically from the function host which is [here ](https://github.com/Azure/azure-webjobs-sdk/blob/b6d5b52da5d2fb457efbf359cbdd733186aacf7c/src/Microsoft.Azure.WebJobs.Host/Extensions/IConfigurationExtensions.cs#L103-L133).

bachuv: Can we add a comment including the information in this thread? I'm mostly looking for a comment that explains why we added the code in the `Resolve` method.

nytian: Just added the comments inside Resolve method. Let me know if it isn't clear to you :)

src/WebJobs.Extensions.DurableTask/StandardConnectionInfoProvider.cs
```diff
@@ -23,10 +24,40 @@ public StandardConnectionInfoProvider(IConfiguration configuration)
...
+        }
+
         /// <inheritdoc />
         public IConfigurationSection Resolve(string name)
         {
-            return this.configuration.GetSection(name);
+            // This implementation is a replication of the WEbJobsConnectionInfo Provider used for the internal durable client.
+            // The original code can be found at:
+            // https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/WebJobsConnectionInfoProvider.cs#L37.
+            // We need to first check the configuration section with the AzureWebJobs prefix, as this is the default name within the Functions app.
```
bachuv, line 50:
```suggestion
            // This implementation is a replica of the WebJobsConnectionInfo Provider used for the internal durable client.
            // The original code can be found at:
            // https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/WebJobsConnectionInfoProvider.cs#L37.
            // We need to first check the configuration section with the AzureWebJobs prefix, as this is the default name within the Functions app whether it's internal or external.
```

bachuv (approved):
LGTM! Let's make sure that the tests pass before merging.

davidmrdavid (approved):

PR #2861, opened 2024-06-23
Add Injected Durable Client with Identity Based Connection Samples

nytian:

Add samples showing how to use external durable client with identity based connection. 

### Pull request checklist

### DIFF
--- a/samples/durable-client-managed-identity/aspnetcore-app/ToDoList.csproj
+++ b/samples/durable-client-managed-identity/aspnetcore-app/ToDoList.csproj
@@ -7,7 +7,7 @@
 
   <ItemGroup>
     <PackageReference Include="EntityFramework" Version="6.4.0" />
-    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="2.13.4" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="2.13.5" />
     <PackageReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Design" Version="3.1.2" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.Tools" Version="3.1.3">
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
--- a/samples/durable-client-managed-identity/aspnetcore-app/ToDoList.csproj
+++ b/samples/durable-client-managed-identity/aspnetcore-app/ToDoList.csproj
@@ -4,7 +4,6 @@
     <TargetFramework>netcoreapp3.1</TargetFramework>
   </PropertyGroup>
 
-
   <ItemGroup>
     <PackageReference Include="EntityFramework" Version="6.4.0" />
     <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="2.13.5" />
--- a/samples/durable-client-managed-identity/aspnetcore-app/ToDoList.csproj
+++ b/samples/durable-client-managed-identity/aspnetcore-app/ToDoList.csproj
@@ -1,0 +1,21 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+  </PropertyGroup>
+
+
+  <ItemGroup>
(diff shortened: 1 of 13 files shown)


samples/durable-client-managed-identity/aspnetcore-app/ToDoList.csproj
```diff
@@ -0,0 +1,21 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+  </PropertyGroup>
+
```
davidmrdavid, line 6: nit :
```suggestion
```

samples/durable-client-managed-identity/aspnetcore-app/appsettings.json
```diff
@@ -0,0 +1,17 @@
...
+      "Microsoft.Hosting.Lifetime": "Information"
+    }
+  },
+  "AllowedHosts": "*",
+  "TaskHub": "MyTestHub",
+  "MyStorage": {
+    "accountName": "YourStorageAccountName",
+    "clientId": "ClientSecretApplicationClientId",
+    "clientsecret": "ClientSecretApplicationClientSecret",
+    "tenantId": "ClientSecretApplicationTenantId"
```
davidmrdavid, line 15: nit
```suggestion
    "clientId": "<your client id here>",
    "clientsecret": "<your client secret here>",
    "tenantId": "<your tenant id here>"
```

samples/durable-client-managed-identity/functions-app/README.md
```diff
@@ -0,0 +1,33 @@
...
+    - client secret application:
+      - AzureWebJobsStorage__clientId
+      - AzureWebJobsStorage__ClientSecret
+      - AzureWebJobsStorage__tenantId
+
+
+## Notes
+
+- The Azure Functions runtime requires a storage account to start, with the default connection name `Storage`.
+- For Durable Functions initiated by the Durable Client to run successfully, both the Durable Client and the Functions runtime must use the same storage account. If you need to use a custom storage connection name for the Durable Client, ensure you also update the Azure Functions runtime to use this custom name.
```
davidmrdavid, line 32: hmm, I find it surprising that the DF Client and the runtime would need the same account. Why would that be the case? I thought DF could use a different account, always. Am I missing something? Would love @bachuv to chime in in case I'm missing something :)

davidmrdavid (commented):
Some comments, but these look pretty great!

samples/durable-client-managed-identity/functions-app/README.md
nytian: I tried the scenario where df client and the runtime use different storage account. The client could create orchestration instances, but all the instances will remain in Pending state. So to make those orchestrators to start/ go to Running state, we need a runtime at the same storage account

davidmrdavid: Ah, I see. I think that's expected though :) . If the storage account is different, then the expecation is that there is **another app** that is listening to that account that will then process those orchestrators.

In other words, in this scenario you have two apps: A and B.

App A's Functions Host and DF uses storageA
App B's Functions Host and DF uses storageB

Now, app A is allowed to create a DurableClient that references storageB and schedules an orchestrator _there_. That will cause app B to pick up that work and process it. This is what we call an "external client", because app A's client is external to the storage account where it is creating an orchestrator. Does that make sense?

nytian: Thanks! That makes sense to me. I will update the notes.

samples/durable-client-managed-identity/functions-app/DurableClientSampleFunctionApp.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <AzureFunctionsVersion>v3</AzureFunctionsVersion>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.Functions.Extensions" Version="1.1.0" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="2.13.5" />
```
nytian, line 8: This version hasn't been released yet. But this should be the version that starts supporting token credential class.

davidmrdavid (approved):

PR #2922, opened 2024-10-01
Retry platform-level errors in the isolated process for .NET isolated

davidmrdavid:
Supersedes: https://github.com/Azure/azure-functions-durable-extension/pull/2902
Joint work with @andystaples.

**Problem:** In .NET isolated today, platform-level errors **in the isolated process** like OOMs, Timeouts, and sudden process exits are not automatically retried by DF. Instead, those cause the orchestrator to be marked as failed. This is an unintended behavior change for .NET isolated. This PR aims to bring back robustness against those errors.

Most of the changes related to this exist in the `OutOfProcMiddleware` and `RemoteOrchestrationContext` files.

**"Extra scope":** In trying to add automatic E2E tests to validate these improvements, I had to go down a rabbithole of .NET isolated testing issues in our smoke testing infrastructure today.

Long story short, our docker-based test runner for .NET isolated wasn't testing runtime correctness (it was passed the `-NoValidation` flag) because of intermittent race conditions when publishing .NET isolated apps to Docker. I didn't get far enough into debugging this to root cause the issue (@jviau tells me that upgrading the SDK may automatically solve them) but I did spend enough hours on it to conclude that, for the sake of unblocking this item, it would be practical to abandon the Docker-based testing for .NET isolated and instead use the GitHub actions CI environment directly.

Therefore, this PR changes the definition of the .NET isolated smoke testing GitHub action to instead run the tests directly on the "ubuntu-latest" image. It also introduces a new test runner powershell script that's well-suited for simulating platform-level errors. I hope the comments in the code itself clarify the intent, but please let me know if not.

**Ask for reviewers:** the key shippable changes here are in  `OutOfProcMiddleware` and `RemoteOrchestrationContext` . Please look at those carefully. I've added comments explaining some of the quirky bits.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteOrchestratorContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteOrchestratorContext.cs
@@ -120,24 +120,20 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             this.SetResultInternal(result);
         }
 
-        /// <summary>
-        /// Recursively inspect the FailureDetails of the failed orchestrator and throw if a platform-level exception is detected.
-        /// </summary>
-        /// <remarks>
-        /// Today, this method only checks for <see cref="SessionAbortedException"/>. In the future, we may want to add more cases.
-        /// Other known platform-level exceptions, like timeouts or process exists due to `Environment.FailFast`, do not yield
-        /// a `OrchestratorExecutionResult` as the isolated invocation is abruptly terminated. Therefore, they don't need to be
-        /// handled in this method.
-        /// However, our tests reveal that OOMs are, surprisngly, caught and returned as a `OrchestratorExecutionResult`
-        /// by the isolated process, and thus need special handling.
-        /// It's unclear if all OOMs are caught by the isolated process (probably not), and also if there are other platform-level
-        /// errors that are also caught in the isolated process and returned as a `OrchestratorExecutionResult`. Let's add them
-        /// to this method as we encounter them.
-        /// </remarks>
-        /// <param name="failureDetails">The failure details of the orchestrator.</param>
-        /// <exception cref="SessionAbortedException">If an OOM error is detected.</exception>
         private void ThrowIfPlatformLevelException(FailureDetails failureDetails)
         {
+            // Recursively inspect the FailureDetails of the failed orchestrator and throw if a platform-level exception is detected.
+            //
+            // Today, this method only checks for <see cref="OutOfMemoryException"/>. In the future, we may want to add more cases.
+            // Other known platform-level exceptions, like timeouts or process exists due to `Environment.FailFast`, do not yield
+            // a `OrchestratorExecutionResult` as the isolated invocation is abruptly terminated. Therefore, they don't need to be
+            // handled in this method.
+            // However, our tests reveal that OOMs are, surprisngly, caught and returned as a `OrchestratorExecutionResult`
+            // by the isolated process, and thus need special handling.
+            //
+            // It's unclear if all OOMs are caught by the isolated process (probably not), and also if there are other platform-level
+            // errors that are also caught in the isolated process and returned as a `OrchestratorExecutionResult`. Let's add them
+            // to this method as we encounter them.
             if (failureDetails.InnerFailure?.IsCausedBy<OutOfMemoryException>() ?? false)
             {
                 throw new SessionAbortedException(failureDetails.ErrorMessage);
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteOrchestratorContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteOrchestratorContext.cs
@@ -71,7 +71,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             return hasError;
         }
 
-        internal void SetResult(IEnumerable<OrchestratorAction> actions, string customStatus)
+        internal void TrySetResult(IEnumerable<OrchestratorAction> actions, string customStatus)
         {
             var result = new OrchestratorExecutionResult
             {
@@ -79,7 +79,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 Actions = actions,
             };
 
-            this.SetResultInternal(result);
+            this.TrySetResultInternal(result);
         }
 
         // TODO: This method should be considered deprecated because SDKs should no longer be returning results as JSON.
@@ -117,10 +117,39 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     innerException: jsonReaderException);
             }
 
-            this.SetResultInternal(result);
+            this.TrySetResultInternal(result);
         }
 
-        private void SetResultInternal(OrchestratorExecutionResult result)
+        /// <summary>
+        /// Recursively inspect the FailureDetails of the failed orchestrator and throw if a platform-level exception is detected.
+        /// </summary>
+        /// <remarks>
+        /// Today, this method only checks for <see cref="OutOfMemoryException"/>. In the future, we may want to add more cases.
+        /// Other known platform-level exceptions, like timeouts or process exists due to `Environment.FailFast`, do not yield
+        /// a `OrchestratorExecutionResult` as the isolated invocation is abruptly terminated. Therefore, they don't need to be
+        /// handled in this method.
+        /// However, our tests reveal that OOMs are, surprisngly, caught and returned as a `OrchestratorExecutionResult`
+        /// by the isolated process, and thus need special handling.
+        /// It's unclear if all OOMs are caught by the isolated process (probably not), and also if there are other platform-level
+        /// errors that are also caught in the isolated process and returned as a `OrchestratorExecutionResult`. Let's add them
+        /// to this method as we encounter them.
+        /// </remarks>
+        /// <param name="failureDetails">The failure details of the orchestrator.</param>
+        /// <exception cref="OutOfMemoryException">If an OOM error is detected.</exception>
+        private void ThrowIfPlatformLevelException(FailureDetails failureDetails)
+        {
+            if (failureDetails.InnerFailure?.IsCausedBy<OutOfMemoryException>() ?? false)
+            {
+                throw new OutOfMemoryException(failureDetails.ErrorMessage);
+            }
+
+            if (failureDetails.InnerFailure != null)
+            {
+                this.ThrowIfPlatformLevelException(failureDetails.InnerFailure);
+            }
+        }
+
+        private void TrySetResultInternal(OrchestratorExecutionResult result)
         {
             // Look for an orchestration completion action to see if we need to grab the output.
             foreach (OrchestratorAction action in result.Actions)
@@ -133,6 +162,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
                     if (completeAction.OrchestrationStatus == OrchestrationStatus.Failed)
                     {
+                        // If the orchestrator failed due to a platform-level error in the isolated process,
+                        // we should re-throw that exception in the host (this process) invocation pipeline,
+                        // so the invocation can be retried.
+                        if (completeAction.FailureDetails != null)
+                        {
+                            this.ThrowIfPlatformLevelException(completeAction.FailureDetails);
+                        }
+
                         string message = completeAction switch
                         {
                             { FailureDetails: { } f } => f.ErrorMessage,
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteOrchestratorContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteOrchestratorContext.cs
@@ -71,7 +71,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             return hasError;
         }
 
-        internal void TrySetResult(IEnumerable<OrchestratorAction> actions, string customStatus)
+        internal void SetResult(IEnumerable<OrchestratorAction> actions, string customStatus)
         {
             var result = new OrchestratorExecutionResult
             {
@@ -79,7 +79,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 Actions = actions,
             };
 
-            this.TrySetResultInternal(result);
+            this.SetResultInternal(result);
         }
 
         // TODO: This method should be considered deprecated because SDKs should no longer be returning results as JSON.
@@ -117,14 +117,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     innerException: jsonReaderException);
             }
 
-            this.TrySetResultInternal(result);
+            this.SetResultInternal(result);
         }
 
         /// <summary>
         /// Recursively inspect the FailureDetails of the failed orchestrator and throw if a platform-level exception is detected.
         /// </summary>
         /// <remarks>
-        /// Today, this method only checks for <see cref="OutOfMemoryException"/>. In the future, we may want to add more cases.
+        /// Today, this method only checks for <see cref="SessionAbortedException"/>. In the future, we may want to add more cases.
         /// Other known platform-level exceptions, like timeouts or process exists due to `Environment.FailFast`, do not yield
         /// a `OrchestratorExecutionResult` as the isolated invocation is abruptly terminated. Therefore, they don't need to be
         /// handled in this method.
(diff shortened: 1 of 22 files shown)


test/SmokeTests/OOProcSmokeTests/DotNetIsolated/FaultyOrchestrators.cs
```diff
@@ -0,0 +1,155 @@
...
+            [OrchestrationTrigger] TaskOrchestrationContext context)
+        {
+            // this orchestrator is not deterministic, on purpose.
+            // we use the non-determinism to force an OOM exception on only the first replay
+            
+            // check if a file named "replayEvidence" exists in source code directory, create it if it does not.
+            // From experience, this code runs in `<sourceCodePath>/bin/output/`, so we store the file two directories above.
+            // We do this because the /bin/output/ directory gets overridden during the build process, which happens automatically
+            // when `func host start` is re-invoked.
+            string evidenceFile = System.IO.Path.Combine(System.IO.Directory.GetCurrentDirectory(), "..", "..", "replayEvidence");
```
andystaples, line 23: Would suggest giving these replayEvidence files test-specific names to avoid conflicts should multiple retry tests run simultaneously. Even if today they are run sequentially, this may not always be true

andystaples (approved):
Left minor suggestion, otherwise, looks good

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/host.json
```diff
@@ -7,5 +7,14 @@
         "excludedTypes": "Request"
       }
     }
-  }
+  },
+  "extensions": {
+    "durableTask": {
+      "storageProvider": {
+        "maxQueuePollingInterval": "00:00:01",
+        "controlQueueVisibilityTimeout": "00:01:00"
```
davidmrdavid, line 15: this just helps make sure we retry orchestrator replays fast enough after a platform-level error.

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/host.json
```diff
@@ -7,5 +7,14 @@
...
+      }
+    }
+  },
+  "functionTimeout": "00:00:30"
```
davidmrdavid, line 19: We don't time out sooner because our OOM-based test needs time to run out of memory.

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -159,6 +164,19 @@ await this.LifeCycleNotificationHelper.OrchestratorStartingAsync(
...
+                // we abort the invocation on "platform level errors" such as:
+                // - a timeout
+                // - an out of memory exception
+                // - a worker process exit
+                if (functionResult.Exception is Host.FunctionTimeoutException
+                    || functionResult.Exception?.InnerException is OutOfMemoryException // see RemoteOrchestrationContext.TrySetResultInternal for details on OOM-handling
+                    || (functionResult.Exception?.InnerException?.GetType().ToString().Contains("WorkerProcessExitException") ?? false))
+                {
+                    // TODO: the `WorkerProcessExitException` type is not exposed in our dependencies, it's part of WebJobs.Host.Script.
+                    // Should we add that dependency or should it be exposed in WebJobs.Host?
```
davidmrdavid, line 177: I don't love the string-based comparison, but it'll be string-based unless we add `WebJobs.Host.Script` to our dependencies. Open to a discussion here, but please note we have a customer waiting so let's look to be practical. I have no strong feelings, just trying to deploy this fix.

src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteOrchestratorContext.cs
```diff
@@ -71,15 +71,15 @@ internal bool TryGetOrchestrationErrorDetails(out Exception? failure)
             return hasError;
         }
 
-        internal void SetResult(IEnumerable<OrchestratorAction> actions, string customStatus)
+        internal void TrySetResult(IEnumerable<OrchestratorAction> actions, string customStatus)
```
cgillum, line 74: Non-blocking nit: Any method named `TryXXX` is usually expected to return a `bool` value indicating whether it succeeded or failed.

src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteOrchestratorContext.cs
```diff
@@ -117,10 +117,39 @@ internal void SetResult(string orchestratorResponseJsonText)
...
+        /// errors that are also caught in the isolated process and returned as a `OrchestratorExecutionResult`. Let's add them
+        /// to this method as we encounter them.
+        /// </remarks>
+        /// <param name="failureDetails">The failure details of the orchestrator.</param>
+        /// <exception cref="OutOfMemoryException">If an OOM error is detected.</exception>
+        private void ThrowIfPlatformLevelException(FailureDetails failureDetails)
+        {
+            if (failureDetails.InnerFailure?.IsCausedBy<OutOfMemoryException>() ?? false)
+            {
+                throw new OutOfMemoryException(failureDetails.ErrorMessage);
```
cgillum, line 143: Manually throwing `OutOfMemoryException` is a bit of a red flag. It's really only supposed to be thrown by the runtime. It's partly problematic because there are various pieces of code in .NET and elsewhere that treat OOM and similar "fatal" exceptions differently. In some cases, I think it's supposed to crash the process.

I strongly recommend we throw something else, like `SessionAbortedException`, which is designed for this exact scenario.

src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask.xml
```diff
@@ -475,6 +475,24 @@
             </summary>
             <param name="result">The result.</param>
         </member>
+        <member name="M:Microsoft.Azure.WebJobs.Extensions.DurableTask.RemoteOrchestratorContext.ThrowIfPlatformLevelException(DurableTask.Core.FailureDetails)">
```
cgillum, line 478: Any idea why this is showing up in our public documentation? I thought these APIs were internal?

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/FaultyOrchestrators.cs
```diff
@@ -0,0 +1,155 @@
...
+                // force the process to run out of memory
+                List<byte[]> data = new List<byte[]>();
+
+                for (int i = 0; i < 10000000; i++)
+                {
+                    data.Add(new byte[1024 * 1024 * 1024]);
+                }
+            }
+            
+            // assuming the orchestrator survived the OOM, delete the evidence file and return
```
cgillum, line 38: I'm not sure I understand this comment, and it's making me question whether I understand what this test is supposed to be doing. I'll ask simply: why would we expect an orchestrator to survive an OOM? It sounds like you're saying there was an OOM but it was handled and we kept running in spite of the OOM. Are you actually saying it *recovered from a worker process crash* (caused by the OOM)? Maybe OOM implies crash, but it's not 100% clear that's what you meant.

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/FaultyOrchestrators.cs
```diff
@@ -0,0 +1,155 @@
...
+            bool isTheFirstReplay = !System.IO.File.Exists(evidenceFile);
+            if (isTheFirstReplay)
+            {
+                System.IO.File.Create(evidenceFile).Close();
+
+                // simulate sudden crash
+                Environment.FailFast("Simulating crash!");
+            }
+            
+            // assuming the orchestrator survived the OOM, delete the evidence file and return
```
cgillum, line 64: Does this comment need to be updated?

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/FaultyOrchestrators.cs
```diff
@@ -0,0 +1,155 @@
...
+        public static async Task<HttpResponseData> HttpStartTimeoutOrchestrator(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post")] HttpRequestData req,
+            [DurableClient] DurableTaskClient client,
+            FunctionContext executionContext)
+        {
+            ILogger logger = executionContext.GetLogger("durable_HttpStartTimeoutOrchestrator");
+
+            // Function input comes from the request content.
+            string instanceId = await client.ScheduleNewOrchestrationInstanceAsync(
+                nameof(TimeoutOrchestrator));
```
cgillum, line 146: Rather than replicating the same HTTP function once for each scenario, it might make more sense to have a single HTTP trigger that takes an orchestrator function name as a parameter.

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/run-smoke-tests.ps1
```diff
@@ -0,0 +1,115 @@
...
+            $success = $false
+            $testIsRunning = $false
+            break
+        }
+
+        # If the orchestrator did not complete yet, wait for a bit before checking again
+        Start-Sleep -Seconds 2
+        $retryCount = $retryCount + 1
+
+    } catch {
```
cgillum, line 82: When is it expected that we would enter this `catch` block?

cgillum (commented):
This is a great PR! I have some questions and comments.

jviau:
Summarizing some offline discussion:

1. Handling timeouts, worker crashes, and **host** out of memories is good
2. We should not handle _worker_ out of memory. This is language dependent and each language should handle OOM their own way
    - This should be left for future work. We will need to see if there is an existing mechanism for a language worker to signal an orchestration should be retried. If not, that mechanism will need to be added.
3. I will follow up on the host side to see if there is an existing way to detect worker crashes or other platform-level failures (instead of checking exception type by string). If not, we can file an issue for that ask (and proceed with the string type check for now)

src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteOrchestratorContext.cs
davidmrdavid: yes, you're right. II chose `Try` to signal somehow that the method may throw, but I know the convention makes this confusing. I'll revert this change.

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/FaultyOrchestrators.cs
davidmrdavid: I don't feel strongly here, but my personal preference is to specific HTTP endpoints for each orchestrator. When testing locally, it saves me time to just copy-paste the entire URL to trigger the orchestrator. 

But it's no big deal. If you feel strongly, I'll generalize it. Just noting this was a deliberate choice.

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/run-smoke-tests.ps1
davidmrdavid: I added comments to the script clarifying this here: https://github.com/Azure/azure-functions-durable-extension/pull/2922/commits/2ff4b79d571e2ec288e5e2da17db92c8370b23a1

Long story short - we expect to enter this catch block if any HTTP request to the host fails. This can happen if the host has not started yet, or if has become unresponsive for whatever reason. Recall that we're inducing isolated worker process crashes, and in my experience that can make the host unresponsive for a bit.

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/FaultyOrchestrators.cs
davidmrdavid: Good catch, this is the result of copy-pasting.
Updated: https://github.com/Azure/azure-functions-durable-extension/pull/2922/commits/a70d0146fae6771f6cfd834a374407045043a9ec

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/FaultyOrchestrators.cs
davidmrdavid: You're right the wording here was not great.

I do not expect an orchestration invocation to continue running beyond an OOM.

What I mean by survive was:
- orchestrator invocation A runs
- invocation A OOMs
- DF extension abandons the invocation
- DF extension retries the orchestrator invocation (now invocation B, every invocation is unique)
- invocation does not OOM now, and completes

So it "survived"  because it retried.

I'm hoping the new comments here make this clearer: https://github.com/Azure/azure-functions-durable-extension/pull/2922/commits/a70d0146fae6771f6cfd834a374407045043a9ec

src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteOrchestratorContext.cs
davidmrdavid: Thanks for the pointer. I'm now throwing `SessionAbortedException`, see: https://github.com/Azure/azure-functions-durable-extension/pull/2922/commits/090f8e729a0c584af3863209154f4ecd9ecd4c9c

src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteOrchestratorContext.cs
davidmrdavid: fixed here: https://github.com/Azure/azure-functions-durable-extension/pull/2922/commits/090f8e729a0c584af3863209154f4ecd9ecd4c9c

src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask.xml
davidmrdavid: My guess is that since I added method-level comments to the private method, they're made it to the xml. I think this is probably just to power the intellisense? Not sure, just a guess.

davidmrdavid:
Thanks @jviau. I'll delegate this bit to you then:

>I will follow up on the host side to see if there is an existing way to detect worker crashes or other platform-level failures (instead of checking exception type by string). If not, we can file an issue for that ask (and proceed with the string type check for now)

As for the rest, I'll see what we can do. OOM is the key case we're trying to unblock via this PR (the rest are related "freebies"). @andystaples will start an internal thread about this.

src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask.xml
cgillum: It's not just intellisense, but also the public API docs that we publish, like this: https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.webjobs.extensions.durabletask?view=azure-dotnet. If it's not meant to be published, then we should remove it from this xml file. Might be worth doing a bit of research as when when/why non-public APIs are getting included in this file.

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/FaultyOrchestrators.cs
cgillum: I don't feel strongly so I'll leave it up to you, but it would enable all the usual benefits of code sharing (easier to add new tests, better consistency b/c only one place to make updates, etc.)

test/SmokeTests/OOProcSmokeTests/DotNetIsolated/run-smoke-tests.ps1
cgillum: Got it, I didn't realize that `Invoke-RestMethod` raised an error on non-success status codes.

jviau:
I think all of the points I brought up can also be addressed after this PR, but it is something we should commit to doing ASAP.

1. Adopting a more 'official' way to identify platform issues / worker crashes
2. Implementing/consuming a way for workers to elect to abandon a session (which they can then use for OOM scenarios)

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -159,6 +164,19 @@ await this.LifeCycleNotificationHelper.OrchestratorStartingAsync(
...
                     this.HostLifetimeService.OnStopping.ThrowIfCancellationRequested();
                 }
+
+                // we abort the invocation on "platform level errors" such as:
+                // - a timeout
+                // - an out of memory exception
+                // - a worker process exit
+                if (functionResult.Exception is Host.FunctionTimeoutException
+                    || functionResult.Exception?.InnerException is SessionAbortedException // see RemoteOrchestrationContext.TrySetResultInternal for details on OOM-handling
+                    || (functionResult.Exception?.InnerException?.GetType().ToString().Contains("WorkerProcessExitException") ?? false))
```
jviau, line 174: Did you verify that `WorkerProcessExitException` is actually available here? Looking at the function host code, I don't see the exception actually _thrown_: https://github.com/Azure/azure-functions-host/blob/1088f24c3ae3a6275f18cbf091fa525c2477be91/src/WebJobs.Script/Workers/ProcessManagement/WorkerProcess.cs#L179-L184

davidmrdavid: Yes, that's how we stumbled upon this error type in the first place. We found this error by forcing a process exit (`Environment.FailFast`) and placing a breakpoint at this position. 

But yeah I admit that I also don't see the explicit throw here. We can investigate further, or I can show you the runtime behavior through the debugger.

src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask.xml
davidmrdavid: Seems that adding the method-level comments was the culprit.
I removed them (moved them to the method body), and the xml change was undone: https://github.com/Azure/azure-functions-durable-extension/pull/2922/commits/c92726846c351aeb341c80fd3f0705e6b93653c4

jviau (approved):

Issue #2810, opened 2024-05-07
CreateHttpManagementPayload is missing for Isolated functions in .net framework 8


well0549:
### Description
The method CreateHttpManagementPayload is missing in 8.0. I used to call this method to get the management payload for a queue triggered function. 

### Expected behavior
A way to get the management payload (as in 6.0) when a Durable function is not started with a HTTP Request. One of the first activities (in 6.0) was to get the management payload. 

```
HttpManagementPayload managementInfo = starter.CreateHttpManagementPayload(functionInput.ServiceBusMessageId);
```

### Actual behavior
The method is completely missing.  Now there is no way to get the management url's anymore if the function is not triggered with a HTTP request. Besides that, the Management URLS where stored in the payload of the function. So I need to call the method AFTER the orchestration was started. Not as a response to a HTTP Request. I need to get this information AFTER the orchestration is created, because in 6.0 the fields were not always populated correct after starting an orchestration. (only under high load). Getting the Management Payload after the orchestration was started, (always first activity) would fix that problem.

### Known workarounds
None

### App Details

```
	<PropertyGroup>
		<TargetFramework>net8.0</TargetFramework>
		<AzureFunctionsVersion>v4</AzureFunctionsVersion>
		<_FunctionsSkipCleanOutput>true</_FunctionsSkipCleanOutput>
		<OutputType>Exe</OutputType>
	</PropertyGroup>
	<ItemGroup>
		<FrameworkReference Include="Microsoft.AspNetCore.App" />
		<PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.22.0" />
		<PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.ServiceBus" Version="5.18.0" />
		<PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Timer" Version="4.3.0" />
		<PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.17.3-preview1" />
		<PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore" Version="1.2.1" />
		<PackageReference Include="Microsoft.Azure.Functions.Worker.ApplicationInsights" Version="1.2.0" />
   		<PackageReference Include="Microsoft.ApplicationInsights.WorkerService" Version="2.22.0" />
	</ItemGroup>
	<ItemGroup>
[...]

PR #2929, opened 2024-10-07
Implement CreateHttpManagementPayload API in Durable Worker Extension

nytian:

Fix issue #2810 

This PR adds the implementation of the CreateHttpManagementPayload API in the Durable Worker Extension. The implementation works as follows:

- If an HttpRequestData object is provided, the HTTP management payload is created using the base URL derived from this request data.
- If no HttpRequestData is available, the base URL is obtained from the Functions Host configuration, which is sent and converted by DurableTaskClient.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
@@ -137,7 +137,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             /// This URL is sent by the client binding object to the Durable Worker extension,
             /// allowing the extension to know the host's base URL for constructing management URLs.
             /// </summary>
-            [JsonProperty("HttpBaseUrl")]
+            [JsonProperty("httpBaseUrl")]
             public string? HttpBaseUrl { get; set; }
         }
     }
--- a/src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
@@ -135,7 +135,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             /// <summary>
             /// The base URL of the Azure Functions host, used in the out-of-proc model.
             /// This URL is sent by the client binding object to the Durable Worker extension,
-            /// allowing the extension to know the host's base URL for constructing complete URLs.
+            /// allowing the extension to know the host's base URL for constructing management URLs.
             /// </summary>
             [JsonProperty("HttpBaseUrl")]
             public string? HttpBaseUrl { get; set; }
--- a/src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
@@ -43,7 +43,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     ConnectionName = attr.ConnectionName,
                     RpcBaseUrl = localRpcAddress,
                     RequiredQueryStringParameters = this.config.HttpApiHandler.GetUniversalQueryStrings(),
-                    BaseUrl = this.config.HttpApiHandler.GetBaseUrl(),
+                    HttpBaseUrl = this.config.HttpApiHandler.GetBaseUrl(),
                 });
             }
 
@@ -131,6 +131,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             /// </summary>
             [JsonProperty("rpcBaseUrl")]
             public string? RpcBaseUrl { get; set; }
+
+            /// <summary>
+            /// The base URL of the Azure Functions host, used in the out-of-proc model.
+            /// This URL is sent by the client binding object to the Durable Worker extension,
+            /// allowing the extension to know the host's base URL for constructing complete URLs.
+            /// </summary>
+            [JsonProperty("HttpBaseUrl")]
+            public string? HttpBaseUrl { get; set; }
         }
     }
 }
--- a/src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
@@ -43,6 +43,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     ConnectionName = attr.ConnectionName,
                     RpcBaseUrl = localRpcAddress,
                     RequiredQueryStringParameters = this.config.HttpApiHandler.GetUniversalQueryStrings(),
+                    BaseUrl = this.config.HttpApiHandler.GetBaseUrl(),
                 });
             }
 
--- a/src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
+++ b/src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
@@ -139,14 +139,7 @@ public static class DurableTaskClientExtensions
             throw new ArgumentException("InstanceId cannot be null or empty.", nameof(instanceId));
         }
 
-        try
-        {
-            return SetHeadersAndGetPayload(client, request, null, instanceId);
-        }
-        catch (InvalidOperationException ex)
-        {
-            throw new InvalidOperationException("Failed to create HTTP management payload. " + ex.Message, ex);
-        }
+        return SetHeadersAndGetPayload(client, request, null, instanceId);
     }
 
     private static HttpManagementPayload SetHeadersAndGetPayload(
@@ -181,7 +174,7 @@ public static class DurableTaskClientExtensions
 
         if (baseUrl == null)
         {
-            throw new InvalidOperationException("Base URL is null. Either use Functions bindings or provide an HTTP request to create the HttpPayload.");
+            throw new InvalidOperationException("Failed to create HTTP management payload as base URL is null. Either use Functions bindings or provide an HTTP request to create the HttpPayload.");
         }
         
         bool isFromRequest = request != null;
--- a/src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
+++ b/src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
@@ -129,7 +129,7 @@ public static class DurableTaskClientExtensions
     /// <returns>An object containing instance control URLs.</returns>
     /// <exception cref="ArgumentException">Thrown when instanceId is null or empty.</exception>
     /// <exception cref="InvalidOperationException">Thrown when a valid base URL cannot be determined.</exception>
-    public static object CreateHttpManagementPayload(
+    public static HttpManagementPayload CreateHttpManagementPayload(
         this DurableTaskClient client,
         string instanceId,
         HttpRequestData? request = null)
@@ -149,7 +149,7 @@ public static class DurableTaskClientExtensions
         }
     }
 
-    private static object SetHeadersAndGetPayload(
+    private static HttpManagementPayload SetHeadersAndGetPayload(
         DurableTaskClient client, HttpRequestData? request, HttpResponseData? response, string instanceId)
     {
         static string BuildUrl(string url, params string?[] queryValues)
@@ -198,15 +198,15 @@ public static class DurableTaskClientExtensions
             response.Headers.Add("Content-Type", "application/json");
         }
 
-        return new
+        return new HttpManagementPayload
         {
-            id = instanceId,
-            purgeHistoryDeleteUri = BuildUrl(instanceUrl, commonQueryParameters),
-            sendEventPostUri = BuildUrl($"{instanceUrl}/raiseEvent/{{eventName}}", commonQueryParameters),
-            statusQueryGetUri = BuildUrl(instanceUrl, commonQueryParameters),
-            terminatePostUri = BuildUrl($"{instanceUrl}/terminate", "reason={{text}}", commonQueryParameters),
-            suspendPostUri =  BuildUrl($"{instanceUrl}/suspend", "reason={{text}}", commonQueryParameters),
-            resumePostUri =  BuildUrl($"{instanceUrl}/resume", "reason={{text}}", commonQueryParameters)
+            Id = instanceId,
+            PurgeHistoryDeleteUri = BuildUrl(instanceUrl, commonQueryParameters),
+            SendEventPostUri = BuildUrl($"{instanceUrl}/raiseEvent/{{eventName}}", commonQueryParameters),
+            StatusQueryGetUri = BuildUrl(instanceUrl, commonQueryParameters),
+            TerminatePostUri = BuildUrl($"{instanceUrl}/terminate", "reason={{text}}", commonQueryParameters),
+            SuspendPostUri =  BuildUrl($"{instanceUrl}/suspend", "reason={{text}}", commonQueryParameters),
+            ResumePostUri =  BuildUrl($"{instanceUrl}/resume", "reason={{text}}", commonQueryParameters)
         };
     }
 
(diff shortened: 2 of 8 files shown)


src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
```diff
@@ -143,12 +172,25 @@ static string BuildUrl(string url, params string?[] queryValues)
         //       request headers into consideration and generate the base URL accordingly.
         //       More info: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded.
         //       One potential workaround is to set ASPNETCORE_FORWARDEDHEADERS_ENABLED to true.
-        string baseUrl = request.Url.GetLeftPart(UriPartial.Authority);
+        string? baseUrl = (request != null) ? request.Url.GetLeftPart(UriPartial.Authority) : GetBaseUrl(client);
+        bool isFromRequest = request != null;
+        
+        if (baseUrl == null)
+        {
+            throw new InvalidOperationException("Base URL is null. Either use Functions bindings or provide an HTTP request to create the HttpPayload.");
```
nytian, line 180: The BaseUrl could be null when 1. the http request provided is not valid; or 2. the DurableTaskClient is not `FunctionsDurableTaskClient` from which we get the host url information.

test/Worker.Extensions.DurableTask.Tests/FunctionsDurableTaskClientTests.cs
```diff
@@ -53,5 +55,93 @@ public async void TerminateDoesNotThrow()
...
+            Assert.Equal($"{BaseUrl}/instances/{instanceId}/suspend?reason={{{{text}}}}", payload.suspendPostUri);
+            Assert.Equal($"{BaseUrl}/instances/{instanceId}/resume?reason={{{{text}}}}", payload.resumePostUri);
+        }
+    }
+
+    /// <summary>
+    /// A minimal implementation of FunctionContext for testing purposes.
+    /// It's used to create a TestHttpRequestData instance, which requires a FunctionContext.
+    /// </summary>
+    public class TestFunctionContext : FunctionContext
```
nytian, line 107: I’ve added these classes to simulate `HttpResponseData` for testing the `CreateHttpManagementPayload` method with a provided `HttpResponseData` instance, specifically for the CreateHttpManagementPayload_WithHttpRequestData() test. Let me know if you think this approach is overly complex or if there’s a more efficient way to handle this testing. I’m open to removing these classes if needed.:)

cgillum: This is the manual way to do it, which is fine. A way to do it with less code would be to use `Moq` (which can generate these kinds of fakes dynamically), but no need to change what you've already done.

test/Worker.Extensions.DurableTask.Tests/FunctionsDurableTaskClientTests.cs
```diff
@@ -53,5 +55,93 @@ public async void TerminateDoesNotThrow()
...
+        public override IDictionary<object, object> Items { get; set; } = new Dictionary<object, object>();
+        public override IInvocationFeatures Features => throw new NotImplementedException();
+    }
+
+    /// <summary>
+    /// A minimal implementation of IServiceProvider for testing purposes.
+    /// </summary>
+    public class EmptyServiceProvider : IServiceProvider
+    {
+        public object GetService(Type serviceType) => null;
```
cgillum, line 125: To make the warning go away, change `null` to `null!`.
```suggestion
        public object GetService(Type serviceType) => null!;
```

cgillum (approved):
LGTM! A few small suggestions.

src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
```diff
@@ -130,6 +131,14 @@ private class OrchestrationClientInputData
             /// </summary>
             [JsonProperty("rpcBaseUrl")]
             public string? RpcBaseUrl { get; set; }
+
+            /// <summary>
+            /// The base URL of the Azure Functions host, used in the out-of-proc model.
+            /// This URL is sent by the client binding object to the Durable Worker extension,
+            /// allowing the extension to know the host's base URL for constructing complete URLs.
```
davidmrdavid, line 138: nit - we can make this more concrete
```suggestion
            /// allowing the extension to know the host's base URL for constructing management URLs.
```

src/Worker.Extensions.DurableTask/FunctionsDurableTaskClient.cs
```diff
@@ -17,15 +17,16 @@ internal sealed class FunctionsDurableTaskClient : DurableTaskClient
...
         : base(inner.Name)
     {
         this.inner = inner;
         this.QueryString = queryString;
+        this.HttpBaseUrl = httpBaseUrl;
     }
 
     public string? QueryString { get; }
-
+    public string? HttpBaseUrl { get; }
```
davidmrdavid, line 29: Is it allowed for this string to ever be null? If not, I'd recommend removing the `?` from the `string` type.

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
```diff
@@ -120,8 +120,37 @@ public static HttpResponseData CreateCheckStatusResponse(
...
+        }
+
+        try
+        {
+            return SetHeadersAndGetPayload(client, request, null, instanceId);
+        }
+        catch (InvalidOperationException ex)
+        {
+            throw new InvalidOperationException("Failed to create HTTP management payload. " + ex.Message, ex);
+        }
```
davidmrdavid, line 149: Instead of catching and re-throwing the InvalidOperationException. Why don't we just throw a single exception inside of `SetHeadersAndGetPayload` and let that exception propagate out of this method? We may need to move the "Failed to create HTTP management payload" message to "SetHeadersAndGetPayload", but that seems 'ok' to me.

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
```diff
@@ -143,12 +172,31 @@ static string BuildUrl(string url, params string?[] queryValues)
...
+            : $"{baseUrl}/instances/{formattedInstanceId}";
         string? commonQueryParameters = GetQueryParams(client);
-        response.Headers.Add("Location", BuildUrl(instanceUrl, commonQueryParameters));
-        response.Headers.Add("Content-Type", "application/json");
+        
+        if (response != null)
+        {
+            response.Headers.Add("Location", BuildUrl(instanceUrl, commonQueryParameters));
+            response.Headers.Add("Content-Type", "application/json");
+        }
```
davidmrdavid, line 199: why is `response` possibly `null` now?

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
```diff
@@ -143,12 +172,31 @@ static string BuildUrl(string url, params string?[] queryValues)
...
+        // 2. There's no valid HttpRequestData provided
+        string? baseUrl = ((request != null) ? request.Url.GetLeftPart(UriPartial.Authority) : GetBaseUrl(client))
+            ?? throw new InvalidOperationException("Base URL is null. Either use Functions bindings or provide an HTTP request to create the HttpPayload.");
+        bool isFromRequest = request != null;
+
         string formattedInstanceId = Uri.EscapeDataString(instanceId);
-        string instanceUrl = $"{baseUrl}/runtime/webhooks/durabletask/instances/{formattedInstanceId}";
+
+        // The baseUrl differs depending on the source. Eg:
+        // - From request: http://localhost:7071/
```
davidmrdavid, line 187: I'm not sure I understand what it means to be "from request". Can you please clarify?

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
```diff
@@ -143,12 +172,31 @@ static string BuildUrl(string url, params string?[] queryValues)
...
         //       More info: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded.
         //       One potential workaround is to set ASPNETCORE_FORWARDEDHEADERS_ENABLED to true.
-        string baseUrl = request.Url.GetLeftPart(UriPartial.Authority);
+
+        // If HttpRequestData is provided, use its URL; otherwise, get the baseUrl from the DurableTaskClient.
+        // The base URL could be null if:
+        // 1. The DurableTaskClient isn't a FunctionsDurableTaskClient (which would have the baseUrl from bindings)
+        // 2. There's no valid HttpRequestData provided
+        string? baseUrl = ((request != null) ? request.Url.GetLeftPart(UriPartial.Authority) : GetBaseUrl(client))
+            ?? throw new InvalidOperationException("Base URL is null. Either use Functions bindings or provide an HTTP request to create the HttpPayload.");
```
davidmrdavid, line 181: this is nitpicky, but I feel this is a bit difficult to read due the combination of ternary expressions the `<condition> ? <true-case> : <false-case>` syntax with the nullable syntax ` <value> ?? <code-to-execute-if-value-is-null>`.

I just feel that there's a lot happening in very little space :) . I would using either the ternary expression, or the nullable syntax, but not both at the same time. The other can be expressed as a simple if-statement.

Not a blocking concern, just a stylistic nit. Feel free to dismiss.

davidmrdavid (commented):
A few questions, looks great. I appreciate the test!

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
```diff
@@ -120,8 +120,37 @@ public static HttpResponseData CreateCheckStatusResponse(
...
+    /// <summary>
+    /// Creates an HTTP management payload for the specified orchestration instance.
+    /// </summary>
+    /// <param name="client">The <see cref="DurableTaskClient"/>.</param>
+    /// <param name="instanceId">The ID of the orchestration instance.</param>
+    /// <param name="request">Optional HTTP request data to use for creating the base URL.</param>
+    /// <returns>An object containing instance control URLs.</returns>
+    /// <exception cref="ArgumentException">Thrown when instanceId is null or empty.</exception>
+    /// <exception cref="InvalidOperationException">Thrown when a valid base URL cannot be determined.</exception>
+    public static object CreateHttpManagementPayload(
```
davidmrdavid, line 132: Quick nit - I noticed this returns `object` which is very general. In the WebJobs extension, we used to return `HttpManagementPayload` as the type, see here: https://github.com/Azure/azure-functions-durable-extension/blob/8a5db71b7a1a67edbc72e55f339b4cc6c090ffcb/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs#L152

Should we not also introduced a worker extension version of this type, so that we can return a more precise return type?

davidmrdavid: cc/ @cgillum, to confirm. Curious if it makes sense to return `object` for this particular API type, not sure if there's something I'm missing

cgillum: Oh, that's a good point. It made sense to use `object` and anonymous types in the case where we were serializing the object into the HTTP response. However, for the purposes of this new, more general purpose API, it's better to create a new `HttpManagementPayload` class so that users of this method have a strongly-typed object they can work with. Thanks @davidmrdavid for bringing this to my attention.

src/Worker.Extensions.DurableTask/FunctionsDurableTaskClient.cs
nytian: If  the `DurableTaskClient` isn’t an instance of `FunctionsDurableTaskClient`, then HttpBaseUrl could be null, as it’s only an attribute of `FunctionsDurableTaskClient`. I’m not entirely sure in what scenario the DurableTaskClient wouldn’t be a FunctionsDurableTaskClient—perhaps it could happen if bindings aren’t being used. That’s why I added an exception for cases where the base URL creation fails.

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
nytian: `SetHeadersAndGetPayload` is also invoked by the existing method [CreateCheckStatusResponse](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs#L65), which provides an HTTP response to `SetHeadersAndGetPayload`. In the new API, `CreateHttpManagementPayload`, there won’t be an HTTP response parameter. Therefore, when `SetHeadersAndGetPayload` is called by `CreateHttpManagementPayload`, the response will be null; however, if it’s called by `CreateCheckStatusResponse`, the response will not be null.

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
nytian: There are two ways to obtain the BaseUrl: (1) from HttpRequestData, which refers to the request in this context. And this usually happen when cx already have a Http bindings, or (2) from the Functions host, if HttpRequestData is not provided, as this information is passed through the DurableClient bindings.

Let me try to update this comment to be more clear. 

Eg, calling CreateHttpManagementPayload with HttpRequestData like this:

image

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
nytian: Haha I just used this so that I could save more space. I will change this to if-statement to make it more clear. Thanks!

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
davidmrdavid: Awesome. @nytian, in that case, do you mind introducing an `HttpManagementPayload` type here, so we can unblock this PR? Thanks!

src/Worker.Extensions.DurableTask/HttpManagementPayload.cs
```diff
@@ -0,0 +1,96 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Functions.Worker;
```
nytian, line 10: This class is identical to the one we have in [WebJobs.Extensions.DurableTask](https://github.com/Azure/azure-functions-durable-extension/tree/dev/src/WebJobs.Extensions.DurableTask)(https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/HttpManagementPayload.cs)

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
nytian: Thanks for the suggestion and context! Just added :)

src/Worker.Extensions.DurableTask/HttpManagementPayload.cs
```diff
@@ -0,0 +1,96 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
```
davidmrdavid, line 4:
```suggestion

// This is a copy of: https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/HttpManagementPayload.cs
```

davidmrdavid (approved):

test/Worker.Extensions.DurableTask.Tests/FunctionsDurableTaskClientTests.cs
```diff
@@ -53,5 +53,51 @@ public async void TerminateDoesNotThrow()
...
+            var mockFunctionContext = new Mock<FunctionContext>();
+            var mockHttpRequestData = new Mock<HttpRequestData>(mockFunctionContext.Object);
+            mockHttpRequestData.SetupGet(r => r.Url).Returns(new Uri(requestUrl));
+
+            dynamic payload = client.CreateHttpManagementPayload(instanceId, mockHttpRequestData.Object);
+
+            AssertHttpManagementPayload(payload, "http://localhost:7075/runtime/webhooks/durabletask", instanceId);
+        }
+
+        private static void AssertHttpManagementPayload(dynamic payload, string BaseUrl, string instanceId)
```
cgillum: Should we replace `dynamic` with `HttpManagementPayload` here, and in the calling functions?

cgillum (approved):
Just the one suggestion on the test code, but otherwise LGTM!

PR #2953, opened 2024-11-07
Update MaxQueuePollingInterval default for Flex Consumption apps

bachuv:
This PR updates the `MaxQueuePollingInterval` to a default of 1 second instead of 30 seconds if the app is running on a Flex Consumption plan to help with cold start latency specific to Durable Functions Flex Consumption apps.

### Pull request checklist


cgillum (approved):
The change looks good to me! I think we just need to add some code comments explaining *why* we have this new logic (to remind our future selves and/or future maintainers).

cgillum:
Oh, and let's make sure we update the release notes with this information so that we have this change nicely documented for people to see.

src/WebJobs.Extensions.DurableTask/Options/AzureStorageOptions.cs
```diff
@@ -162,7 +163,27 @@ public string TrackingStoreConnectionStringName
         /// Gets or sets the maximum queue polling interval.
         /// </summary>
         /// <value>Maximum interval for polling control and work-item queues.</value>
-        public TimeSpan MaxQueuePollingInterval { get; set; } = TimeSpan.FromSeconds(30);
+        public TimeSpan MaxQueuePollingInterval
+        {
+            get
```
AnatoliB, line 172: If I'm reading this implementation correctly, this `get` will _always_ override the `maxQueuePollingInterval` field with the default based on the env variable, so `set` will not even matter, which is probably not the intent. Consider adding a unit test that confirms that the value passed to `set` is actually returned by the next `get`. On the implementation side, consider using `Lazy<TimeSpan>` for `maxQueuePollingInterval`.

cgillum: Ah, I totally missed that. Good call - we need to check to see whether the value has been initialized already and not initialize it again. Another option to consider is defaulting to a dummy value, like `TimeSpan.Zero` which has the benefit of being a non-breaking change.

bachuv: I added a change to check for if `this.maxQueuePollingInterval == TimeSpan.Zero` (since `TimeSpan.Zero` is the default value for `TimeSpan`) before setting the queue polling default setting. If it's not `TimeSpan.Zero`, then it just returns the current value of `this.maxQueuePollingInterval.` Is this a reasonable fix or should I go with another approach (e.g. using `Lazy<TimeSpan>`)?

I also added unit tests for `AzureStorageOptions`.

AnatoliB: The `Lazy<TimeSpan>` idea will not actually work because `Lazy` doesn't have a setter, and I forgot about it. Your approach makes sense, thank you.

AnatoliB (approved):

PR #2967, opened 2024-11-15
Make scalers into per-task hub singletons

### LINKED ISSUE #2981: Unexplained rise in storage activity causing higher costs
We have multiple deployments of a Functions App across several tenants. Since the 1st November we have seen a sudden rise in transactions across all storage accounts for these apps, without any clear cause, including on deployments which are idle. This has cause a large increase in costs. The bug has affected different versions of the app, with different runtime versions, in similar ways

Image

Image

The fall in transactions shows where I have stopped the relevant function app.

The issue seems similar to those described in https://github.com/Azure/Azure-Functions/issues/2231 and in https://github.com/Azure/Azure-Functions/issues/2417, although in our case the majority of transactions are queue transactions. 

The apps shown above are v4, .Net 6 (LTS) In-Process, running on Windows, with individual Consumption plans. The problem also occurred with a version of the app in v3. The apps do not use the AzureWebJobsDashboard setting.

cgillum:
This PR improves the efficiency of scale logic with the Azure Storage backend by creating a singleton scaler object for an entire task hub rather than having one per function, which is redundant since each scaler does the exact same work (regardless of function).

Resolves issue where storage costs increase dramatically when Scale Controller v3 is in use (same for when runtime-driven scaling is in use).

FYI @AnatoliB 

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
@@ -35,6 +35,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         private readonly JObject storageOptionsJson;
         private readonly ILogger logger;
 
+        private readonly object initLock = new object();
+
+#if !FUNCTIONS_V1
+        private DurableTaskScaleMonitor singletonScaleMonitor;
+#endif
+
+#if FUNCTIONS_V3_OR_GREATER
+        private DurableTaskTargetScaler singletonTargetScaler;
+#endif
+
         public AzureStorageDurabilityProvider(
             AzureStorageOrchestrationService service,
             IStorageAccountProvider storageAccountProvider,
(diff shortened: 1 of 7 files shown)


src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
```diff
@@ -242,16 +251,22 @@ public override bool TryGetScaleMonitor(
...
-            scaleMonitor = new DurableTaskScaleMonitor(
-                functionId,
-                functionName,
-                hubName,
-                storageAccount,
-                this.logger,
-                metricsProvider);
-            return true;
+            lock (this.initLock)
+            {
```
davidmrdavid, line 255: a bit nervous about a lock in the ScaleController (and I'm not sure it is really needed, I suspect other instances of the scaler class will be garbage collected away so long as the use the same ID), but the implementation seems safe.

cgillum: I'm not sure if it's needed either, but any particular thing that makes you nervous? I'm fine with removing it.

bachuv (approved):

PR #2974, opened 2024-11-26
Make scalers into per-task hub singletons (cherry pick #2967)

cgillum:
Cherry pick of PR #2967 into the `v3.x` branch.

Needed to resolve several merge conflicts due to the different Azure Storage SDK types, but the changes were pretty straightforward.

I also updated the nuget package versions as part of this PR so that we don't have to do a separate PR for this.


nytian (approved):

nytian:
One test, `DurableTaskEndToEndTests.DurableEntity_AsyncIO`, continues to fail in the CI , returning a 403 status code instead of the expected 200. I tested this locally and it worked fine, so it seems like there might be a connection issue here. Since the code in this PR doesn't affect anything related to this test, I'll go ahead and merge the PR first.

PR #2998, opened 2024-12-17
Correlate client and orchestrator functions for DF .NET Isolated Distributed Tracing

bachuv:
These changes help correlate client and orchestrator functions in Durable Functions .NET Isolated. 

Gantt chart (shows HTTP trigger and orchestrator function in one trace)
image

### Pull request checklist


src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -154,8 +155,22 @@ public override Task<Empty> Hello(Empty request, ServerCallContext context)
             {
                 try
                 {
+                    ActivitySource activityTraceSource = new ActivitySource("DurableTask.WebJobs");
```
bachuv: For context, the source name is "DurableTask.WebJobs" because we load the `DurableTelemetryModule` and it looks for sources that start with "DurableTask" ([related code link](https://github.com/Azure/durabletask/blob/93b2dde58366da95165fe8c8512b577c84628813/src/DurableTask.ApplicationInsights/DurableTelemetryModule.cs#L43)). I'm working on updating the listeners to add a new one ("WebJobs.Extensions.DurableTask") when we initialize the module so we can remove this "DurableTask.WebJobs" that's just used to test for now.

src/Worker.Extensions.DurableTask/FunctionsDurableTaskClient.cs
```diff
@@ -76,6 +77,12 @@ public override Task<string> ScheduleNewOrchestrationInstanceAsync(
         StartOrchestrationOptions? options = null,
         CancellationToken cancellation = default)
     {
+        Activity? currActivity = Activity.Current;
+        if (options == null && options?.ParentTraceId == null && currActivity != null)
+        {
+            options = new StartOrchestrationOptions(ParentTraceId: currActivity?.Id?.ToString());
+        }
```
bachuv: I added the `ParentTraceId` information to the `StartOrchestrationOptions`, but I'm wondering if there's a better place to store this information since `StartOrchestrationOptions` can be configured by the customer and we don't necessarily need customers to have access to `ParentTraceId`.

jviau (changes requested):
I'd like to avoid having a new activity source and span added. We should try and having the trace ID captured from the worker directly added to the enqueued orchestration. I think to accomplish this we will want to hydrate a shim activity which has its ID set to the trace context from the gRPC payload.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -154,8 +156,34 @@ public override Task<Empty> Hello(Empty request, ServerCallContext context)
...
+                        string traceId = traceParentContextSplit[1];
+                        string spanId = traceParentContextSplit[2];
+
+                        // reflection to set trace id and span id of newActivity
+                        if (newActivity != null)
+                        {
+                            typeof(Activity).GetField("_traceId", BindingFlags.Instance | BindingFlags.NonPublic)?.SetValue(newActivity, traceId);
+                            typeof(Activity).GetField("_spanId", BindingFlags.Instance | BindingFlags.NonPublic)?.SetValue(newActivity, spanId);
+                        }
+                    }
```
bachuv: I used reflection here to set the `Activity`'s trace ID and span ID, but don't set the rehydrated `Activity`s parent Ids since that information isn't passed through the `CreateInstanceRequest` type which ends up not correlating the client and orchestration functions in the distributed trace. I think we need to pass additional information to successfully correlate the client and orchestrator functions.

As a reference, the code below is how we rehydrated an `Activity` for .NET In-Proc in DTFx:
1. Get the parent context which is saved in the `ExecutionStartedEvent` and set it as the `parentContext`

```
if (!startEvent.TryGetParentTraceContext(out ActivityContext activityContext))
...
Activity? activity = ActivityTraceSource.StartActivity(
          activityName,
          kind: activityKind,
          parentContext: activityContext,
          startTime: startTime);
```
https://github.com/Azure/durabletask/blob/d7f33b07107eef66898538f195c184378bee8745/src/DurableTask.Core/Tracing/TraceHelper.cs#L78

2. Set the rehydrated `Activity` with the original `Activity`s trace ID and span ID:
```
if (startEvent.ParentTraceContext.Id != null && startEvent.ParentTraceContext.SpanId != null)
{
    activity.SetId(startEvent.ParentTraceContext.Id!);
    activity.SetSpanId(startEvent.ParentTraceContext.SpanId!);
}
```

https://github.com/Azure/durabletask/blob/d7f33b07107eef66898538f195c184378bee8745/src/DurableTask.Core/Tracing/TraceHelper.cs#L107

In the example above, we had the following information:
- ParentTraceContext to point to the parent
- Id for the rehydrated `Activity`
- spanId for the rehydrated `Activity`

With the .NET Isolated approach, we only have the last two items, not the information about its parent. **Would it be worth updating [`CreateInstanceRequest`](https://github.com/microsoft/durabletask-protobuf/blob/0d2af05978e59509359e1b9a97166728790ceabe/protos/orchestrator_service.proto#L264) to have the following fields?**
- TraceContext parentTraceContext**
- TraceContext currentTraceContext

[`TraceContext`](https://github.com/microsoft/durabletask-protobuf/blob/0d2af05978e59509359e1b9a97166728790ceabe/protos/orchestrator_service.proto#L62) also has a `spanId` field which would be another approach to help pass this information since we could create the Id using the traceId  (from the parent) and spanId, but it's deprecated. **Is it possible to add that field back or create a new one?**

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -154,8 +156,34 @@ public override Task<Empty> Hello(Empty request, ServerCallContext context)
             {
                 try
                 {
+                    Activity? newActivity = new Activity("gRPC start orchestration");
```
cgillum: Does this span name match the convention we're using for other spans?

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -154,8 +156,34 @@ public override Task<Empty> Hello(Empty request, ServerCallContext context)
...
+
+                    string traceParentContext = request.ParentTraceContext.TraceParent.ToString();
```
cgillum: Naming: Should this variable be named differently? It doesn't seem to be the trace context. Rather, it seems to be the traceparent value.
```suggestion
                    string traceParentString = request.ParentTraceContext.TraceParent.ToString();
```

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
cgillum: I wonder if we might be over complicating things a bit. Instead of reconstructing the trace activity we got from the client using reflection, can we start a new trace activity and have the incoming trace context be its parent?  Something like this (this code won't compile, but is just illustrative):

```csharp
Activity? activity = ActivityTraceSource.StartActivity(
          "gRPC create orchestration",
          kind: activityKind,
          parentContext: request.ExecutionStarted.ParentTraceContext,
          startTime: DateTime.UtcNow);
```

I feel like this would be cleaner and would help capture the fact that there is a gRPC network hop between the language worker and our host extension.

jviau: I wouldn't expect you to need the parent information here. The incoming trace-context _is_ the parent information.

The scenario is, we are about to call `TaskHubClient` to schedule an orchestration. In that call, `TaskHubClient` is going to create a `send_orchestration` activity with kind=client. At time of this activity creation, we want to make sure it has access to whatever span was active when the customer's code asked to schedule this orchestration. This is done by ensuring `Activity.Current` is set where its id (traceId, spanId) are set to what we want the _parent_ of the mentioned activity to be. This process should **not** look at `Activity.Current`'s parent information.

Creating a new `Activity.Current` with traceId and spanId set _should_ be sufficient. We shouldn't need parent information. If this isn't working, we may need debug into `TaskHubClient` and its activity creation to see what is going on.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
bachuv: I removed this new Activity.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
bachuv: I removed this.

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
```diff
@@ -45,13 +46,24 @@ public TelemetryActivator(IOptions<DurableTaskOptions> options, INameResolver na
...
+            if (this.webJobsTelemetryModule != null)
+            {
+                this.webJobsTelemetryModule.DisposeAsync();
+            }
+
+            return default;
         }
 
         /// <inheritdoc/>
         public void Dispose()
```
jviau, line 58: You can switch this implementation to be `this.DisposeAsync().AsTask().GetAwaiter().GetResult();`

src/WebJobs.Extensions.DurableTask/Correlation/TraceActivityConstants.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Correlation
+{
+    internal class TraceActivityConstants
```
jviau, line 6: Consider combining this into `Schema`. You could even make it helper functions.

``` CSharp
internal static class Schema
{
    internal static class SpanNames
    {
        internal static string CreateOrchestration(string name, string? version)
           => FormatName("create_orchestration", name, version);

        static string FormatName(string prefix, string name, string? version)
            => string.IsNullOrEmpty(version) ? $"{prefix}:{name}" : $"{prefix}:{name}@{version}";
    }
}
```

Now it is used as `Schema.SpanNames.CreateOrchestration(name, version)`

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -154,27 +159,53 @@ public override Task<Empty> Hello(Empty request, ServerCallContext context)
...
-                    }
-                    else
+                        Name = request.Name,
+                        Version = request.Version,
+                        OrchestrationInstance = instance,
+                        ScheduledStartTime = request.ScheduledStartTimestamp?.ToDateTime(),
+                    };
+
+                    // Get the parent trace context from CreateInstanceRequest
+                    string? traceParent = request.ParentTraceContext.TraceParent;
```
jviau: Recommend a null-check just to be safe

```suggestion
                    string? traceParent = request.ParentTraceContext?.TraceParent;
```

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -154,27 +159,53 @@ public override Task<Empty> Hello(Empty request, ServerCallContext context)
...
+                    string? traceState = request.ParentTraceContext.TraceState;
+
+                    // Create a new activity with the parent context
+                    ActivityContext.TryParse(traceParent, traceState, out ActivityContext parentActivityContext);
+                    Activity? scheduleOrchestrationActivity = StartActivityForNewOrchestration(executionStartedEvent, parentActivityContext);
```
jviau: Rely on dispose for stopping an activity. You can remove the `.Stop()` call with this change,
```suggestion
                    using Activity? scheduleOrchestrationActivity = StartActivityForNewOrchestration(executionStartedEvent, parentActivityContext);
```

jviau (approved):
Looks good, but some additional things to check:

1. With the move away from `TaskHubClient`, do we have other usage in this class? If so, can we create a follow up work item to replace those with direct `IOrchestrationService` usage and then cleanup methods around `TaskHubClient`.
2. Do a pass over `TaskHubClient` and see if there are other scenarios that are now 'missed' by no longer using it, and ensure they are handled directly here.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -154,27 +159,53 @@ public override Task<Empty> Hello(Empty request, ServerCallContext context)
...
+
+                        return new P.CreateInstanceResponse
+                        {
+                            InstanceId = instance.InstanceId,
+                        };
                     }
-
-                    // TODO: should this not include the ExecutionId and other elements of the taskhubClient response?
-                    return new P.CreateInstanceResponse
+                    catch (InvalidOperationException)
```
cgillum: Isn't this already handled by the `catch` statements in the outer try/catch block? Seems like you should be able to remove this inner try/catch.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -195,6 +226,64 @@ public override Task<Empty> Hello(Empty request, ServerCallContext context)
                 }
             }
 
+            private DurabilityProvider GetDurabilityProviderToScheduleOrchestration(ServerCallContext context)
```
cgillum: Is this new method needed? Could we instead use the existing `GetClient(context)` method?

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -154,27 +159,53 @@ public override Task<Empty> Hello(Empty request, ServerCallContext context)
...
+                        await this.GetDurabilityProviderToScheduleOrchestration(context).CreateTaskOrchestrationAsync(
+                            new TaskMessage
+                            {
+                                Event = executionStartedEvent,
+                                OrchestrationInstance = instance,
+                            },
+                            this.GetStatusesNotToOverride());
+
+                        // Stop the activity
+                        scheduleOrchestrationActivity?.Stop();
```
cgillum: Should this be done in a finally block? Or do we only want to do it if the call succeeds?

cgillum (commented):
Overall looks good, but it would be great if we could apply Jacob's feedback, which I agree with.

Also, looks like the CI might be failing. Do we understand why?

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
jviau: If switched to a `using` statement, that will stop as part of disposal.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
bachuv: I replaced it with a call to the existing `GetDurabilityProvider(context)` method.

bachuv:
The CI is failing with the following error since the protobuf changes haven't been incorporated yet. This is expected and will go away once the related protobuf PR is merged and released.

`error CS1061: 'CreateInstanceRequest' does not contain a definition for 'ParentTraceContext' and no accessible extension method 'ParentTraceContext' accepting a first argument of type 'CreateInstanceRequest' could be found (are you missing a using directive or an assembly reference?) `

cgillum:
Looks good. One more thing: can we add an end-to-end test with @andystaples framework to verify that this is working? I was hoping we could do something really simple, like the following:

1. From a test code, start a new trace activity (I have some code [here](https://dev.azure.com/msazure/One/_git/AAPT-DTMB?path=/src/Backend/Microsoft.DTMB.Tests.Integration/DurableTaskFrameworkTests.cs&version=GC1f163c2219338a6056e8ab531c7350d682efb2ce&line=66&lineEnd=70&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) which might be useful as a reference)
2. Start an orchestration using HttpClient (presumably the trace ID will get picked up automatically) with no input.
3. The orchestration calls a single activity called "GetDistributedTraceId"
4. Return the value of `Activity.Current` from the activity function as the output, and return that as the output of the orchestrator
5. In the test code, wait for the orchestration to complete and then compare the output value (the trace ID of the activity) with the trace ID that was generated in step 1.

Do we have enough pieces in place to make this end-to-end test work?

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
```diff
@@ -75,12 +75,12 @@ public void Initialize(TelemetryConfiguration configuration)
                 if (this.options.Tracing.Version == Options.DurableDistributedTracingVersion.V2)
                 {
                     DurableTelemetryModule module = new DurableTelemetryModule();
-                    module.Initialize(telemetryConfiguration);
-                    this.telemetryModule = module;
+                    module.Initialize(configuration);
+                    this.TelemetryModule = module;
 
                     WebJobsTelemetryModule webJobsModule = new WebJobsTelemetryModule();
```
jviau, line 81: Small design suggestion: we could reduce down to needing only 1 wrapped telemetry module (in this class) with a minor change: `WebJobsTelemetryModule` can wrap/include the `DurableTelemetryModule` internally, and forward the `Initialize` call to it.

cgillum (approved):
I love that we have an end-to-end test for this now!

test/e2e/Tests/Tests/DistributedTracingTests.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            ShouldListenTo = source => source.Name == "DistributedTracingTests",
+            Sample = (ref ActivityCreationOptions<ActivityContext> _) => ActivitySamplingResult.AllDataAndRecorded,
+            ActivityStarted = activity => { /* Handle activity started */ },
+            ActivityStopped = activity => { /* Handle activity stopped */ }
+        };
+        ActivitySource.AddActivityListener(_activityListener);
+    }
+
+    [Theory]
```
andystaples: This can probably be a [Fact] if it's only one test that follows this pattern

test/e2e/Tests/Tests/DistributedTracingTests.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        using HttpResponseMessage response = await HttpHelpers.InvokeHttpTrigger(functionName, "");
+
+        string statusQueryGetUri = await DurableHelpers.ParseStatusQueryGetUriAsync(response);
+        var orchestrationDetails = await DurableHelpers.GetRunningOrchestrationDetailsAsync(statusQueryGetUri);
+        string output = orchestrationDetails.Output;
+        ActivityContext.TryParse(output, null, out ActivityContext activityContext);
+
+        Assert.Equal(activity?.TraceId.ToString(), activityContext.TraceId.ToString());
+    }
```
andystaples, line 51: Are there other tests that are applicable here? Making sure that logs emitted by the function, or internal spans created by the function are all linked correctly?

bachuv: We won't need to check any logs emitted by the function for distributed tracing. We also don't need to check for any internal spans to check if the spans are linked correctly because the trace Id gives us that information. We started an `Activity` in the test and then checked if our Durable Activity function has the same trace Id as the `Activity` that was started in the test (this means that they are linked in the same trace).

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
bachuv: Thanks for the suggestion! I'll make this change in a follow up PR. I tracked this change in GitHub issue https://github.com/Azure/azure-functions-durable-extension/issues/3038.

jviau: You don't have to make the change at all if you don't want.

cgillum (approved):
Protobuf changes look good to me.

PR #3000, opened 2024-12-18
Add integration test framework for isolated

andystaples:
- Add integration testing project
- Methods for interacting with "func", "azurite", etc
- One app with basic "Hello Cities" scenario
- Build script that 
  - installs a local copy of Core Tools
  - builds and deploys the extensions to the function app
  - builds the function app
  - installs and launches azurite

### Pull request checklist


cgillum (commented):
I suggest a slightly revised file structure for this, basically putting all end-to-end related code under a `test/e2e` directory.

/test/e2e/Apps/DotNetIsolated/App.csproj
/test/e2e/Apps/DotNetIsolated/...
/test/e2e/Tests/E2ETests.csproj
/test/e2e/Tests/...

In the future, we can add other language-specific apps under the /test/e2e/Apps/ directory too.

test/DotnetIsolatedE2ETests/Fixtures/FixtureHelpers.cs
```diff
@@ -0,0 +1,89 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.Durable.Tests.DotnetIsolatedE2E
+
+{
+    public static class FixtureHelpers
+    {
+        public static Process GetFuncHostProcess(bool enableAuth = false)
+        {
+            var funcProcess = new Process();
```
cgillum: nit: let's wait to create the `Process` object until the time when we're actually use it. This is especially important for disposable objects, like `Process`, which can cause memory/handle leaks if they are created but not disposed (which will happen if you create it and we throw an exception in line 25 or line 39).

test/DotnetIsolatedE2ETests/Fixtures/FixtureHelpers.cs
```diff
@@ -0,0 +1,89 @@
...
+            string? e2eHostJson = Directory.GetFiles(e2eAppBinPath, "host.json", SearchOption.AllDirectories).FirstOrDefault();
+
+            if (e2eHostJson == null)
+            {
+                throw new InvalidOperationException($"Could not find a built worker app under '{e2eAppBinPath}'");
+            }
+
+            var e2eAppPath = Path.GetDirectoryName(e2eHostJson);
+
+            var cliPath = Path.Combine(rootDir, @"test/DotnetIsolatedE2ETests/Azure.Functions.Cli/func");
```
cgillum: This seems to imply that the func.exe tool is checked into the repo, which I assume is not your intention. I recommend putting it into a temporary directory instead.

andystaples: func isn't checked into the repo, but running the build-e2e-test.ps1 script will install it into this location. Useful for several reasons - 
1) Durable "power users" often modify the Core Tools in the PATH, we want an unmodified copy running for these tests
2) Pipeline automation will have to take this step anyway, might as well do it in this script

andystaples: I can modify the install directory - but for the pipeline scenario it is most useful to have it inside the repo folder structure somewhere. Dotnet worker does it in the root.

test/DotnetIsolatedE2EApps/MainApp/app.csproj
```diff
@@ -0,0 +1,32 @@
...
+    <AzureFunctionsVersion>v4</AzureFunctionsVersion>
+    <OutputType>Exe</OutputType>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.21.0" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="3.1.2" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.DurableTask" Version="1.3.2" />
```
bachuv: Is it possible to add a project reference to the Worker Extensions package like [this example](https://github.com/Azure/azure-functions-durable-extension/blob/53abf0f5d893340e2ebeb24e352377c5d0c15393/test/SmokeTests/OOProcSmokeTests/DotNetIsolated/DotNetIsolated.csproj#L24) instead of referencing a specific version?

test/DotnetIsolatedE2ETests/Fixtures/FunctionAppFixture.cs
bachuv, line 1: Can we resolve the warnings that are showing up in this PR?

test/DotnetIsolatedE2ETests/DotnetIsolatedE2ETests.csproj
bachuv, line 1: Can you add a simple README for this project explaining the scripts that we have and how the fixtures and helper files are used?

bachuv (commented):
Thanks for opening this PR!

test/DotnetIsolatedE2ETests/Fixtures/FixtureHelpers.cs
andystaples: Done

test/DotnetIsolatedE2ETests/Fixtures/FixtureHelpers.cs
andystaples: Moved the install location for Core Tools and the start location for azurite to a owned folder in [System.IO.Path]::GetTempPath()/similar.

test/DotnetIsolatedE2EApps/MainApp/app.csproj
andystaples: Agreed, this would solve several issues with the current solution, including: 
- Dynamic modifying of app.csproj with the build script will lead to unnecessary noise in commits and/or work excluding these files from unrelated PRs
- Would remove the requirement of manually bumping extension versions to run test locally
- Would simplify the build script 

Need to explore this further. The extension build process in isolated has some gotchas but it may be possible to work around

test/DotnetIsolatedE2ETests/Fixtures/FunctionAppFixture.cs
andystaples: Should be done - let me know if you see specific new ones or ones I missed

test/DotnetIsolatedE2ETests/DotnetIsolatedE2ETests.csproj
andystaples: Done

andystaples: I focused more on the build scripts and practical usage steps, happy to add some info on fixtures and helpers if needed but for the most part, the goal is to abstract them to where adding or updating tests will not require interacting with those files at all

test/e2e/Tests/Fixtures/FunctionAppFixture.cs
```diff
@@ -0,0 +1,165 @@
...
+                FixtureHelpers.StartProcessWithLogging(this._funcProcess, this._logger);
+
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    // ensure child processes are cleaned up
+                    _jobObjectRegistry = new JobObjectRegistry();
+                    _jobObjectRegistry.Register(this._funcProcess);
+                }
+
+                var httpClient = new HttpClient();
```
cgillum: Minor: It's important to dispose `HttpClient` objects when we're doing with them to avoid resource leaks. A simple `using` on this line will ensure that this object gets cleaned up at the end of this method invocation.
```suggestion
                using var httpClient = new HttpClient();
```

test/e2e/Tests/Helpers/HttpHelpers.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Durable.Tests.DotnetIsolatedE2E
+{
+    class HttpHelpers
+    {
+        public static async Task<HttpResponseMessage> InvokeHttpTrigger(string functionName, string queryString = "")
+        {
+            // Basic http request
+            HttpRequestMessage request = GetTestRequest(functionName, queryString);
```
cgillum: Same suggestion about disposing disposable objects to prevent resource leaks.
```suggestion
            using HttpRequestMessage request = GetTestRequest(functionName, queryString);
```

test/e2e/Tests/Helpers/HttpHelpers.cs
```diff
@@ -0,0 +1,78 @@
...
+            request.Content.Headers.ContentType = new MediaTypeHeaderValue(mediaType);
+            request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue(mediaType));
+            return await GetResponseMessage(request);
+        }
+
+        public static async Task<bool> InvokeHttpTrigger(string functionName, string queryString, HttpStatusCode expectedStatusCode, string expectedMessage, int expectedCode = 0)
+        {
+            string uri = $"{Constants.FunctionsHostUrl}/api/{functionName}{queryString}";
+            using (var request = new HttpRequestMessage(HttpMethod.Get, uri))
+            {
```
cgillum: nit: for this case, you don't need the curly braces with the using statement. Rather, you can save some indenting (which makes the code a bit cleaner) by removing the `{}`. In that case, line 34 would simply be:
```csharp
using var request = new HttpRequestMessage(HttpMethod.Get, uri);
```

test/e2e/Tests/Helpers/HttpHelpers.cs
```diff
@@ -0,0 +1,78 @@
...
+                request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("text/plain"));
+                var response = await GetResponseMessage(request);
+
+                Console.WriteLine(
+                    $"InvokeHttpTrigger: {functionName}{queryString} : {response.StatusCode} : {response.ReasonPhrase}");
```
cgillum: Will console logs be captured in the test output? Normally we write logs in Xunit to `ITestOutputHelper`.

test/e2e/Tests/Helpers/HttpHelpers.cs
```diff
@@ -0,0 +1,78 @@
...
+                {
+                    return false;
+                }
+
+                if (!string.IsNullOrEmpty(expectedMessage))
+                {
+                    string actualMessage = await response.Content.ReadAsStringAsync();
+                    Console.WriteLine(
+                        $"InvokeHttpTrigger: expectedMessage : {expectedMessage}, actualMessage : {actualMessage}");
+                    return actualMessage.Contains(expectedMessage);
```
cgillum: Could/should we use `Assert.Contains(...)` here instead?

test/e2e/Tests/Fixtures/FixtureHelpers.cs
```diff
@@ -0,0 +1,79 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.Durable.Tests.DotnetIsolatedE2E
```
cgillum: tip: to reduce indentation, we should use file-scoped namespaces and drop the curly braces. Just need to add a `;` at the end of the `namespace` declaration. I would like us to do this for all new code we write.

```suggestion
namespace Microsoft.Azure.Durable.Tests.DotnetIsolatedE2E;
```

test/e2e/Tests/Tests/HelloCitiesTest.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            _fixture = fixture;
+            _fixture.TestLogs.UseTestLogger(testOutputHelper);
+        }
+
+        [Theory]
+        [InlineData("HelloCities_HttpStart", "", HttpStatusCode.Accepted, "")]
+        public async Task HttpTriggerTests(string functionName, string queryString, HttpStatusCode expectedStatusCode, string expectedMessage)
+        {
+            HttpResponseMessage response = await HttpHelpers.InvokeHttpTrigger(functionName, queryString);
```
cgillum:
```suggestion
            using HttpResponseMessage response = await HttpHelpers.InvokeHttpTrigger(functionName, queryString);
```

test/e2e/Tests/Tests/HelloCitiesTest.cs
```diff
@@ -0,0 +1,37 @@
...
+            string actualMessage = await response.Content.ReadAsStringAsync();
+
+            Assert.Equal(expectedStatusCode, response.StatusCode);
+
+            if (!string.IsNullOrEmpty(expectedMessage))
+            {
+                Assert.False(string.IsNullOrEmpty(actualMessage));
+                //Assert.Contains(expectedMessage, actualMessage);
```
cgillum: Why is this assert commented out?

test/e2e/Tests/Constants.cs
```diff
@@ -0,0 +1,16 @@
...
+
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Azure.Durable.Tests.DotnetIsolatedE2E
+{
+    internal class Constants
+    {
+        public static IConfiguration Configuration = TestUtility.GetTestConfiguration();
+
+        internal static string FunctionsHostUrl = Configuration["FunctionAppUrl"] ?? "http://localhost:7071";
```
cgillum: nit: since these are in a "Constants" file, we should probably make them readonly.

```suggestion
        internal static readonly string FunctionsHostUrl = Configuration["FunctionAppUrl"] ?? "http://localhost:7071";
```

test/e2e/Tests/build-e2e-test.ps1
```diff
@@ -0,0 +1,167 @@
...
+    exit $LastExitCode 
+  }
+}
+
+$FUNC_CLI_DIRECTORY = Join-Path $ProjectTemporaryPath 'Azure.Functions.Cli'
+if($SkipCoreTool -or (Test-Path $FUNC_CLI_DIRECTORY))
+{
+  Write-Host "---Skipping Core Tools download---"  
+}
+elseif (Test-Path $FUNC_CLI_DIRECTORY)
```
cgillum: Seems this `elseif` is redundant with the check above.

test/e2e/Tests/e2e-tests-readme.md
```diff
@@ -0,0 +1,59 @@
...
+## Prerequisites
+
+- Visual Studio IDE
+- PowerShell
+- npm/Node
+- .NET SDK
+
+## Running the E2E Tests
+
+### Step 1: Increment the host and worker package versions
```
cgillum: Is this step necessary if I just want to run the tests without making any changes?

test/e2e/Apps/BasicDotNetIsolated/app.csproj
```diff
@@ -0,0 +1,32 @@
...
+    <AzureFunctionsVersion>v4</AzureFunctionsVersion>
+    <OutputType>Exe</OutputType>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.21.0" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="3.0.2" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.DurableTask" Version="1.2.2" />
```
cgillum: I wonder why we wouldn't just use a project reference here instead of a package reference? That way we never have to worry about keeping this version number up to date.

test/e2e/Apps/BasicDotNetIsolated/app.csproj
```diff
@@ -0,0 +1,32 @@
...
+    <TargetFramework>net8.0</TargetFramework>
+    <AzureFunctionsVersion>v4</AzureFunctionsVersion>
+    <OutputType>Exe</OutputType>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.21.0" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="3.0.2" />
```
cgillum: Is there a particular reason to include the WebJobs package reference here? Normally .NET Isolated apps should never reference WebJobs packages directly.

test/e2e/Tests/e2e-tests-readme.md
```diff
@@ -0,0 +1,59 @@
...
+1. Installing a copy of Core Tools into your system's temp directory to ensure an unmodified Core Tools. This is necessary, as the tests will not attempt to use the "func" referenced in PATH 
+2. Ensure the test app(s) are running the correct extension code by: 
+    * Building the host and worker extensions from their projects within this repo
+    * Packing the extensions into local NuGet packages
+    * Copying the built packages into the test app's local nuget source folder as configured in nuget.config
+    * Updating the test app's .csproj files to reference the local package version
+    * Building the test app projects
+3. Install and start azurite emulator using Node
+
+NOTE: It should not be necessary to run start-emulators.ps1 manually, as it should be called by the build script. If you have a instance of Azurite already running, it will recognize and skip this step.
```
cgillum, line 42: Nice

test/e2e/Tests/e2e-tests-readme.md
```diff
@@ -0,0 +1,59 @@
+# End-to-End Test Project
+
+This document provides instructions on how to use the end-to-end (E2E) test project for the Azure Functions Durable Extension.
+
+## Prerequisites
+
+- Visual Studio IDE
```
cgillum: Is Visual Studio required or just recommended? Based on what I'm seeing, it doesn't look like we strictly require an IDE to run tests.

test/e2e/Tests/start-emulators.ps1
```diff
@@ -0,0 +1,198 @@
...
+
+# Reassigning $IsWindows no longer allowed - might need to refactor this logic
+# if (!$IsWindows -and !$IsLinux -and !$IsMacOs)
+# {
+#     # For pre-PS6
+#     Write-Host "Could not resolve OS. Assuming Windows."
+#     $IsWindows = $true
+# }
+
+if (!$IsWindows)
```
cgillum: Consider adding `-and $StartCosmosDBEmulator` so that we don't write this message about skipping if the caller never asked for the cosmos DB emulator.

.gitignore
```diff
@@ -297,3 +297,9 @@ functions-extensions/
 .vscode/
 .ionide/
 /src/WebJobs.Extensions.DurableTask/Microsoft.Azure.WebJobs.Extensions.DurableTask.xml
+
+# E2E Tests build output
+/test/DotnetIsolatedE2ETests/Azure.Functions.Cli/*
```
cgillum: Is this where we put the CLI or does it go into a temp directory (in which case we don't need this .gitignore line)?

cgillum (commented):
Looking great. I've added some comments, but I don't think we necessarily need to block this PR on any of them.

test/DotnetIsolatedE2EApps/MainApp/app.csproj
cgillum: It would definitely be better to do a project reference as I mentioned in a separate comment, but we can do that as part of a follow up PR if it's not a simple fix.

test/e2e/Tests/Helpers/HttpHelpers.cs
andystaples: Deleting this whole method as it is never used and - IMO - we should rely on the Task<HttpResponseMessage> method and validate responses in tests or methods defined in test classes

test/e2e/Tests/Helpers/HttpHelpers.cs
andystaples: Deleting this whole method as it is never used and - IMO - we should rely on the Task<HttpResponseMessage> method and validate responses in tests or methods defined in test classes

test/e2e/Tests/Tests/HelloCitiesTest.cs
andystaples: Carryover from the worker test code this code was derived from. Since the response here is a set of API URLs, we don't do any validation yet. Need to write a set of Durable test methods to parse out the StatusQueryGetAPI URL, invoke until orchestration completion, and parse the "actual" function output as a future work item. 
Deleting this line

test/e2e/Tests/Helpers/HttpHelpers.cs
andystaples: Done

test/e2e/Tests/Fixtures/FixtureHelpers.cs
andystaples: Done

test/e2e/Tests/build-e2e-test.ps1
andystaples: Removed it

test/e2e/Tests/e2e-tests-readme.md
andystaples: Will address with package reference work - this will also involve clearing the local NuGet cache which will make this step redundant

test/e2e/Apps/BasicDotNetIsolated/app.csproj
andystaples: Will address along with the work to update to package references

test/e2e/Tests/e2e-tests-readme.md
andystaples: Removed this line

test/e2e/Tests/start-emulators.ps1
andystaples: Done

.gitignore
andystaples: Removed this and the last line from .gitignore, not needed now

test/DotnetIsolatedE2EApps/MainApp/app.csproj
andystaples: Done!

test/e2e/Apps/BasicDotNetIsolated/app.csproj
andystaples: Done - now the test app has a reference to the worker extension project only. The webjobs extension is still packed and put in a local nuget packages folder alongside the app, so that the WorkerExtensions.csproj can get the version it needs

test/e2e/Tests/build-e2e-test.ps1
```diff
@@ -111,34 +116,26 @@ Write-Host "Moving nupkg from WebJobs extension to $E2EAppProjectDirectory/packa
...
+    $cachedVersionFolders | ForEach-Object {
+      if ($_.Name -eq $webJobsExtensionVersion)
+      {
+        Write-Host "Removing cached version $webJobsExtensionVersion from nuget cache"
+        Remove-Item -Recurse -Force $_.FullName -ErrorAction Stop
+      }
+    }
+    
+    # Write-Host "Updating WebJobs.Extensions.DurableTask version to $webJobsExtensionVersion"
+    # dotnet add app.csproj package Microsoft.Azure.WebJobs.Extensions.DurableTask --version $webJobsExtensionVersion
```
cgillum: Can these commented-out lines be removed?

cgillum (approved):
LGTM! As a follow up to this PR, let's create another PR to run these tests as part of the CI.

test/DotnetIsolatedE2ETests/DotnetIsolatedE2ETests.csproj
bachuv: Thanks!

bachuv (approved):

PR #3009, opened 2025-01-14
Replace ITelemetryActivator with ITelemetryModule

### LINKED ISSUE #2870: Durable Tracing V2 issue when using Application Insights with  Entra Authentication
### Description

We have encountered a problem regarding the combination of Durable Tracing V2 and using ManagedIdentity with Application Insights.

We found that when disabling `local authentication` in our Application Insights (see [Microsoft Entra authentication for Application Insights](https://learn.microsoft.com/en-us/azure/azure-monitor/app/azure-ad-authentication?tabs=net)) and relying on the ManagedIdentity of the Azure Function Host to authenticate to Application Insights, we start losing telemetry data related to the V2 distributed tracing.

Looking into the code, it seems like the Durable Function Extension configures its own TelemetryClient which does not take the ManagedIdentity of the Function Host into account.

### Expected behavior

When using Managed Identity there should be a way to set the Managed Identity credentials  of the Azure Function Hosts when this TelemetryClient is created, similar to how this is handled in the `azure-webjobs-sdk` with the `APPLICATIONINSIGHTS_AUTHENTICATION_STRING` (see [code here](https://github.com/Azure/azure-webjobs-sdk/blob/64219e4e7e47a64a34185613e9cb0e6b1d9842e7/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs#L322))

### Actual behavior

This TelemetryClient only sets the `APPINSIGHTS_INSTRUMENTATIONKEY` and the `APPLICATIONINSIGHTS_CONNECTION_STRING`, but does not have a way to set credentials when creating the TelemetryConfiguration (see [full code reference here](https://github.com/Azure/azure-functions-durable-extension/blob/829777ae7efb3dbbbab396f4bd4d0f76bc3c0955/src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs#L141) or simplified snippet below)

```csharp
private TelemetryConfiguration SetupTelemetryConfiguration()
        {
            TelemetryConfiguration config = TelemetryConfiguration.CreateDefault();
            if (this.OnSend != null)
            {
                config.TelemetryChannel = new NoOpTelemetryChannel { OnSend = this.OnSend };
            }

            string resolvedInstrumentationKey = this.nameResolver.Resolve("APPINSIGHTS_INSTRUMENTATIONKEY");
            string resolvedConnectionString = this.nameResolver.Resolve("APPLICATIONINSIGHTS_CONNECTION_STRING");

            bool instrumentationKeyProvided = !string.IsNullOrEmpty(resolvedInstrumentationKey);
            bool connectionStringProvided = !string.IsNullOrEmpty(resolvedConnectionString);

            if (instrumentationKeyProvided && connectionStringProvided)
            {
                this.endToEndTraceHelper.ExtensionWarningEvent(...);
            }

            if (!instrumentationKeyProvided && !connectionStringProvided)
            {
                this.endToEndTraceHelper.ExtensionWarningEvent(...);
            }

            if (instrumentationKeyProvided)
            {
                this.endToEndTraceHelper.ExtensionInformationalEvent(...);

#pragma warning disable CS0618 // Type or member is obsolete
                config.InstrumentationKey = resolvedInstrumentationKey;
#pragma warning restore CS0618 // Type or member is obsolete
            }

            if (connectionStringProvided)
            {
                this.endToEndTraceHelper.ExtensionInformationalEvent(...);

                config.ConnectionString = resolvedConnectionString;
            }

            return config;
        }

```

### App Details

- **Durable Functions extension version (e.g. v1.8.3)**: >=2.4.1
- **Programming language used**: Python

tagging: @ransonjb

bachuv:
This PR replaces `ITelemetryActivator` with `ITelemetryModule` so we can have more flexibility with configuration. For example, with this change we can set Managed Identity credentials instead of only setting Application Insights relevant settings.

For context:
- `ITelemetryActivator`: Used to activate or create telemetry objects. It also initializes the telemetry client for distributed tracing, allowing settings to be verified and adjusted during initialization
- `ITelemetryModule`: Used to initialize and configure telemetry modules from `TelemetryConfiguration`. It ensures that telemetry modules are properly set up to collect and send data using the same configuration.

Also, in this PR, we remove some log statements and checks. With `ITelemetryActivator`, we checked if the customer was providing `APPINSIGHTS_INSTRUMENTATIONKEY` and/or `APPLICATIONINSIGHTS_CONNECTION_STRING` and logged different warnings based on the configuration. With `ITelemetryModule`, it automatically has the configuration in the `TelemetryConfiguration` object, so we no longer need to check for these configurations.

Resolves #2870 

### Pull request checklist


jviau:
This looks good to me overall. But is it possible to add some unit tests for the `Initialize` method? Also, have you verified this works manually? That initialize is called and durable tracing is configured?

Also are all the failing builds of concern or transient?

test/FunctionsV2/PlatformSpecificHelpers.FunctionsV2.cs
```diff
@@ -89,7 +90,7 @@ public static ITestHost CreateJobHost(
 
                         if (onSend != null)
                         {
-                            serviceCollection.AddSingleton<ITelemetryActivator>(serviceProvider =>
+                            serviceCollection.AddSingleton<ITelemetryModule>(serviceProvider =>
```
bachuv: @jviau - after updating this line to register an `ITelemetryModule` for tests, I noticed that the lambda expression here doesn't get evaluated anymore and therefore the `TelemetryActivator`s constructor and `Intialize()` method don't get called. I thought it was because the new `Initialize()` method takes a `TelemetryConfiguration` as an argument so I registered a `TelemetryConfiguration` as a singleton before line 91, but it's still not initializing `TelemetryActivator`. Do you know if there are any additional services I should register to get `TelemetryActivator` to register?

For context, the changes in this PR work when I manually test them in a sample DF app. The tests are just not passing because of the issue I described earlier.

bachuv:
> This looks good to me overall. But is it possible to add some unit tests for the `Initialize` method? Also, have you verified this works manually? That initialize is called and durable tracing is configured?
> 
> Also are all the failing builds of concern or transient?

I'm working on adding unit tests for the `Initialize` method. I verified that this works manually.

The failing build includes some existing tests that are failing now because of the changes so I'm looking into that. I left a comment in the PR that explains the issue.

test/FunctionsV2/PlatformSpecificHelpers.FunctionsV2.cs
jviau: `ITelemetryModule`'s are resolved and invoked by the AppInsights SDK. Does the test code have the AppInsights SDK hooked up?

bachuv: The test code references `Microsoft.ApplicationInsights` through the package reference to the WebJobs extension. It doesn't have any code to hook up Application Insights though. I tried adding `services.AddApplicationInsightsTelemetry()`, but that ended up giving me an error that it needs an `IHostingEnvironment` so I wasn't sure if that was correct.

bachuv: @jviau, were you thinking of an approach different from `services.AddApplicationInsightsTelemetry()`?

jviau: Possibly? I haven't looked at the exact test failures yet, but all of this is just a _convenience_ part of the AppInsights SDK to have a defined hook for code to run as part of the SDK starting up. If that doesn't work for your tests, doing the same thing manually -- instantiating and calling the `ITelemetryModule` with the correct config instance is perfectly acceptable. What that looks like exactly I can't say without taking a deeper look.

bachuv: I was able to get rid of the `IHostingEnvironment` error by adding `loggingBuilder.AddApplicationInsightsWebJobs();`, but the tests are still failing. 

I also tried registering `TelemetryConfiguration` before the call to register `ITelemetryModule`, tried to make `services.AddApplicationInsightsTelemetry()` work, and tried to instantiate and call the `ITelemetryModule`, but nothing worked.

I might try to go back to using `ITelemetryActivator`, check for if the customer provided any managed identity specific environment variables/config, then manually set those since getting managed identity working with Application Insights was the original request that the customer reported in their GitHub issue.

@jviau, let me know if there are any other approaches you recommend that I try.

jviau: @bachuv I don't recommend reverting to `ITelemetryActivator`. Doing that puts diverges durable from how AppInsights is configured from the rest of the product. Using the `ITelemetryModule` allows durable to be agnostic for how app insights is configured. If the host adds support for more forms of identitiy, does durable really want to have to copy that work let alone even be aware of it? This is far less code to maintain. I especially don't recommend going back as this is a test setup issue, not an implementation issue.

test/FunctionsV2/CorrelationEndToEndTests.cs
```diff
@@ -219,109 +219,6 @@ internal async Task<Tuple<List<OperationTelemetry>, List<ExceptionTelemetry>>>
...
-        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
-        [InlineData(false, false, false)]
-        [InlineData(false, false, true)]
-        [InlineData(true, false, false)]
-        [InlineData(true, false, true)]
-        [InlineData(false, true, false)]
-        [InlineData(false, true, true)]
-        [InlineData(true, true, false)]
-        [InlineData(true, true, true)]
-        public void TelemetryClientSetup_AppInsights_Warnings(bool instrumentationKeyIsSet, bool connStringIsSet, bool extendedSessions)
```
jviau, line 237: Good call deleting this. Not something durable should be enforcing - particularly with OTel coming along which will not always be app-insights backed.

test/FunctionsV2/PlatformSpecificHelpers.FunctionsV2.cs
jviau: Looking at the test code, it will need refactoring as part of this to keep the same test validation. The current test code has 2 main gaps after theis change:

1. Reliance on using a test-only property `TelemetryActivator.OnSend` to shim in custom test hooks. That is no longer hooked up at all. 
2. DI changes. `TelemetryActivator` was instantiated and initialized as part of starting the Durable Webjobs extension. This will not longer be the case as we are relying on AppInsights SDK to perform this call. But appinsights SDK is not configured for this test host.

I think the easiest route for this change is to:

1. As part of test setup, just don't add `TelemetryActivator` at all!
2. Remove `OnSend` property
3. You can just instantiate your own `TelemetryActivator` as part of tests and pass in a custom `TelemetryConfiguration` to `Initialize`.

bachuv: Thanks for the pointers! I kept `ITelemetryModule` and updated the tests to instantiate `TelemetryActivator`. Please take a look and let me know if I can make any improvements.

test/FunctionsV2/PlatformSpecificHelpers.FunctionsV2.cs
```diff
@@ -86,20 +87,6 @@ public static ITestHost CreateJobHost(
                         {
                             serviceCollection.AddSingleton(serializerSettingsFactory);
                         }
-
-                        if (onSend != null)
```
jviau, line 90: The `onSend` parameter can also be removed from here and any other functions passing it on this method.

test/FunctionsV2/CorrelationEndToEndTests.cs
```diff
@@ -191,134 +196,88 @@ internal async Task<Tuple<List<OperationTelemetry>, List<ExceptionTelemetry>>>
...
             {
                 await host.StartAsync();
                 var client = await host.StartOrchestratorAsync(orchestratorFunctionNames[0], input, this.output);
                 var status = await client.WaitForCompletionAsync(this.output, timeout: TimeSpan.FromSeconds(90));
                 await host.StopAsync();
             }
 
-            var sendItemList = this.ConvertTo(sendItems);
+            // Flush the telemetry client
+            TelemetryClient telemetryClient = new TelemetryClient(telemetryConfiguration);
```
jviau: recommend moving this above, along with the creation of the channel and config.

test/FunctionsV2/CorrelationEndToEndTests.cs
```diff
@@ -191,134 +196,88 @@ internal async Task<Tuple<List<OperationTelemetry>, List<ExceptionTelemetry>>>
             };
             DurableTaskOptions options = new DurableTaskOptions();
             options.Tracing = traceOptions;
-            var sendAction = new Action<ITelemetry>(
```
jviau, line 194: can `sendItems` declaration above be removed now?

test/FunctionsV2/CorrelationEndToEndTests.cs
```diff
@@ -191,134 +196,88 @@ internal async Task<Tuple<List<OperationTelemetry>, List<ExceptionTelemetry>>>
...
+
+            TelemetryConfiguration telemetryConfiguration = new TelemetryConfiguration();
+            string connStringValue = "InstrumentationKey=xxxx;IngestionEndpoint =https://xxxx.applicationinsights.azure.com/;LiveEndpoint=https://xxxx.livediagnostics.monitor.azure.com/";
+            telemetryConfiguration.ConnectionString = connStringValue;
+
+            // Set up the custom telemetry channel
+            TestTelemetryChannel testTelemetryChannel = new TestTelemetryChannel();
+            telemetryConfiguration.TelemetryChannel = testTelemetryChannel;
+
+            TelemetryActivator activator = new TelemetryActivator(optionsWrapper, mockNameResolver.Object);
```
jviau: This might have impact on other tests if you don't dispose it.

```suggestion
            using TelemetryActivator activator = new TelemetryActivator(optionsWrapper, mockNameResolver.Object);
```

jviau (approved):

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
```diff
@@ -17,13 +17,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Correlation
...
-    public class TelemetryActivator : ITelemetryActivator, IAsyncDisposable, IDisposable
+    public class TelemetryActivator : ITelemetryModule, IAsyncDisposable, IDisposable
     {
         private readonly DurableTaskOptions options;
         private readonly INameResolver nameResolver;
-        private EndToEndTraceHelper endToEndTraceHelper;
         private TelemetryClient telemetryClient;
-        private IAsyncDisposable telemetryModule;
+#pragma warning disable SA1401 // Fields should be private
+        internal IAsyncDisposable TelemetryModule;
```
cgillum: To resolve this warning, can we make this a property instead of a field? (if we do, please move it right below the constructor)

cgillum (approved):
LGTM! Please resolve Jacob's remaining feedback items plus the one that I added and then go ahead and merge.

PR #3021, opened 2025-01-30
Add a test for PurgeAllInstancesAsync

### LINKED ISSUE #2702: PurgeAllInstancesAsync throwing Grpc.Core.RpcException
### Description

```
    public class HttpTriggerTest
    {
        [Function("HttpTriggerTest")]
        public async Task<HttpResponseData> Run([HttpTrigger(AuthorizationLevel.Function, "get", "post")] HttpRequestData req, [DurableClient] DurableTaskClient client)
        {
            var result = await client.PurgeAllInstancesAsync(new PurgeInstancesFilter(null, DateTime.Today.AddDays(-30),
                new List<OrchestrationRuntimeStatus>
                {
                    OrchestrationRuntimeStatus.Completed,
                    OrchestrationRuntimeStatus.Failed,
                    OrchestrationRuntimeStatus.Terminated
                })
            );

            var response = req.CreateResponse(HttpStatusCode.OK);
            return response;
        }
    }
```

Exception: Grpc.Core.RpcException: Status(StatusCode="Unknown", Detail="Exception was thrown by handler.")
[2023-12-22T08:10:04.660Z]    at Microsoft.DurableTask.
[2023-12-22T08:10:04.660Z]    at Microsoft.DurableTask.Client.Grpc.GrpcDurableTaskClient.PurgeInstancesCoreAsync(PurgeInstancesRequest request, CancellationToken cancellation)

I wouldn't expect this exception. This is happening on a new function app on .NET 8 isolated with the simple http trigger as above with the package versions below. I get this error locally running via azurite

### App Details
.NET 8 Isolated
Microsoft.Azure.Functions.Worker 1.20.0
Microsoft.Azure.Functions.Worker.Sdk 1.16.4
Microsoft.Azure.Functions.Worker.Extensions.DurableTask 1.1.0

andystaples:
- Adds an E2E test for PurgeAllInstancesAsync
- Fix bug with PurgeAllInstancesAsync relating to passing a null start date

resolves #2702

### Pull request checklist


test/e2e/Tests/Tests/PurgeInstancesTests.cs
```diff
@@ -0,0 +1,49 @@
...
+    [Theory]
+    [InlineData("PurgeOrchestrationHistory", HttpStatusCode.OK, @"^Purged [0-9]* records$")]
+    public async Task HttpTriggerTests(string functionName, HttpStatusCode expectedStatusCode, string responseRegex)
+    {
+        using HttpResponseMessage response = await HttpHelpers.InvokeHttpTrigger(functionName, "");
+        string actualMessage = await response.Content.ReadAsStringAsync();
+        Assert.Matches(responseRegex, actualMessage);
+        Assert.Equal(expectedStatusCode, response.StatusCode);
+    }
+}
```
bachuv, line 101: Are there any other test cases that we can add? Maybe a test to cover the changes you made in ProtobufUtils.cs?

andystaples: This test actually does cover that case - starttime is null in the function app code for this test, and before my change, it threw instead of returning. 
Would love to add more tests, though. I think a good set to start with would be 
- A test with empty end date
- A test with a start and end date
- Purging from (now + 1 minute) to (now + 1 day) should delete 0 records
- Starting an orchestration, then purging from >1 day ago should not purge the new orchestration's history
- Starting an orchestration, waiting for it to complete, then purging all history should delete that orchestration's history

Does this seem reasonable? Other scenarios I should cover? @bachuv @cgillum

nytian: That sounds good to me. For the last three bullets, I remember Chris has a sample test for reference, I also referred to this when adding purge test at DTS

https://github.com/microsoft/durabletask-mssql/blob/ee4a99c31eff04d8f9595fcfeb5b5dd0a873a99f/test/DurableTask.SqlServer.AzureFunctions.Tests/CoreScenarios.cs#L311

bachuv: This sounds reasonable to me!

src/WebJobs.Extensions.DurableTask/ProtobufUtils.cs
```diff
@@ -441,7 +441,7 @@ internal static PurgeInstanceFilter ToPurgeInstanceFilter(P.PurgeInstancesReques
             // This ternary condition is necessary because the protobuf spec __insists__ that CreatedTimeFrom may never be null,
             // but nonetheless if you pass null in function code, the value will be null here
             return new PurgeInstanceFilter(
-                request.PurgeInstanceFilter.CreatedTimeFrom == null ? DateTime.MinValue : request.PurgeInstanceFilter.CreatedTimeTo.ToDateTime(),
+                request.PurgeInstanceFilter.CreatedTimeFrom == null ? DateTime.MinValue : request.PurgeInstanceFilter.CreatedTimeFrom.ToDateTime(),
```
cgillum, line 444: Yikes

test/e2e/Tests/Tests/PurgeInstancesTests.cs
```diff
@@ -23,27 +23,79 @@ public PurgeInstancesTests(FunctionAppFixture fixture, ITestOutputHelper testOut
...
-        {
-            _output.WriteLine(message);
-        }
-        catch
-        {
-            // Ignore
-        }
+        DateTime purgeStartTime = DateTime.MinValue;
+        DateTime purgeEndTime = DateTime.UtcNow;
+        string queryParams = $"?purgeStartTime={purgeStartTime.ToString("o")}&purgeEndTime={purgeEndTime.ToString("o")}";
```
cgillum: tip: this can be simplified further by removing the `ToString()` method call but leaving the format specifier:
```suggestion
        string queryParams = $"?purgeStartTime={purgeStartTime:o}&purgeEndTime={purgeEndTime:o}";
```

test/e2e/Tests/Tests/PurgeInstancesTests.cs
```diff
@@ -0,0 +1,101 @@
...
+
+    [Fact]
+    public async Task PurgeOrchestrationHistory_StartAndEnd_Succeeds()
+    {
+        DateTime purgeStartTime = DateTime.MinValue;
+        DateTime purgeEndTime = DateTime.UtcNow;
+        string queryParams = $"?purgeStartTime={purgeStartTime.ToString("o")}&purgeEndTime={purgeEndTime.ToString("o")}";
+        using HttpResponseMessage response = await HttpHelpers.InvokeHttpTrigger("PurgeOrchestrationHistory", queryParams);
+        string actualMessage = await response.Content.ReadAsStringAsync();
+        Assert.Matches(@"^Purged [0-9]* records$", actualMessage);
```
cgillum, line 34: Looks like we may or may not have any actual data that gets purged in this test and in most of the other tests. How should I think about what "success" means in these cases?

andystaples: Mostly I was ensuring that the method does not throw an exception - like it did before passing a null CreatedTimeFrom before this PR.
There is at least one test that invokes an orchestrator and checks for non-zero records purged. Happy to update them all to do this, if we think it would improve coverage. 
I also made a deliberate decision not to check for an exact number of rows, due to this being an implementation detail of the backend and not particularly important to the end user. Happy to reevaluate this decision as needed, too

test/e2e/Tests/Tests/PurgeInstancesTests.cs
andystaples: Doen

cgillum (approved):

PR #3028, opened 2025-02-04
Add suspend/resume tests

andystaples:
Adds testing for suspend/resume in Durable

### Pull request checklist

* [x] My changes **should** be added to **v2.x** branch.

### DIFF
--- a/test/e2e/Tests/Helpers/DurableHelpers.cs
+++ b/test/e2e/Tests/Helpers/DurableHelpers.cs
@@ -52,6 +52,21 @@ internal class DurableHelpers
         return new OrchestrationStatusDetails(statusQueryResponseString);
     }
 
+    internal static async Task WaitForOrchestrationState(string statusQueryGetUri, string desiredState, int maxTimeoutSeconds)
+    {
+        DateTime timeoutTime = DateTime.Now + TimeSpan.FromSeconds(maxTimeoutSeconds);
+        while (DateTime.Now < timeoutTime)
+        {
+            var currentStatus = await GetRunningOrchestrationDetailsAsync(statusQueryGetUri);
+            if (currentStatus.RuntimeStatus == desiredState)
+            {
+                return;
+            }
+            await Task.Delay(100);
+        }
+        throw new TimeoutException($"Orchestration did not reach {desiredState} status within {maxTimeoutSeconds} seconds.");
+    }
+
     private static string TokenizeAndGetValueFromKeyAsString(string? json, string key)
     {
         if (string.IsNullOrEmpty(json))
--- a/test/e2e/Tests/Helpers/DurableHelpers.cs
+++ b/test/e2e/Tests/Helpers/DurableHelpers.cs
@@ -52,7 +52,7 @@ internal class DurableHelpers
         return new OrchestrationStatusDetails(statusQueryResponseString);
     }
 
-    internal static async Task WaitForOrchestrationState(string statusQueryGetUri, string desiredState, int maxTimeoutSeconds)
+    internal static async Task WaitForOrchestrationStateAsync(string statusQueryGetUri, string desiredState, int maxTimeoutSeconds)
     {
         DateTime timeoutTime = DateTime.Now + TimeSpan.FromSeconds(maxTimeoutSeconds);
         while (DateTime.Now < timeoutTime)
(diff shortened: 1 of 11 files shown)


test/e2e/Tests/Tests/SuspendResumeTests.cs
```diff
@@ -0,0 +1,197 @@
...
+        using HttpResponseMessage response = await HttpHelpers.InvokeHttpTrigger("LongOrchestrator_HttpStart", "");
+
+        Assert.Equal(HttpStatusCode.Accepted, response.StatusCode);
+        string instanceId = await DurableHelpers.ParseInstanceIdAsync(response);
+        string statusQueryGetUri = await DurableHelpers.ParseStatusQueryGetUriAsync(response);
+
+        Thread.Sleep(1000);
+
+        var orchestrationDetails = await DurableHelpers.GetRunningOrchestrationDetailsAsync(statusQueryGetUri);
+        Assert.Equal("Running", orchestrationDetails.RuntimeStatus);
```
cgillum: Rather than sleeping and hoping that the status will change after the sleep interval, we can make these checks a bit more reliable and potentially faster if we use a polling pattern to check for the state change. This would address the very common flakey test problem where engineers need to increase sleep durations because things ran slower than we expected. Consider creating a helper method that can wait for an orchestration to transition from one state to another (Pending --> Running, Running --> Suspended, Suspended --> Running, etc.)

test/e2e/Tests/Tests/SuspendResumeTests.cs
```diff
@@ -0,0 +1,197 @@
...
+
+        Assert.Equal(HttpStatusCode.Accepted, response.StatusCode);
+        string instanceId = await DurableHelpers.ParseInstanceIdAsync(response);
+        string statusQueryGetUri = await DurableHelpers.ParseStatusQueryGetUriAsync(response);
+
+        Thread.Sleep(1000);
+        try
+        {
+            using HttpResponseMessage suspendResponse = await HttpHelpers.InvokeHttpTrigger("SuspendInstance", $"?instanceId={instanceId}");
+            await AssertRequestFailsAsync(suspendResponse);
```
cgillum, line 126: A couple things:
1. The method `AssertRequestFailsAsync` assumes that the request is a termination request, but you're sending a suspend request.
2. What's the expected behavior for in-proc? Do we throw if you try to suspend a suspended orchestration? I was hoping we wouldn't, but if we are, then checking for a failure message is fine.

andystaples: 1. The behavior is the same, but I've updated the parameters and variable names to reflect this. 
2. Yes, we throw in the same way for in-proc (except with a more verbose exception message, as described in the bug linked in the comments)

test/e2e/Tests/Tests/SuspendResumeTests.cs
andystaples: Done. Feel free to critique my implementation

test/e2e/Tests/Helpers/DurableHelpers.cs
```diff
@@ -52,6 +52,21 @@ internal static async Task<OrchestrationStatusDetails> GetRunningOrchestrationDe
...
+    internal static async Task WaitForOrchestrationStateAsync(string statusQueryGetUri, string desiredState, int maxTimeoutSeconds)
+    {
+        DateTime timeoutTime = DateTime.Now + TimeSpan.FromSeconds(maxTimeoutSeconds);
+        while (DateTime.Now < timeoutTime)
+        {
+            var currentStatus = await GetRunningOrchestrationDetailsAsync(statusQueryGetUri);
+            if (currentStatus.RuntimeStatus == desiredState)
+            {
+                return;
+            }
```
cgillum, line 71: As discussed, let's also handle the case where the orchestration has completed/terminated/failed, in which case additional polling would be futile.  Something like this:
```suggestion
            }
            else if (currentStatus.RuntimeStatus == "Completed" || currentStatus.RuntimeStatus == "Terminated" || currentStatus.RuntimeStatus == "Failed")
            {
                throw new InvalidOperationException($"The orchestration ended unexpectedly in a {currentStatus.RuntimeStatus} state!");
            }
```

andystaples: I did add this, on the next lines
```C#
            if (finalStates.Contains(currentStatus.RuntimeStatus))
            {
                throw new TaskCanceledException($"Orchestration reached {currentStatus.RuntimeStatus} state when test was expecting {desiredState}");
            }
```

cgillum (approved):

PR #3034, opened 2025-02-06
Improve exception handling in .NET isolated

### LINKED ISSUE #2711: [out-of-proc] `RetryContext.LastFailure` incorrectly captures thrown exception
### Description
When inspecting `RetryContext.LastFailure` during activity retry cycle, I noticed that `.ErrorType` and `.ErrorMessage` are populated from `Exception.Message`, with `.ErrorType` being the first part of `.Message`, until a `':'`, while the remaining part of the `.Message` is copied to `.ErrorMessage`.

As a result of this, `RetryContext.LastFailure.IsCausedBy<TException>()` does not work, because it relies on `.ErrorType`, which is malformed.

If I drop `':'` from my exception message, then `.ErrorType` is `"(unknown)"`

### Expected behavior

- `RetryContext.LastFailure.ErrorType` is set to correct the exception type.
- `RetryContext.LastFailure.ErrorMessage` is set to `.Message` of the captured exception.
- `RetryContext.LastFailure.IsCausedBy<TException>()` works

### Actual behavior
- `RetryContext.LastFailure.ErrorType` seems to be set to `exception.Message` sliced till first `':'` or `"(unknown)"` if `':'` is absent
-  `RetryContext.LastFailure.ErrorMessage` seems to be set to `exception.Message` sliced from first `':'` or the beginning of the message
-  `RetryContext.LastFailure.InnerFailure` is `null` (cannot judge if this is useful or not)
- `RetryContext.LastFailure.IsCausedBy<TException>()` does not work

Had to scramble some info here. 
Here is the exception I throw from my activity:
Screenshot 2024-01-08 141740

And here is what I observe in the `AsyncRetryHandler`:
Screenshot 2024-01-08 141810

### Known workarounds
I originally intended to append `Reason: ...` to exception messages as a workaround, but this lead me to discovery that `':'` is treated specially (which made things even worse).
I am considering (ab-)using this feature to simulate the intended behavior.

### App Details
- `Microsoft.Azure.Functions.Worker.Extensions.DurableTask 1.1.0`
- `Microsoft.Azure.Functions.Worker.Sdk 1.16.4`
- Azure Functions v4
- .NET 8

andystaples:
Improves exception handling in .NET isolated by serializing the exception details in the worker middleware, then deserializing them back in host middleware. 
This may introduce breaking changes to the way that the Functions Host logs exceptions from Durable activities. 
Also adds testing for error handling, activity retry, and activity timeout scenarios in dotnet-isolated

Resolves #2711 

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
+++ b/src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
@@ -691,7 +691,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             try
             {
-                string serializedMessage = exception.Split('\n')[0];
+                if (exception[0] != '{')
+                {
+                    details = null;
+                    return false;
+                }
+
+                int newlineIndex = exception.IndexOf('\n');
+                string serializedMessage = newlineIndex < 0 ? exception : exception.Substring(0, newlineIndex).Trim();
                 P.TaskFailureDetails? taskFailureDetails = JsonConvert.DeserializeObject<P.TaskFailureDetails>(serializedMessage);
                 if (taskFailureDetails != null)
                 {
--- a/src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
+++ b/src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
@@ -13,6 +13,7 @@ using DurableTask.Core.Exceptions;
 using DurableTask.Core.History;
 using DurableTask.Core.Middleware;
 using Microsoft.Azure.WebJobs.Host.Executors;
+using Newtonsoft.Json;
 using P = Microsoft.DurableTask.Protobuf;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
@@ -606,6 +607,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 Exception rpcException = e.InnerException;
                 if (TryGetRpcExceptionFields(rpcException.Message, out string? exception, out string? stackTrace))
                 {
+                    if (TryExtractSerializedFailureDetailsFromException(exception, out FailureDetails? details) && details is not null)
+                    {
+                        return details;
+                    }
+
                     if (TrySplitExceptionTypeFromMessage(exception, out string? exceptionType, out string? exceptionMessage))
                     {
                         return new FailureDetails(exceptionType, exceptionMessage, stackTrace, innerFailure: null, isNonRetriable: false);
@@ -622,6 +628,21 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
             // Don't recognize this exception - return it as-is
             return new FailureDetails(e);
+        }
+
+        private static FailureDetails? GetFailureDetails(P.TaskFailureDetails taskFailureDetails)
+        {
+            if (taskFailureDetails is null)
+            {
+                return null;
+            }
+
+            return new FailureDetails(
+                taskFailureDetails.ErrorType ?? string.Empty,
+                taskFailureDetails.ErrorMessage ?? string.Empty,
+                taskFailureDetails.StackTrace,
+                GetFailureDetails(taskFailureDetails.InnerFailure),
+                taskFailureDetails.IsNonRetriable);
         }
 
         private static bool TryGetRpcExceptionFields(
@@ -664,6 +685,27 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             }
 
             return true;
+        }
+
+        private static bool TryExtractSerializedFailureDetailsFromException(string exception, out FailureDetails? details)
+        {
+            try
+            {
+                string serializedMessage = exception.Split('\n')[0];
+                P.TaskFailureDetails? taskFailureDetails = JsonConvert.DeserializeObject<P.TaskFailureDetails>(serializedMessage);
+                if (taskFailureDetails != null)
+                {
+                    details = GetFailureDetails(taskFailureDetails);
+                    return true;
+                }
+            }
+            catch (Exception)
+            {
+                // Apparently the exception message was not serialized by the worker middleware, continue
+            }
+
+            details = null;
+            return false;
         }
 
         private static bool TrySplitExceptionTypeFromMessage(
--- a/src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
+++ b/src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
@@ -584,6 +584,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     isReplay: false,
                     scheduledEvent.EventId);
 
+                bool detailsParsedFromSerializedException;
+
                 activityResult = new ActivityExecutionResult
                 {
                     ResponseEvent = new TaskFailedEvent(
@@ -591,8 +593,17 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                         taskScheduledId: scheduledEvent.EventId,
                         reason: $"Function '{functionName}' failed with an unhandled exception.",
                         details: null,
-                        GetFailureDetails(result.Exception)),
+                        GetFailureDetails(result.Exception, out detailsParsedFromSerializedException)),
                 };
+
+                if (!detailsParsedFromSerializedException)
+                {
+                    this.TraceHelper.ExtensionWarningEvent(
+                        this.Options.HubName,
+                        functionName.Name,
+                        instance.InstanceId,
+                        "Failure details not parsed from serialized exception details, worker failed to serialize exception");
+                }
             }
 
             // Send the result of the activity function to the DTFx dispatch pipeline.
@@ -600,8 +611,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             dispatchContext.SetProperty(activityResult);
         }
 
-        private static FailureDetails GetFailureDetails(Exception e)
+        private static FailureDetails GetFailureDetails(Exception e, out bool fromSerializedException)
         {
+            fromSerializedException = false;
             if (e.InnerException != null && e.InnerException.Message.StartsWith("Result:"))
             {
                 Exception rpcException = e.InnerException;
@@ -609,6 +621,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 {
                     if (TryExtractSerializedFailureDetailsFromException(exception, out FailureDetails? details) && details is not null)
                     {
+                        fromSerializedException = true;
                         return details;
                     }
 
@@ -708,7 +721,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             }
             catch (Exception)
             {
-                // Apparently the exception message was not serialized by the worker middleware, continue
+                // Apparently the exception message was not serialized by the worker middleware, this will be logged in CallActivityAsync()
             }
 
             details = null;
--- a/src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
+++ b/src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
@@ -596,7 +596,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                         GetFailureDetails(result.Exception, out detailsParsedFromSerializedException)),
                 };
 
-                if (!detailsParsedFromSerializedException)
+                if (!detailsParsedFromSerializedException && this.extension.PlatformInformationService.GetWorkerRuntimeType() == WorkerRuntimeType.DotNetIsolated)
                 {
                     this.TraceHelper.ExtensionWarningEvent(
                         this.Options.HubName,
--- a/src/Worker.Extensions.DurableTask/DurableTaskFunctionsMiddleware.cs
+++ b/src/Worker.Extensions.DurableTask/DurableTaskFunctionsMiddleware.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Diagnostics.CodeAnalysis;
 using System.Threading.Tasks;
+using Microsoft.Azure.Functions.Worker.Extensions.DurableTask.Exceptions;
 using Microsoft.Azure.Functions.Worker.Middleware;
 using Microsoft.DurableTask.Worker.Grpc;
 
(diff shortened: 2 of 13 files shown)


test/e2e/Tests/Tests/ErrorHandlingTests.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        await DurableHelpers.WaitForOrchestrationStateAsync(statusQueryGetUri, "Failed", 30);
+
+        var orchestrationDetails = await DurableHelpers.GetRunningOrchestrationDetailsAsync(statusQueryGetUri);
+        Assert.StartsWith("Microsoft.DurableTask.TaskFailedException", orchestrationDetails.Output);
+        Assert.Contains("This activity failed", orchestrationDetails.Output);
+    }
+
+    [Fact]
+    public async Task OrchestratorWithCaughtActivityException_ShouldSucced()
```
cgillum:
```suggestion
    public async Task OrchestratorWithCaughtActivityException_ShouldSucceed()
```

test/e2e/Tests/Tests/ErrorHandlingTests.cs
```diff
@@ -0,0 +1,99 @@
...
+        string actualMessage = await response.Content.ReadAsStringAsync();
+
+        Assert.Equal(HttpStatusCode.Accepted, response.StatusCode);
+        string statusQueryGetUri = await DurableHelpers.ParseStatusQueryGetUriAsync(response);
+
+        await DurableHelpers.WaitForOrchestrationStateAsync(statusQueryGetUri, "Failed", 30);
+
+        var orchestrationDetails = await DurableHelpers.GetRunningOrchestrationDetailsAsync(statusQueryGetUri);
+        Assert.StartsWith("Microsoft.DurableTask.TaskFailedException", orchestrationDetails.Output);
+        Assert.Contains("This activity failed", orchestrationDetails.Output);
```
cgillum, line 35: Is there anything more we can validate here? For example, the type of the inner exception?

test/e2e/Tests/Tests/ErrorHandlingTests.cs
cgillum: Looks like the same typo is in multiple methods.

test/e2e/Tests/Tests/ErrorHandlingTests.cs
```diff
@@ -0,0 +1,99 @@
...
+        _fixture = fixture;
+        _fixture.TestLogs.UseTestLogger(testOutputHelper);
+        _output = testOutputHelper;
+    }
+
+    [Fact]
+    public async Task OrchestratorWithUncaughtActivityException_ShouldFail()
+    {
+        using HttpResponseMessage response = await HttpHelpers.InvokeHttpTrigger("RethrowActivityException_HttpStart", "");
+        string actualMessage = await response.Content.ReadAsStringAsync();
```
cgillum: Looks like the `actualMessage` variable isn't being used in any of these tests.

test/e2e/Tests/Tests/TimeoutTests.cs
```diff
@@ -0,0 +1,39 @@
...
+        _output = testOutputHelper;
+    }
+
+    [Theory]
+    [InlineData(2, "The activity function timed out")]
+    [InlineData(10, "The activity function completed successfully")]
+    public async Task TimeoutFunction_ShouldTimeoutWhenAppropriate(int timeoutSeconds, string expectedOutput)
+    {
+        using HttpResponseMessage response = await HttpHelpers.InvokeHttpTrigger("TimeoutOrchestrator_HttpStart", $"?timeoutSeconds={timeoutSeconds}");
+        string actualMessage = await response.Content.ReadAsStringAsync();
```
cgillum, line 29: `actualMessage` is not used here either.

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -666,6 +687,27 @@ private static bool TryGetRpcExceptionFields(
             return true;
         }
 
+        private static bool TryExtractSerializedFailureDetailsFromException(string exception, out FailureDetails? details)
+        {
+            try
+            {
+                string serializedMessage = exception.Split('\n')[0];
```
cgillum: Two things:
1. `string.Split` can be a bit expensive. It's faster and less memory intense to do `IndexOf` and `Substring`.
2. Let's not attempt to parse the message unless we have some clue that it's JSON to avoid parsing things that obviously aren't JSON.

Suggestion:
```csharp
if (exception[0] != '{') // expecting JSON
{
    details = null;
    return false;
}

int newlineIndex = input.IndexOf('\n');
string failureDetailsJson = newlineIndex < 0 ? exception : exception.Substring(0, newlineIndex).Trim();
P.TaskFailureDetails? taskFailureDetails = JsonConvert.DeserializeObject<P.TaskFailureDetails>(failureDetailsJson);
// ...
```

src/Worker.Extensions.DurableTask/Exceptions/DurableSerializationException.cs
```diff
@@ -0,0 +1,42 @@
...
+
+    public override string ToString()
+    {
+        TaskFailureDetails? failureDetails = ExceptionToTaskFailureDetailsRecursive(this.FromException);
+        return JsonConvert.SerializeObject(failureDetails);
+    }
+
+    public override string? StackTrace => this.FromException.StackTrace;
+
+    private static TaskFailureDetails? ExceptionToTaskFailureDetailsRecursive(Exception? fromException)
```
cgillum: Do we have a shared version of this we can use? Or is it in some inaccessible place, so we have to copy it? I'm a tiny bit worried about keeping multiple copies up-to-date if we decide to add richer information to `TaskFailureDetails` (which is very likely).

src/Worker.Extensions.DurableTask/DurableTaskFunctionsMiddleware.cs
```diff
@@ -94,4 +103,33 @@ static async Task RunEntityAsync(
...
+    {
+        try
+        {
+            await next(functionContext);
+            return;
+        }
+        catch (Exception ex)
+        {
+            throw new DurableSerializationException(ex);
+        }
```
cgillum, line 130: Would it be beneficial to catch and do this for orchestration and entity triggers as well?

test/e2e/Tests/Tests/ErrorHandlingTests.cs
andystaples: Fixed

test/e2e/Tests/Tests/TimeoutTests.cs
andystaples: Fixed

test/e2e/Tests/Tests/ErrorHandlingTests.cs
andystaples: Fixed

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
andystaples: Implemented

src/Worker.Extensions.DurableTask/Exceptions/DurableSerializationException.cs
andystaples: Moved to a new TaskFailureDetailsConverter

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -666,6 +687,34 @@ private static bool TryGetRpcExceptionFields(
...
+                P.TaskFailureDetails? taskFailureDetails = JsonConvert.DeserializeObject<P.TaskFailureDetails>(serializedMessage);
+                if (taskFailureDetails != null)
+                {
+                    details = GetFailureDetails(taskFailureDetails);
+                    return true;
+                }
+            }
+            catch (Exception)
+            {
+                // Apparently the exception message was not serialized by the worker middleware, continue
```
bachuv: Do we want to log a message saying what's in this comment? Will this be useful for debugging issues in the future?

test/e2e/Tests/Tests/ErrorHandlingTests.cs
andystaples: Added code to double check the inner exception type in the test functionapp code, and am also validating it in the log output in tests.

src/Worker.Extensions.DurableTask/DurableTaskFunctionsMiddleware.cs
andystaples: I will suggest that no, for now, no change is needed for these function types. 

Orchestrations never attempt to parse exception details returned from the host into FailureDetails. 

Entitites do, however, since the entity data is returned through the TaskEntityDispatcher, error message is already parsed by the worker and returned in an intelligent way. 

There is some difference in end result for entities. Here is one example: activity functions that throw exceptions have and continue to throw the same exception type in the orchestration. By contrast, entities throw an EntityOperationFailedException which has a TaskFailureDetails property containing the outermost user code exception. 

Additionally, after the outermost exception thrown by the user code, no inner exceptions are returned when an entity throws. This appears to be deliberate, as inner user exception details were stripped in both the webjobs and worker extensions. We can re-evaluate this decision, but I believe it to be beyond the scope of this PR.

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
andystaples: Done - the actual warning is logged using TraceHelper.ExtensionWarningEvent in CallActivityAsync based on whether the GetFailureDetails call reports successful deserialization of the new message. 
If there is a better way to log this, please, let me know

andystaples: Oh also, a consideration: GetFailureDetails is used for all languages that use OutOfProcMiddleware, so I believe Java/others will now log this warning on every activity failure, as the worker extensions for these languages do not know to serialize the exception details in this way. 
We may eventually want to update the other extensions in a similar way - happy to remove/comment this warning until we expect all languages to support exception serialization, or I can try to check the language and only warn for dotnet, or leave the warning in as-is, if we think that the warning will not be distracting for other languages.

andystaples: Update - added a check for dotnet isolated

src/Worker.Extensions.DurableTask/Exceptions/DurableSerializationException.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using Microsoft.DurableTask.Protobuf;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.DurableTask.Exceptions;
+
+internal class DurableSerializationException : Exception
+{
+    private Exception FromException;
```
cgillum: Fields should be lowercased and, whenever possible, should be `readonly`.
```suggestion
    private readonly Exception fromException;
```

test/e2e/Apps/BasicDotNetIsolated/ActivityErrorHandling.cs
```diff
@@ -0,0 +1,163 @@
...
+        return await client.CreateCheckStatusResponseAsync(req, instanceId);
+    }
+
+    [Function("CatchActivityException_HttpStart")]
+    public static async Task<HttpResponseData> CatchHttpStart(
+        [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post")] HttpRequestData req,
+        [DurableClient] DurableTaskClient client,
+        FunctionContext executionContext)
+    {
+        ILogger logger = executionContext.GetLogger("RethrowActivityException_HttpStart");
```
cgillum: Should this be:
```suggestion
        ILogger logger = executionContext.GetLogger("CatchHttpStart");
```

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
bachuv: > as the worker extensions for these languages do not know to serialize the exception details in this way.

Can you clarify what you mean here?

test/e2e/Apps/BasicDotNetIsolated/ActivityErrorHandling.cs
```diff
@@ -0,0 +1,163 @@
...
+using Microsoft.DurableTask;
+using Microsoft.DurableTask.Client;
+using Microsoft.DurableTask.Entities;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.Durable.Tests.E2E;
+
+public static class ActivityErrorHandling
+{
+    private static ConcurrentDictionary<string, int> retryCount = new ConcurrentDictionary<string, int>();
```
cgillum: nit: consider naming this `globalRetryCount` to make it clear that it's a global (static) value.

test/e2e/Apps/BasicDotNetIsolated/TimeoutOrchestration.cs
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    [Function(nameof(LongActivity))]
+    public static string LongActivity([ActivityTrigger] string instanceId, FunctionContext executionContext)
+    {
+        for (int i = 0; i < 5; i++) 
+        {
+            Thread.Sleep(1000);
+        }
```
cgillum: What's the purpose of having a sleep in a for-loop vs. just doing `Thread.Sleep(5000)`?

cgillum (approved):
LGTM - just adding a few more comments/questions.

src/Worker.Extensions.DurableTask/Exceptions/DurableSerializationException.cs
andystaples: Done

test/e2e/Apps/BasicDotNetIsolated/ActivityErrorHandling.cs
andystaples: Fixed

test/e2e/Apps/BasicDotNetIsolated/ActivityErrorHandling.cs
andystaples: Done

test/e2e/Apps/BasicDotNetIsolated/TimeoutOrchestration.cs
andystaples: Carryover from testing, updated

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
andystaples: As in, the Java worker doesn't have the worker-side logic implemented in this PR - to wrap the exception message in Json so that the WebJobs extension can parse. I've added a task in Java to look at this here https://github.com/microsoft/durabletask-java/issues/205

PR #3051, opened 2025-02-25
Enable Central Package Management 

nytian:
Move to [Central Package Management](https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management)

- Added two  Directory.Packages.props : One for the whole project, another one is just for the samples.
- Remove all versions at .csproj files
-  Updated some test files to remove build warnings from Xunit. 

### Pull request checklist


samples/Directory.Packages.props
andystaples, line 1: I might suggest we don't use CPM for the samples - if customers want to download and adapt a single sample project, they will have to re-add the versions for the packages in that project manually. Open to discussion, though

nytian: That's a good point. While it's fine to me since many packages in the samples are the same, and several are somewhat outdated over time. This allows us to manage them all more easily.

cgillum: I could go either way on this. I definitely like the idea of making it easy to copy/paste samples easily, and I agree that CPM makes this harder. But on the other hand, there are a lot of samples and a lot of dependencies to keep up to date. From reviewing this PR, it's clear to me that a LOT of them are VERY out of date!

In terms of which way to go, the main thing I'm not sure about is how common it is for customers to copy/paste our samples and build independent projects out of them. Given that the dependencies were already super old and nobody mentioned anything, I'm assuming that folks may not be using them in this way. Thus, I'm currently leaning towards making them CPM to lower the maintenance costs of these samples.

Part of me also thinks we may want to get rid of some of these samples anyways, since they're largely for in-proc, which is being phased out. We may want to migrate the useful samples to the E2E test platform.

bachuv: Should we add some comments or a README in the samples that explain CPM so customers are aware of how they can override the version if they want to?

test/Common/DurableTaskEndToEndTests.cs
bachuv, line 1: Do we know why the `DurableEntity_AsyncIO` test is failing in the CI? Does it work for you locally?

bachuv (commented):
Thanks for opening this PR and fixing the warnings in the tests!

test/Common/DurableTaskEndToEndTests.cs
nytian: It works locally. So I think it should be fine. This test has been tricky all the time :(

samples/Directory.Packages.props
nytian: Thanks, everyone, for the discussion. I’ve added comments to all .csproj files in samples, directing the customer to refer to samples/Directory.Packages.props for package versions.

test/Common/DurableTaskEndToEndTests.cs
andystaples: I have been experimenting with this test - I think adding a small delay makes it more reliable, but I'm still testing different approaches for this. 
See this PR for reference 
https://github.com/Azure/azure-functions-durable-extension/pull/3052

andystaples (approved):

samples/VSSample.Tests/VSSample.Tests.csproj
```diff
@@ -5,6 +5,7 @@
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
+  <!-- We Move to central package management: For version details, please refer to samples/Directory.packages.props -->
```
bachuv: Can we change "We Move to central" to "We use central..."?

Also, can we give instructions about how they can override the versions?

bachuv (approved):

PR #3061, opened 2025-03-02
Relax type check in `OrchestrationInputConverter`

### LINKED ISSUE #3035: Orchestration input `IInputConverter` does not handle colletion iterfaces
### Description
Orchestrations receive `null` instead of an input if input binding type is anything other than `List<T>` or `T[]` (perhaps some other select cases).
If input type is e.g. `IEnuemrable<T>` or `IReadOnlyCollection<T>`, input silently gets converted into `null`.

### Expected behavior
Input is set to the value that was passed to the orchestration.

### Actual behavior
We get `null`

Test setup looks like this:

```csharp
public static class Endpoints
{
    [Function(nameof(PassIntCollectionOrchestration))]
    public static async Task PassIntCollectionOrchestration([OrchestrationTrigger] TaskOrchestrationContext context)
    {
        var input = Enumerable.Range(0, 10).ToList();

        await context.CallSubOrchestratorAsync(nameof(CheckForNullOrchestration), input);
    }

    [Function(nameof(CheckForNullOrchestration))]
    public static async Task CheckForNullOrchestration([OrchestrationTrigger] TaskOrchestrationContext context, IEnumerable<int> input)
    {
        if (input is null)
        {
            throw new InvalidOperationException("Got null where collection was expected");
        }

        await Task.CompletedTask;
    }

    [Function(nameof(HttpEntryPoint))]
    public static async Task HttpEntryPoint([HttpTrigger(AuthorizationLevel.Anonymous, "GET", Route = "debug")] HttpRequestData req, [DurableClient] DurableTaskClient client)
    {
        await client.ScheduleNewOrchestrationInstanceAsync(nameof(PassIntCollectionOrchestration));
    }
}
```

Now I dug deep into the framework and found the following:
- The input is correctly serialized, at least when executing locally backed by an instance of Azurite. Verified using `Durable Functions Monitor` extension for VSCode.
- While traversing the execution stack with debugger, I found `OrchestrationInputConverter` from `Microsoft.Azure.Functions.Worker.Extensions.DurableTask`, which I believe to me the main source of problems, since it returns `ConversionResult.Unhandled` when in reality it should return `Success`.
- I also discovered `DefaultFunctionInputBindingFeature` from `Microsoft.Azure.Functions.Worker.Context.Features`, which fall-backs to setting default (i.e. `null`) values to bindings that failed conversion with all available input converters (i.e. if all returned `ConversionResult.Unhandled`).

https://github.com/Azure/azure-functions-durable-extension/blob/85d4c87a76184a9cd78bc30f49ec31169dcc9aa7/src/Worker.Extensions.DurableTask/OrchestrationInputConverter.cs#L51-L73

Why is this wrong (in my opinion)? JSON (de-)serializer deserializes a collection (e.g., `[1, 2, 3, 4]`) into a `List<T>`, so `value.GetType()` is actually `List<T>`. However, the target type comes from the function signature! And it can be, e.g., an `IEnumerable<T>`! Naturally, `typeof(List<T>) != typeof(IEnumerable<T>)` and conversion fails. I believe the last check in condition should be replaced with `context.TargetType.IsInstanceOfType(value)`, which will allow passing `List<T>` as `IEnumerable<T>` (or practically any other meaningful collection as long as interfaces match). I did not check if target type is used for deseriazliation, but I believe it is the expected behavior when deserializing into collection interfaces -- you always get some sort of `List<T>` or `Dictionary<TK, TV>`.

Alternatively, a special branch could be introduced to only handle `IEnumerable<T>`-based types, if `context.TargetType.IsInstanceOfType(value)` is deemed too broad for general use cases.

The second issue is here:

https://github.com/Azure/azure-functions-dotnet-worker/blob/5366211c497e10e7c1534010069f79a9d06106ca/src/DotNetWorker.Core/Context/Features/DefaultFunctionInputBindingFeature.cs#L71-L92

When all conversions fail, the input binding is set to `null` because `IEnumerable<T>` is a reference type. Personally I would expect this branch to hard fail -- we have an input that is unbound and has no default parameter values, hence there is no way to populate it. That way we will get a much more concise and informative exception rather than dealing with wild `NullReferenceExceptions` because our collection input is suddenly `null`.

### Known workarounds
Yup, steal `OrchestrationInputConverter` code and tweak the condition mentioned above:

<details>
<summary>ModifiedOrchestrationInputConverter .cs</summary>

```csharp
using Microsoft.Azure.Functions.Worker;
using Microsoft.Azure.Functions.Worker.Converters;

public sealed class ModifiedOrchestrationInputConverter : IInputConverter
{
    private const string OrchestrationInputKey = "__orchestrationInput__";

    public static InputContext GetInputContext(FunctionContext context)
    {
        if (context is null)
        {
        }

        FunctionDefinition definition = context.FunctionDefinition;
        foreach (FunctionParameter parameter in definition.Parameters)
        {
            if (!IgnoredInputType(parameter.Type)
                && !definition.InputBindings.ContainsKey(parameter.Name)
                && !definition.OutputBindings.ContainsKey(parameter.Name))
            {
                // We take the first parameter we encounter which is not an explicitly ignored type
                // and has exactly 0 bindings (in or out) as our input candidate.
                return new(parameter.Type, context);
            }
        }

        return new(null, context);
    }

    public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
    {
        if (context is null)
        {
        }

        // We only convert if:
        // 1. The "Source" is null - IE: there are no declared binding parameters.
        // 2. We have a cached input.
        // 3. The TargetType matches our cached type.
        // If these are met, then we assume this parameter is the orchestration input.
        if (context.Source is null
            && context.FunctionContext.Items.TryGetValue(OrchestrationInputKey, out object? value)
// >>>>>>>HERE IS THE MODIFIED LINE<<<<<<<<<<<<<<<
            && context.TargetType.IsInstanceOfType(value))
        {
            // Remove this from the items so we bind this only once.
            context.FunctionContext.Items.Remove(OrchestrationInputKey);
            return new(ConversionResult.Success(value));
        }

        return new(ConversionResult.Unhandled());
    }

    private static bool IgnoredInputType(Type type)
    {
        // These are input types we know other converters handle.
        // TODO: is there a more concrete way we can determine if a type is already handled?
        return type == typeof(FunctionContext) || type == typeof(CancellationToken);
    }

    public class InputContext
    {
        private readonly FunctionContext context;
        private readonly Type? type;

        public InputContext(Type? type, FunctionContext context)
        {
            this.type = type;
            this.context = context;
        }

        public Type Type => this.type ?? typeof(object);

        public void PrepareInput(object? input)
        {
            // Short circuit if there is no input for our converter to handle, or if we have already prepared it.
            if (this.type is null || input is null || this.context.Items.ContainsKey(OrchestrationInputKey))
            {
                return;
            }

            this.context.Items[OrchestrationInputKey] = input;
        }
    }
}
```
</details>

Which can be then globally injected as 

```csharp
var host = new HostBuilder()
    .ConfigureFunctionsWorkerDefaults(
        builder => {  }, 
        opts =>
        {
            opts.InputConverters.RegisterAt<ModifiedOrchestrationInputConverter>(0);
        }
    )
    .Build();
```
### App Details

- `<PackageReference Include="Microsoft.Azure.Functions.Worker" Version="2.0.0" />`
- `<PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.DurableTask" Version="1.2.2" />`
- `<PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.2.0" />`
- `<PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="2.0.0" />`

### Screenshots

Image

Ilia-Kosenkov:
`OrchestrationInputConverter` responsible for processing orchestration inputs in isolated model has a strict type check that verifies that deserialized value's type equals _exactly_ to the target type in the function prototype. While it is correct for the majority of cases, it is not so when the target type is a collection interface. Standard behavior of, e.g., `System.Text.Json` serializer to produce a `List<T>` when target deserialization type is, e.g., `IEnumerable<T>`. Now, `List<T>` implements `IEnumerable<T>` and this can safely be passed to the orchestration, however since `typeof(List<T>) != typeof(IEnumerable<T>)`, the check fails and orchestration receives a `null` input.

This PR addresses this issue by slightly relaxing the type check, reporting a successful conversion of actual deserialized value can be assigned to target type due to inheritance or interface implementation. This unblocks collection interfaces _and_ custom json serializers that might return a more concrete type than target deserialization type (i.e. custom polymorphic deserialization).

resolves #3035 

### Pull request checklist

* [x] My changes **should** be added to **v2.x** branch.

### DIFF
--- a/src/Worker.Extensions.DurableTask/OrchestrationInputConverter.cs
+++ b/src/Worker.Extensions.DurableTask/OrchestrationInputConverter.cs
@@ -58,11 +58,13 @@ internal class OrchestrationInputConverter : IInputConverter
         // We only convert if:
         // 1. The "Source" is null - IE: there are no declared binding parameters.
         // 2. We have a cached input.
-        // 3. The TargetType matches our cached type.
+        // 3. The TargetType is either:
+        //    a. TargetType is in the inheritance hierarchy of the cached input (i.e. an ancestor type),
+        //    b. TargetType is an interface implemented by the cached input.
         // If these are met, then we assume this parameter is the orchestration input.
         if (context.Source is null
             && context.FunctionContext.Items.TryGetValue(OrchestrationInputKey, out object? value)
-            && context.TargetType == value?.GetType())
+            && context.TargetType.IsInstanceOfType(value))
         {
             // Remove this from the items so we bind this only once.
             context.FunctionContext.Items.Remove(OrchestrationInputKey);
--- a/test/Worker.Extensions.DurableTask.Tests/OrchestrationInputConverterTests.cs
+++ b/test/Worker.Extensions.DurableTask.Tests/OrchestrationInputConverterTests.cs
@@ -1,0 +1,72 @@
+﻿using Microsoft.Azure.Functions.Worker.Converters;
+using Microsoft.Azure.Functions.Worker.Extensions.DurableTask;
+using Moq;
+
+namespace Microsoft.Azure.Functions.Worker.Tests;
+
+public sealed class TestCollectionItem
+{
+}
+
+public sealed class OrchestrationInputConverterTests
+{
+    /// <summary>
+    /// JSON serializer typically deserialize collections as List{T} instances.
+    /// An instance of List{T} can be converted to any of the following interfaces or types:
+    /// </summary>
+    private static readonly Type[] GenericCollectionTypes =
+    {
+        typeof(IEnumerable<>),
+        typeof(IList<>),
+        typeof(ICollection<>),
+        typeof(IReadOnlyList<>),
+        typeof(IReadOnlyCollection<>),
+        typeof(List<>)
+    };
+
+    /// <summary>
+    /// System under test.
+    /// </summary>
+    private readonly IInputConverter inputConverter = new OrchestrationInputConverter();
+
+    private readonly Mock<FunctionContext> functionContextMock = new();
+
+    private readonly Mock<ConverterContext> converterContextMock = new();
+
+    public OrchestrationInputConverterTests()
+    {
+        // Default setup for ConvertAsync method:
+        //   - converterContext.Source should be null.
+        //   - converterContext.FunctionContext should be functionContextMock.Object.
+        converterContextMock.SetupGet(convCtx => convCtx.Source).Returns(default(object));
+        converterContextMock.SetupGet(convCtx => convCtx.FunctionContext).Returns(functionContextMock.Object);
+    }
+
+    /// <summary>
+    /// Generate test cases for each of the generic collection types.
+    /// </summary>
+    /// <param name="type">Collection item type.</param>
+    /// <returns>Test data.</returns>
+    public static TheoryData<Type> GenerateConcreteTestCollectionTypesFor(Type type)
+    {
+        return new TheoryData<Type>(GenericCollectionTypes.Select(ciType => ciType.MakeGenericType(type)));
+    }
+
+    [Theory(DisplayName = "ConvertAsync: Deserialized value is List<T> and target type is collection or collection interface of T")]
+    [MemberData(nameof(GenerateConcreteTestCollectionTypesFor), typeof(TestCollectionItem))]
+    public async Task ConvertAsync_WhenDeserializedValueIsListOfT_AndTargetTypeIsCollectionInterfaceOfT_ReturnsConversionResultSuccess(Type collectionType)
+    {
+        var inputData = new List<TestCollectionItem> {new(), new(), new()};
+        var functionContextItems = new Dictionary<object, object> {{"__orchestrationInput__", inputData}};
+        functionContextMock.SetupGet(funcCtx => funcCtx.Items).Returns(functionContextItems);
+        converterContextMock.SetupGet(convCtx => convCtx.TargetType).Returns(collectionType);
+
+        ConversionResult result = await inputConverter.ConvertAsync(converterContextMock.Object);
+
+        // Assert conversion is successful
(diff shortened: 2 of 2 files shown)


Ilia-Kosenkov:
@lilyjma , hello, here is my suggestion for #3035 . If this is acceptable, I probably will need some help dealing with necessary ceremonies.
I added unit tests, ensured that they fail on current `dev` branch -> added fix -> ensured tests are green.
If other tests are required (or if I should put these someplace else, please let me know).

I also produced a test nuget, referenced it locally in my test project, and verified that now the particular issue mentioned in #3035 is fixed.

Thanks in advance.

Ilia-Kosenkov:
@microsoft-github-policy-service agree company="Sievo"

lilyjma:
thanks @Ilia-Kosenkov for the contribution. I'll ask someone on the team to review your PR.

Ilia-Kosenkov:
Hey @lilyjma , a friendly reminder about this PR. The change is small, it would be nice to get an assessment whether it aligns with the direction the framework is taking right now. And to us, it is a small but nice quality of life improvement. Thanks!

cgillum:
@jviau do you see any issues with accepting this PR?

jviau (approved):

Ilia-Kosenkov:
So what happens now? The smoke test seems to fail due to test setup issue, can it be ignored for this specific PR?

cgillum:
Yeah, I think there's something wrong with the MSSQL test configuration. We can ignore it and merge.

PR #3065, opened 2025-03-06
Add E2E test for DTS backend

nytian:

Enable E2E test with DTS/AzureManaged backend

### Pull request checklist

### DIFF
--- a/.github/workflows/E2ETest.yml
+++ b/.github/workflows/E2ETest.yml
@@ -69,7 +69,7 @@ jobs:
       - name: Setup E2E tests
         shell: pwsh
         run: |
-          $env:MSSQL_SA_PASSWORD = "Str0ngP@ss!2025"
+          $env:MSSQL_SA_PASSWORD = "TEST12_$([Guid]::NewGuid())!"
           .\test\e2e\Tests\build-e2e-test.ps1 -StartMSSqlContainer
 
       - name: Build
--- a/.github/workflows/E2ETest.yml
+++ b/.github/workflows/E2ETest.yml
@@ -66,10 +66,13 @@ jobs:
         with:
           dotnet-version: 8.0.x
 
+      - name: Initialize Environment Variables
+        run: | 
+          echo "MSSQL_SA_PASSWORD=TEST12_$(echo $RANDOM)!" >> $GITHUB_ENV
+
       - name: Setup E2E tests
         shell: pwsh
         run: |
-          $env:MSSQL_SA_PASSWORD = "TEST12_$([Guid]::NewGuid())!"
           .\test\e2e\Tests\build-e2e-test.ps1 -StartMSSqlContainer
 
       - name: Build
--- a/.github/workflows/E2ETest.yml
+++ b/.github/workflows/E2ETest.yml
@@ -69,7 +69,7 @@ jobs:
       - name: Setup E2E tests
         shell: pwsh
         run: |
-          $env:MSSQL_SA_PASSWORD = "$([Guid]::NewGuid())"
+          $env:MSSQL_SA_PASSWORD = "Str0ngP@ss!2025"
           .\test\e2e\Tests\build-e2e-test.ps1 -StartMSSqlContainer
 
       - name: Build
--- a/.github/workflows/E2ETest.yml
+++ b/.github/workflows/E2ETest.yml
@@ -83,7 +83,7 @@ jobs:
   e2e-dts:
     runs-on: ubuntu-latest
     env:
-      E2E_TEST_DURABLE_BACKEND: "DTS"
+      E2E_TEST_DURABLE_BACKEND: "azureManaged"
     steps:
       - uses: actions/checkout@v4
 
@@ -108,4 +108,4 @@ jobs:
 
       - name: Run E2E tests
         working-directory: test/e2e/Tests
-        run: dotnet test --filter DTS!=Skip
+        run: dotnet test --logger "console;verbosity=detailed" --filter DTS!=Skip
--- a/.github/workflows/E2ETest.yml
+++ b/.github/workflows/E2ETest.yml
@@ -79,3 +79,33 @@ jobs:
       - name: Run E2E tests
         working-directory: test/e2e/Tests
         run: dotnet test --filter MSSQL!=Skip
+
+  e2e-dts:
+    runs-on: ubuntu-latest
+    env:
+      E2E_TEST_DURABLE_BACKEND: "DTS"
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup .NET Core
+        uses: actions/setup-dotnet@v3
+        with:
+          dotnet-version: 6.0.x
+
+      - name: Setup .NET Core
+        uses: actions/setup-dotnet@v3
+        with:
+          dotnet-version: 8.0.x
+
+      - name: Setup E2E tests
+        shell: pwsh
+        run: |
+          .\test\e2e\Tests\build-e2e-test.ps1 -StartDTSContainer
+
+      - name: Build
+        working-directory: test/e2e/Tests
+        run: dotnet build
+
+      - name: Run E2E tests
+        working-directory: test/e2e/Tests
+        run: dotnet test --filter DTS!=Skip
--- a/test/e2e/Tests/E2ETests.csproj
+++ b/test/e2e/Tests/E2ETests.csproj
@@ -5,6 +5,8 @@
     <LangVersion>latest</LangVersion>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
+    <!-- Only add the following line when you want to locally debug
+    <RunSettingsFilePath>$(MSBuildProjectDirectory)\test.runsettings</RunSettingsFilePath> -->   
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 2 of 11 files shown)


test/e2e/Tests/Fixtures/FixtureHelpers.cs
```diff
@@ -99,6 +99,12 @@ internal static void AddDurableBackendEnvironmentVariables(Process funcProcess,
                 funcProcess.StartInfo.EnvironmentVariables["AzureFunctionsJobHost__extensions__durableTask__storageProvider__connectionStringName"] = "SQLDB_Connection";
                 funcProcess.StartInfo.EnvironmentVariables["AzureFunctionsJobHost__extensions__durableTask__storageProvider__createDatabaseIfNotExists"] = "true";
                 return;
+            case "azuremanaged":
```
andystaples, line 102: This should agree with the value in line#86 of E2ETest.yml

test/e2e/Tests/Tests/ErrorHandlingTests.cs
```diff
@@ -32,7 +32,7 @@ public async Task OrchestratorWithUncaughtActivityException_ShouldFail()
         await DurableHelpers.WaitForOrchestrationStateAsync(statusQueryGetUri, "Failed", 30);
 
         var orchestrationDetails = await DurableHelpers.GetRunningOrchestrationDetailsAsync(statusQueryGetUri);
-        Assert.StartsWith("Microsoft.DurableTask.TaskFailedException", orchestrationDetails.Output);
+       // Assert.StartsWith("Microsoft.DurableTask.TaskFailedException", orchestrationDetails.Output);
```
andystaples: If commenting these lines is necessary, let's document why - have we opened a bug against DTS to fix whatever is causing the difference in exception logging?

test/e2e/Tests/test.runsettings
```diff
@@ -9,7 +9,10 @@
             <MSSQL_SA_PASSWORD>{{YourPasswordHere}}</MSSQL_SA_PASSWORD>
             <E2E_TEST_DURABLE_BACKEND>MSSQL</E2E_TEST_DURABLE_BACKEND>
             -->
-            <E2E_TEST_DURABLE_BACKEND>AzureStorage</E2E_TEST_DURABLE_BACKEND>
+			<!-- Enable this if you want to run tests against DTS
```
andystaples: nit. consistent comment spacing with the above set of comments

andystaples (commented):
Left a few minor comments

test/e2e/Tests/E2ETests.csproj
```diff
@@ -5,7 +5,8 @@
     <LangVersion>latest</LangVersion>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
-    <RunSettingsFilePath>$(MSBuildProjectDirectory)\test.runsettings</RunSettingsFilePath>
+    <!-- Only add the following line when you want to locally debug
+    <RunSettingsFilePath>$(MSBuildProjectDirectory)\test.runsettings</RunSettingsFilePath> -->
```
nytian, line 9: This has to be commented out. Otherwise, environment variables set in the E2ETest.yml will be overwritten by this file and thus all the E2E CI will use AzureStorage backend

test/e2e/Tests/Tests/ErrorHandlingTests.cs
```diff
@@ -32,7 +32,11 @@ public async Task OrchestratorWithUncaughtActivityException_ShouldFail()
         await DurableHelpers.WaitForOrchestrationStateAsync(statusQueryGetUri, "Failed", 30);
 
         var orchestrationDetails = await DurableHelpers.GetRunningOrchestrationDetailsAsync(statusQueryGetUri);
-        Assert.StartsWith("Microsoft.DurableTask.TaskFailedException", orchestrationDetails.Output);
+        
+        // In DTS, the output message is:
+        // "Task 'RaiseException' (#0) failed with an unhandled exception: This activity failed"
+        // Assert.StartsWith("Microsoft.DurableTask.TaskFailedException", orchestrationDetails.Output);
+
```
nytian: DTS return differs slightly with the AS backend. And I've attached the details here. But @andystaples we want all backends to remain consistent. I'm not sure if this difference will be updated further, so please let me know if I should avoid commenting this out and instead skip the test for DTS.

test/e2e/Tests/Tests/PurgeInstancesTests.cs
```diff
@@ -48,6 +48,7 @@ public async Task PurgeOrchestrationHistory_Start_Succeeds()
     }
 
     [Fact]
+    [Trait("DTS", "Skip")] // Skip this test as there is a bug with current DTS backend, the createdTimeTo couldn't be null.
```
nytian, line 51: This bug is fixed. And we should enable the test once the latest dp rolls out.

.github/workflows/E2ETest.yml
```diff
@@ -69,7 +69,7 @@ jobs:
       - name: Setup E2E tests
         shell: pwsh
         run: |
-          $env:MSSQL_SA_PASSWORD = "$([Guid]::NewGuid())"
+          $env:MSSQL_SA_PASSWORD = "Str0ngP@ss!2025"
```
nytian: The GUID doesn't guarantee that the generated output meets the SQL password requirements every time. If it doesn't, the db will automatically shut down and thus cause the tests failed. I updated it to be like this, and it worked good with my local end. But I still see `sa log in failed` failures in the CI logs. Could you check this?@andystaples

cgillum: Hardcoded passwords are likely to be flagged. How about something like this?
```suggestion
          $env:MSSQL_SA_PASSWORD = "TEST12_$([Guid]::NewGuid())!"
```

test/e2e/Tests/Tests/ErrorHandlingTests.cs
cgillum: Do we know why the message is different? Could it be a bug that we should fix?

test/e2e/Tests/Tests/ErrorHandlingTests.cs
```diff
@@ -110,6 +118,7 @@ public async Task OrchestratorWithRetriedActivityException_ShouldSucceed()
 
     [Fact]
     [Trait("MSSQL", "Skip")] // Durable Entities are not supported in MSSQL/Dotnet Isolated, see https://github.com/microsoft/durabletask-mssql/issues/205
+    [Trait("DTS", "Skip")] // DTS will fail this test becasue the exception  type is `InvalidOperationException` instead of `EntityOperationFailedException`
```
cgillum: Can we open a bug for this? It would be better if all backends return a consistent exception type so that customer code doesn't need to be changed when switching to DTS.

test/e2e/Tests/Tests/OrchestrationQueryTests.cs
```diff
@@ -23,6 +23,7 @@ public OrchestrationQueryTests(FunctionAppFixture fixture, ITestOutputHelper tes
 
 
     [Fact]
+    [Trait("DTS", "Skip")] // Skip this test for now as there is a bug in DTS with GetAllInstances
```
cgillum: Can you provide a GitHub or ADO link? We should track these bugs and make sure they get fixed.

test/e2e/Tests/build-e2e-test.ps1
```diff
@@ -192,6 +195,26 @@ function StartMSSQLContainer($mssqlPwd) {
   docker ps
 }
 
+function StartDTSContainer() {
+  Write-Host "Pulling down the durabletaskspublic.azurecr.io/dts-emulator:latest-amd64 image..."
+  docker pull durabletaskspublic.azurecr.io/dts-emulator:latest-amd64
+
+  # Start the DTS Server docker container with the specified edition
+  Write-Host "Starting DTS docker container on port 8080" -ForegroundColor DarkYellow
+  docker run -itP -p 8080:8080 -p 8082:8082 -e ClientAuth__DisableAuthentication=true -d durabletaskspublic.azurecr.io/dts-emulator:latest-amd64
```
cgillum: We have an official MCR distribution of this container now, and it fixes some bugs, like the need for the extra environment variable to disable auth.
```suggestion
  docker run -itP -p 8080:8080 -p 8082:8082 -d mcr.microsoft.com/dts/dts-emulator
```

test/e2e/Tests/build-e2e-test.ps1
```diff
@@ -192,6 +195,26 @@ function StartMSSQLContainer($mssqlPwd) {
   docker ps
 }
 
+function StartDTSContainer() {
+  Write-Host "Pulling down the durabletaskspublic.azurecr.io/dts-emulator:latest-amd64 image..."
+  docker pull durabletaskspublic.azurecr.io/dts-emulator:latest-amd64
```
cgillum:
```suggestion
  docker pull mcr.microsoft.com/dts/dts-emulator
```

cgillum (commented):
Really glad to see this! A few minor comments.

test/e2e/Tests/Tests/ErrorHandlingTests.cs
andystaples: Let's do the same as the other failing tests - file a bug against DTS, mark as skipped for now, with a comment + link to the bug

test/e2e/Tests/E2ETests.csproj
andystaples: Good catch! Looks like there are some mssql tests failing now, we'll have to address those, but it can be done separately

cgillum (approved):

PR #3066, opened 2025-03-08
Distributed Tracing for Entities (In-Process)

sophiatev:
This PR adds distributed tracing for entities, specifically for

- Calling entities (from an orchestration)
- Signaling entities (from an orchestration, a client, or another entity)
- An entity starting an orchestration

The following files had tracing added for the above cases

- `DurableClient`, for a client signaling an entity 
- `DurableEntityContext`, for an entity signaling another entity or an entity starting an orchestration
- `DurableOrchestrationContext`, for an orchestration calling or signaling an entity
- `TaskEntityShim`, for an entity processing a request

Other file changes are
- In order to correctly correlate traces in the OOProc case, the OOProc languages will pass trace information in the headers of their HTTP signal entity requests. This trace information is parsed by the `HttpApiHandler`, which uses it to "recreate" the `Activity` from the OOProc side via `TraceHelper.StartActivityUsingTraceContext`. This `Activity` includes only the minimal information necessary such that `Activity.Current.Context` is correctly set to contain the trace context information passed in from OOProc in the HTTP headers. In this way, the `Activity` for signaling the entity that is ultimately created by `DurableClient.SignalEntityAsync` (which uses `Activity.Current.Context` as the parent trace context) will correctly be correlated as the child of the `Activity` from the OOProc side. 
-  `EntityTriggerAttributeBindingProvider` is modified to pass extra information to the OOProc languages - specifically, the request ID is passed in the case of signaling an entity. This is necessary for if the signaled entity itself ends up signaling another entity. Since the entity-to-entity signals are processed as a batch in `TaskEntityShim`, we need a way to correlate these entity-to-entity signals to the original signal entity request, such that traces are correctly correlated.
- `TraceHelper`, `Schema`, and `TraceActivityConstants` all contain code assisting in creating the traces. `DiagnosticActivityExtensions` which mimics the same [file ](https://github.com/Azure/durabletask/blob/main/src/DurableTask.Core/Tracing/DiagnosticActivityExtensions.cs) in DurableTask.Core, is used to create custom setters for the span id, trace state, and trace id of an `Activity`. This is used by `TraceHelper.StartActivityUsingTraceContext` when `HttpApiHandler` "recreates" the OOProc `Activity` based on the information passed in the HTTP headers.
-  `RequestMessage` is modified to contain additional trace and request information that is used by `TaskEntityShim` to correctly correlate the trace corresponding to an entity processing an invocation as the child of the trace for the signal/call entity request.
- `DurableTaskExtension` is modified to tell DurableTask.Core that entities are enabled (see this [PR](https://github.com/Azure/durabletask/pull/1190)) 

Here is an example of the new tracing logic for a simple orchestration. The orchestration first signals an entity, then signals an entity again, with this second signal request itself triggering the entity to signal another entity. Finally the orchestration calls the two entities and returns. 

image

The resulting traces are

image

Each signal request has type `ActivityKind.Producer` and each call request has type `ActivityKind.Client` (an entity starting an orchestration is also of type `ActivityKind.Producer`). When an entity actually processes the request, for a signal the span has type `ActivityKind.Producer` and for a call the span has type `ActivityKind.Server`. 

If instead of starting the orchestration via an HTTP request we signal an entity to start the orchestration, the trace would look like this

image

### Pull request checklist

    * [x] Otherwise: This will be done in a future PR


src/WebJobs.Extensions.DurableTask/TraceHelper.cs
```diff
@@ -0,0 +1,116 @@
...
+
+#nullable enable
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
+{
+    /// <summary>
+    /// Provides helper methods for tracing orchestration activities.
+    /// </summary>
+    internal class TraceHelper
+    {
+        private const string Source = "DurableTask.Core";
```
bachuv: This should be `WebJobs.Extensions.DurableTask`

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -328,10 +328,14 @@ private async Task SignalEntityAsyncInternal(DurableClient durableClient, string
             var guid = Guid.NewGuid(); // unique id for this request
             var instanceId = EntityId.GetSchedulerIdFromEntityId(entityId);
             var instance = new OrchestrationInstance() { InstanceId = instanceId };
+
+            using var signalEntityActivity = TraceHelper.StartActivityForCallingOrSignalingEntity(instanceId, entityId.EntityName, operationName, true, Activity.Current?.Context);
+            var parentTraceContext = new ActivityContext(signalEntityActivity.TraceId, signalEntityActivity.SpanId, signalEntityActivity.ActivityTraceFlags, signalEntityActivity.TraceStateString);
```
bachuv: Does `ActivityContext()` take null parameters? Let's make sure that we have the appropriate null checks in case `signalEntityActivity` is null. Should we add[ null conditional operators](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-) here (e.g. signalEntityActivity?.TraceId)?

(and same comment for the other context implementations)

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
```diff
@@ -424,10 +426,14 @@ private void SignalEntityInternal(EntityId entity, DateTime? scheduledTimeUtc, s
             {
                 InstanceId = EntityId.GetSchedulerIdFromEntityId(entity),
             };
+
+            using var signalEntityActivity = TraceHelper.StartActivityForCallingOrSignalingEntity(target.InstanceId, entity.EntityName, operation, true, Activity.Current?.Context, this.InstanceId);
+            var parentTraceContext = new ActivityContext(signalEntityActivity.TraceId, signalEntityActivity.SpanId, signalEntityActivity.ActivityTraceFlags, signalEntityActivity.TraceStateString);
```
cgillum: It looks like we call this constructor with a long list of parameters in multiple places. I haven't examined the code super closely yet, but could we remove some of the redundancy a bit by creating a method that allows us to do something like `var parentTraceContext = signalEntityActivity.ToActivityContext()`?

src/WebJobs.Extensions.DurableTask/Correlation/TraceActivityConstants.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System;
```
cgillum: Looks like this can be removed?

src/WebJobs.Extensions.DurableTask/Correlation/TraceActivityConstants.cs
```diff
@@ -10,8 +12,12 @@ internal class TraceActivityConstants
         public const string Activity = "activity";
         public const string Event = "event";
         public const string Timer = "timer";
+        public const string Entity = "entity";
 
         public const string CreateOrchestration = "create_orchestration";
         public const string OrchestrationEvent = "orchestration_event";
+
+        public const string CallEntity = "CallEntity";
+        public const string SignalEntity = "SignalEntity";
```
cgillum: Looking at other constants in this file, it seems we're using snake_case for the values. For example `create_orchestration` and `orchestration_event`. Should we be doing the same for these new constants you're defining?

cgillum (commented):
Added a few minor comments.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -4,15 +4,19 @@
...
 using System.Net.Http;
 using System.Reflection;
 using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
+using Azure.Core;
 using DurableTask.Core;
 using DurableTask.Core.Exceptions;
+using DurableTask.Core.Tracing;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask.Correlation;
```
bachuv: Can we remove `using Azure.Core` and `using Microsoft.Azure.WebJobs.Extensions.DurableTask.Correlation`?

src/WebJobs.Extensions.DurableTask/Correlation/Schema.cs
bachuv, line 1: There were some comments about updating the schema in the durabletask PR. Please update this PR if there are any updates in the durabletask PR.

src/WebJobs.Extensions.DurableTask/Bindings/EntityTriggerAttributeBindingProvider.cs
```diff
@@ -183,7 +183,8 @@ private static string EntityContextToString(DurableEntityContext arg)
                     operationBatch.Add(new JObject(
                         new JProperty("name", operation.Operation),
                         new JProperty("input", operation.Input),
-                        new JProperty("isSignal", operation.IsSignal)));
+                        new JProperty("isSignal", operation.IsSignal),
+                        new JProperty("id", operation.Id)));
```
bachuv, line 187: Where is this `id` used?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -327,6 +335,11 @@ private async Task SignalEntityAsyncInternal(DurableClient durableClient, string
             var guid = Guid.NewGuid(); // unique id for this request
             var instanceId = EntityId.GetSchedulerIdFromEntityId(entityId);
             var instance = new OrchestrationInstance() { InstanceId = instanceId };
+
+            using var signalEntityActivity = TraceHelper.StartActivityForCallingOrSignalingEntity(instanceId, entityId.EntityName, operationName, signalEntity: true, scheduledTimeUtc, this.ParentTraceContext ?? Activity.Current?.Context);
+
+            // Reset state so that subsequent requests do not use the same parent trace context
+            this.ParentTraceContext = null;
```
bachuv: Why do we need this line? Can you give an example?

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -1152,6 +1117,19 @@ private async Task<HttpResponseMessage> HandlePostEntityOperationRequestAsync(
                 operationName = request.GetQueryNameValuePairs()["op"] ?? string.Empty;
             }
 
+            if (client is DurableClient durableClient)
+            {
+                string traceParent = GetHeaderValueFromHeaders("traceparent", request.Headers);
+                string traceState = GetHeaderValueFromHeaders("tracestate", request.Headers);
+
+                // Note that we break the pattern here of not creating the Activity for signaling the entity if we cannot successfully parse the parent trace context (since DurableClient will just use Activity.Current.Context instead)
+                // In that case the Activity will not correctly attach to the HTTP request, but will correctly propagate all the other Activities following the signal entity request
```
bachuv: Are you saying here that signaling an entity uses `durableClient.ParentTraceContext` to correlate the traces?

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -552,17 +558,22 @@ private async Task ProcessOperationRequestAsync(RequestMessage request)
...
+            processEntityInvocationActivity?.Dispose();
+            callEntityActivity?.Dispose();
         }
 
         private async Task ExecuteOutOfProcBatch()
         {
             OutOfProcResult outOfProcResult = null;
 
+            // We are crudely setting the start time of all the entity invocations to be the same since they are executed together in a batch
+            // var startTime = DateTimeOffset.UtcNow;
```
bachuv: Should we remove this line that's commented out?

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -761,7 +828,13 @@ public struct OperationResult
...
                 /// </summary>
                 [JsonProperty("duration")]
-                public double DurationInMilliseconds { get; set; }
+                public double DurationInMilliseconds { get; set; }
+
+                /// <summary>
+                /// The start time of this operation's execution, in milliseconds, since January 1st 1970 midnight in UTC.
+                /// </summary>
+                [JsonProperty("startTime")]
+                public long StartTimeInMilliseconds { get; set; }
```
bachuv: Should we rename this property to include "execution"? Let's keep it consistent/similar to the property name that is in the related Python Entities tracing PR.

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
bachuv, line 1: The diff in this file seems to include some lines that were unchanged like `this.context.CurrentOperationResponse = new ResponseMessage();`. Do you know if we can remove those unchanged lines from the diff? I'm not sure why they are showing up. Feel free to reach out to me about this and we can look at it together.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
sophiatev: One situation where it could be useful is if we have two client requests to signal an entity via a Python app, for example. Say that the parent trace context passed from the Python repo to WebJobs in the second case is malformed, and so `HttpApiHandler` is unable to parse and use it. In this case, since the parent trace context from the first signal entity request is still attached to the `DurableClient`, the second signal to the entity would incorrectly use it as its parent trace. 

Although, right now, what will happen is that the second signal to the entity would use `Activity.Current` as the parent instead, which means that it will have no parent at all. `DurableClient.SignalEntity` is used in both OOProc and non-OOProc, where for OOProc we want to manually attach the parent trace context via the `ParentTraceContext` field, and for non-OOProc we just want to use `Activity.Current` as the parent. The current logic checks if `ParentTraceContext` has been set, and if not, just uses `Activity.Current`. If we can't parse the parent trace context in `HttpApiHandler`, then `ParentTraceContext` will not be set, and so `Activity.Current` is used in OOProc in that situation too. I was conflicted as to what to do here since in all other non-OOProc cases, if we can't parse the parent trace, we just don't create the `Activity`. (This is what I tried to address in the comment on line 1125 in `HttpApiHandler`). I couldn't find a clean way to avoid creating the `Activity` for OOProc since even if `HttpApiHandler` doesn't set `this.ParentTraceContext`, the `Activity` is still be created using `Activity.Current` as the parent. I think to avoid creating the `Activity` in this case we would have to add another field to `DurableClient`, indicating if the call is OOProc or not. So it's basically a conflict between consistency across all cases vs. "cleanness", and I wasn't sure what the best option is.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
sophiatev: Sorry, this comment didn't do a good job explaining what I meant. But I actually address this in this [reply ](https://github.com/Azure/azure-functions-durable-extension/pull/3066#discussion_r2078252142) here (and I'm not sure what the best option is, actually).

src/WebJobs.Extensions.DurableTask/Bindings/EntityTriggerAttributeBindingProvider.cs
sophiatev: I tried to address this in the PR description but let me know if I should add a comment in the code as well :) 

"`EntityTriggerAttributeBindingProvider` is modified to pass extra information to the OOProc languages - specifically, the request ID is passed in the case of signaling an entity. This is necessary for if the signaled entity itself ends up signaling another entity. Since the entity-to-entity signals are processed as a batch in `TaskEntityShim`, we need a way to correlate these entity-to-entity signals to the original signal entity request, such that traces are correctly correlated."

src/WebJobs.Extensions.DurableTask/Correlation/DiagnosticActivityExtensions.cs
bachuv, line 1: Can you add a file header to this file?

```
```

bachuv, line 1: Can you take a look at the errors/warnings in this file?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -689,19 +693,49 @@ internal async Task<TResult> CallDurableTaskFunctionAsync<TResult>(
...
+                            request.RequestTime = DateTimeOffset.UtcNow;
+                        }
+
+                        // We still want to attach the current Activity as the parent trace context to the request in the case of a call to an entity so that when we create the Activity for the call this information is available.
+                        // In the case of signaling the entity, Activity.Current will be set to the signalEntityActivity just started.
+                        if (Activity.Current is { } activity)
+                        {
+                            request.ParentTraceContext = new DTCore.Tracing.DistributedTraceContext(activity.Id, activity.TraceStateString);
+                        }
+                    }
```
bachuv: Can we move this code to a helper method?

src/WebJobs.Extensions.DurableTask/EntityScheduler/RequestMessage.cs
```diff
@@ -81,7 +82,19 @@ internal class RequestMessage
...
 
         [JsonIgnore]
-        public bool IsLockRequest => this.LockSet != null;
+        public bool IsLockRequest => this.LockSet != null;
+
+        /// <summary>
+        /// Parent trace context of this request message.
+        /// </summary>
+        [JsonProperty(PropertyName = "parentTraceContext", DefaultValueHandling = DefaultValueHandling.Ignore)]
+        public DistributedTraceContext ParentTraceContext { get; set; }
```
bachuv, line 91: Should this be a nullable property?

```suggestion
        public DistributedTraceContext? ParentTraceContext { get; set; }
```

bachuv (approved):
LGTM. Let's merge this PR once the related DT.Core changes are released so the CI passes.

bachuv (approved):

Issue #2689, opened 2023-12-05
Failure details from sub-orchestrators come back as '(unknown)'


PaulBDavies:
### Description
When a sub-orchestration fails, it throws a 'TaskFailedException' in the parent orchestrator. However when I catch this error, the FailureDetails come out as '(unknown)' for both ErrorType and ErrorMessage. I'm assuming this isn't intended behaviour, since when I try the same thing with activity functions I'm able to retrieve useful information from these fields.

### Expected behavior
The FailureDetails of the TaskFailedException contains the error message/type of the exception that caused the failure.

### Actual behavior
The FailureDetails of the TaskFailedException both contain the string '(unknown)'

### Relevant source code snippets
Here is some pseudocode similar to my own, I'm performing the same operation for multiple items in a list. If one fails I still want to perform the operation for the remainder of the items in the list, but I want to keep track of any errors that happened along the way.
```csharp
foreach(var item in list)
{
    try
    {
        await context.CallSubOrchestratorAsync(nameof(FailingFunction), item);
    }
    catch(TaskFailedException ex)
    {
        var error = ex.FailureDetails?.ErrorMessage ?? "Unknown Error";
        errors.Add(error);
    }
}
// I'd expect 'errors' here to contain a list of error messages, which I can log/throw as an aggregate exception
// or something, but they are all just '(unknown)'
```

### Known workarounds
N/A

### App Details

- **Durable Functions extension version**: v1.0.0 (although I did try upgrading to v1.1.0 and saw the same issue) - This is the package version on nuget and doesn't seem to line up with the tags on this repo?
- **Azure Functions runtime version**: v4
- **Programming language used**: C#

### Screenshots
N/A

### If deployed to Azure
N/A

JohanKlijn:
Any news on this issue?

We use a lot of sub-orchestrators and our main orchestrator is handling/logging all the exceptions and now we all the relevant information for debugging issues are not available any more.

cgillum:
@JohanKlijn a new set of nuget packages were released yesterday and have some fixes related to exception propagation. Can you try upgrading to the latest to see if this issue is resolved for you?

PaulBDavies:
This seems to have resolved the issue for me

andrew-xu-asurion:
we're having the same issue, the details of exception thrown at sub orchestration cannot be captured at orchestration. Are there any solutions?

asos-gurpreetsingh:
Having same issues

[...]

PR #3070, opened 2025-03-19
Improve Error Type Accuracy in TaskFailedException FailureDetails with .NET-Isolated app

nytian:
partialy solves #2689  

This PR ensures that when orchestrators catch exceptions, the FailureDetails.ErrorType will match the type of the original exception thrown by the activity or sub-orchestration function.

For example, if an activity function throws an InvalidOperationException, the FailureDetails.ErrorType in the resulting TaskFailedException should be "InvalidOperationException" .

### Pull request checklist

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -64,7 +64,7 @@
     <PackageVersion Include="Microsoft.CodeAnalysis" Version="3.9.0" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Workspaces.MSBuild" Version="3.9.0" />
     <PackageVersion Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="2.0.65" />
-    <PackageVersion Include="Microsoft.DurableTask.Abstractions" Version="1.8.0" />
+    <PackageVersion Include="Microsoft.DurableTask.Abstractions" Version="1.8.1" />
     <PackageVersion Include="Microsoft.DurableTask.Generators" Version="1.0.0-preview.1" />
     <PackageVersion Include="Microsoft.DurableTask.SqlServer.AzureFunctions" Version="1.5.0" />
     <PackageVersion Include="Microsoft.Extensions.Configuration" Version="9.0.0" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -64,7 +64,7 @@
     <PackageVersion Include="Microsoft.CodeAnalysis" Version="3.9.0" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Workspaces.MSBuild" Version="3.9.0" />
     <PackageVersion Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="2.0.65" />
-	  <PackageVersion Include="Microsoft.DurableTask.Abstractions" Version="1.8.0" />
+    <PackageVersion Include="Microsoft.DurableTask.Abstractions" Version="1.8.0" />
     <PackageVersion Include="Microsoft.DurableTask.Generators" Version="1.0.0-preview.1" />
     <PackageVersion Include="Microsoft.DurableTask.SqlServer.AzureFunctions" Version="1.5.0" />
     <PackageVersion Include="Microsoft.Extensions.Configuration" Version="9.0.0" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -64,6 +64,7 @@
     <PackageVersion Include="Microsoft.CodeAnalysis" Version="3.9.0" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Workspaces.MSBuild" Version="3.9.0" />
     <PackageVersion Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="2.0.65" />
+	  <PackageVersion Include="Microsoft.DurableTask.Abstractions" Version="1.8.0" />
     <PackageVersion Include="Microsoft.DurableTask.Generators" Version="1.0.0-preview.1" />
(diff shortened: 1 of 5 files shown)


test/e2e/Apps/BasicDotNetIsolated/ActivityErrorHandling.cs
```diff
@@ -96,11 +112,26 @@ public static async Task<string> CatchActivityException(
...
         }
         catch (TaskFailedException ex)
-        {
+        {  
             return ex.Message;
         }
     }
 
+    [Function(nameof(CatchActivityExceptionFailreDetails))]
+    public static async Task<TaskFailureDetails> CatchActivityExceptionFailreDetails(
```
andystaples: Let's make this Task<TaskFailureDetails?> to make the compiler happy

test/e2e/Apps/BasicDotNetIsolated/ActivityErrorHandling.cs
```diff
@@ -96,11 +112,26 @@ public static async Task<string> CatchActivityException(
             return output;
         }
         catch (TaskFailedException ex)
-        {
+        {  
             return ex.Message;
         }
     }
 
+    [Function(nameof(CatchActivityExceptionFailreDetails))]
```
andystaples: nit. Typo: CatchActivityExceptionFailureDetails

src/Worker.Extensions.DurableTask/Exceptions/DurableSerializationException.cs
```diff
@@ -11,14 +11,20 @@ internal class DurableSerializationException : Exception
     // We set the base class properties of this exception to the same as the parent, 
     // so that methods in the worker after this can still (typically) access the same information vs w/o
     // this exception type. 
-    internal DurableSerializationException(Exception fromException) : base(fromException.Message, fromException.InnerException)
+    internal DurableSerializationException(Exception fromException) : base(CreateExceptionMessage(fromException), fromException.InnerException)
```
andystaples, line 14: Just want to understand why this change was made - the reason that I only did the JSON serialization in ToString() originally was because I didn't want to alter the actual properties of the exception - before this change, if the functions worker needed to access the properties of the user code exception, they were functionally unmodified, and it wasn't until the exception was converted to a string for gRPC transmission that the serialization happened. 
Can you help me understand what was the negative effect of this decision and what does serializing the message property improve?

andystaples (commented):
Left some comments

src/Worker.Extensions.DurableTask/Exceptions/DurableSerializationException.cs
nytian: Customers want more detailed information about exceptions within the `FailureDetails` of a TaskFailedEvent, including the specific error type. Currently, this detail is missing, and thus cx will always get `unknown` type inside FailureDetails.

Why this is missing, is because we only send the exception message at DurableSerializationException. For some reason, the ToString() is only called by grpc log, while the grpc worker only pass exception message and inner Failure. (Check class[ here](https://github.com/Azure/azure-functions-host/blob/cb16901e58789c7aed5db8e8c913d71b9c8fc2a6/src/WebJobs.Script/Workers/Rpc/RpcException.cs#L9)) 

Thus, when DurableSerializationException went to the OutOfProcMiddleware [GetFailureDetails](https://github.com/Azure/azure-functions-durable-extension/blob/bb01165b968fe40f7615e66f9991276cec316b45/src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs#L614), we will fail for TryExtractSerializedFailureDetailsFromException and TrySplitExceptionTypeFromMessage, and cause the type becomes `unknown`. 

I don't believe I've altered the original user exception; I'm only updating DurableSerializationException, which I assume is used for passing errors via gRPC, if I understand your question correctly. My main concern is that serializing the entire failure details might be excessive and redundant—for example, the error trace and inner exception will be included in the message while also being part of the full exception. However, since I suppose `TryExtractSerializedFailureDetailsFromException` should be used, I'll leave it as is.

andystaples: Understood. 
I think what I was missing was this:
> For some reason, the ToString() is only called by grpc log, while the grpc worker only pass exception message and inner Failure

Very surprising, TBH. 

When I was talking about the "properties of the user code exception" I meant the properties of the exception exposed to the ooproc worker - before my change that added DurableSerializationException, the native user code exception type was thrown all the way up until the gRPC logging step. I was trying to be extra careful not to change behavior of the worker, especially for "special case" exception types like OutOfMemoryException. 

Anyway, based on the test case (which looks great, btw) and your own experimentation, I'm good to approve this PR. Thanks for investigating this!

andystaples (approved):
Approved

andystaples:
BTW - I see the failure in mssql E2E tests - NBD for me. I'll need to investigate this at some point, but definitely not related to these changes.

Directory.Packages.props
```diff
@@ -64,6 +64,7 @@
     <PackageVersion Include="Microsoft.CodeAnalysis" Version="3.9.0" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Workspaces.MSBuild" Version="3.9.0" />
     <PackageVersion Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="2.0.65" />
+    <PackageVersion Include="Microsoft.DurableTask.Abstractions" Version="1.8.0" />
```
cgillum: Should this be `1.8.1`, which is the latest version?

cgillum (approved):
LGTM!

PR #3072, opened 2025-03-24
Enable version-aware orchestrations (.NET in-proc)

AnatoliB:
### Goal

The goal of this PR is to allow updating orchestration logic without requiring users to drain or abandon in-flight orchestration instances. Developers will be able to maintain backward compatibility in their orchestrator function code by checking which version of the app created the current orchestration instance and, based on that, executing either the new or old logic. This preserves the intent of the original orchestrator function and avoids non-determinism issues during function replay.

This PR enables this capability for .NET in-process Durable Function apps. Support for other runtimes is coming soon.

### Details

- The user will be able to assign a `defaultVersion` property to their durable app in `host.json`. All orchestration instances created by this app will automatically have this version value permanently associated with them.
- Orchestrator functions will have access to a `context.Version` property containing the version value that was set when the orchestration was created.

### Example

Suppose that the original app contains:

**host.json**
```json
{
  "extensions": {
    "durableTask": {
      "defaultVersion": "1.0"
    }
  }
}
```

**Orchestrator function**
```csharp
[FunctionName("MyOrchestration")]
public static async Task RunOrchestrator(
{
    await context.CallActivityAsync<string>("ActivityA", ...));
    await context.CallActivityAsync<string>("ActivityB", ...));
    await context.CallActivityAsync<string>("ActivityC", ...));
}
```

In the next version, we want to invoke `ActivityD` instead of `ActivityB`. However, we expect some orchestrations to be still running during the app upgrade, and some of them may have have already invoked `ActivityB`. Simply replacing `ActivityB` invocation with `ActivityD` could cause non-determinism issues when replaying history with the new orchestrator code.

Here is what we can do to keep our orchestrator backward compatible:

**host.json**
```json
{
  "extensions": {
    "durableTask": {
      "defaultVersion": "2.0"
    }
  }
}
```

**Orchestrator function**
```csharp
[FunctionName("MyOrchestration")]
public static async Task RunOrchestrator([OrchestrationTrigger] IDurableOrchestrationContext context)
{
    await context.CallActivityAsync<string>("ActivityA", ...));
    if (context.Version == "1.0") {
        // legacy code path
        await context.CallActivityAsync<string>("ActivityB", ...));
    } else {
        // new code path
        await context.CallActivityAsync<string>("ActivityD", ...));
    }
    await context.CallActivityAsync<string>("ActivityC", ...));
}
```

As a result, orchestrations created before the app upgrade will follow the old code path and invoke `ActivityB`, while new orchestrations will invoke `ActivityD` instead.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
@@ -706,7 +706,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     {
                         var dummyTask = innerContext.CreateSubOrchestrationInstance<object>(
                           fireAndForgetMessage.FunctionName,
-                          DurableOrchestrationContext.DefaultVersion,
+                          innerContext.Version,
                           fireAndForgetMessage.InstanceId,
                           fireAndForgetMessage.Input,
                           new Dictionary<string, string>() { { OrchestrationTags.FireAndForget, "" } });
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
@@ -706,7 +706,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     {
                         var dummyTask = innerContext.CreateSubOrchestrationInstance<object>(
                           fireAndForgetMessage.FunctionName,
-                          innerContext.Version,
+                          this.Config.Options.DefaultVersion,
                           fireAndForgetMessage.InstanceId,
                           fireAndForgetMessage.Input,
                           new Dictionary<string, string>() { { OrchestrationTags.FireAndForget, "" } });
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -559,6 +559,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 }
             }
 
+            // Propagate the default version to orchestrators.
+            // TODO: Decide whether we want to propagate the default version to actitities and entities as well.
+            string version = (functionType == FunctionType.Orchestrator)
+                             ? this.Config.Options.DefaultVersion
+                             : string.Empty;
+
             this.Config.ThrowIfFunctionDoesNotExist(functionName, functionType);
 
             Task<TResult> callTask = null;
@@ -572,22 +578,19 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     System.Diagnostics.Debug.Assert(instanceId == null, "The instanceId parameter should not be used for activity functions.");
                     System.Diagnostics.Debug.Assert(operation == null, "The operation parameter should not be used for activity functions.");
                     System.Diagnostics.Debug.Assert(!oneWay, "The oneWay parameter should not be used for activity functions.");
+                    if (retryOptions == null)
                     {
-                        string version = string.Empty;
-                        if (retryOptions == null)
-                        {
-                            this.IncrementActionsOrThrowException();
-                            callTask = this.InnerContext.ScheduleTask<TResult>(functionName, version, input);
-                        }
-                        else
-                        {
-                            this.IncrementActionsOrThrowException();
-                            callTask = this.InnerContext.ScheduleWithRetry<TResult>(
-                                functionName,
-                                version,
-                                retryOptions.GetRetryOptions(),
-                                input);
-                        }
+                        this.IncrementActionsOrThrowException();
+                        callTask = this.InnerContext.ScheduleTask<TResult>(functionName, version, input);
+                    }
+                    else
+                    {
+                        this.IncrementActionsOrThrowException();
+                        callTask = this.InnerContext.ScheduleWithRetry<TResult>(
+                            functionName,
+                            version,
+                            retryOptions.GetRetryOptions(),
+                            input);
                     }
 
                     break;
@@ -608,46 +611,43 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                         throw new ArgumentException(nameof(instanceId), "Orchestration instance ids must not start with @");
                     }
 
+                    if (oneWay)
                     {
-                        string version = this.Config.Options.DefaultVersion;
-                        if (oneWay)
+                        this.IncrementActionsOrThrowException();
+                        var dummyTask = this.InnerContext.CreateSubOrchestrationInstance<TResult>(
+                                functionName,
+                                version,
+                                instanceId,
+                                input,
+                                new Dictionary<string, string>() { { OrchestrationTags.FireAndForget, "" } });
+
+                        System.Diagnostics.Debug.Assert(dummyTask.IsCompleted, "task should be fire-and-forget");
+                    }
+                    else
+                    {
+                        if (this.ContextLocks != null)
+                        {
+                            throw new LockingRulesViolationException("While holding locks, cannot call suborchestrators.");
+                        }
+
+                        if (retryOptions == null)
                         {
                             this.IncrementActionsOrThrowException();
-                            var dummyTask = this.InnerContext.CreateSubOrchestrationInstance<TResult>(
-                                    functionName,
-                                    version,
-                                    instanceId,
-                                    input,
-                                    new Dictionary<string, string>() { { OrchestrationTags.FireAndForget, "" } });
-
-                            System.Diagnostics.Debug.Assert(dummyTask.IsCompleted, "task should be fire-and-forget");
+                            callTask = this.InnerContext.CreateSubOrchestrationInstance<TResult>(
+                                functionName,
+                                version,
+                                instanceId,
+                                input);
                         }
                         else
                         {
-                            if (this.ContextLocks != null)
-                            {
-                                throw new LockingRulesViolationException("While holding locks, cannot call suborchestrators.");
-                            }
-
-                            if (retryOptions == null)
-                            {
-                                this.IncrementActionsOrThrowException();
-                                callTask = this.InnerContext.CreateSubOrchestrationInstance<TResult>(
-                                    functionName,
-                                    version,
-                                    instanceId,
-                                    input);
-                            }
-                            else
-                            {
-                                this.IncrementActionsOrThrowException();
-                                callTask = this.InnerContext.CreateSubOrchestrationInstanceWithRetry<TResult>(
-                                    functionName,
-                                    version,
-                                    instanceId,
-                                    retryOptions.GetRetryOptions(),
-                                    input);
-                            }
+                            this.IncrementActionsOrThrowException();
+                            callTask = this.InnerContext.CreateSubOrchestrationInstanceWithRetry<TResult>(
+                                functionName,
+                                version,
+                                instanceId,
+                                retryOptions.GetRetryOptions(),
+                                input);
                         }
                     }
 
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -140,7 +140,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private MessageSorter MessageSorter => this.messageSorter ?? (this.messageSorter = new MessageSorter());
 
-        public string Version => "";
+        public string Version => this.InnerContext.Version;
 
         /// <summary>
         /// Returns the orchestrator function input as a raw JSON string value.
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -27,8 +27,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         DurableOrchestrationContextBase // for v1 legacy compatibility.
 #pragma warning restore 618
     {
-        public const string DefaultVersion = "";
-
         private readonly Dictionary<string, IEventTaskCompletionSource> pendingExternalEvents =
             new Dictionary<string, IEventTaskCompletionSource>(StringComparer.OrdinalIgnoreCase);
 
@@ -561,8 +559,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 }
             }
 
-            // TODO: Support for versioning
-            string version = DefaultVersion;
+            string version = ((IDurableOrchestrationContext)this).Version;
             this.Config.ThrowIfFunctionDoesNotExist(functionName, functionType);
 
             Task<TResult> callTask = null;
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
@@ -559,7 +559,6 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 }
             }
 
-            string version = ((IDurableOrchestrationContext)this).Version;
             this.Config.ThrowIfFunctionDoesNotExist(functionName, functionType);
 
             Task<TResult> callTask = null;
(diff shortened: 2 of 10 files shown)


ltouro:
It this any different than having a 'version' property on the workflow input? 

I'm trying to figure how far this feature will go to help avoid any version issues.

AnatoliB:
@ltouro Right now, the only advantage is that there will be no need to pass a version explicitly (which can be cumbersome if you have many orchestrators and sub-orchestrators). Short-term, consider this a quality-of-life improvement.

However, this is just the first step. The next step will be for the Functions platform to make sure orchestrations are routed to the worker versions that are capable of handling these orchestrations. As a result, we will be able to run old and new workers side-by-side, which can be used for rolling upgrades and true zero-downtime deployment. But, in order to do that, the platform should be version-aware, so the version cannot be just your own input parameter.

test/Common/TestOrchestrations.cs
```diff
@@ -1549,5 +1549,32 @@ public static async Task<string> ReplaySafeLogger_OneLogMessage([OrchestrationTr
...
+        }
+
+        public static async Task<string> GetOrchestrationVersion_SubOrchestrator([OrchestrationTrigger] IDurableOrchestrationContext ctx, ILogger log)
+        {
+            return await Task.FromResult(VersionToString(ctx.Version));
+        }
+
+        private static string VersionToString(string version)
+        {
+            return version == null ? "null" : $"'{version}'";
```
andystaples: Does this evaluate to 'version' or version? Just checking

andystaples (commented):
One tiny code question left below. I just want to make sure I understand this correctly - am I correct that in this implementation, the orchestration version is not stored in the storage backend - meaning every time the orchestrator is loaded for replay, the DurableOrchestrationContext will be re-created using the version set in host.json? If so, does this create possible non-deterministic exceptions in the sample code provided in the PR description, where the first two replays executing orchestrations A and B happens on version 1.0, but then the app code is upgraded in-place before the replay proceeding orchestration C and the replay path branches to D?
Not that this is a new issue, that would be the behavior today if the code change simply replaced B for D, just making sure I understand the behavior well.

AnatoliB:
> One tiny code question left below. I just want to make sure I understand this correctly - am I correct that in this implementation, the orchestration version is not stored in the storage backend - meaning every time the orchestrator is loaded for replay, the DurableOrchestrationContext will be re-created using the version set in host.json? If so, does this create possible non-deterministic exceptions in the sample code provided in the PR description, where the first two replays executing orchestrations A and B happens on version 1.0, but then the app code is upgraded in-place before the replay proceeding orchestration C and the replay path branches to D? Not that this is a new issue, that would be the behavior today if the code change simply replaced B for D, just making sure I understand the behavior well.

@andystaples Thank you for this question, it is important. The orchestration version _is_ in fact stored on the backend side. For example, the Azure Storage backend stores it in the Version column of the instance table. During replay, the `context.Version` property is populated from the storage, not from host.json, and this is why the example works as intended (ABC for old orchestrations, ADC for new orchestrations), without non-determinism issues.

This persistence is already implemented on the DTFx side and at least the Azure Storage backend, my PR is not adding that, and this is why you don't see any new code related to that. DTFx users already can take advantage of this property. The only thing DF users are currently missing is a way to specify a version on orchestration creation, and this is what my PR is adding.

test/Common/TestOrchestrations.cs
AnatoliB: If `version` is not `null`, this code returns the version string surrounded by single quotes. I just wanted to be very explicit in the tests and clearly distinguish between `null`, an empty string, and a non-empty string that accidentally happens to contain `"null"`.

Technically, all of these work and lead to different results. I was thinking about introducing some validation there (perhaps prohibit either nulls, or empty strings, or both), but decided against that for now. `null` can be caused by `appVersion` not specified in `host.json` and an empty string can be caused by an env variable reference that evaluated to an empty string because the variable is missing. Both are potentially valid use cases, so I want to make sure they lead to reasonable behavior.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
```diff
@@ -706,7 +706,7 @@ internal void SendOutbox(OrchestrationContext innerContext, bool writeBackSucces
                     {
                         var dummyTask = innerContext.CreateSubOrchestrationInstance<object>(
                           fireAndForgetMessage.FunctionName,
-                          DurableOrchestrationContext.DefaultVersion,
+                          innerContext.Version,
```
nytian: Just to confirm: the sub-orchestrations scheduled by entity functions will inherit the current app version as well by this, correct?

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
```diff
@@ -19,6 +19,11 @@ public interface IDurableOrchestrationContext
         /// </summary>
         string Name { get; }
 
+        /// <summary>
+        /// Gets the version of the current orchestration function.
+        /// </summary>
+        string Version => string.Empty;
```
nytian, line 25: I wonder why this always return string.Empty?  According to the description for `AppVersion` on `DurableTaskOptions`, I thought this should be the same value as AppVersion, is there anything that I am missing?

test/Common/OrchestrationVersionTests.cs
```diff
@@ -0,0 +1,88 @@
...
+                Assert.Equal(OrchestrationRuntimeStatus.Completed, status.RuntimeStatus);
+                var expectedOutput = $"Orchestration: {expectedContextVersion}; Sub-orchestration: {expectedContextVersion}";
+                Assert.Equal(expectedOutput, status.Output.ToString());
+                await host.StopAsync();
+            }
+        }
+
+        [Fact]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        public async Task OrchestrationVersionIsImmutable()
```
nytian: This test seems to verify that in-flight orchestratiors keep their version even after the durable functions app version changes. Can we rename it to somethign like `ExistingOrchestrationVersionIsImmutable` for better clarity?

test/Common/OrchestrationVersionTests.cs
```diff
@@ -0,0 +1,88 @@
...
+            await host1.StopAsync();
+
+            // Resume the same orchestration on a host with appVersion set to 2.0.
+            using ITestHost host2 = GetJobHost(taskHubName, appVersion: "2.0");
+            await host2.StartAsync();
+            await client.RaiseEventAsync("Resume", this.output);
+            status = await client.WaitForCompletionAsync(this.output, timeout: TimeSpan.FromMinutes(1));
+            Assert.Equal(OrchestrationRuntimeStatus.Completed, status.RuntimeStatus);
+            await host2.StopAsync();
+
```
nytian, line 72: Can we also add an assert here checking the host or the functions app is already in version 2.0?

test/Common/OrchestrationVersionTests.cs
```diff
@@ -0,0 +1,88 @@
...
+        public OrchestrationVersionTests(ITestOutputHelper output)
+        {
+            this.output = output;
+            this.loggerProvider = new TestLoggerProvider(output);
+        }
+
+        [Theory]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        [InlineData(null, "null")]
```
nytian: Sorry I am a little bit confused here, I saw your reply below, but what will be the actual version value when cx define null, "", "null" in real cases?

nytian (commented):
Thanks for the PR! Left a few comments and questions. :)

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -23,6 +23,13 @@ public class DurableTaskOptions
         private string resolvedHubName;
         private string defaultHubName;
 
+        /// <summary>
+        /// The AppVersion value provided here will be automatically assigned to every orchestration
+        /// instance created by this app, and this value will be available on IDurableOrchestrationContext
+        /// passed to each orchestrator function replay invocation of this orchestration instance.
+        /// </summary>
+        public string AppVersion { get; set; }
```
cgillum: Should we instead name this `DefaultVersion`? This is what we named it in the [portable SDK changes](https://github.com/microsoft/durabletask-dotnet/blob/e32fa7a4ff9e569120957299508a58e88c115398/src/Client/Core/DurableTaskClientOptions.cs#L22) and it would be ideal if we were consistent between the two app models for the sake of documentation and clarity. I decided I liked this name because calling it a "default" allows us to introduce per-orchestration version *overrides* in the future without creating semantic confusion/ambiguity.

A disadvantage of `AppVersion` is that it might lead developers to believe that app and orchestration versions are two completely different things, when in reality we're actually creating a strong coupling between the version configured in host.json with the orchestration versions. I recognize that I originally suggested "appVersion" as the name, but as I thought more about the implementation details, I became a bit worried about it.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -558,8 +559,7 @@ internal async Task<TResult> CallDurableTaskFunctionAsync<TResult>(
                 }
             }
 
-            // TODO: Support for versioning
-            string version = DefaultVersion;
+            string version = ((IDurableOrchestrationContext)this).Version;
```
cgillum: The `version` here is passed to both orchestrations and activities. I don't think we've really thought through how/whether to version activities. Should we isolate the version assignment to just orchestrations to be safe?

src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationContext.cs
cgillum: This is the default implementation for the interface, but the actual class that implements this interface (see DurableOrchestrationContext.cs) will return the value from `DurableTaskOptions`.

cgillum (commented):
I think we also need to make a change to LocalGrpcListener.cs [here](https://github.com/Azure/azure-functions-durable-extension/blob/a6e44d966f1df4652dd202694834fd6c773a1aae/src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs#L173) to ensure that .NET Isolated and Java apps also inherit the default (app) version that's specified in host.json. Otherwise, I think it will work in JavaScript and Python but not in Java or .NET Isolated.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
AnatoliB: Yes, I've been thinking about this as well, I will rename to `DefaultVersion`.

test/Common/OrchestrationVersionTests.cs
AnatoliB: Done.

test/Common/OrchestrationVersionTests.cs
AnatoliB: @nytian Sorry, I don't quite understand this. The host version here is a test configuration, not a result of the code that we are testing here. The `host2` object is configured with version "2.0" during initialization, I don't think there is a need to assert that.

If what you had in mind is validating that the version specified in the host configuration also shows up in the context passed to the orchestration code, this is what the `OrchestrationVersionIsDeterminedByHostDefaultVersion` test is validating. But please let me know if you meant something else.

test/Common/OrchestrationVersionTests.cs
AnatoliB: In the orchestrator function code, `ctx.Version` will contain exactly what the `DefaultVersion` in `host.json` contains (or `null` if it is not specified). In this test, I want to confirm that this is indeed the case, so that, for example, nulls are not silently replaced by empty strings or the other way around. I want to do it by looking at the string returned from the `GetOrchestrationVersion_AfterExternalEvent` orchestrator, so I need a way to reliably distinguish between `null`, an empty string, a string containing `"null"`, etc.

But I can see that it's causing more confusion than I anticipated, so I'll do something to make it clearer.

AnatoliB: Ok, I'm serializing this value to JSON now, I hope it clarifies things.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
AnatoliB: @nytian This is a great question. Are you actually asking about new orchestrations started with `IDurableEntityContext.StartNewOrchestration` (as I don't see an API to schedule a _sub-orchestration_ from an entity)? Something like this: orchestration 1 calls an entity, and this entity starts orchestration 2.

No version is currently propagated. To me, two behaviors can make sense here:
1. Propagate the version from orchestration 1 to orchestration 2. This is reasonable because orchestration 2 is conceptually a part of the same workflow.
2. Make orchestration 2 use `DefaultVersion` instead. This is also reasonable because this is technically a new orchestration.

I like (2) a little bit more, but I can easily imagine meaningful use cases for both. @cgillum, do you have any opinion?

In any case, this deserves a test, and my test currently shows that neither (1) nor (2) is currently happening, so something is missing.

AnatoliB:
> I think we also need to make a change to LocalGrpcListener.cs [here](https://github.com/Azure/azure-functions-durable-extension/blob/a6e44d966f1df4652dd202694834fd6c773a1aae/src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs#L173) to ensure that .NET Isolated and Java apps also inherit the default (app) version that's specified in host.json. Otherwise, I think it will work in JavaScript and Python but not in Java or .NET Isolated.

@cgillum Thank you for the pointer. This PR is focused on .NET in-proc though. I can still do it here if you insist, but I think it would be cleaner to address this in a separate PR, as .NET Isolated and non-.NET languages may require even more changes and tests.

cgillum:
> > I think we also need to make a change to LocalGrpcListener.cs [here](https://github.com/Azure/azure-functions-durable-extension/blob/a6e44d966f1df4652dd202694834fd6c773a1aae/src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs#L173) to ensure that .NET Isolated and Java apps also inherit the default (app) version that's specified in host.json. Otherwise, I think it will work in JavaScript and Python but not in Java or .NET Isolated.
> 
> @cgillum Thank you for the pointer. This PR is focused on .NET in-proc though. I can still do it here if you insist, but I think it would be cleaner to address this in a separate PR, as .NET Isolated and non-.NET languages may require even more changes and tests.

I won't insist, but if this change already affects more than just in-proc, wouldn't it make sense to go ahead and add support for all out-of-proc languages instead of just a subset of them?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
cgillum: I don't recall if we ever had a conversation about how sub-orchestration versioning should work, so I feel a bit unprepared to answer this. :) My intuition is that any new orchestration should always use the default (app) version, which I assume is a variant of (2). That's an easy policy to understand, it ensures that the latest code is being run when possible and is not at risk for non-determinism problems. Option (1) worries me because of the extra complexity it introduces (i.e., the implicit coupling between the version of the client and the version of the orchestration).

AnatoliB: @cgillum What do you think about regular sub-orchestrations launched by orchestrations directly with `context.CallSubOrchestratorAsync`, _without entities involved_? I don't think we had a conversation about this case, but (1) is my intuitive expectation, as I assume potential tight coupling between sub-orchestrations and parent orchestrations. This is what the current PR does.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -23,6 +23,13 @@ public class DurableTaskOptions
         private string resolvedHubName;
         private string defaultHubName;
 
+        /// <summary>
+        /// The DefaultVersion value provided here will be automatically assigned to every orchestration
+        /// instance created by this app, and this value will be available on IDurableOrchestrationContext
+        /// passed to each orchestrator function replay invocation of this orchestration instance.
+        /// </summary>
+        public string DefaultVersion { get; set; }
```
halspang, line 31: In the portable SDK, we have versioning broken down into two different options, the client and the worker. 

Do we want to have this in a `Client` subsection to match that? It seems like in durable functions, all the options share the same options file.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
AnatoliB: I've just realized that the distinction between "sub-orchestrations" and "new orchestrations" that I brought up above is irrelevant from the implementation perspective, because starting an orchestration from an entity leads to an `OrchestrationContext.CreateSubOrchestrationInstance` call anyway. So, let's decide what behavior we want from the user perspective, and perhaps apply the same policy to both cases.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
AnatoliB: Thank you @halspang. I'm not yet clear on how Functions users would benefit from this separation. @cgillum, do you have any opinion?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableEntityContext.cs
AnatoliB: After discussing and thinking, decided to stop propagating the parent orchestration's version to sub-orchestrations, whether to invoked directly or via an entity. All orchestrations and sub-orchestrations will be associated with the currently configured `DefaultVersion`.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
AnatoliB: Agreed, not passing the version to activities anymore.

AnatoliB:
> I won't insist, but if this change already affects more than just in-proc, wouldn't it make sense to go ahead and add support for all out-of-proc languages instead of just a subset of them?

Discussed offline, decided not to do the out-of-proc part yet.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
cgillum: We don't have a client/worker split in Durable Functions, so I think it makes sense to leave this as-is.

cgillum (approved):

PR #3076, opened 2025-03-28
Distributed Tracing for Entities (Isolated)

sophiatev:
This PR adds support for distributed tracing for entities in the .NET isolated framework. The changes mostly focus on correctly propagating information between the durabletask and durabletask-dotnet repos. In order to do so, the proto definition of objects related to signaling entities or entities starting orchestrations had to be populated with extra information (like the parent trace context of the request, or the time of the request, etc. For more details see this [PR](https://github.com/microsoft/durabletask-protobuf/pull/40) in the protobuf repo). The changes in the `ProtobufUtils` class focus on adding this extra information. The changes in `LocalGrpcListener` are to:
- Create a trace in the case of a client signaling an entity
- Add the parent trace context created for a client starting an orchestration to the corresponding `ExecutionStartedEvent`, because I noticed that traces were not correctly being created in this case otherwise (this change is unrelated to tracing for entities). 

The various other PRs related to this effort are
- [durabletask-dotnet](https://github.com/microsoft/durabletask-dotnet/pull/404)
- [protobuf](https://github.com/microsoft/durabletask-protobuf/pull/40)
- [durabletask](https://github.com/Azure/durabletask/pull/1198)

### Pull request checklist


src/WebJobs.Extensions.DurableTask/Correlation/TraceHelper.cs
```diff
@@ -18,13 +18,14 @@ internal class TraceHelper
 
         private static readonly ActivitySource ActivityTraceSource = new ActivitySource(Source);
 
-        internal static Activity? StartActivityForNewOrchestration(ExecutionStartedEvent startEvent, ActivityContext parentTraceContext)
+        internal static Activity? StartActivityForNewOrchestration(ExecutionStartedEvent startEvent, ActivityContext parentTraceContext, DateTimeOffset? startTime = null)
```
bachuv: Can we assign `startTime` to `default` here instead of `null`?

src/WebJobs.Extensions.DurableTask/Correlation/TraceHelper.cs
```diff
@@ -60,7 +61,7 @@ internal class TraceHelper
                 Schema.SpanNames.CallOrSignalEntity(entityName, operationName),
                 kind: signalEntity ? ActivityKind.Producer : ActivityKind.Client,
                 parentContext: parentTraceContext.Value,
-                startTime: startTime);
+                startTime: startTime ?? default);
```
bachuv: Can we remove this change since we added the default value in the `startTime` optional parameter?

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -755,7 +755,7 @@ private async Task ExecuteOutOfProcBatch()
                         request.IsSignal,
                         request.ScheduledTime,
                         parentTraceContext,
-                        request.RequestTime.Value);
+                        request.RequestTime);
```
bachuv, line 758: Why do we need this change?

src/WebJobs.Extensions.DurableTask/Correlation/TraceHelper.cs
sophiatev: Unfortunately no, the issue is that I had to make the startTime parameter a `DateTimeOffset?` rather than a `DateTimeOffset` even with the default value because of how this method is called in the [LocalGrpcListener](https://github.com/Azure/azure-functions-durable-extension/blob/ea9f27b4e3d5974eee9eb010015f3b7e57e6ff43/src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs#L189) (the issue being that we could get null when trying to extract the request time from the `CreateInstanceRequest`) and so that's why I had to add this change to the `TraceHelper` method

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
sophiatev: This is due to having to make the `startTime` parameter in the TraceHelper method signature a `DateTimeOffset?` rather than `DateTimeOffset` as discussed in our convo [above](https://github.com/Azure/azure-functions-durable-extension/pull/3076#discussion_r2092073142), and so we no longer have to extract the `Value` here and can just directly pass in `request.RequestTime` (which has type `DateTimeOffset?`)

bachuv (approved):

bachuv (dismissed):

bachuv (approved):

PR #3083, opened 2025-04-09
Support X-Forwarded-Host and X-Forwarded-Proto headers in Durable Functions HTTP management APIs

### LINKED ISSUE #1026: Support Custom Domains for Status/Admin endpoints
I'd like to be able to supply a custom domain to be used when creating the CheckStatusResponse message. Currently it appears to use the host from the incoming HttpRequestMessage's RequestUri.

The Location header for the 202 Accepted is then populated with the host name of the actual app service rather than the custom domain we have assigned to our Application Gateway.

Allow an Environment Variable to be supplied that will override the host name used for the status endpoint and other management endpoints. 

I've tried intercepting the incoming request in code and changing the host to be what I want. This works fine for the first response, but subsequent calls to the status endpoint will return the app service host name as I can't intercept and override the Request URI in there.

**Additional context**
I'm calling a Durable Function via an Application Gateway. The App Gateway uses Path Based routing so the URLs look like:
- https://mygateway.com/api >> backend API service
- https://mygateway.com/functions >> backend Function service

rukasakurai:
### Context

This PR addresses #1026, which requests proper support for Durable Functions being hosted behind a reverse proxy such as **Azure Front Door** or **Application Gateway**.

By default, Durable Functions generates management URLs (e.g., `statusQueryGetUri`) using the values from the incoming HTTP request. As a result, when Durable Functions is deployed behind a reverse proxy, the request received by Durable Functions often reflects the **internal routing address** (e.g., `myapp.azurewebsites.net`) rather than the **client-facing address** (e.g., `mycustomdomain.com`). This results in management URLs that are not usable by the client.

This PR adds support for honoring `X-Forwarded-Host` and `X-Forwarded-Proto` headers, allowing Durable Functions to return URLs that reflect what the client actually used when making the request.

Without this fix, developers must implement workarounds like [this one](https://github.com/rukasakurai/durable_test_app/pull/7/files) to rewrite management URLs manually.

---

Resolves #1026

---

### Description

This PR introduces support for the `X-Forwarded-Host` and `X-Forwarded-Proto` headers in Durable Functions' HTTP response URLs (e.g., `statusQueryGetUri`, `terminatePostUri`).

- A new config setting, `HttpOptions.HttpSettings.UseForwardedHost`, enables this feature.
- When enabled, the management URIs use values from `X-Forwarded-Host` and `X-Forwarded-Proto` instead of the default `Host` and scheme.
- Unit tests have been added to verify behavior under different header scenarios.

---

### Pull request checklist

### DIFF
--- a/test/Common/HttpApiHandlerTests.cs
+++ b/test/Common/HttpApiHandlerTests.cs
@@ -1592,12 +1592,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
             Assert.Equal(HttpStatusCode.Accepted, actualResponse.StatusCode);
         }
 
-        [Theory]
-        [InlineData("example.com", "https")]
+        [Fact]
         [Trait("Category", PlatformSpecificHelpers.TestCategory)]
-        public void GetClientResponseLinks_Uses_Forwarded_Headers_When_Enabled(
-            string forwardedHost,
-            string forwardedProto)
+        public async void GetClientResponseLinks_Uses_Forwarded_Headers_When_Enabled()
         {
             // Arrange
             var options = new DurableTaskOptions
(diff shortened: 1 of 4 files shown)


test/Common/HttpApiHandlerTests.cs
```diff
@@ -1592,6 +1592,98 @@ public async Task SignalEntity_Is_Success(bool hasKey, bool hasOp, bool hasConte
             Assert.Equal(HttpStatusCode.Accepted, actualResponse.StatusCode);
         }
 
+        [Theory]
+        [InlineData("example.com", null, "example.com", null)]
+        [InlineData("example.com", "https", "example.com", "https")]
+        [InlineData("custom.domain.com", null, "custom.domain.com", null)]
+        [InlineData("custom.domain.com", "https", "custom.domain.com", "https")]
```
cgillum: Non-blocking, but since `forwardedHost` and `expectedHost` are always the same (same with `forwardedProto` and `expectedProto`) you don't actually need to define both of them. Just use the same value for both the forwarded and expected values.

cgillum (approved):
Thanks for this PR!

cgillum:
@rukasakurai looks like there is a build failure in the test code you added. Can you take a look?

> HttpApiHandlerTests.cs(1634,17): error CS1503: Argument 4: cannot convert from 'System.Net.Http.HttpRequestMessage' to 'bool'
> HttpApiHandlerTests.cs(1676,17): error CS1503: Argument 4: cannot convert from 'System.Net.Http.HttpRequestMessage' to 'bool'

rukasakurai:
@cgillum ,
Two tests failing right now that I'm working on fixing.
One is the below. I don't believe the updated code touches passwords, so I'm wondering if this is due to the code change in this PR or something else.
```
ERROR: Unable to set system administrator password: Password validation failed. The password does not meet SQL Server password policy requirements because it is too short. The password must be at least 8 characters..
```
[Smoke Test - .NET in-proc w/ MSSQL on Functions V4 / build (pull_request)](https://github.com/Azure/azure-functions-durable-extension/actions/runs/14403735022/job/40395085041?pr=3083)

cgillum:
@rukasakurai you can ignore the MSSQL test failure. There's something wrong with that pipeline that needs to be fixed separate from your PR.

We just need to fix your unit tests, which run as part of the "Validate Build (except E2E tests)" pipeline.

cgillum (approved):

PR #3084, opened 2025-04-11
Fix scaling when using a non-default Azure Storage account

AnatoliB:
When using a non-default Azure Storage account (specified at `extensions/durableTask/storageProvider/connectionName` in `host.json`), Target-Based Scaling on Consumption and Elastic Premium plans does not work as expected because the Scale Controller tries to use the default Function App connection instead. As a result:

- Durable triggers are not considered when making scaling decisions.
- The default storage account may experience a high volume of transactions trying to find the non-existent task hub, which increases storage cost.

It turns out that, _when running in the Scale Controller context_, the `azureStorageDurabilityProviderFactory.DefaultConnectionName` property (that the `DurableTaskTriggersScaleProvider` constructor relies on to determine the connection name) does not contain the connection name configured in `host.json`, and instead contains the very generic `"Storage"` value. This PR fixes the problem by making the `DurableTaskTriggersScaleProvider` constructor look for this configuration in the `triggerMetadata` parameter instead.

Alternatively, we may want to figure out why `azureStorageDurabilityProviderFactory.DefaultConnectionName` does not contain the expected value to begin with. I don't have an answer or a working fix for that yet, though. The fix proposed in this PR is tested.

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Scale/DurableTaskTriggersScaleProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/Scale/DurableTaskTriggersScaleProvider.cs
@@ -5,7 +5,6 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
-using Grpc.Core;
 using Microsoft.Azure.WebJobs.Host.Scale;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
--- a/src/WebJobs.Extensions.DurableTask/Scale/DurableTaskTriggersScaleProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/Scale/DurableTaskTriggersScaleProvider.cs
@@ -42,9 +42,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Scale
             IDurabilityProviderFactory durabilityProviderFactory = this.GetDurabilityProviderFactory();
             DurabilityProvider defaultDurabilityProvider = durabilityProviderFactory.GetDurabilityProvider();
 
-            string? connectionName = durabilityProviderFactory is AzureStorageDurabilityProviderFactory azureStorageDurabilityProviderFactory
-                ? azureStorageDurabilityProviderFactory.DefaultConnectionName
-                : null;
+            string? connectionName = GetConnectionName(durabilityProviderFactory, this.options);
 
             this.targetScaler = ScaleUtils.GetTargetScaler(
                 defaultDurabilityProvider,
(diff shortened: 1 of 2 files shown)


src/WebJobs.Extensions.DurableTask/Scale/DurableTaskTriggersScaleProvider.cs
```diff
@@ -61,6 +65,26 @@ public DurableTaskTriggersScaleProvider(
                 this.options.HubName);
         }
 
+        private static string? GetConnectionName(IDurabilityProviderFactory durabilityProviderFactory, DurableTaskOptions options)
+        {
+            if (durabilityProviderFactory is AzureStorageDurabilityProviderFactory azureStorageDurabilityProviderFactory)
+            {
+                // First, look for the connection name in the options initialized from the trigger metadata
```
cgillum: The comment is talking about trigger metadata, but I don't see anything trigger specific here. Does the comment need to be updated?

src/WebJobs.Extensions.DurableTask/Scale/DurableTaskTriggersScaleProvider.cs
```diff
@@ -61,6 +65,26 @@ public DurableTaskTriggersScaleProvider(
...
+                var azureStorageOptions = new AzureStorageOptions();
+                if (options != null && options.StorageProvider != null)
+                {
+                    JsonConvert.PopulateObject(JsonConvert.SerializeObject(options.StorageProvider), azureStorageOptions);
+                }
+
+                // If the connection name is not found in the options, use the default connection name from the factory
+                return azureStorageOptions.ConnectionName ?? azureStorageDurabilityProviderFactory.DefaultConnectionName;
```
cgillum, line 81: I'm not understanding the behavior change. Doesn't the `AzureStorageDurabilityProviderFactory` constructor already do this exact logic and provide the correct storage account connection name in the `DefaultConnectionName` property?

src/WebJobs.Extensions.DurableTask/Scale/DurableTaskTriggersScaleProvider.cs
AnatoliB: @cgillum I'll find a better way to explain this. The `options` parameter contains data (`StorageProvider`) populated from `triggerMetadata` passed to the `DurableTaskTriggersScaleProvider` constructor (line 40).

src/WebJobs.Extensions.DurableTask/Scale/DurableTaskTriggersScaleProvider.cs
AnatoliB: @cgillum The `AzureStorageDurabilityProviderFactory` constructor does have the same logic, but it receives a different input, so it produces a different result. When this constructor is invoked _in the Scale Controller context_, the `options` parameter contains an empty `StorageProvider` collection. I don't yet understand why, and, more importantly, whether this is intentional or not. I suspect this has something to do with how DI instantiates these objects, and this may require a fix in the Scale Controller.

However, when the `DurableTaskTriggersScaleProvider` constructor is invoked, the `triggerMetadata` parameter does contain the correct `StorageProvider` collection, and this is why my fix works.

cgillum (commented):
Have we validated whether runtime-driven scaling using an alternate connection string continues to work with this fix?

src/WebJobs.Extensions.DurableTask/Scale/DurableTaskTriggersScaleProvider.cs
AnatoliB: I've updated the comments, I hope it's clear now

AnatoliB:
> Have we validated whether runtime-driven scaling using an alternate connection string continues to work with this fix?

Yes, it still works.

cgillum (approved):
I get it now. The key thing is that the correct configuration comes from `TriggerMetadata` when run in a scale controller context. It seems the connection name is the one thing we weren't getting from that context, which is why we're seeing this issue.

I think this change makes sense.

PR #3092, opened 2025-04-16
Support 202 Automatic Polling at Worker.Extensions.DurableTask

### LINKED ISSUE #2781: TaskOrchestrationContext.CallHttpAsync` automated polling of "202 Accepted" responses.
### Description
TaskOrchestrationContext.CallHttpAsync` extension method method in dotnet-isolated Durable Functions does not appear to support automatic polling of 202 responses.

I am using this functionality to call a Http triggered function on a different Function app that starts its own durable orchestration and uses the `IDurableClient.CreateCheckStatusResponse` method to return the status monitoring endpoints.

I've also opened discussion: https://github.com/Azure/azure-functions-durable-extension/discussions/2775 but as it didn't get a lot of traction, figured it was worth raising here. 

### Expected behavior
As I understand the documentation [here](https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-http-features?tabs=csharp#consuming-http-apis) this extension method should automatically poll the uri in the location header upon recieving a 202 http response, meaning there is no need to explicitly consume and poll the `statusQueryGetUri` endpoint.

Noting the example appears to be for an in-process work, I'm unsure if this is even fully supported by isolated workers. If there is a different pattern I should be using, please point it out.

### Actual behavior
The 'durableResponse' after awaiting is a 202 status code with the standard `CreateCheckStatusResponse` payload as its body,

```csharp
var request = new DurableHttpRequest(HttpMethod.Post, new Uri({{url-to-azure-function}}))
{ 
      Content = JsonSerializer.Serialize({{payload-as-object}}),
      AsynchronousPatternEnabled = true,
      Headers = new Dictionary<string, Microsoft.Extensions.Primitives.StringValues> { { "Content-type", "application/json" } }
};

// Send off a call to the durable function.
var durableResponse = await context.CallHttpAsync(request);

// durableResponse.StatusCode -> 202
```

### Known workarounds
None known 

### App Details

- **Durable Functions extension version (e.g. v1.8.3)**: 1.15.1
- **Azure Functions runtime version (1.0 or 2.0)**: isolated worker 
- **Programming language used**: C# .net 8

### If deployed to Azure
Not applicable - running locally.

nytian:

resolves #2781
- Update `CallHttpAsync` at TaskOrchestrationContextExtensionMethods.cs to support automatic polling of 202 responses
- E2E test added 
- **This PR won't work unless changes in https://github.com/microsoft/durabletask-dotnet/pull/415 is released.**

### Pull request checklist


src/Worker.Extensions.DurableTask/TaskOrchestrationContextExtensionMethods.cs
```diff
@@ -20,14 +24,47 @@ public static class TaskOrchestrationContextExtensionMethods
...
+
+            DateTime fireAt = default(DateTime);
+
+            if (headersDictionary.TryGetValue("Retry-After", out StringValues retryAfter))
+            {
+                fireAt = context.CurrentUtcDateTime.AddSeconds(int.Parse(retryAfter));
+            }
+            else
+            {
+                // TODO: use the configuration DefaultAsyncRequestSleepTimeMilliseconds from durabletaskextension
```
nytian: The context here is from [TaskOrchestrationContext](https://github.com/microsoft/durabletask-dotnet/blob/main/src/Abstractions/TaskOrchestrationContext.cs), and it doesn't have property saving config from DurableExtension. We can do this but that might require quite a lot changes, including DurableTask.Abstractions. Thus I am not sure if we should do that, or if there is any other easy way to do that.

test/e2e/Tests/Tests/HTTPFeatureTests.cs
```diff
@@ -0,0 +1,39 @@
...
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Azure.Durable.Tests.DotnetIsolatedE2E;
+
+[Collection(Constants.FunctionAppCollectionName)]
+public class HTTPFeatureTests
+{
+    private readonly FunctionAppFixture _fixture;
+    private readonly ITestOutputHelper _output;
```
cgillum: Looks like this code is not following our style guidelines. For example:
```suggestion
    readonly ITestOutputHelper output;
```
Please update accordingly.

test/e2e/Tests/Tests/HTTPFeatureTests.cs
```diff
@@ -0,0 +1,39 @@
...
+
+    [Fact]
+    public async Task HTTPAutomaticPollingTests()
+    {
+        using HttpResponseMessage response = await HttpHelpers.InvokeHttpTrigger("HttpStart_HTTPPollingOrchestrator");
+
+        Assert.Equal(HttpStatusCode.Accepted, response.StatusCode);
+        string statusQueryGetUri = await DurableHelpers.ParseStatusQueryGetUriAsync(response);
+
+        await DurableHelpers.WaitForOrchestrationStateAsync(statusQueryGetUri, "Completed", 150);
```
cgillum, line 38: What is the `150` parameter and what are it's units? Please use a named parameter to make the code easier to understand.

test/e2e/Tests/Tests/HTTPFeatureTests.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        var orchestrationDetails = await DurableHelpers.GetRunningOrchestrationDetailsAsync(statusQueryGetUri);
+
+        // Verify that the output contains the LongRunningOrchestrator's result,
+        // ensuring the orchestrator completed and did not just return a 202 Accepted.
+        Assert.Contains("Hello Tokyo", orchestrationDetails.Output);
```
cgillum: We should also assert the the orchestration ran for the expected length of time.

Also, is there any way we can confirm how many times the implementation did polling, perhaps by examining logs?

src/Worker.Extensions.DurableTask/TaskOrchestrationContextExtensionMethods.cs
cgillum: Hmm. I'm not sure the right approach for this. One option would be to introduce a generic property bag (`Dictionary<string, object>`) in `TaskOrchestrationContext`, and have that be populated using data from the trigger context that we get from the WebJobs extension. For example, adding a a key named `df.http.defaultAsyncRequestSleepTimeMilliseconds` and have the value come from host.json. We still have to add the property bag to the DurableTask.Abstractions, but we only need to make that one change and we can easily add other properties to the dictionary in the future without changing DurableTask.Abstractions a second time. Let me know if I can help explain this idea better.

src/Worker.Extensions.DurableTask/TaskOrchestrationContextExtensionMethods.cs
```diff
@@ -20,14 +24,47 @@ public static class TaskOrchestrationContextExtensionMethods
...
 
-        return context.CallActivityAsync<DurableHttpResponse>(Constants.HttpTaskActivityReservedName, request);
+        DurableHttpResponse response = await context.CallActivityAsync<DurableHttpResponse>(Constants.HttpTaskActivityReservedName, request);
+        
+        while (response.StatusCode == HttpStatusCode.Accepted && request.AsynchronousPatternEnabled )
+        {
+            // If Headers is null or missing, we can't poll the Location URL, so return the response.
+            if (response.Headers is null)
+            {
+                break;
```
bachuv, line 48: Can we add a log here that will let us know that the headers were null or missing? It might help with debugging.

test/e2e/Tests/Tests/HTTPFeatureTests.cs
bachuv: Do we emit any logs about the activity named `HttpTaskActivityReservedName`?

src/Worker.Extensions.DurableTask/TaskOrchestrationContextExtensionMethods.cs
```diff
@@ -53,8 +53,10 @@ public static async Task<DurableHttpResponse> CallHttpAsync(this TaskOrchestrati
             }
             else
             {
-                // TODO: use the configuration DefaultAsyncRequestSleepTimeMilliseconds from durabletaskextension
-                fireAt = context.CurrentUtcDateTime.AddMilliseconds(30000);
+                // Gets configuration DefaultAsyncRequestSleepTimeMilliseconds from durabletaskextension
+                int defaultAsyncRequestSleepTimeMilliseconds = context.Properties.TryGetValue("df.http.defaultAsyncRequestSleepTimeMilliseconds", out var value) && value is double d
```
jviau: Can you extract this to a constant? Same with the other location of this. I get they are different assemblies with no common library to put it in - so two different constants are fine.

nuget.config
```diff
@@ -1,8 +1,9 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 
 <configuration>
   <packageSources>
-    <clear/>
+    <clear />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+    <add key="Package source" value="C:\Users\naiyuantian\localNuget" />
```
jviau: Make sure to remove this before checking in.

src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteOrchestratorContext.cs
```diff
@@ -39,6 +43,9 @@ public RemoteOrchestratorContext(OrchestrationRuntimeState runtimeState, TaskOrc
         [JsonProperty("upperSchemaVersion")]
         internal int UpperSchemaVersion { get; } = 4;
 
+        [JsonProperty("configurations")]
+        public Dictionary<string, object?> Configurations { get; private set; }
```
jviau: This doesn't have to be a dictionary - you could have a POCO here. It will serialize all the same.

test/e2e/Tests/Tests/HTTPFeatureTests.cs
nytian: I added comments for this -- that's be cause this orchestration will invoke another orchestration which takes 2 min to finish.

test/e2e/Tests/Tests/HTTPFeatureTests.cs
nytian: I added logs at `TaskOrchestrationContextExtensionMethods` where we implement the polling. And in the test, I just checked this logs exists to ensure we did polling.  But note that since I added the logger with OrchestrationContext, so customers will see the logs too and they can't disable that.

src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteOrchestratorContext.cs
nytian: If I switch this to a POCO, that means I’ll need to convert it into a dictionary first, since this is defined as a map in the Protobuf. I am worried that this is a bit tedious—just wondering if there’s a cleaner or more efficient way to handle this?

src/Worker.Extensions.DurableTask/TaskOrchestrationContextExtensionMethods.cs
nytian: Sorry I am a little bit lost, which one are you referring to?

test/e2e/Tests/Tests/HTTPFeatureTests.cs
nytian: The tricky part to assert how many times we polls 202 response is that it looks like the configured wait time is only used in the first one or two polling attempts. After that, a `Retry-After` header (usually 5 seconds) will be included in the response, so the polling interval will shorten. This makes the number of logs a bit inconsistent when testing—especially since the orchestration sometimes runs longer than the timer (e.g., I set 1 min, but it can exceed that). 

Also, does anyone know where the Retry-After header is set? I couldn't find the code for it.

src/Worker.Extensions.DurableTask/TaskOrchestrationContextExtensionMethods.cs
cgillum: I think Jacob's asking if the key `"df.http.defaultAsyncRequestSleepTimeMilliseconds"` can be stored in a constant.

src/WebJobs.Extensions.DurableTask/ProtobufUtils.cs
```diff
@@ -617,5 +618,41 @@ internal static P.PurgeInstancesResponse CreatePurgeInstancesResponse(PurgeResul
                     throw new NotSupportedException($"Deserialization of {operationResult.ResultTypeCase} is not supported.");
             }
         }
+
+        /// <summary>
+        /// Converts a <see cref="Dictionary{TKey, TValue}"/> of <c>&lt;string, object?&gt;</c> to a <see cref="MapField{String, Value}"/> for gRPC transmission.
+        /// </summary>
+        /// <param name="input">The dictionary to convert.</param>
+        /// <returns>The corresponding <see cref="MapField{String, Value}"/>, or <c>null</c> if input is <c>null</c>.</returns>
+        internal static MapField<string, Value> ConvertDictionaryToStructMap(Dictionary<string, object?> input)
```
cgillum: nit: consider renaming this method to match the return type.

```suggestion
        internal static MapField<string, Value> ConvertDictionaryToMapField(Dictionary<string, object?> input)
```

src/Worker.Extensions.DurableTask/TaskOrchestrationContextExtensionMethods.cs
```diff
@@ -14,20 +19,65 @@ namespace Microsoft.DurableTask;
...
     /// <param name="context">The task orchestration context.</param>
     /// <param name="request">The DurableHttpRequest used to make the HTTP call.</param>
     /// <returns>DurableHttpResponse</returns>
-    public static Task<DurableHttpResponse> CallHttpAsync(this TaskOrchestrationContext context, DurableHttpRequest request)
+    public static async Task<DurableHttpResponse> CallHttpAsync(this TaskOrchestrationContext context, DurableHttpRequest request)
     {
         if (context is null)
         {
             throw new ArgumentNullException(nameof(context));
         }
```
cgillum, line 37: It's important to ensure that all logger names have a prefix so that customers can easily configure the visibility across multiple related loggers. I looked at the .NET Isolated worker repo and found out that they have a convention for this (see [here](https://github.com/Azure/azure-functions-dotnet-worker/blob/ab45bacd09effcc40d046d1451451e74f50d1019/extensions/Worker.Extensions.Http.AspNetCore/src/Infrastructure/ExtensionTrace.cs#L15)). So I suggest we name this something like `Microsoft.Azure.Functions.Worker.Extensions.DurableTask.CallHttp`.

src/Worker.Extensions.DurableTask/TaskOrchestrationContextExtensionMethods.cs
```diff
@@ -45,8 +95,20 @@ public static Task<DurableHttpResponse> CallHttpAsync(this TaskOrchestrationCont
         {
             Content = content,
             HttpRetryOptions = retryOptions,
+            AsynchronousPatternEnabled = true,
```
cgillum: This will be a behavior change for existing customers using this API. Should we instead default to `false` and make it settable?

test/e2e/Apps/BasicDotNetIsolated/HTTPFeature.cs
```diff
@@ -0,0 +1,83 @@
...
+using Microsoft.Azure.Functions.Worker.Extensions.DurableTask.Http;
+using Microsoft.DurableTask;
+using Microsoft.DurableTask.Client;
+using Microsoft.Extensions.Logging;
+using System.Net;
+using System.Text.Json.Nodes;
+
+namespace Microsoft.Azure.Durable.Tests.E2E;
+
+public static class HTTPFeature
```
cgillum: nit: Common .NET naming conventions use `Http` instead of `HTTP`.

src/Worker.Extensions.DurableTask/TaskOrchestrationContextExtensionMethods.cs
nytian: Thanks, I will make it settable. But the default value is true at [in-proc model ](https://github.com/Azure/azure-functions-durable-extension/blob/2883a516805218a6fda2dd0e1109755ea6c5d0b2/src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs#L35). If changing to false here then it will be inconsistent.

src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteOrchestratorContext.cs
jviau: You have a custom method converting dictionary to protobuf map. You could also make a custom method to convert from your POCO to a protobuf map. And I don't see any protobuf deserialization, so not sure if that direction would be needed.

src/Worker.Extensions.DurableTask/TaskOrchestrationContextExtensionMethods.cs
cgillum: I suppose we have a problem either way. For existing .NET Isolated customers, this is a behavior change because .NET Isolated did not previously do polling. For existing in-proc customers, it will not be a change in behavior, as you mentioned. I guess I'm mostly worried about the existing .NET Isolated customers who upgrade to this newer nuget package. They might not expect any behavior changes. In-proc customers, on the other hand, already have to do a massive migration, and behavior changes are already expected (and can be documented).

src/Worker.Extensions.DurableTask/TaskOrchestrationContextExtensionMethods.cs
```diff
@@ -72,7 +73,7 @@ public static async Task<DurableHttpResponse> CallHttpAsync(this TaskOrchestrati
 
             DurableHttpRequest newHttpRequest = CreateLocationPollRequest(request, locationUrl);
 
-            logger.LogInformation($"Polling HTTP status at location: {locationUrl}");
+            logger.LogInformation($"Polling Http status at location: {locationUrl}");
```
cgillum: Non-blocking, but my comment about HTTP vs Http only applies to .NET method and type names. Not to strings and comments, for which HTTP is more appropriate.

src/Worker.Extensions.DurableTask/TaskOrchestrationContextExtensionMethods.cs
```diff
@@ -38,15 +89,34 @@ public static Task<DurableHttpResponse> CallHttpAsync(this TaskOrchestrationCont
...
+    /// <param name="asynchronousPatternEnabled">Boolean controls Whether Durable HTTP should automatically handle async HTTP patterns like 202 with polling. Default to true. </param>
     /// <returns>A <see cref="Task{DurableHttpResponse}"/>Result of the HTTP call.</returns>
-    public static Task<DurableHttpResponse> CallHttpAsync(this TaskOrchestrationContext context, HttpMethod method, Uri uri, string? content = null, HttpRetryOptions? retryOptions = null)
+    public static Task<DurableHttpResponse> CallHttpAsync(
+        this TaskOrchestrationContext context, 
+        HttpMethod method,
+        Uri uri,
+        string? content = null,
+        HttpRetryOptions? retryOptions = null,
+        bool asynchronousPatternEnabled = true)
```
cgillum: Changing the parameter list for an existing method is a binary breaking change. We need to instead add a new method overload that has the new parameter.

src/Worker.Extensions.DurableTask/Worker.Extensions.DurableTask.csproj
```diff
@@ -29,7 +29,7 @@
     <AssemblyOriginatorKeyFile>..\..\sign.snk</AssemblyOriginatorKeyFile>
 
     <!-- Version information -->
-    <VersionPrefix>1.4.0</VersionPrefix>
+    <VersionPrefix>1.4.1</VersionPrefix>
```
cgillum: This needs to be a minor version update since we're making an API change (adding a new parameter to `CallHttpAsync`).

src/Worker.Extensions.DurableTask/TaskOrchestrationContextExtensionMethods.cs
cgillum: Also, per our other discussion, I think the default value should be `false` to avoid a behavior change for existing .NET Isolated users who may be calling this method today and not expecting async polling. We can document this difference from in-proc in the Durable Functions Versions official docs.

src/Worker.Extensions.DurableTask/TaskOrchestrationContextExtensionMethods.cs
```diff
@@ -38,15 +89,35 @@ public static Task<DurableHttpResponse> CallHttpAsync(this TaskOrchestrationCont
...
+    /// <param name="asynchronousPatternEnabled">Boolean controls Whether Durable HTTP should automatically handle async HTTP patterns like 202 with polling. Default to false. </param>
     /// <returns>A <see cref="Task{DurableHttpResponse}"/>Result of the HTTP call.</returns>
-    public static Task<DurableHttpResponse> CallHttpAsync(this TaskOrchestrationContext context, HttpMethod method, Uri uri, string? content = null, HttpRetryOptions? retryOptions = null)
+    public static Task<DurableHttpResponse> CallHttpAsync(
+        this TaskOrchestrationContext context, 
+        HttpMethod method,
+        Uri uri,
+        string? content = null,
+        HttpRetryOptions? retryOptions = null,
+        bool asynchronousPatternEnabled = false)
```
cgillum, line 119: Adding a new parameter to an existing public method is a binary breaking change. We need to instead create a new overload with this parameter.

src/WebJobs.Extensions.DurableTask/Bindings/OrchestrationTriggerAttributeBindingProvider.cs
```diff
@@ -172,6 +172,7 @@ public Task<ITriggerData> BindAsync(object? value, ValueBindingContext context)
                         NewEvents = { remoteContext.NewEvents.Select(ProtobufUtils.ToHistoryEventProto) },
                         EntityParameters = remoteContext.EntityParameters.ToProtobuf(),
                     };
+                    orchestratorRequest.Properties.Add(ProtobufUtils.ConvertPocoToProtoMap(remoteContext.Configurations));
```
cgillum, line 175: Non-blocking: I'm not sure I understand why we need to convert from a POCO to a map. Wouldn't it make more sense to convert from a POCO to a custom protobuf message type? Was it a mistake to use a map in the protobuf contract? What we're doing now seems like it's creating unnecessary overhead, especially now that we have to use reflection like this for every orchestrator request.

nytian: Oh, that’s because of the comment [here](https://github.com/Azure/azure-functions-durable-extension/pull/3092#discussion_r2056409143). I updated the configuration dictionary used in `RemoteOrchestratorContext` to be a POCO, which is now defined in the new added `RemoteOrchestratorConfiguration` file. I think it should be fine, since we’ll only be adding new properties to RemoteOrchestratorConfiguration for values we want to pass through the bindings. But let me know if you have other concerns.

cgillum (approved):

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
bachuv, line 1: Can we remove this change or update it to 3.3.0?

src/Worker.Extensions.DurableTask/AssemblyInfo.cs
bachuv, line 1: Same with this change -- update the version number to v3.3.0 or remove the change.

bachuv (approved):

bachuv (approved):

PR #3095, opened 2025-04-22
Introduce worker versioning into durable functions

halspang:
This commit adds the worker versioning options to the base DurableTaskOptions which are read through the host.json file. This change also includes a few changes for wiring the values through as well as testing.

> [!IMPORTANT]
> This change depends on multiple external changes/new versions:
> - A release containing the changes in [DurableTask.Core](https://github.com/Azure/durabletask/commit/63d45d2c9217a8760d03b0151978879f4aff7e29)
> - A PR and subsequent release for [DurableTask.Abstractions](https://github.com/microsoft/durabletask-dotnet/pull/416)

As is, the tests in this PR will fail due to the lack of appropriate references. Once those releases are made, this PR will have to be updated. However, the code/tests are all complete in their implementation and are OK to be reviewed. This just cannot be merged until that point.


test/e2e/Tests/Tests/VersioningTests.cs
```diff
@@ -0,0 +1,59 @@
...
+        _fixture = fixture;
+        _fixture.TestLogs.UseTestLogger(testOutputHelper);
+        _output = testOutputHelper;
+    }
+
+    [Theory]
+    [InlineData("")] // Represents a non-versioned case.
+    [InlineData("1.0")]
+    [InlineData("2.0")]
+    public async Task TestVersionedOrchestration_OKWithMatchingVersion(string version)
```
AnatoliB: Please consider adding a test that would check which version is passed to suborchestrations. In fact, there seems to be a problem here: in my manual tests, `context.Version` is empty in suborchestrations, but I expected it to match `defaultVersion` in `host.json`. See the discussion here: https://github.com/Azure/azure-functions-durable-extension/pull/3072#discussion_r2034075880

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -30,6 +31,16 @@ public class DurableTaskOptions
         /// </summary>
         public string DefaultVersion { get; set; }
 
+        /// <summary>
+        /// The strategy that will be used for matching versions when running an orchestration. See <see cref="VersioningSettings.VersionMatchStrategy"/> for more information.
+        /// </summary>
+        public VersioningSettings.VersionMatchStrategy VersionMatchStrategy { get; set; } = VersioningSettings.VersionMatchStrategy.None;
```
AnatoliB: Let's think about this default a bit more. I understand that `None` looks intuitively safe. However, let's consider the following important use case:

1. There is a Function app (V1) that was deployed before any versioning support existed, or before the user became interested in it. This app obviously doesn't have anything in `host.json`.
2. This app is active, it has long-running orchestrations in flight.
3. At some point, the user wants to change orchestrator logic. They are willing to learn how versioning works and make all the necessary config and code changes in V2, so they set `defaultVersion` in `host.json` to a non-empty value. However, they are not willing to take any downtime, so they want to perform a rolling update, which means that V1 and V2 workers will be active at the same time.
4. A V2 worker creates an orchestration marked as V2.
5. A V1 worker may try to act on this orchestration.

With the currently proposed default (`VersionMatchStrategy.None`), any V1 worker can just try and handle this orchestration. As a result, in order to upgrade without downtime, the user will have to make one more deployment that just sets `VersionMatchStrategy` without making any other changes and, most importantly, they must _know_ that this is what they have to do.

If we change the default to `VersionMatchStrategy.CurrentOrOlder`, the V1 workers will reject this orchestration, which is what the user needs. The upgrade story becomes simpler, with less opportunities to break things.

I also suspect that `CurrentOrOlder` will become the most popular choice for those interested in versioning anyway. But this is not just about the convenience: I believe it's also a much safer and non-destructive choice.

AnatoliB:
This problem is probably not introduced nor fixable in this PR, but I discovered an issue with orchestration rejection when testing this E2E, so I'll mention it here.

I have an orchestrator like this:

```csharp
[Function(nameof(WaitForEventOrchestrator))]
public static async Task<string> WaitForEventOrchestrator(
    [OrchestrationTrigger] TaskOrchestrationContext context)
{
    context.SetCustomStatus("Waiting for Continue event...");
    await context.WaitForExternalEvent<object>("Continue");
    context.SetCustomStatus("Continue event received");
    return context.Version;
}
```

Versioning config in `host.json`:

```json
"defaultVersion": "2.0",
"versionMatchStrategy": "CurrentOrOlder",
"versionFailureStrategy": "Reject",
```

When I start an orchestration and it gets paused waiting for an external event, as expected. Next, I shut the worker down, change `defaultVersion` to `1.0` to emulate the situation when an old worker is still around, start the worker again, and raise the external event. This worker is unable to complete the orchestration, but I'm observing two unexpected side effects:

1. The persisted orchestration history keeps growing, adding `EventRaised` and `orchestratorStarted` to the table repeatedly and indefinitely.
2. The custom status disappears.

Both can be significant issues for the user in some contexts.

test/e2e/Tests/Tests/VersioningTests.cs
halspang: Added tests (and the required code) for this use case.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
halspang: Based off our offline conversation, updated this to match your model.

halspang:
Updated the code to handle the sub-orchestration version use case you brought up @AnatoliB.

Note that this PR now also depends on this change: https://github.com/microsoft/durabletask-dotnet/pull/437

So that means the dependencies are for a DurableTask.Core release and the packages in the PR above ^

test/e2e/Tests/Tests/VersioningTests.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        Assert.Equal(HttpStatusCode.Accepted, response.StatusCode);
+        string statusQueryGetUri = await DurableHelpers.ParseStatusQueryGetUriAsync(response);
+
+        await DurableHelpers.WaitForOrchestrationStateAsync(statusQueryGetUri, "Completed", 30);
+
+        var orchestrationDetails = await DurableHelpers.GetRunningOrchestrationDetailsAsync(statusQueryGetUri);
+        if (version != null)
+        {
+            Assert.Equal($"Parent Version: '2.0' | Sub Version: '{version}'", orchestrationDetails.Output);
```
cgillum: I'm confused. Shouldn't both the parent and the child have the same explicitly-specified version?

This is what I was expecting:
```suggestion
            Assert.Equal($"Parent Version: '{version}' | Sub Version: '{version}'", orchestrationDetails.Output);
```

cgillum: Perhaps the test code just needs to be clarified. Instead of labeling the sub-orchestration version parameter as `version` here and in the HTTP trigger function, it should be named `subOrchestrationVersion` or something similar to better clarify how it's intended to be used.

src/Worker.Extensions.DurableTask/Worker.Extensions.DurableTask.csproj
```diff
@@ -56,4 +56,7 @@
       <PackagePath>content/SBOM</PackagePath>
     </Content>
   </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\..\..\durabletask-dotnet\src\Abstractions\Abstractions.csproj" />
```
cgillum: I'm surprised to see that we don't already have this kind of project reference. Looking up at lines 42-43, I'm also surprised that those aren't already project references. This seems suspicious to me, and I wonder if this was an oversight or if it was done on purpose. Either way, it seems dangerous to have a mix of package references and project references here because the version numbers could get out of sync. I would expect that we'd be using project references for everything defined locally in this repo.

cgillum (commented):
Some initial feedback. Haven't gone through the whole thing again yet.

src/Worker.Extensions.DurableTask/Worker.Extensions.DurableTask.csproj
halspang: No, that shouldn't be there anymore. I've been having a weird issue where my VS says that I've changed/updated the code but then it doesn't appear to be reflected in git. I will fix and repush this :)

src/Worker.Extensions.DurableTask/FunctionsOrchestrationContext.cs
```diff
@@ -87,6 +89,31 @@ public override Task<TResult> CallSubOrchestratorAsync<TResult>(
...
+                subOptions = new SubOrchestrationOptions
+                {
+                    InstanceId = subOrchestrationOptions.InstanceId,
+                    Retry = subOrchestrationOptions.Retry,
+                    Version = subOrchestrationOptions.Version?.Version ?? defaultVersion,
+                };
+            }
+            else
+            {
+                subOptions = new SubOrchestrationOptions
```
cgillum: I'm predicting ongoing maintanance problems with this ad-hoc copy mechanism. For example, this code doesn't preserve `Tags` which is something that Bill's PR is adding. It might be easier to make sure we preserve existing configuration in the user-provided `options` argument if we define a copy-constructor in the `SubOrchestrationOptions` class and call it here when adding our Version configuration.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -225,7 +225,7 @@ public override Task<Empty> Hello(Empty request, ServerCallContext context)
                     ExecutionStartedEvent executionStartedEvent = new ExecutionStartedEvent(-1, request.Input)
                     {
                         Name = request.Name,
-                        Version = request.Version,
+                        Version = request.Version != null ? request.Version : this.extension.Options.DefaultVersion,
```
cgillum, line 228: Just a heads-up that @nytian is also making big refactoring changes to this class, and we should make sure that these changes don't get lost.

cgillum (commented):
Finished reviewing. Just a couple more things...

src/Worker.Extensions.DurableTask/Worker.Extensions.DurableTask.csproj
cgillum: Oh - that project isn't even part of this repo. That explains my confusion. 😂

test/e2e/Tests/Tests/VersioningTests.cs
halspang: Updated it to be `subOrchestrationVersion`. The point of this test is to show we can individually version the sub-orchestration and that it can also take on the default. The parent is only using the default in this case, hence the constant `2.0`.  Hopefully it is clearer now with the new variable :)

src/Worker.Extensions.DurableTask/FunctionsOrchestrationContext.cs
halspang: Looking into this a little bit I found we already have a constructor that takes `TaskOptions`. So as long as `TaskOptions` is where `Tags` lives, we should be good here (provided the constructor is updated when tags are added). 

If you'd prefer I rewrite it to have a copy constructor from another `SubOrchestrationOptions`, I can do that but it'd require another package release. Happy to do it if you feel we need it though!

cgillum: Given the priority of the versioning work, I don't want to block this specific PR on this, but could we open an issue and add a TODO here to track updating this and do it as a fast follow?

cgillum (approved):

src/Worker.Extensions.DurableTask/FunctionsOrchestrationContext.cs
halspang: Sounds like a plan to me! Here's the issue, `TODO` coming in a moment.

https://github.com/microsoft/durabletask-dotnet/issues/440

cgillum (approved):

AnatoliB (approved):

Issue #2862, opened 2023-11-18
Durable Function - dotnet-isolated - Grpc.Core.RpcException: Status(StatusCode="Unknown", Detail="Exception was thrown by handler.")


gui28347:
- Trying to migrate a function running in-process to out-of-process dotnet-isolated 7.

- Same code works perfectly in in-process mode. Causing this exception in dotnet-isolated 7 after migrating the required code to run as dotnet-isolatd.

Stack: 

```LOG
[2023-11-18T04:13:01.793Z] Function 'MainActivityVM', Invocation id 'b81a7a79-437c-4d76-8ae2-64e48fbbb62e': An exception was thrown by the invocation.
[2023-11-18T04:13:01.794Z] Result: Function 'MainActivityVM', Invocation id 'b81a7a79-437c-4d76-8ae2-64e48fbbb62e': An exception was thrown by the invocation.
Exception: System.AggregateException: One or more errors occurred. (RunMain Exception: Grpc.Core.RpcException: Status(StatusCode="Unknown", Detail="Exception was thrown by handler.")
[2023-11-18T04:13:01.795Z]    at Microsoft.DurableTask.Client.Grpc.GrpcDurableTaskClient.RaiseEventAsync(String instanceId, String eventName, Object eventPayload, CancellationToken cancellation)
[2023-11-18T04:13:01.795Z]    at TEST99.SampleFunction.RunMain(FunctionContext context, DurableTaskClient dclient, String vmname, String vmtype) in C:\temp\investigate\AzureAutomation_PersonalDOTNET8\TEST99\SampleFunction.cs:line 82)
[2023-11-18T04:13:01.796Z]  ---> System.Exception: RunMain Exception: Grpc.Core.RpcException: Status(StatusCode="Unknown", Detail="Exception was thrown by handler.")
[2023-11-18T04:13:01.797Z]    at Microsoft.DurableTask.Client.Grpc.GrpcDurableTaskClient.RaiseEventAsync(String instanceId, String eventName, Object eventPayload, CancellationToken cancellation)
[2023-11-18T04:13:01.797Z]    at TEST99.SampleFunction.RunMain(FunctionContext context, DurableTaskClient dclient, String vmname, String vmtype) in C:\temp\investigate\AzureAutomation_PersonalDOTNET8\TEST99\SampleFunction.cs:line 82
[...]

PR #3097, opened 2025-04-30
Add Explicit Exception Types to RaiseEventAsync and Make Exception Throwing controllable via Settings

nytian:
Solves #2862 , and https://github.com/microsoft/durabletask-dotnet/issues/248

This PR improves exception handling in RaiseEventAsync by:

- Returning RpcException with details error message instead of an unknown gRPC error.
- Make Configuration throwStatusExceptionsOnRaiseEvent setting to control if the above RpcException will throw at DurableTaskClient.RaiseEventAsync.

### Pull request checklist


src/Worker.Extensions.DurableTask/FunctionsDurableTaskClient.cs
```diff
@@ -61,6 +64,11 @@ public override Task<PurgeResult> PurgeInstanceAsync(
     public override Task RaiseEventAsync(
         string instanceId, string eventName, object? eventPayload = null, CancellationToken cancellation = default)
     {
+        if (this.ThrowStatusExceptionsOnRaiseEvent == false && string.IsNullOrEmpty(instanceId))
```
nytian, line 67: Since we're using a gRPC client, the event name and instance ID must not be null—otherwise the gRPC message wouldn't be sent at all.  Also, according to the IDurableClient.RaiseEventAsync documentation, the configuration only controls whether an ArgumentException is thrown when the instance ID is null, and an InvalidOperationException when the instance is completed. It does not control behavior when the event name is null. That’s why I added it this way.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -256,7 +256,33 @@ private OrchestrationStatus[] GetStatusesNotToOverride()
...
+                bool throwStatusExceptionsOnRaiseEvent = this.extension.Options.ThrowStatusExceptionsOnRaiseEvent ?? this.extension.DefaultDurabilityProvider.CheckStatusBeforeRaiseEvent;
+                try
+                {
+                    await this.GetClient(context).RaiseEventAsync(request.InstanceId, request.Name, Raw(request.Input));
+                }
+                catch (ArgumentException)
+                {
+                    if (throwStatusExceptionsOnRaiseEvent)
+                    {
+                        throw new RpcException(new Status(StatusCode.InvalidArgument, "The instance id is invalid."));
```
cgillum: I think we want to throw this exception regardless of the setting. The setting only controls the case where there is a status mismatch.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -256,7 +256,33 @@ private OrchestrationStatus[] GetStatusesNotToOverride()
...
+                    if (throwStatusExceptionsOnRaiseEvent)
+                    {
+                        throw new RpcException(new Status(StatusCode.FailedPrecondition, "The orchestration instance with the provided instance id is not running."));
+                    }
+                }
+                catch (Exception ex)
+                {
+                    if (throwStatusExceptionsOnRaiseEvent)
+                    {
+                        throw new RpcException(new Status(StatusCode.Unknown, ex.Message));
```
cgillum: I don't think you need to check for `throwStatusExceptionOnRaiseEvent` here either. We should always return this more helpful RPC exception in this case.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
nytian: Thanks, Chris! I'm a bit confused about the [description ](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs#L228)of this setting —so it only controls the case where the instance status isn't found, for which we would throw an ArgumentException) [here,](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs#L811)  For other cases like ArgumentNullException due to missing instanceId, eventName, or taskName, we’ll still always throw, right?

cgillum: > it only controls the case where the instance status isn't found, for which we would throw an ArgumentException

Yes, this is correct.

test/e2e/Apps/BasicDotNetIsolated/ExternalEventOrchestration.cs
```diff
@@ -0,0 +1,85 @@
...
+    public static async Task<HttpResponseData> SendExternalEvent_HttpStart(
+        [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post")] HttpRequestData req,
+        [DurableClient] DurableTaskClient client,
+        FunctionContext executionContext)
+    {
+        var response = req.CreateResponse();
+
+        try
+        {
+            await client.RaiseEventAsync("ExternalEventTest", "Approval", true);
```
cgillum: Should we instead use a randomly generated instance ID, which is passed into this HTTP trigger via a parameter? Using a hardcoded instance ID runs the risk of creating conflicts due to previously created instances in the same task hub.

test/e2e/Tests/Fixtures/FixtureHelpers.cs
```diff
@@ -100,12 +100,14 @@ internal static void AddDurableBackendEnvironmentVariables(Process funcProcess,
                 funcProcess.StartInfo.EnvironmentVariables["AzureFunctionsJobHost__extensions__durableTask__storageProvider__connectionStringName"] = "SQLDB_Connection";
                 funcProcess.StartInfo.EnvironmentVariables["AzureFunctionsJobHost__extensions__durableTask__storageProvider__createDatabaseIfNotExists"] = "true";
                 funcProcess.StartInfo.EnvironmentVariables["AzureFunctionsJobHost__extensions__durableTask__MaxGrpcMessageSizeInBytes"] = "6291456";
+                funcProcess.StartInfo.EnvironmentVariables["AzureFunctionsJobHost__extensions__durableTask__ThrowStatusExceptionsOnRaiseEvent"] = "true";
```
cgillum, line 103: Why are we setting this only on a subset of backends (instead of all backends)?

test/e2e/Tests/Tests/ExternalEventTests.cs
```diff
@@ -0,0 +1,61 @@
...
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Azure.Durable.Tests.DotnetIsolatedE2E;
+
+[Collection(Constants.FunctionAppCollectionName)]
+public class ExternalEventTests
+{
+    private readonly FunctionAppFixture _fixture;
+    private readonly ITestOutputHelper _output;
```
cgillum: nit: Unlike the Functions team code, we don't really use `_`-prefixed field names in our code bases. Rather, we use normal caselCasing and use `this.` for referencing.

cgillum (commented):
Just a few relatively small things.

test/e2e/Tests/Tests/ExternalEventTests.cs
andystaples: This is consistent with the other files in this test suite. @nytian if you make this change, can you do the same in the other test files as well?

test/e2e/Tests/Fixtures/FixtureHelpers.cs
nytian: Only Azure Storage sets [CheckStatusBeforeRaiseEvent](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs#L61C30-L61C57) to true, and ThrowStatusExceptionsOnRaiseEvent depends on that. SQL and DTS do not implement this setting, so their default is false. So at these two backends, no exception will be thrown unless the setting is explicitly specified.

cgillum (approved):

cgillum (approved):

PR #3098, opened 2025-04-30
Add check for ACA for Linux logging

andystaples:
Adds a check that will initialize Linux logging if the environment is detected as Managed Environment (Azure Container Apps). This will lead to telemetry being emitted with the prefix MS_DURABLE_FUNCTION_EVENTS_LOGS for consumption by ACA logging pipelines that will ultimately land in the DurableFunctionsEvents table for internal diagnostics. 

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -415,6 +415,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             // Determine host platform
             bool inConsumption = this.PlatformInformationService.IsInConsumptionPlan();
+            bool isManagedAppEnvironment = this.PlatformInformationService.IsManagedAppEnvironment();
 
             string tenant = this.PlatformInformationService.GetLinuxTenant();
             string stampName = this.PlatformInformationService.GetLinuxStampName();
@@ -422,7 +423,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
             // in linux consumption, logs are emitted to the console.
             // In other linux plans, they are emitted to a logfile.
-            var linuxLogger = new LinuxAppServiceLogger(writeToConsole: inConsumption, containerName, tenant, stampName);
+            var linuxLogger = new LinuxAppServiceLogger(writeToConsole: inConsumption || isManagedAppEnvironment, containerName, tenant, stampName);
 
             // The logging service for linux works by capturing EventSource messages,
             // which our linux platform does not recognize, and logging them via a
--- a/src/WebJobs.Extensions.DurableTask/IPlatformInformation.cs
+++ b/src/WebJobs.Extensions.DurableTask/IPlatformInformation.cs
@@ -84,9 +84,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         bool IsInConsumptionPlan();
 
         /// <summary>
-        /// Determine the underlying plan is Consumption or not.
+        /// Gets a value indicating whether the application is running in Managed App environment.
         /// </summary>
-        /// <returns> True if the plan is Consumption. Otherwise, False.</returns>
+        /// <returns><see cref="true"/> if running in Managed App environment; otherwise, false.</returns>
         bool IsManagedAppEnvironment();
 
         /// <summary>
--- a/src/WebJobs.Extensions.DurableTask/IPlatformInformation.cs
+++ b/src/WebJobs.Extensions.DurableTask/IPlatformInformation.cs
@@ -84,6 +84,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         bool IsInConsumptionPlan();
 
         /// <summary>
+        /// Determine the underlying plan is Consumption or not.
+        /// </summary>
+        /// <returns> True if the plan is Consumption. Otherwise, False.</returns>
+        bool IsManagedAppEnvironment();
+
+        /// <summary>
         /// Determine the underlying operating system.
         /// </summary>
         /// <returns>An OperatingSystem enum.</returns>
--- a/release_notes.md
+++ b/release_notes.md
@@ -4,17 +4,7 @@
 
 ### New Features
 
-- Fail fast if extendedSessionsEnabled set to 'true' for the worker type that doesn't support extended sessions (https://github.com/Azure/azure-functions-durable-extension/pull/2732).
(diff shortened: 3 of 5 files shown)


andystaples:
@jviau All this platform logic is unique to the Durable extension, but it seems like we should be able to get at the IEnvironment defined by the host and use that instead. 
https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Environment/IEnvironment.cs
Any idea if this is possible from an extension, and how?
Edit: the EnvironmentExtensions specifically
https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Environment/EnvironmentExtensions.cs#L317C28-L317C51

src/WebJobs.Extensions.DurableTask/IPlatformInformation.cs
```diff
@@ -83,6 +83,12 @@ public interface IPlatformInformation
         /// <returns> True if the plan is Consumption. Otherwise, False.</returns>
         bool IsInConsumptionPlan();
 
+        /// <summary>
+        /// Determine the underlying plan is Consumption or not.
+        /// </summary>
+        /// <returns> True if the plan is Consumption. Otherwise, False.</returns>
+        bool IsManagedAppEnvironment();
```
cgillum, line 90: I suggest we call this `IsManagedContainerAppsEnvironment()` since this is specific to ACA + Functions. Just calling it "managed" is a bit too ambiguous.

Also, the copy/paste comments need to be updated.

andystaples: Fixed the comments
The naming is copied from Host's implementation of the same check: 
https://github.com/Azure/azure-functions-host/blob/32c0e71e73aafdc720da234cce96ef69985ddcdf/src/WebJobs.Script/Environment/EnvironmentExtensions.cs#L317C28-L317C51
Though since this is our implementation, we can change it if it matters

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -415,14 +415,15 @@ private void InitializeLinuxLogging()
...
+            bool isManagedAppEnvironment = this.PlatformInformationService.IsManagedAppEnvironment();
 
             string tenant = this.PlatformInformationService.GetLinuxTenant();
             string stampName = this.PlatformInformationService.GetLinuxStampName();
             string containerName = this.PlatformInformationService.GetContainerName();
 
             // in linux consumption, logs are emitted to the console.
             // In other linux plans, they are emitted to a logfile.
-            var linuxLogger = new LinuxAppServiceLogger(writeToConsole: inConsumption, containerName, tenant, stampName);
+            var linuxLogger = new LinuxAppServiceLogger(writeToConsole: inConsumption || isManagedAppEnvironment, containerName, tenant, stampName);
```
krishnajaju, line 426: `CONTAINER_NAME` is not set in case of FA on ACA. I can share an alternative app setting, if you can let me know what `CONTAINER_NAME` environment variable is set to in Consumption SKU.

krishnajaju: Similarly, can you let me know how tenant and stampName are used today in logging, so I can provide alternatives for FA on ACA, if necessary.

jviau:
> @jviau All this platform logic is unique to the Durable extension, but it seems like we should be able to get at the IEnvironment defined by the host and use that instead. https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Environment/IEnvironment.cs Any idea if this is possible from an extension, and how? Edit: the EnvironmentExtensions specifically https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Environment/EnvironmentExtensions.cs#L317C28-L317C51

You will need to duplicate that code yourself. `IEnvironment` is just an abstraction around `System.Environment` at runtime. I wonder if you can also use `IConfiguration`, but I am not sure if in linux the way we apply env vars during specialization will refresh `IConfiguration` - will need testing.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
krishnajaju: okay, I see if these values are not set, these params will be ignored in the logger. I think this behavior should be fine, we can fill the `tenant`, `roleinstance` and other columns in the internal logging pipelines. 

Since `tenant` will not be set here, should we add `!string.IsNullOrEmpty(stampName)` check in this [code](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs#L65), based on [this comment](https://github.com/Azure/azure-functions-durable-extension/blob/9dddafccd4b290e24c272072c93c82f29b26e272/src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs#L57-L63).?

andystaples: For an app I set up in East US 2: 

Env var name | C# var name | DurableFunctionsEvents log table name | Env var value | log value

CONTAINER_NAME | containerName | RoleInstance | 78C4427D-638816376417622016 | App-78C4427D-638816376417622016
WEBSITE_STAMP_DEPLOYMENT_ID | tenant | Tenant | waws-prod-bn1-083 | waws-prod-bn1-083
WEBSITE_HOME_STAMPNAME | stampName | EventStampName | null | null

Looks like the Container Name is just a GUID representing the functions Linux container itself. tenant represents the Antares stamp the container is running on. stampName is null but I doubt this is intentional, FunctionsLogs table has EventStampName identical to Tenant.

krishnajaju: Thanks for checking this. I think it is fine for these values to be empty for ACA case as well, and we can look into setting this in the logging pipeline.

Since tenant will not be set here, should we add `!string.IsNullOrEmpty(stampName)` check in this [code](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs#L65), based on [this comment](https://github.com/Azure/azure-functions-durable-extension/blob/9dddafccd4b290e24c272072c93c82f29b26e272/src/WebJobs.Extensions.DurableTask/LinuxAppServiceLogger.cs#L57-L63).?

krishnajaju (approved):

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
andystaples: I don't think it entirely matters, since it'll be null either way, but I've added the check just for consistency.

vivekjilla (approved):

cgillum (approved):

PR #3106, opened 2025-05-08
Add New Configuration GrpcHttpClientTimeout for DurableTaskClient 

nytian:

resolves https://github.com/orgs/Azure/projects/745/views/1?pane=issue&itemId=99899741&issue=microsoft%7Cdurabletask-dotnet%7C268

Main changes at this PR 
1. Add new configuration `GrpcHttpClientTimeout` which can set the HTTP client timeout at DurableTaskClient. Note that this only applies to .NET 6 or higher
2. Update DurableTaskClient.PurgeInstaceAsync for better grpc exception handling

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
+++ b/src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
@@ -249,6 +249,14 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// </summary>
         public int MaxGrpcMessageSizeInBytes { get; set; } = 4194304;
 
+        /// <summary>
+        /// Sets the timeout for the HTTP client used by the gRPC client, which is used by Durable Funsiont C# Isolated and Java.
+        /// If the server does not respond within this period, the HTTP request will time out.
+        /// Default is 100 seconds.
+        /// This settings only applies when .NET 6 or greater is used.
+        /// </summary>
+        public TimeSpan? GrpcHttpClientTimeout { get; set; } = TimeSpan.FromSeconds(100);
+
         // Used for mocking the lifecycle notification helper.
         internal HttpMessageHandler NotificationHandler { get; set; }
 
--- a/src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
+++ b/src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
@@ -243,14 +243,15 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         public AppLeaseOptions AppLeaseOptions { get; set; } = AppLeaseOptions.DefaultOptions;
 
         /// <summary>
-        /// Option to control the receive message size in bytes of the gRPC client, which is used by Durable Funsiont C# Isolated and Java.
+        /// Option to control the receive message size in bytes of the gRPC client, which is used by Durable Functions C# Isolated and Java (and potentially more languages in the future).
+        /// If the server does not respond within this period, the HTTP request will time out.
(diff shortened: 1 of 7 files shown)


src/Worker.Extensions.DurableTask/FunctionsDurableClientProvider.netstandard.cs
```diff
@@ -10,7 +11,7 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.DurableTask;
 
 internal partial class FunctionsDurableClientProvider
 {
-    private static Channel CreateChannel(ClientKey key, int? maxGrpcMessageSize)
+    private static Channel CreateChannel(ClientKey key, int? maxGrpcMessageSize, TimeSpan grpcHttpClientTimeout)
```
nytian, line 14: when using Grpc.Core in .NET Standard, there isn't a default timeout enforced on gRPC requests. Unlike Grpc.Net.Client, which uses HttpClient and has a default timeout of 100 seconds  So basically we don't need to set any timeout here, we pass this argument only to align with the other one at NET 6.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -249,6 +249,14 @@ public string HubName
         /// </summary>
         public int MaxGrpcMessageSizeInBytes { get; set; } = 4194304;
 
+        /// <summary>
+        /// Sets the timeout for the HTTP client used by the gRPC client, which is used by Durable Funsiont C# Isolated and Java.
```
cgillum:
```suggestion
        /// Sets the timeout for the HTTP client used by the gRPC client, which is used by Durable Functions C# Isolated and Java (and potentially more languages in the future).
```

src/Worker.Extensions.DurableTask/FunctionsDurableClientProvider.cs
```diff
@@ -94,8 +94,9 @@ public async ValueTask DisposeAsync()
     /// <param name="taskHub">The name of the task hub this client is for.</param>
     /// <param name="connectionName">The name of the connection to use for the task-hub.</param>
     /// <param name="maxGrpcMessageSizeInBytes">Max message size in bytes grpc channel can receive.</param>
+    /// <param name="grpcHttpClientTimeout">Timeout for the underlying HTTP client used by the gRPC channel.Only applies when .NET 6 or greater is used.</param>
```
cgillum:
```suggestion
    /// <param name="grpcHttpClientTimeout">Timeout for the underlying HTTP client used by the gRPC channel. Only applies when .NET 6 or greater is used.</param>
```

cgillum (approved):

cgillum:
Let's also create a docs PR for this new host.json setting.

cgillum (approved):

Alan-Hinton:
I've noticed 2 things about the GrpcHttpClientTimeout setting:
1) It needs to be in the host.json file, I originally had it in the environment variables but that didn't have any effect
2) The value needs to be timespan string (e.g. "00:09:59"), I originally entered an numeric value in seconds (i.e. 599) and this caused an exception in a seemingly totally unrelated place that took me ages to track down


Hopefully this will help anyone else using this feature.  

It might be worth improving the error handling/reporting when parsing timespans from the host.json file to make it easier for devs to spot this kind of error. You could also update the [documentation ](https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-bindings?tabs=python-v2%2Cin-process%2C2x-durable-functions&pivots=programming-language-csharp#hostjson-settings) so that it shows how the value should be formatted.

PR #3111, opened 2025-05-14
Fix byte[] Input Handling in Activity Functions at .NET Isolated 

### LINKED ISSUE #2746: ActivityTrigger with byte[] as input results in null reference exception after upgrade to Microsoft.Azure.Functions.Worker.Extensions.DurableTask 1.1.1
### Description
After upgrading from `Microsoft.Azure.Functions.Worker.Extensions.DurableTask` **1.0.4** to **1.1.1** my ActivityTrigger functions throw an exception on a `byte[]` binding parameter.

I have an Azure Durable Function and my orchestrator invokes an `ActivityTrigger` Function which returns a `byte[][]`. I then use a fan-out/fan-in pattern to process each `byte[]` in this array in parallel. With package version 1.0.4 this works, but after upgrading to 1.1.1 calling the ActivityTrigger throws an exception.

I reported this before in https://github.com/Azure/azure-functions-durable-extension/issues/2678 which was closed as duplicate and I hoped it would be resolved with version 1.1.1, but unfortunately the issue still occurs in the latest version.

### Expected behavior
The Function with ActivityTrigger should accept `byte[]` as valid input.

### Actual behavior
The following exception is thrown in the autogenerated `DirectFunctionExecutor`

```
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=FunctionApp
  StackTrace:
   at FunctionApp.DirectFunctionExecutor.<ExecuteAsync>d__3.MoveNext() in C:\Users\[redacted]\source\repos\Sandbox\FunctionApp\Microsoft.Azure.Functions.Worker.Sdk.Generators\Microsoft.Azure.Functions.Worker.Sdk.Generators.FunctionExecutorGenerator\GeneratedFunctionExecutor.g.cs:line 50

  This exception was originally thrown at this call stack:
    FunctionApp.DirectFunctionExecutor.ExecuteAsync(Microsoft.Azure.Functions.Worker.FunctionContext) in GeneratedFunctionExecutor.g.cs
```

Create a new empty Function App. Change the Program.cs to the following:

```csharp
using Microsoft.Azure.Functions.Worker;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Hosting;

var host = new HostBuilder()
    .ConfigureFunctionsWorkerDefaults()
    .ConfigureServices(services =>
    {
        services.AddApplicationInsightsTelemetryWorkerService();
        services.ConfigureFunctionsApplicationInsights();
    })
    .Build();

host.Run();
```

And create class `FanOutByteArrayOrchestrator.cs` with the below content:

```csharp
using Microsoft.Azure.Functions.Worker;
using Microsoft.Azure.Functions.Worker.Http;
using Microsoft.DurableTask;
using Microsoft.DurableTask.Client;
using System.Text;
using System.Text.Json;

namespace FunctionApp
{
    public class FanOutByteArrayOrchestrator
    {
        [Function("FanOutByteArrayOrchestratorApi")]
        public async Task<HttpResponseData> RunApi(
            [HttpTrigger] HttpRequestData request,
            [DurableClient] DurableTaskClient durableTaskClient)
        {
            var orchestrationInstanceId = await durableTaskClient
                .ScheduleNewOrchestrationInstanceAsync("FanOutByteArrayOrchestrator")
                .ConfigureAwait(true);

            return durableTaskClient.CreateCheckStatusResponse(request, orchestrationInstanceId);
        }

        [Function("FanOutByteArrayOrchestrator")]
        public async Task Run(
            [OrchestrationTrigger] TaskOrchestrationContext taskOrchestrationContext)
        {
            var events = await taskOrchestrationContext
                .CallActivityAsync<byte[][]>("CreateEvents")
                .ConfigureAwait(true);

            // Fan-out/fan-in
            var eventTasks = events
                .Select(x => taskOrchestrationContext.CallActivityAsync("ProcessEvent", input: x))
                .ToList();
            await Task.WhenAll(eventTasks).ConfigureAwait(true);
        }

        [Function("CreateEvents")]
        public Task<byte[][]> CreateEvents([ActivityTrigger] CancellationToken cancellationToken)
        {
            var cars = new List<Car>
            {
                new() { HorsePower = 500, Name = "Tesla Model 3"},
                new() { HorsePower = 1000, Name = "Red Bull Formula 1 Car"},
            };

            var byteArray = cars.Select(x => Encoding.UTF8.GetBytes(JsonSerializer.Serialize(x)));

            return Task.FromResult(byteArray.ToArray());
        }

        [Function("ProcessEvent")]
        public Task ProcessEvent([ActivityTrigger] byte[] eventToProcess, CancellationToken cancellationToken)
        {
            var jsonString = Encoding.UTF8.GetString(eventToProcess);
            var car = JsonSerializer.Deserialize<Car>(jsonString);

            Console.WriteLine($"Name: {car?.Name} has {car?.HorsePower} horse power.");

            return Task.CompletedTask;
        }
    }

    internal class Car
    {
        public int HorsePower { get; set; }
        public string? Name { get; set; }
    }
}
```

Content of the .csproj file:

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>net8.0</TargetFramework>
    <AzureFunctionsVersion>v4</AzureFunctionsVersion>
    <OutputType>Exe</OutputType>
    <Nullable>enable</Nullable>
  </PropertyGroup>
  <ItemGroup>
    <FrameworkReference Include="Microsoft.AspNetCore.App" />
    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.21.0" />
    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.DurableTask" Version="1.1.1" />
    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http" Version="3.1.0" />
    <!--<PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore" Version="1.2.1" />-->
    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.17.0" />
    <PackageReference Include="Microsoft.ApplicationInsights.WorkerService" Version="2.22.0" />
    <PackageReference Include="Microsoft.Azure.Functions.Worker.ApplicationInsights" Version="1.2.0" />
  </ItemGroup>
  <ItemGroup>
    <None Update="host.json">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
    </None>
    <None Update="local.settings.json">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
      <CopyToPublishDirectory>Never</CopyToPublishDirectory>
    </None>
  </ItemGroup>
  <ItemGroup>
    <Using Include="System.Threading.ExecutionContext" Alias="ExecutionContext" />
  </ItemGroup>
</Project>
```

### Screenshot
image

### Known workarounds
Downgrade to Microsoft.Azure.Functions.Worker.Extensions.DurableTask 1.0.4.

### If deployed to Azure
Same behavior is seen when deployed to Azure - in my case an isolated azure function app.

nytian:

- resolves #2746 by adding a special case when activity inputs are byte[]
- Add e2e test that cover different type inputs of activity functions. 

### Pull request checklist

### DIFF
--- a/test/e2e/Apps/BasicDotNetIsolated/ActivityInputType.cs
+++ b/test/e2e/Apps/BasicDotNetIsolated/ActivityInputType.cs
@@ -93,7 +93,7 @@ public static class ActivityInputType
     {
         if (input.Data?.GetType() != typeof(byte[]))
         {
-            return $"Error: Expected Data to be byte[] but got {input.Data.GetType().Name}";
+            return $"Error: Expected Data to be byte[] but got {input.Data!.GetType().Name}";
         }
 
         return $"Received CustomClass: {{Name: {input.Name}, Age: {input.Age}, Duration: {input.Duration}, Data: [{string.Join(", ", input.Data)}]}}";
@@ -116,22 +116,22 @@ public static class ActivityInputType
     {
         for (int i = 0; i < input.Length; i++)
         {
-            if (input[i].Data.GetType() != typeof(byte[]))
+            if (input[i].Data!.GetType() != typeof(byte[]))
             {
-                return $"Error: Expected Data to be byte[] but got {input[i].Data.GetType().Name}";
+                return $"Error: Expected Data to be byte[] but got {input[i].Data!.GetType().Name}";
             }
         }
 
         var items = input.Select(item => 
-            $"{{Name: {item.Name}, Age: {item.Age}, Duration: {item.Duration}, Data: [{string.Join(", ", item.Data)}]}}");
+            $"{{Name: {item.Name}, Age: {item.Age}, Duration: {item.Duration}, Data: [{string.Join(", ", item.Data!)}]}}");
         return $"Received CustomClass[]: [{string.Join(", ", items)}]";
     }
 }
 
 public class CustomClass
 {
-    public string Name { get; set; }
+    public string? Name { get; set; }
     public int Age { get; set; }
-    public byte[] Data { get; set; }
+    public byte[]? Data { get; set; }
     public TimeSpan Duration { get; set; }
 }
--- a/test/e2e/Apps/BasicDotNetIsolated/ActivityInputType.cs
+++ b/test/e2e/Apps/BasicDotNetIsolated/ActivityInputType.cs
@@ -1,0 +1,137 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Grpc.Core;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Azure.Functions.Worker.Http;
+using Microsoft.DurableTask;
+using Microsoft.DurableTask.Client;
+using Microsoft.Extensions.Logging;
+using System.Net;
+using System.Text;
+using System.Linq;
+
+namespace Microsoft.Azure.Durable.Tests.E2E;
+
+public static class ActivityInputType
+{
+    [Function("ActivityInputType_HttpStart")]
+    public static async Task<HttpResponseData> ActivityInputType_HttpStart(
+        [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post")] HttpRequestData req,
+        [DurableClient] DurableTaskClient client,
+        FunctionContext executionContext)
+    {
+        ILogger logger = executionContext.GetLogger("ActivityInputType_HttpStart");
+
+        string instanceId = await client.ScheduleNewOrchestrationInstanceAsync(
+            nameof(ActivityInputTypeOrchestrator));
+
+        return await client.CreateCheckStatusResponseAsync(req, instanceId);
+    }
+
+    [Function(nameof(ActivityInputTypeOrchestrator))]
+    public static async Task<List<string>> ActivityInputTypeOrchestrator(
+        [OrchestrationTrigger] TaskOrchestrationContext context)
+    {
+        var output = new List<string>();
+
+        // Test byte array input
+        byte[] byteArrayInput = new byte[] { 1, 2, 3, 4, 5 };
+        output.Add(await context.CallActivityAsync<string>(nameof(ByteArrayInput), byteArrayInput));
+
+        // Test empty byte array input
+        byte[] emptyByteArray = new byte[0];
+        output.Add(await context.CallActivityAsync<string>(nameof(ByteArrayInput), emptyByteArray));
+
+        // Test single byte input
+        byte singleByteInput = 42;
+        output.Add(await context.CallActivityAsync<string>(nameof(SingleByteInput), singleByteInput));
+
+        // Test custom class input
+        var customClassInput = new CustomClass
+        {
+            Name = "Test",
+            Age = 25,
+            Data = new byte[] { 1, 2, 3 },
+            Duration = TimeSpan.FromHours(1)
+        };
+        output.Add(await context.CallActivityAsync<string>(nameof(CustomClassInput), customClassInput));
+
+        // Test int array input
+        int[] intArrayInput = new int[] { 1, 2, 3, 4, 5 };
+        output.Add(await context.CallActivityAsync<string>(nameof(IntArrayInput), intArrayInput));
+
+        // Test string input
+        string stringInput = "Test string input";
+        output.Add(await context.CallActivityAsync<string>(nameof(StringInput), stringInput));
+
+        // Test array of custom class input
+        var complexInput = new CustomClass[]
+        {
+            new CustomClass { Name = "Test1", Age = 25, Data = new byte[] { 1, 2, 3 }, Duration = TimeSpan.FromMinutes(30) },
+            new CustomClass { Name = "Test2", Age = 30, Data = new byte[0], Duration = TimeSpan.FromMinutes(45) }
+        };
+        output.Add(await context.CallActivityAsync<string>(nameof(CustomClassArrayInput), complexInput));
+
+        return output;
+    }
+
+    [Function(nameof(ByteArrayInput))]
+    public static string ByteArrayInput([ActivityTrigger] byte[] input, FunctionContext executionContext)
+    {
+        return $"Received byte[]: [{string.Join(", ", input)}]";
+    }
+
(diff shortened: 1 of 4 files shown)


nytian:
I am not sure if this is the best way to fix this tbh. Usually for oop process( .net isoalted and java), activity functions input target type is string, -- the detonation type should always be string [here](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs#L125). And that's also what we expect at [Activity Input Converter ](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/Worker.Extensions.DurableTask/ActivityInputConverter.cs#L40)at worker extension. 

But somehow when input type is byte[], the destination type is byte[] instead of string, which caused the issue here. As we will always return string from[ DurableActivityContext](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActivityContext.cs#L132), which will cause an error [here ](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/Bindings/ObjectValueProvider.cs#L17) as string doesn't match byte[]. 

So that's how I updated in this PR : return a byte[] at DurableActivityContex instead of string to match the destination type which we can't change, and add a addition case at [Activity Input Converter ](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/Worker.Extensions.DurableTask/ActivityInputConverter.cs#L40). 

But I doubt if this is more like a host/worker issue as the destination type we got is wrong.

cgillum (commented):
Hey @nytian. I apologize, but I'm not understanding the description of the issue. You mentioned that ooproc always expects string, but [the code you referenced](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs#L125) doesn't seem to be directly related to expectations about strings.

I wonder if maybe OOProc in Azure Functions supports both `string` and `byte[]` natively, and we just need to be able to handle both?

It seems that you're doubting your solution and I'm trying to better understand why you are doubting it. Maybe you can clarify a bit more what your concerns are?

nytian:
My apologies, and thank you for the review! @cgillum 

My main question is: based on our design, it seems that regardless of the actual input type for activity functions, the [destination type](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs#L125) at binding providers should always be `string` in the modern OOP model (OutOfProcOrchestrationProtocol.MiddlewarePassthrough ). That means, after calling GetInput(see [here ](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs#L142)and [here](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableActivityContext.cs#L130)), we typically expect to receive a string. And then we match the string content and destination type [here](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs#L145). And that also where we throw the error now, string doesn't match byte[]

The issue arises specifically with byte[] inputs — unlike other types, we're not getting a string as destination type, but keep it as byte[]. And that will make the match above caused [error.](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/Bindings/ObjectValueProvider.cs#L21) From what I understand, how this value is determined is determined by the Functions host, as [this method](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs#L55) should be called by host. And I didn't find how they determine this, also I’m wondering if this might actually be something that should be addressed on the host side, rather than within our layer.

My current fix works by adding an exception specifically for byte[] input, but I’m not sure this is the most ideal or scalable solution.

test/e2e/Apps/BasicDotNetIsolated/ActivityInputType.cs
```diff
@@ -0,0 +1,137 @@
...
+
+        var items = input.Select(item => 
+            $"{{Name: {item.Name}, Age: {item.Age}, Duration: {item.Duration}, Data: [{string.Join(", ", item.Data)}]}}");
+        return $"Received CustomClass[]: [{string.Join(", ", items)}]";
+    }
+}
+
+public class CustomClass
+{
+    public string Name { get; set; }
```
andystaples: This warning should be fixable, either by making Name nullable or setting a default value (String.Empty?)

test/e2e/Apps/BasicDotNetIsolated/ActivityInputType.cs
```diff
@@ -0,0 +1,137 @@
...
+    public int Age { get; set; }
+    public byte[] Data { get; set; }
```
andystaples: Same for Data

andystaples (commented):
With or without the fix for byte[], this is a good e2e test to add for isolated. 
Am I understanding the discussion so far that we think the reason byte[] isn't being serialized to string like other types is due to some specific logic around byte[] in the host?

test/e2e/Apps/BasicDotNetIsolated/ActivityInputType.cs
nytian: Thanks for the review!  yeah that;s my main question: I don't know why byte[] is not treated at string like other types in host ,and thus I don't know if my fix is the ideal way. And I will update all of these warnings

cgillum (approved):
These are great tests. I don't know if there's a better way to fix this particular issue, but my hope is that it puts us in a better position than we were in before, and that future compatibility issues can be addressed as they are discovered.

PR #3118, opened 2025-05-28
Fix the IOException on gRPC server close

andystaples:
Adds exception checking around the new gRPC server close logic. 
Also attempts to check if a port is in use before using it with the gRPC server, to further prevent issues stemming from opening multiple servers from the same process. 

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
@@ -5,8 +5,8 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
-using System.Reflection;
-using System.Reflection.Metadata.Ecma335;
+using System.Linq;
+using System.Net.NetworkInformation;
 using System.Threading;
 using System.Threading.Tasks;
 using DurableTask.Core;
@@ -17,10 +17,8 @@ using DurableTask.Core.Query;
 using DurableTask.Core.Serializing.Internal;
 using Google.Protobuf.WellKnownTypes;
 using Grpc.Core;
-using Microsoft.ApplicationInsights.Extensibility.Implementation;
 using Microsoft.Azure.WebJobs.Extensions.DurableTask.Correlation;
 using Microsoft.Extensions.Hosting;
-using Microsoft.Extensions.Logging;
 using DTCore = DurableTask.Core;
 using P = Microsoft.DurableTask.Protobuf;
 
@@ -65,15 +63,34 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     new ChannelOption(ChannelOptions.MaxSendMessageLength, int.MaxValue),
                 };
 
-                if (this.grpcServer != null)
+                if (this.grpcServer is not null)
                 {
-                    await this.grpcServer.ShutdownAsync();
+                    try
+                    {
+                        await this.grpcServer.ShutdownAsync();
+                    }
+                    catch (IOException)
+                    {
+                        // Do nothing, IOException is a known exception type when trying to shutdown a server
+                        // when its port was already in use
+                    }
+                    catch (Exception ex)
+                    {
+                        this.extension.TraceHelper.ExtensionWarningEvent(
+                            this.extension.Options.HubName,
+                            functionName: string.Empty,
+                            instanceId: string.Empty,
+                            message: $"Unexpected error when closing gRPC server. Exception details: {ex}");
+                    }
                 }
 
                 this.grpcServer = new Server(options);
                 this.grpcServer.Services.Add(P.TaskHubSidecarService.BindService(new TaskHubGrpcServer(this)));
 
-                int listeningPort = numAttempts == 1 ? DefaultPort : this.GetRandomPort();
+                // Attempt to get an unused port. Note that while unlikely, it is possible that the port returned by this method
+                // may be utilized by another process between this call and the gRPC server create below, hence we still need to
+                // guard against port conflicts.
+                int listeningPort = this.GetLikelyAvailablePort();
                 int portBindingResult = this.grpcServer.Ports.Add("localhost", listeningPort, ServerCredentials.Insecure);
                 if (portBindingResult != 0)
                 {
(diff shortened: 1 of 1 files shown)


src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -65,15 +63,34 @@ public async Task StartAsync(CancellationToken cancelToken = default)
                     new ChannelOption(ChannelOptions.MaxSendMessageLength, int.MaxValue),
                 };
 
-                if (this.grpcServer != null)
+                if (this.grpcServer is not null)
```
andystaples, line 66: Not sure this whole ShutdownAsync section is still necessary - given we know that it throws an IOException when 4001 is conflicted anyway, I doubt it is having any effect on global resources as originally theorized. Let me know if I should remove it alltogether

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -120,18 +137,28 @@ public async Task StopAsync(CancellationToken cancelToken = default)
...
 
-        private int GetRandomPort()
+        private int GetLikelyAvailablePort()
         {
-            // Get a random port that has not already been attempted so we don't waste time trying
-            // to listen to a port we know is not free.
+            // Get an available port for use in the gRPC server. Try 4001 first, then select a random open port
+            // in the 30000-31000 range.
+            IPGlobalProperties ipGlobalProperties = IPGlobalProperties.GetIPGlobalProperties();
+            TcpConnectionInformation[] tcpConnInfoArray = ipGlobalProperties.GetActiveTcpConnections();
```
andystaples: Not sure how expensive these calls are - if there is a better way to check if a port is in use, let me know - I saw [Jacob's implementation in his AspNetCore PR](https://github.com/Azure/azure-functions-durable-extension/blob/e1662a1f2cc91bc9c26f3b47c0897ce91da0567d/src/WebJobs.Extensions.DurableTask/Grpc/AspNetCoreLocalGrpcListener.cs#L106) and I'm sure I could do something similar with fixed ports to preserve the 4001 default behavior, but it also seemed expensive

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -120,18 +137,28 @@ public async Task StopAsync(CancellationToken cancelToken = default)
...
                 randomPort = this.portGenerator.Next(MinPort, MaxPort);
+                if (!tcpConnInfoArray.Where(x => x.LocalEndPoint.Port == randomPort).Any())
+                {
+                    return randomPort;
+                }
             }
-            while (this.attemptedPorts.Contains(randomPort));
 
-            return randomPort;
+            return 0;
```
andystaples: Probably need to either throw here, or guard against this in the calling method. Any preference?

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
cgillum: I think you're probably right, TBH. However, it definitely looks suspicious that we don't do anything to dispose of resources. I suggest we keep the shutdown call but with the added exception handling, as you have in the current commit. That will at least remove all doubts.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
cgillum: I think if it can run in less than a second or two it should be fine. But I'm not familiar with these particular APIs. I'm more familiar with the approach Jacob uses, so I do wonder if that might be a safer approach (since it's a bit more understood).

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
cgillum: Yeah, we should throw so that we don't get into some unknown state. But I wonder if we should have a larger number of attempts. Inevitably, throwing will restart the host, which triggers more retries anyways.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
andystaples: Refactored to use this approach. Wondering now if we should keep this 30000-31000 requirement here or if we can replace this loop and random port generator checking specific ports with Jacob's method of defaulting to TcpListener's "find a free port" behavior instead.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
andystaples: Bumped attempts to 50 and added an exception if no port is found in 50 attempts. Might be able to remove all of this if we abandon the 30000-31000 requirement as detailed above

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -120,18 +137,46 @@ public async Task StopAsync(CancellationToken cancelToken = default)
...
+                    return randomPort;
+                }
             }
-            while (this.attemptedPorts.Contains(randomPort));
 
-            return randomPort;
+            throw new Exception(string.Format("Failed to get free port for local gRPC server after {0} attempts", numAttempts));
+        }
+
+        private bool IsTcpPortFree(int port)
```
cgillum, line 164: nit: any method that doesn't reference instance-members should be marked as `static`

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -120,18 +137,46 @@ public async Task StopAsync(CancellationToken cancelToken = default)
...
+        {
+            var listener = new TcpListener(IPAddress.Loopback, port);
+            try
+            {
+                listener.Start();
+                return true;
+            }
+            catch (SocketException)
+            {
+                return false;
```
cgillum, line 179: Should we log a message if we fail to bind to a particular port? Previously we were doing this as part of the failure handling for the gRPC server, but it seems we're no longer likely to go down that path with this new implementation.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -120,18 +137,46 @@ public async Task StopAsync(CancellationToken cancelToken = default)
...
                 randomPort = this.portGenerator.Next(MinPort, MaxPort);
+                if (this.IsTcpPortFree(randomPort))
+                {
+                    return randomPort;
+                }
             }
-            while (this.attemptedPorts.Contains(randomPort));
 
-            return randomPort;
+            throw new Exception(string.Format("Failed to get free port for local gRPC server after {0} attempts", numAttempts));
```
cgillum: nit: it's more conventional to do the following:
```suggestion
            throw new InvalidOperationException($"Failed to get free port for local gRPC server after {numAttempts} attempts");
```

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -120,18 +137,46 @@ public async Task StopAsync(CancellationToken cancelToken = default)
             }
         }
 
-        private int GetRandomPort()
+        private int GetLikelyAvailablePort()
```
cgillum: Probably best to call this `GetAvailablePort()` since we've already checked its availability prior to returning.

cgillum (commented):
A few things I think we should clean up quickly before merging.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
andystaples: The log statement added to this method now uses an instance-member reference

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
andystaples: Added a log for this with separate log text than the warning if we try to bind a gRPC server and get a conflict.

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
andystaples: Done

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
andystaples: Fair enough. Was thinking that the name was appropriate due to the chance the port becomes used after the check but before the gRPC server starts, but this is explained in a comment when the method is called.

cgillum (approved):

PR #3121, opened 2025-06-02
Fix OOProc bug for starting an orchestration with an instance id

bachuv:
This PR fixes an OOProc bug where instance ids were being created in the extension even if an instance id was provided when starting an orchestration in the Python, JS, and PowerShell SDKs. This PR fixes that bug and user specified Instances Ids are now read.

This change also includes updates to the smoke tests to run orchestrations with and without specifying instance ids when starting orchestrations.

### Runtime Enhancements:
* Updated `HttpApiHandler.cs` to use a provided instance ID if available; otherwise, it generates a new one

### Smoke Test Updates:
* Added a `-TestWithCustomInstanceId` flag to `run-smoke-tests.ps1` and `e2e-test.ps1` to validate orchestrations with custom instance IDs. The tests ensure that the returned instance ID matches the requested one.
* **C#:** Updated `HelloCitiesUntyped.cs` to parse and use a custom instance ID from query parameters when starting orchestrations.
* **JavaScript:** Enhanced `durableHello.js` to accept an `instanceId` query parameter and pass it to the orchestration client.
* **Java:** Updated `AzureFunctions.java` to handle an `instanceId` query parameter and use it for orchestration if provided.
* **Python:** Modified `__init__.py` to retrieve and use a custom instance ID from query parameters when starting orchestrations.

### Pull request checklist


sophiatev (approved):
LGTM

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
```diff
@@ -217,7 +217,7 @@ public override Task<Empty> Hello(Empty request, ServerCallContext context)
                     // Create the orchestration instance
                     var instance = new OrchestrationInstance
                     {
-                        InstanceId = request.InstanceId ?? Guid.NewGuid().ToString("N"),
+                        InstanceId = string.IsNullOrEmpty(request.InstanceId) ? Guid.NewGuid().ToString("N") : request.InstanceId,
```
nytian, line 220: Does this issue also apply to .NET Isolated and Java? Since I remember we already assign the value for InstanceId at isolated sdks

nytian (approved):
Thanks for the PR and the tests. Just left a small question but that's not a blocker

src/WebJobs.Extensions.DurableTask/LocalGrpcListener.cs
bachuv: I didn't find the bug when testing with .NET Isolated or Java, but I just added the check here so it's consistent.

PR #3134, opened 2025-06-20
Add E2E for PowerShell

andystaples:
Adds E2E coverage for the new external PowerShell SDK
Does not cover all scenarios - mostly focused on updating build scripts and test code to support multiple languages

### Pull request checklist


test/e2e/Apps/BasicPowerShell/extensions.csproj
andystaples, line 1: Debating whether or not to include this file in the repo. We can .gitignore it and generate it only during the test runs using build-e2e-test.ps1, if we prefer

test/e2e/Tests/Fixtures/FunctionAppFixture.cs
```diff
@@ -12,166 +9,47 @@ namespace Microsoft.Azure.Durable.Tests.DotnetIsolatedE2E;
...
     }
 
-    public async Task InitializeAsync()
+    public Task InitializeAsync()
     {
-        // start host via CLI if testing locally
-        if (Constants.FunctionsHostUrl.Contains("localhost"))
+        string? e2eTestLanguageEnvVarValue = Environment.GetEnvironmentVariable("E2E_TEST_FUNCTIONS_LANGUAGE");
+        _logger.LogInformation("E2E_TEST_FUNCTIONS_LANGUAGE set to " + e2eTestLanguageEnvVarValue);
+        switch ((e2eTestLanguageEnvVarValue ?? "").ToLowerInvariant())
```
andystaples, line 30: Instead of this approach, where the fixture uses the env var to select which language to launch and at what location, it may be possible to do a "chimera" approach where we launch many copies of func all at once on different ports, one for each app. Then the tests can run against all of the apps on each run... needs more investigation

test/e2e/Apps/BasicPowerShell/extensions.csproj
cgillum: What are the pros and cons of either option?

.github/workflows/E2ETest.yml
```diff
@@ -76,13 +92,29 @@ jobs:
...
+      - name: Setup Environment for PowerShell
+        shell: pwsh
+        run: |
+          echo "E2E_TEST_FUNCTIONS_LANGUAGE=powershell" >> $GITHUB_ENV
+          echo "TEST_APP_NAME=BasicPowerShell" >> $GITHUB_ENV
+
+      - name: Run E2E tests (PowerShell)
         working-directory: test/e2e/Tests
-        run: dotnet test --filter AzureStorage!=Skip
+        run: dotnet test --filter 'AzureStorage!=Skip&PowerShell!=Skip'
```
cgillum, line 117: It looks like we're running these sequentially. Have you considered running them in parallel? Probably not important now since there's only one test, but curious to know if there are any technical complexities with parallel execution across languages.

andystaples: Yes I made a note of this r.e. the "Chimera" approach. I think it should be possible to run all the language apps at once, using different ports for Core Tools, and then run all the tests at once too. Need to explore with this more, though. Any thoughts on whether there might be Task Hub conflicts in any of the backends doing this?

test/e2e/Apps/BasicPowerShell/extensions.csproj
andystaples: Leaving it in risks having it as a "changed" file by Git every time we rev the version of the extension. Leaving it out adds complexity to the build

andystaples: I have removed extensions.csproj and automated generating it into build-e2e-test.ps1

test/e2e/Tests/Fixtures/FunctionAppFixture.cs
andystaples: Will implement in future PR

.github/workflows/E2ETest.yml
andystaples: Will implement in future PR

cgillum: Responding to your previous question, we'd want to make sure that each app has its own task hub.

test/e2e/Tests/Fixtures/IsolatedFunctionAppProcess.cs
```diff
@@ -0,0 +1,33 @@
...
+
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.Durable.Tests.DotnetIsolatedE2E;
+
+public class IsolatedFunctionAppProcess : FunctionAppProcess
+{
+    public IsolatedFunctionAppProcess(ILogger logger, TestLoggerProvider TestLogs) : base(logger, TestLogs) { }
+
+    internal override string GetAppPath()
```
cgillum: I'm curious why this logic would be any different across app types. For example, I see that there's a lot more code for IsolatedFunctionAppProcess compared to PowerShellFunctionAppProcess, but fundamentally they both attempt to find `$"../../../../../../test/e2e/Apps/{this._appName}"`. I guess one difference is whether or not there's a `bin` directory, but even so, I wonder if most of this logic can live in the common FunctionAppProcess base class?

test/e2e/Tests/Helpers/DurableHelpers.cs
```diff
@@ -9,7 +9,7 @@ internal class DurableHelpers
 {
     static readonly HttpClient _httpClient = new HttpClient();
 
-    static readonly List<string> finalStates = new List<string>()
+    static readonly List<string> _finalStates = new List<string>()
```
cgillum: nit: we don't normally use the `_` prefix in this codebase, particularly since it's redundant with our `this.` prefix convention. I'd prefer we not establish a new convention unless there's a good reason to.

test/e2e/Tests/Tests/ErrorHandlingTests.cs
```diff
@@ -36,13 +36,14 @@ public async Task OrchestratorWithUncaughtActivityException_ShouldFail()
 
         var orchestrationDetails = await DurableHelpers.GetRunningOrchestrationDetailsAsync(statusQueryGetUri);
         
-        Assert.StartsWith("Microsoft.DurableTask.TaskFailedException", orchestrationDetails.Output);
+        Assert.StartsWith(this.fixture._functionLanguageLocalizer?.GetLocalizedStringValue("RethrownActivityException.ErrorMessage"), orchestrationDetails.Output);
```
cgillum: Optional: Since strings like these are test-specific, I wonder if it would be more maintainable to just hardcode the localized string value directly in the test code, choosing which one to use using a switch statement or some other helper function that wraps the switch statement. That would make it much more obvious to see differences in behavior between languages for a particular scenario, which I think is valuable. The current `ITestLanguageLocalizer` pattern works, but also means I need to look at two different files to understand what kind of messages a particular language worker will return for a particular scenario, which seems less than ideal, ergonimically.

cgillum (commented):
Finished reviewing all files

test/e2e/Tests/Fixtures/IsolatedFunctionAppProcess.cs
andystaples: Isolated cannot start Core Tools at the project root - it must be started in bin/<target>/<netversion> - this is what the extra code for isolated is doing, with added complexity of searching for host.json and using it's parent to avoid the code depending on a specific target. All the other out-of-process apps are run from the project root. 
I can see two paths forward, tell me which you prefer: 
1. Keeping these paths separate, and generalizing PowerShellFunctionAppProcess to OutOfProcFunctionAppProcess if no other Python/Java/Node specific logic is needed in the startup. 
2. Removing all this generalization stuff and just checking in the structure if /bin/ exists. If it does, startup there, else startup in project root. 
While #2 is simpler, I have some concerns, namely that keeping IFunctionAppProcess allows us to generalize if other language-specific startup logic is needed, and that we can provide more specific errors for users running this project for the first time who may not have run ./build-e2e-tests.ps1 to generate the /bin/ folder in BasicDotNetIsolated

test/e2e/Tests/Helpers/DurableHelpers.cs
andystaples: Good to know - I will clean this up everywhere

test/e2e/Tests/Tests/ErrorHandlingTests.cs
andystaples: Just for context, there will be a lot more than two of these cases once we are done with this project - I already have another half-dozen for the other PowerShell tests, and that doesn't include
- Durable tests that are not applicable in PowerShell due to feature gaps
- Outputs that are identical in PowerShell and .NET that might not be the same in other ooproc due to the external PS SDK using durabletask-dotnet components

I think this depends on what our goals are for standardization between languages. If it is our intent that, eventually, all outputs from Durable should be identical no matter what language customers are using, then I think this makes sense - introduce deliberately clunky statements in the tests so that we are forced to look at them and increase consistency over time. 
If, however, we recognize that there are certain outputs that _must_ differ between langauges (which I belive there should be - e.g. exception formatting) I prefer abstracting this logic away from the test code, just for readability in the specific tests. 
I do agree that `ITestLanguageLocalizer` makes it hard to see the specifics of differences that we want to eventually address. Perhaps a hybrid approach makes sense, where `ITestLanguageLocalizer` is used for gaps that we acknowledge and don't want to fix, and inline hardcoded values for scenarios that we do intend to standardize.

AnatoliB (commented):
Nit: if all your _HttpStart functions are identical, consider merging them into one and passing the orchestrator name to the invocation requests, so we have less clutter:

```json
  "bindings": [
    {
      "type": "httpTrigger",
      "direction": "in",
      "route": "orchestrators/{FunctionName}",
      ...
    },
```

```powershell
$FunctionName = $Request.Params.FunctionName
$InstanceId = Start-DurableOrchestration -FunctionName $FunctionName
```

Unless you _need_ them to be different, now or soon.

andystaples:
> Nit: if all your _HttpStart functions are identical, consider merging them into one and passing the orchestrator name to the invocation requests, so we have less clutter:
> Unless you _need_ them to be different, now or soon.

Unfortunately we need the routes to be the same as the Isolated ones (and leave room for the other languages to also have the same routes), so /orchestrators/{orchestratorName} will not suffice. Is there another way we can combine the functions while still allowing for a consistent routing strategy between all languages? @AnatoliB

test/e2e/Apps/BasicPowerShell/RaiseComplexException/run.ps1
```diff
@@ -0,0 +1,17 @@
+#
+# Copyright (c) Microsoft. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#
+
+param($InstanceId)
+
+Import-Module -Name "MyHelperModule"
+
+if (GetExecutionCount($InstanceId) -gt 0) {
```
AnatoliB: I will strongly encourage you to avoid this usage of parentheses with PowerShell, as it looks deceptively similar to other languages and invites incorrect usage. For example, if someone ever decides to add one more parameter, this code will lead to _surprising_ results: `GetExecutionCount($InstanceId, $somethingElse)`.

Consider rewriting it like this:

```powershell
if ((GetExecutionCount $InstanceId) -gt 0) {
    IncrementExecutionCount $InstanceId
    "Success"
}
else {
    IncrementExecutionCount $InstanceId
    throw [System.InvalidOperationException]::new("This activity failed\r\nMore information about the failure", [System.OverflowException]::new("Inner exception message"))
}
```

test/e2e/Tests/Fixtures/IsolatedFunctionAppProcess.cs
cgillum: I don't have a strong opinion on this, but I'm anticipating that there won't be much difference at all between languages, hence my suggestion to simplify the abstractions. I usually prefer to design systems with the least number of abstractions as possible and then add the abstractions only when they're strictly needed.

test/e2e/Apps/BasicPowerShell/RaiseException/run.ps1
```diff
@@ -0,0 +1,17 @@
+#
+# Copyright (c) Microsoft. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#
+
+param($InstanceId)
+
+Import-Module -Name "MyHelperModule"
+
+if (GetExecutionCount($InstanceId) -gt 0) {
```
AnatoliB: The same issue with parentheses

test/e2e/Tests/Tests/ErrorHandlingTests.cs
cgillum: I think the clunkiness is unavoidable - it either gets scattered across a new language localizer extension, making it harder to discover, or it gets embedded into the test code. My hope was that we could create a helper function or something that minimizes the clunkiness if we went with the latter option. But I think all this is just a matter of preference, hence why I'm calling this optional.

test/e2e/Apps/BasicPowerShell/requirements.psd1
```diff
@@ -0,0 +1,8 @@
+#
+# Copyright (c) Microsoft. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#
+
+@{
+    'AzureFunctions.PowerShell.Durable.SDK' = '1.*'
```
AnatoliB: Let's upgrade to `2.*`

test/e2e/Apps/BasicPowerShell/RaiseComplexException/run.ps1
andystaples: Fixed

test/e2e/Apps/BasicPowerShell/RaiseException/run.ps1
andystaples: Fixed

test/e2e/Apps/BasicPowerShell/requirements.psd1
andystaples: Done

andystaples:
> > Nit: if all your _HttpStart functions are identical, consider merging them into one and passing the orchestrator name to the invocation requests, so we have less clutter:
> > Unless you _need_ them to be different, now or soon.
> 
> Unfortunately we need the routes to be the same as the Isolated ones (and leave room for the other languages to also have the same routes), so /orchestrators/{orchestratorName} will not suffice. Is there another way we can combine the functions while still allowing for a consistent routing strategy between all languages? @AnatoliB

I have updated both isolated and PS to use a generic orchestrator starter function in as many scenarios as possible. This cut a huge amount of app code in both languages - thanks!

test/e2e/Tests/Helpers/DurableHelpers.cs
andystaples: Should be done - resolving, but point out any others you see

test/e2e/Tests/Fixtures/FunctionAppFixture.cs
```diff
@@ -12,166 +9,51 @@ namespace Microsoft.Azure.Durable.Tests.DotnetIsolatedE2E;
...
+                break;
+            case "powershell":
+                this.functionAppProcess = new PowerShellFunctionAppProcess(this.logger, this.TestLogs);
+                this.functionLanguageLocalizer = new PowerShellTestLanguageLocalizer();
+                break;
+            default:
+                this.logger.LogWarning("Environment variable E2E_TEST_FUNCTIONS_LANGUAGE not set, tests configured for dotnet-isolated");
+                this.functionAppProcess = new IsolatedFunctionAppProcess(this.logger, this.TestLogs);
+                this.functionLanguageLocalizer = new IsolatedTestLanguageLocalizer();
+                break;
```
bachuv: Should the default be to not run any tests? I'm wondering if it's misleading to have the default be dotnet-isolated. For example, if we're trying to run the PowerShell tests and E2E_TEST_FUNCTIONS_LANGUAGE isn't set correctly to `powershell` then it seems like it will run .NET Isolated tests? Is that correct? Let me know if the behavior is different.

test/e2e/Tests/Tests/ErrorHandlingTests.cs
AnatoliB: I will vote for this approach:
> Perhaps a hybrid approach makes sense, where ITestLanguageLocalizer is used for gaps that we acknowledge and don't want to fix, and inline hardcoded values for scenarios that we do intend to standardize.

cgillum (approved):
I'm good with the current set of changes as-is. Would appreciate a signoff from @AnatoliB as well before merging.

test/e2e/Tests/Fixtures/FunctionAppFixture.cs
andystaples: You are correct. I went back and forth on this - but given the behavior for the storage backend setup is to default to AzureStorage if that env var is missing, I figured we were already making assumptions. Hopefully the warning is enough to help diagnose issues like the one you mentioned

test/e2e/Tests/Fixtures/IsolatedFunctionAppProcess.cs
andystaples: Saw your approval - had to fix a bug anyway, so I went ahead and removed the FunctionAppProcess variants and inlined that logic. There is a note there that it needs to be improved, which I'll do once I merge this PR and start merging my second branch with the other tests back into these changes. 
I left ILanguageLocalizer in, for now, though. Let's also address that concern in the next PR

test/e2e/Tests/Fixtures/FunctionAppFixture.cs
bachuv: So if "e2e-mssql" runs and passes, then is there a chance that the tests actually ran with the Azure Storage backend? Where would I see the warning that the environment variable is missing in these scenarios?

cgillum (approved):

PR #3136, opened 2025-06-23
Support Token Credential via HTTP feature at .NET-isolated Durable Functions 

### LINKED ISSUE #2948: TokenSource parameter in not available in DurableHttpRequest constructor in Azure Durable function in Isolated mode
The class DurableHTTPRequest does not support managed identity token source when using Azure Durable function in Isolated mode. 
https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/Worker.Extensions.DurableTask/HTTP/DurableHttpRequest.cs

But it is supported in in-process mode.
https://github.com/Azure/azure-functions-durable-extension/blob/4714ad1b0f4534a28d37a6871847543b8d43abc8/src/WebJobs.Extensions.DurableTask/DurableHttpRequest.cs

Please provide the solution for the same. If Isolated mode has lesser features than In-Process why is it recommended to upgrade it?

nytian:

resolves #2948 
This PR adds support for token-based authentication in HTTP calls.

  - Added the `ITokenSource` interface
  - Added `ManagedIdentityTokenSource`  and `ManagedIdentityOptions` .
  - Added `TokenSourceConverter` for JSON serialization of `ITokenSource` instances.
  - Extended the `DurableHttpRequest` class to support token source and added a new overload for `CallHttpAsync` in `TaskOrchestrationContextExtensionMethods`.
  - I wasn't able to add an E2E test for this scenario because DefaultAzureCredential isn't supported in GitHub CI. Instead, I added a unit test for the serialization converter, as well as a test to verify that the raw input passed from `Worker.Extensions.DurableTask` can be correctly deserialized back into a `DurableHttpRequest` in `WebJobs.Extensions.DurableTask`.

### Pull request checklist


test/Common/DurableHttpTests.cs
```diff
@@ -151,8 +151,8 @@ public void SerializeManagedIdentityOptions()
     }
    },
   ""asynchronousPatternEnabled"": true,
-  ""retryOptions"": null,
-  ""timeout"": null
+  ""retryOptions"": null,
+  ""timeout"": null
 }";
```
nytian, line 156: I tried, but adding the new test still triggers these formatting changes. Since it's just formatting, I think it should be fine.

src/Worker.Extensions.DurableTask/ITokenSource.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.DurableTask
+{
+    /// <summary>
+    /// Implementations of this interface can be used to provide authorization tokens for outbound HTTP requests.
+    /// This calss is the same as the one in WebJobs.Extensions.DurableTask/ITokenSource.cs
+    /// The implementation is kept in sync to ensure compatibility.
+    /// </summary>
+    public interface ITokenSource
```
jviau: Including this as a public contract is misleading. No custom implementation of this will ever be usable on the WebJobs side (which is where it is actually used), because whatever assembly that implements it will not be loaded there.

This limitation for out of proc with the current HTTP polling approach was called out when we first ported it.

cgillum: As an alternative, what about an abstract class with an internal constructor to prevent customers from creating their own implementations?

src/Worker.Extensions.DurableTask/HTTP/DurableHttpRequest.cs
```diff
@@ -35,6 +35,38 @@ public DurableHttpRequest(HttpMethod method, Uri uri, IDictionary<string, String
...
+    /// <param name="tokenSource">Token source for authentication.</param>
+    /// <param name="asynchronousPatternEnabled">Specifies whether the DurableHttpRequest should handle the asynchronous pattern.</param>
+    /// <param name="timeout">TimeSpan used for HTTP request timeout.</param>
+    /// <param name="httpRetryOptions">Retry options used for the HTTP request.</param>
+    public DurableHttpRequest(
+        HttpMethod method,
+        Uri uri,
+        IDictionary<string, StringValues>? headers = null,
+        string? content = null,
+        ManagedIdentityTokenSource? tokenSource = null,
```
cgillum: `ManagedIdentityTokenSource` is just one of many different types of token sources that we may want to support. Rather than accepting a specific concrete type, it would be better to accept an interface or abstract base type. That would allow us to support additional token source types in the future (workload identity credentials, Azure CLI credentials, etc.).

src/Worker.Extensions.DurableTask/ManagedIdentityOptions.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// <param name="authorityHost">The host of the Azure Active Directory authority.</param>
+        /// <param name="tenantId">The tenant id of the user to authenticate.</param>
+        public ManagedIdentityOptions(Uri? authorityHost = null, string? tenantId = null)
+        {
+            this.AuthorityHost = authorityHost;
+            this.TenantId = tenantId;
+        }
+
+        /// <summary>
+        /// The host of the Azure Active Directory authority. The default is https://login.microsoftonline.com/.
```
cgillum: Azure Active Directory has been renamed to Entra ID.
```suggestion
        /// The host of the Entra ID authority. The default is https://login.microsoftonline.com/.
```

src/Worker.Extensions.DurableTask/ManagedIdentityTokenSource.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// Token Source implementation for Azure Managed Identities.
+    /// This calss is the same as the one in WebJobs.Extensions.DurableTask/ManagedIdentityTokenSource.cs
+    /// The implementation is kept in sync to ensure compatibility.
+    /// </summary>
+    public class ManagedIdentityTokenSource
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ManagedIdentityTokenSource"/> class.
+        /// </summary>
+        /// <param name="resource">The Azure Active Directory resource identifier of the web API being invoked.</param>
```
cgillum:
```suggestion
        /// <param name="resource">The Entra ID resource identifier of the web API being invoked.</param>
```

src/Worker.Extensions.DurableTask/ManagedIdentityOptions.cs
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.DurableTask
```
cgillum: We should use file-scoped namespaces consistently in all our source files.
```suggestion
namespace Microsoft.Azure.Functions.Worker.Extensions.DurableTask;
```

src/Worker.Extensions.DurableTask/Worker.Extensions.DurableTask.csproj
```diff
@@ -29,7 +29,7 @@
     <AssemblyOriginatorKeyFile>..\..\sign.snk</AssemblyOriginatorKeyFile>
 
     <!-- Version information -->
-    <VersionPrefix>1.5.0</VersionPrefix>
+    <VersionPrefix>1.5.1</VersionPrefix>
```
cgillum: Since we're adding new public APIs, this should be a minor version update. However, it looks like the latest version is still `1.4.0`, so I think we can leave this as `1.5.0`.

test/Worker.Extensions.DurableTask.Tests/TokenSourceConverterTests.cs
```diff
@@ -0,0 +1,61 @@
...
+    public class TokenSourceConverterTests
+    {
+        [Fact]
+        // Test that a ManagedIdentityTokenSource instance can be serialized correctly. 
+        public void TokenSourceConverter_SerializeManagedIdentityTokenSource_ProducesCorrectJson()
+        {
+            var tokenSource = new ManagedIdentityTokenSource("https://management.core.windows.net/.default");
+            var options = new JsonSerializerOptions();
+            options.Converters.Add(new TokenSourceConverter());
+            var converter = new TokenSourceConverter();
```
cgillum: Remove unused variable

test/Worker.Extensions.DurableTask.Tests/TokenSourceConverterTests.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            var managedIdentityOptions = new ManagedIdentityOptions
+            {
+                AuthorityHost = new Uri("https://login.microsoftonline.com"),
+                TenantId = "test-tenant-id"
+            };
+            var tokenSource = new ManagedIdentityTokenSource("https://graph.microsoft.com/.default", managedIdentityOptions);
+            var options = new JsonSerializerOptions();
+            options.Converters.Add(new TokenSourceConverter());
+            var converter = new TokenSourceConverter();
```
cgillum: Removed unused variables

src/Worker.Extensions.DurableTask/HTTP/DurableHttpRequest.cs
```diff
@@ -35,6 +35,38 @@ public DurableHttpRequest(HttpMethod method, Uri uri, IDictionary<string, String
...
+    /// <param name="httpRetryOptions">Retry options used for the HTTP request.</param>
+    public DurableHttpRequest(
+        HttpMethod method,
+        Uri uri,
+        IDictionary<string, StringValues>? headers = null,
+        string? content = null,
+        TokenSource? tokenSource = null,
+        bool asynchronousPatternEnabled = false,
+        TimeSpan? timeout = null,
+        HttpRetryOptions? httpRetryOptions = null)
```
cgillum: It looks like some of the parameters have been reordered compared to the overload above. For example, you've added `content` and `tokenSource` between `headers` and `asynchronousPatternEnabled`, creating an inconsistency. I suggest reordering these parameters as follows:

- method
- uri
- headers
- asynchronousPatternEnabled
- timeout
- content
- httpRetryOptions
- tokenSource

Alternatively, consider removing this overload entirely and instead just rely on customers to set the properties directly rather than using the constructor. Constructors are best used for common cases and for required properties.

src/Worker.Extensions.DurableTask/ManagedIdentityOptions.cs
```diff
@@ -0,0 +1,38 @@
...
+/// Configuration options for Managed Identity.
+/// This calss is the same as the one in WebJobs.Extensions.DurableTask/ManagedIdentityOptions.cs.
+/// The implementation is kept in sync to ensure compatibility.
+/// </summary>
+public class ManagedIdentityOptions
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ManagedIdentityOptions"/> class.
+    /// </summary>
+    /// <param name="authorityHost">The host of the Azure Active Directory authority.</param>
```
cgillum:
```suggestion
    /// <param name="authorityHost">The host of the Entra ID authority.</param>
```

src/Worker.Extensions.DurableTask/ManagedIdentityOptions.cs
```diff
@@ -0,0 +1,38 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.DurableTask;
+
+/// <summary>
+/// Configuration options for Managed Identity.
+/// This calss is the same as the one in WebJobs.Extensions.DurableTask/ManagedIdentityOptions.cs.
```
cgillum: I think we should remove this comment about the WebJobs code. This will not be useful information for .NET Isolated customers, and it might confuse them.

src/Worker.Extensions.DurableTask/ManagedIdentityTokenSource.cs
```diff
@@ -0,0 +1,54 @@
...
+
+using System;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.DurableTask;
+
+/// <summary>
+/// Token Source implementation for Azure Managed Identities.
+/// This calss is the same as the one in WebJobs.Extensions.DurableTask/ManagedIdentityTokenSource.cs
+/// The implementation is kept in sync to ensure compatibility.
```
cgillum: Same comment about removing documentation that references WebJobs code.

src/Worker.Extensions.DurableTask/TokenSource.cs
```diff
@@ -0,0 +1,30 @@
...
+
+/// <summary>
+/// Abstract base class for token sources that can be inherited for future token credential types.
+/// This class provides a common foundation for different token source implementations.
+/// </summary>
+public abstract class TokenSource
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="TokenSource"/> class.
+    /// This constructor is internal to prevent customers from constructing their own token source types.
```
cgillum: This sentence doesn't belong in user-facing documentation comments. Talking about customers should only be done using regular internal facing comments.

src/Worker.Extensions.DurableTask/Worker.Extensions.DurableTask.csproj
```diff
@@ -29,7 +29,7 @@
     <AssemblyOriginatorKeyFile>..\..\sign.snk</AssemblyOriginatorKeyFile>
 
     <!-- Version information -->
-    <VersionPrefix>1.5.0</VersionPrefix>
+    <VersionPrefix>1.6.0</VersionPrefix>
```
cgillum, line 32: The current version is 1.4.0, so we don't need to double-increment this version.

cgillum (changes requested):

src/Worker.Extensions.DurableTask/Worker.Extensions.DurableTask.csproj
nytian: oh because we will release v1.5.0 this week, but this PR won't catch it

cgillum (approved):

PR #3145, opened 2025-06-30
Finish implementing E2E tests for PowerShell

andystaples:
Adds the remainder of the E2E test implementations for PowerShell
- Comments which tests are not supported due to feature gaps
- Improves logic in LanguageLocalizer and FunctionAppProcess

### Pull request checklist


test/e2e/Tests/Tests/ExternalEventTests.cs
```diff
@@ -45,21 +44,30 @@ public async Task RaiseExternalEventTests()
         string responseContent = await resendEventResponse.Content.ReadAsStringAsync();
 
         // Verify the returned exception contains the correct information. 
-        Assert.Contains("FailedPrecondition", responseContent);
-        Assert.Contains("The orchestration instance with the provided instance id is not running.", responseContent);
+        // In dotnet-isolated, this is the StatusCode of the RPC exception. 
+        // In other languages, it is the exception type
+        Assert.Contains(fixture.functionLanguageLocalizer!.GetLocalizedStringValue("ExternalEvent.CompletedInstance.ErrorName"), responseContent);
```
cgillum: nit: we generally shouldn't be using the `!` in new code. Is there a way we can make this field non-nullable?

test/e2e/Tests/Tests/TerminateOrchestratorTests.cs
```diff
@@ -86,18 +84,22 @@ public async Task TerminateTerminatedOrchestration_ShouldFail()
...
-        Assert.Contains($"InvalidOperationException: Cannot terminate the orchestration instance {instanceId} because instance is in the Terminated state.", terminateAgainResponseMessage);
+        Assert.Contains(fixture.functionLanguageLocalizer!.GetLocalizedStringValue("TerminateTerminatedInstance.FailureMessage", instanceId), terminateAgainResponseMessage);
 
         // Give some time for Core Tools to write logs out
         Thread.Sleep(500);
 
-        Assert.Contains(this.fixture.TestLogs.CoreToolsLogs, x => x.Contains("Cannot terminate orchestration instance in the Terminated state.") &&
+        // For some reason, PowerShell does not log these warnings - instead the status code is 410 (Gone) with no log
+        // when the instance is completed
+        if (fixture.functionLanguageLocalizer.GetLanguageType() == LanguageType.DotnetIsolated)
```
cgillum: I'm curious whether this is specific to .NET Isolated or whether it's specific to PowerShell. Given that the comment calls out PowerShell specifically, should we change the if-check to compare against PowerShell instead of .NET Isolated? For example:
```suggestion
        if (fixture.functionLanguageLocalizer.GetLanguageType() != LanguageType.PowerShell)
```

When we add support for more languages (e.g. Python), then we'll know for sure when these checks run, and we can update both the code and the if-check at the same time, as appropriate.

test/e2e/Tests/Tests/VersioningTests.cs
```diff
@@ -25,7 +25,7 @@ public VersioningTests(FunctionAppFixture fixture, ITestOutputHelper testOutputH
     [InlineData("")] // Non-versioned/empty-versioned case.
     [InlineData("1.0")]
     [InlineData("2.0")]
-    [Trait("PowerShell", "Skip")] // Test not yet implemented in PowerShell
+    [Trait("PowerShell", "Skip")] // Orchestration versioning not yet implemented in PowerShell
```
cgillum: It's my understanding that @AnatoliB completed the work for versioning in PowerShell, and that we're just waiting for the standalone SDK update to be deployed to the PS Gallery. Is that correct?

test/e2e/Tests/Helpers/DurableHelpers.cs
```diff
@@ -9,7 +9,7 @@ internal class DurableHelpers
 {
     static readonly HttpClient httpClient = new HttpClient();
 
-    static readonly List<string> finalStates = new List<string>()
+    static readonly List<string> _finalStates = new List<string>()
```
cgillum: I thought I remembered that we moved away from the `_` prefix in the last PR?

test/e2e/Apps/BasicPowerShell/HelloCities_HttpStart/run.ps1
```diff
@@ -0,0 +1,14 @@
+#
+# Copyright (c) Microsoft. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
cgillum, line 3: I noticed some of our run.ps1 files have this copyright header and others don't. Seems like we should be consistent about this.

test/e2e/Apps/BasicPowerShell/HelloCities_HttpStart/run.ps1
```diff
@@ -0,0 +1,14 @@
...
+#
+
+using namespace System.Net
```
cgillum, line 6: Is this `System.Net` using statement needed? I didn't notice any code below that would require it.

test/e2e/Apps/BasicPowerShell/LongActivity/run.ps1
```diff
@@ -0,0 +1,4 @@
+param($name)
+
+Start-Sleep -Milliseconds 5000
```
cgillum, line 11: Dumb question: but what makes 5 seconds "long" in this context? Can we add a comment explaining the significance of this duration a bit more?

test/e2e/Apps/BasicPowerShell/SendExternalEvent_HttpStart/run.ps1
```diff
@@ -0,0 +1,29 @@
...
+    StatusCode = [HttpStatusCode]::OK
+    Body = ""
+}
+
+try {
+    Send-DurableExternalEvent -InstanceId $instanceId -EventName "Approval"
+    $response.Body = "External event sent to $instanceId."
+}
+catch {
+    if ($_.Exception.GetType().Name -eq "HttpResponseException") {
```
cgillum, line 22: I might be missing something obvious, but why do we need a different value for `$response.Body` depending on whether this exception type is HttpResponseException?

test/e2e/Apps/BasicPowerShell/SimulatedWorkActivity/run.ps1
```diff
@@ -0,0 +1,5 @@
+param($sleepMs)
+
+Write-Information ("Sleeping for " + "$sleepMs" + "ms.")
```
cgillum: The syntax here is not what I'd normally expect. Isn't the below more conventional?
```suggestion
Write-Information "Sleeping for ${sleepMs}ms."
```

test/e2e/Apps/BasicPowerShell/SimulatedWorkActivity/run.ps1
```diff
@@ -0,0 +1,5 @@
...
+Start-Sleep -Milliseconds $sleepMs
+("Slept for " + "$sleepMs" + "ms.")
```
cgillum: Similar question, and I'm not sure why this one doesn't include `Write-Information`. I'd expect something like the below:
```suggestion
Write-Information "Slept for ${sleepMs}ms."
```

cgillum (commented):
One of the things I love about this PR is how much easier it is to see the parity differences between languages. This is super insightful.

A few comments below:

test/e2e/Tests/Tests/VersioningTests.cs
andystaples: Actually, it already is deployed, so we can probably add this test now. I'll work with him on it

test/e2e/Tests/Helpers/DurableHelpers.cs
andystaples: Yeah that was before I forked the first commit for this branch off and looks like I didn't catch this one in the rebase
I'll fix it

test/e2e/Apps/BasicPowerShell/SendExternalEvent_HttpStart/run.ps1
andystaples: Technically we don't - HttpResponseException is the expected result in the test case when you try to send an event to an invalid instance ID. We check this output specifically. The second block is for some other exception type, which would be a test failure, but it's still nice to return a status code and message. I could change the status code to SERVER_ERROR or something different from BAD_REQUEST though

test/e2e/Tests/Tests/ExternalEventTests.cs
andystaples: We sure can! Moved the init to constructor, made non-nullable, and removed !

test/e2e/Tests/Tests/TerminateOrchestratorTests.cs
andystaples: Yes - I suspect the same will be true for the other languages, but given the comment language and the fact that we _want_ to know this for sure, I have inverted the check as suggested. This will cause the test to fail on a naive Python implementation if it is not logged, and we can change this code again then if that happens.

test/e2e/Apps/BasicPowerShell/HelloCities_HttpStart/run.ps1
andystaples: I think I got all of them now

test/e2e/Apps/BasicPowerShell/HelloCities_HttpStart/run.ps1
andystaples: Nope! Removed

test/e2e/Apps/BasicPowerShell/LongActivity/run.ps1
andystaples: Comment added

test/e2e/Apps/BasicPowerShell/SimulatedWorkActivity/run.ps1
andystaples: Normally, the syntax would be 
```powershell
"Sleeping for $sleepMs ms." 
``` 
but given the test is specifically checking for an output like "Sleeping for 100ms." (no space between the number and "ms.", the string would look like 
```powershell
"Sleeping for $sleepMsms."
```
which would not produce the expected result. 
```powershell
"Sleeping for ${$sleepMs}ms."
```
should work though! Changed it to that

test/e2e/Apps/BasicPowerShell/SimulatedWorkActivity/run.ps1
andystaples: The write-information line is for logging, the final line is the activity function's result. PowerShell will capture objects sent to the output pipeline of the script (anything that would normally go to stdout in the console without the use of Write-Information, Write-Host, etc) and instead will use those values as the activity result. 
I'll add a comment to make this more clear

cgillum: Ah, right, I forgot about that. :)

test/e2e/Tests/Tests/VersioningTests.cs
AnatoliB: I know I should have added these comments to the previous PR, but it is already merged, so I'll add them here:

- The ability to pass a version to the orchestration creation API is not implemented for non-.NET languages at this point. `defaultVersion` in `host.json` is the only way, and it requires an app restart. So, we'll need to either implement this capability, or make the tests modify `host.json` and restart the app. Accidentally, passing a version to the API _may_ already be possible when using the standalone PowerShell SDK, but this is not something that we intentionally enabled and tested yet.
- The "non-versioned case" comment for the `version == null` case is incorrect: it's in fact versioned, and it's the primary use case.

cgillum (approved):

PR #3147, opened 2025-07-02
Add RestartAsync  at Durable Functions Isolated

### LINKED ISSUE #2903: Restart not support in Isolated mode
### What version of .NET does your existing project use?

.NET 6

### What version of .NET are you attempting to target?

.NET 8

### Description

after migrating to isolated mode, restart feature is not work, is there any solution for this?
image

### Project configuration and dependencies

_No response_

### Link to a repository that reproduces the issue

_No response_

nytian:

resolves #2903 

* Added the new `RestartAsync` support from `DurableTaskClient`. 
* Updated the HTTP management payload to include a `RestartPostUri` for orchestration restart. 
* Add E2E tests for this scenario.

_Note: This PR won't work unless v1.14.0 of dotnet sdk including https://github.com/microsoft/durabletask-dotnet/pull/456 released._

Related PR:

- https://github.com/microsoft/durabletask-dotnet/pull/456
- https://github.com/microsoft/durabletask-protobuf/pull/46

### Pull request checklist


src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
```diff
@@ -273,6 +275,41 @@ private static ObjectSerializer GetObjectSerializer(HttpResponseData response)
...
+        CancellationToken cancellation = default)
+    {
+        // Get the status of the existing instance, including input
+        OrchestrationMetadata? status = await client.GetInstancesAsync(instanceId, getInputsAndOutputs: true, cancellation) 
+            ?? throw new ArgumentException($"An orchestration with the instanceId {instanceId} was not found.");
+        
+        // Deserialize the input.
+        string orchestratorName = status.Name;
+        string? input = status.SerializedInput != null
+            ? System.Text.Json.JsonSerializer.Deserialize<string>(status.SerializedInput)
```
jviau: This won't work for all cases. Customer's may be using a serializer other than STJ. They may also have STJ customizations.

We shouldn't add a public API like this with such a limitation.

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
```diff
@@ -273,6 +275,41 @@ private static ObjectSerializer GetObjectSerializer(HttpResponseData response)
...
 
+    /// <summary>
+    /// Restarts an existing orchestration instance with the original input.
+    /// </summary>
+    /// <param name="client">The <see cref="DurableTaskClient"/>.</param>
+    /// <param name="instanceId">The ID of the orchestration instance to restart.</param>
+    /// <param name="restartWithNewInstanceId">If true, starts with a new instance ID; otherwise, uses the same instance ID.</param>
+    /// <param name="cancellation">A token that signals if the operation should be canceled.</param>
+    /// <returns>The new instance ID.</returns>
+    public static async Task<string> RestartAsync(
```
jviau: I am not entirely sold on this public API. It leverages purely public APIs internally, so a customer could implement it themselves. Meaning this is purely a "convenience" API. To me, convenience APIs must meet a much higher bar to be included.

- Are we actually unblocking any customer scenario here that is impossible to do without this API?
- Why is this added as an extension in the durable functions repo and not directly in `durabletask-dotnet`?

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
```diff
@@ -273,6 +275,41 @@ private static ObjectSerializer GetObjectSerializer(HttpResponseData response)
...
+        
+        // Deserialize the input.
+        string orchestratorName = status.Name;
+        string? input = status.SerializedInput != null
+            ? System.Text.Json.JsonSerializer.Deserialize<string>(status.SerializedInput)
+            : null;
+        
+        if (restartWithNewInstanceId)
+        {
+            return await client.ScheduleNewOrchestrationInstanceAsync(orchestratorName, input, null, cancellation);
```
jviau: This actually isn't restarting the orchestration. The behavior here will entirely depend on the backend. Some may throw if the orchestration already exists. Some may silently ignore this call. Others may actually restart the orchestration.

jviau (dismissed):

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
cgillum: Agreed this is just a convenience API. However, we need it for parity with in-proc. Given it's not adding a ton of value, isn't it better to limit it to DF rather than adding it to the broader `durabletask-dotnet`?

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
cgillum: +1. It's also wasteful to deserialize and reserialize the input, even if we could preserve the custom serialization settings.

A better way to handle this case is to do something like introduce a [`RawInput`](https://github.com/Azure/durabletask/blob/fafc8372e3f15de2eb09ff250575804f8e0d224d/src/DurableTask.Core/Serializing/Internal/RawInput.cs#L26) abstraction like we did for DTFx and use it here. That way we can use the previous input as-is without needing to interpret it.

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
cgillum: @nytian one thing to consider is making this a first-class API and call into the durable extension's implementation of this API ([`IDurableOrchestrationClient.RestartAsync`](https://github.com/Azure/azure-functions-durable-extension/blob/f943321ca9722a033b9416d18e879509f8c430b2/src/WebJobs.Extensions.DurableTask/ContextInterfaces/IDurableOrchestrationClient.cs?plain=1#L330-L338)). That would allow us to apply any backend-specific handling, which we can't do in the SDK layer. That would also solve the input serialization problem mentioned earlier.

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
jviau: I don't mind making this a first-class API. It would actually be preferred, due to the backend differences mentioned in my other comment. This way only backends which support restart can implement this, all others could opt to throw instead.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -1158,6 +1159,14 @@ async Task<string> IDurableOrchestrationClient.RestartAsync(string instanceId, b
                 throw new ArgumentException($"An orchestrastion with the instanceId {instanceId} was not found.");
             }
 
+            bool isInstaceNotCompleted = status.RuntimeStatus == OrchestrationRuntimeStatus.Running || status.RuntimeStatus == OrchestrationRuntimeStatus.Pending || status.RuntimeStatus == OrchestrationRuntimeStatus.Suspended;
```
cgillum: nit: Please wrap this into multiple lines so that it can be read without scrolling.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -12,6 +12,7 @@
 using DurableTask.Core;
 using DurableTask.Core.Entities;
 using DurableTask.Core.History;
+using Microsoft.AspNetCore.Hosting.Server;
```
cgillum: Can this be removed?

test/e2e/Apps/BasicPython/extensions.csproj
```diff
@@ -11,7 +11,7 @@
   <ItemGroup>
     <!-- This reference will be dynamically updated by build-e2e-tests.ps1 as part of test setup. -->
     <!-- Do not commit changes to this line unless necessary -->
-    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="3.3.0" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="3.5.0" />
```
cgillum, line 14: Not related to this PR, but should these be project references instead of package references so that we don't have to remember to update them? @andystaples

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
nytian: Resolved since we’ve made it a first-class API, we no longer need to serialize the input but instead rely directly on the orchestration status's input.

cgillum (approved):
LGTM - no blockers - just a few minor things.

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
nytian: Resolved as it's a first party API now

src/Worker.Extensions.DurableTask/DurableTaskClientExtensions.cs
nytian: To avoid inconsistency across backends, I added a check in DurableClient.RestartAsync. If an orchestration is not yet completed and a restart is requested with the same instance ID, we will now throw an InvalidOperationException.

test/e2e/Apps/BasicPython/extensions.csproj
nytian: I can open a separate PR for this, or @andystaples if you’d prefer. For now, it’s auto-generated by running build-e2e-test.ps1 locally.

jviau (commented):
resetting review

bachuv (approved):

PR #3150, opened 2025-07-10
Support for non-DI entity constructor arguments

cgillum:
I noticed that we're missing a feature from in-proc entities, which is the ability to pass constructor arguments to entities using the `DispatchAsync` method.  This feature is documented [here](https://learn.microsoft.com/azure/azure-functions/durable/durable-functions-dotnet-entities?pivots=isolated) and has an in-proc test demonstrating the usage [here](https://github.com/Azure/azure-functions-durable-extension/blob/48da8ad1ba1ace6c3991ac2c3573b8e03704fe6c/test/FunctionsV2/TestEntityWithDependencyInjectionHelpers.cs?plain=1#L67). This PR adds the feature back.

For example, in in-proc, it's possible to do this:

```csharp
[FunctionName(nameof(BlobBackedEnvironment))]
public static Task BlobEnvironmentFunction(
    [EntityTrigger] IDurableEntityContext context,
    [Blob(BlobContainerPath, System.IO.FileAccess.Read)] CloudBlobContainer blob)
{
    // pass the blob to the entity as a constructor argument
    return context.DispatchAsync<BlobBackedEnvironment>(blob);
}
```

The entity class implementation can then be implemented as follows:

```csharp
[JsonObject(MemberSerialization = MemberSerialization.OptIn)]
public class BlobBackedEnvironment : IWritableEnvironment
{
    private readonly INameResolver nameResolver;
    private readonly CloudBlobContainer blobContainer;

    public BlobBackedEnvironment(INameResolver nameResolver, CloudBlobContainer blobContainer)
    {
        this.nameResolver = nameResolver; // comes from dependency-injection
        this.blobContainer = blobContainer; // comes from the DispatchAsync arguments
    }

    // ...
}
```

For .NET Isolated, we don't have any `DispatchAsync<T>()` overload that takes a parameter that can be used for populating constructor arguments. That's what this PR adds, creating parity between in-proc and .NET Isolated.

### Pull request checklist

    * [ ] Otherwise: I've added my notes to `release_notes.md` (SEEMS LIKE WE DON'T DO THIS ANYMORE?)


test/e2e/Tests/E2ETests.csproj
```diff
@@ -5,6 +5,7 @@
     <LangVersion>latest</LangVersion>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
+    <RootNamespace>Microsoft.Azure.Durable.Tests.DotnetIsolatedE2E</RootNamespace>
```
nytian, line 8: what is this new line is for?

test/e2e/Tests/GlobalSuppressions.cs
```diff
@@ -0,0 +1,13 @@
+﻿// This file is used by Code Analysis to maintain SuppressMessage
+// attributes that are applied to this project.
+// Project-level suppressions either have no target or are given
+// a specific target and scoped to a namespace, type, member, etc.
+
+using System.Diagnostics.CodeAnalysis;
+
+[assembly: SuppressMessage(
```
nytian, line 8: I am not quite familiar with Code Analysis but why add this file twice at different folder?

nytian (approved):
Just left small questions.

test/e2e/Tests/GlobalSuppressions.cs
cgillum: They are for different projects. One is for the function app project, and one is for the test project.

test/e2e/Tests/E2ETests.csproj
cgillum: I added it so that the IDE knows what namespace to use when I add new files. By default, it uses the name of the project, but the convention in this project was to instead use `Microsoft.Azure.Durable.Tests.DotnetIsolatedE2E`. By adding this, people no longer need to manually update the namespace in newly created files.

test/e2e/Apps/BasicDotNetIsolated/ClassBasedEntities.cs
```diff
@@ -0,0 +1,82 @@
...
+using Microsoft.DurableTask.Entities;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Azure.Durable.Tests.E2E;
+
+public class ClassBasedEntities
+{
+    // HTTP trigger to start the orchestration and wait for it to complete.
+    [Function(nameof(StartClassBasedEntityOrchestration))]
+    public static async Task<HttpResponseData> StartClassBasedEntityOrchestration(
```
andystaples: Any particular reason we defined a new HTTP starter function rather than using the generic StartOrchestration starter?

test/e2e/Tests/Tests/ClassBasedEntityTests.cs
```diff
@@ -0,0 +1,45 @@
...
+        string statusQueryGetUri = await DurableHelpers.ParseStatusQueryGetUriAsync(response);
+        await DurableHelpers.WaitForOrchestrationStateAsync(statusQueryGetUri, "Completed", 30);
+
+        // The output of the orchestration should contain the expected string that contains the entity's state.
+        // The entity state is a simple string that shows whether the injected services are available.
+        DurableHelpers.OrchestrationStatusDetails orchestrationDetails =
+            await DurableHelpers.GetRunningOrchestrationDetailsAsync(statusQueryGetUri);
+
+        this.output.WriteLine(
+            $"Orchestration {orchestrationDetails.RuntimeStatus}. Output = {orchestrationDetails.Output}");
```
andystaples: Is this writeline needed?

cgillum: I guess not. If the test fails, then it will tell me what part of the string didn't match.

test/e2e/Apps/BasicDotNetIsolated/ClassBasedEntities.cs
cgillum: Ah, good catch.

test/e2e/Tests/Tests/ClassBasedEntityTests.cs
andystaples: Assert.Equal will do this for you in the test output

andystaples (approved):

PR #3154, opened 2025-07-14
Extended Sessions for Isolated (Orchestrations)

### LINKED ISSUE #2905: What is the Durable function "isolated" mode alternative for "extendedSessionsEnabled" ?
I have a .NET 8 based "isolated" mode durable function which implements the fan out/fan in pattern.
image

Expected F2 count is 100k activity function calls, which is getting executed in batch of 20 F2 calls.
image

Delay between each batch getting increased along with the execution and function app memory also increasing consistently.
image

**This was resolved by enabling extendedSession in "in-proc" mode.**. By enabling the extended session, 20 hours of execution was reduced to 1.5 hours.
`{
  "extensions": {
    "durableTask": {
      "extendedSessionsEnabled": true,
      "extendedSessionIdleTimeoutInSeconds": 30
    }
  }
}`

But in "isolated" mode, extended sessions are not supported.
[https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-perf-and-scale#language-runtime-considerations](url)
image

What is the solution/work around in "isolated" mode?

sophiatev:
This PR introduces enabling extended sessions for orchestrations in the .NET isolated framework. The way this is achieved is via an `IMemoryCache` which maintains the extended sessions state in memory. The main changes in this PR are:

- `FunctionsWorkerApplicationBuilderExtensions`: Create this cache as a singleton upon worker startup
- `FunctionsWorkerApplicationBuilderExtensions`: Pass the cache to the dotnet SDK when executing the orchestration request. 
- `OutOfProcMiddlware`: Extracts information about extended sessions passed in the `dispatchContext` from DT.Core to send to the worker. It also checks for the special edge case of a worker needing an orchestration history when none was attached to the request (`OrchestratorResponse.requiresHistory` is true), and throws a `SessionAbortedException`, in which case DT.Core will end the session and retry the work item, this time indicating that a history should be attached with the request. 
- `RemoteOrchestratorConfiguration` and `RemoteOrchestratorContext`: New fields added to pass additional information to the worker related to extended sessions
- `OrchestrationTriggerAttributeBindingProvider`: Modified to only include past events in the orchestration request sent to the worker if the host indicates that this is necessary. (Past events are not included if this orchestration request is within an extended session, and not its first execution). 
- `DurableTaskOptions`: Modified to not throw an exception in the case that extended sessions are enabled for .NET isolated (but an exception is still thrown for other OOProc languages). 

Other PRs: 
- [DT.Core](https://github.com/Azure/durabletask/pull/1232)
- [DotNet SDK](https://github.com/microsoft/durabletask-dotnet/pull/449)
- [Protobuf](https://github.com/microsoft/durabletask-protobuf/pull/47)

### Open questions
Is there a way to pass information from `DurableTaskOptions` to the worker upon startup? Specifically, the [`FunctionsWorkerApplicationBuilderExtensions`](https://github.com/Azure/azure-functions-durable-extension/pull/3154/files#diff-511b4afa2b8fca921550167126b4179908c8e31f2e928039d42f05bb248bf26e) class. Ideally, we would only make the cache for extended sessions if they are enabled (`DurableTaskOptions.ExtendedSessionsEnabled` is true), and set the cache `ExpirationScanFrequency` to be some function of the user-specified `DurableTaskOptions.ExtendedSessionIdleTimeoutInSeconds` (maybe something along the lines of `ExtendedSessionIdleTimeoutInSeconds` / 5). 

### Design Callout
There are several avenues by which the worker can inform the host that it needs an orchestration history in the case that the worker has ended the extended session before the host. The worker will evict an extended session after the user-specified `extendedSessionIdleTimeoutInSeconds` expires. More specifically, what this means is that if the extended session is not accessed within that timeframe, the worker will evict it from the cache. In the meantime, it has to send the result of the orchestration work item back to the host, the host has to process this and commit it to storage, then wait for new orchestration messages, then send the worker the next work item once new messages arrive. From the host's perspective, the `extendedSessionIdleTimeoutInSeconds` applies only to the amount of time we wait for new orchestration messages to arrive before ending the extended session. All that to say, there could perhaps be an appreciable number of situations where the worker ends the extended session before the host and will require a history.

The current approach is perhaps the simplest - have the worker inform the host via the `requiresHistory` flag that it needs a history, in which case the host throws a `SessionAbortedException` from `OutOfProcMiddleware`. DT.Core already has all the logic to handle a `SessionAbortedException` and retry the work item. The cons of this approach are that the `SessionAbortedException` surfaces to the logs and may alarm customers. There is also an added delay that comes from having DT.Core abort the orchestration work item entirely and try it again later in this approach. All that being said, perhaps the number of situations where this occurs could be reduced by just advising customers to up their `extendedSessionIdleTimeoutInSeconds` count in the isolated model. 

An alternative is to have the worker call [StreamInstanceHistory](https://github.com/microsoft/durabletask-protobuf/blob/fd9369c6a03d6af4e95285e432b7c4e943c06970/protos/orchestrator_service.proto#L702-L703) in the case that it has since ended the extended session and needs a history. This will not surface an exception to the customer, and will lead to less of a delay in processing the work item. But it could be much more complicated to implement as I am not sure how to get a client over to the [`GrpcOrchestrationRunner`](https://github.com/microsoft/durabletask-dotnet/blob/main/src/Worker/Grpc/GrpcOrchestrationRunner.cs). to accomplish this, and may require additional edge-case handling for all the network issues that could arise. 

### Manual testing done thus far
- If extended sessions are not enabled, everything still proceeds as expected (the orchestration history is replayed upon every execution)
- If an extended session expires, a history is sent along with the orchestration request to the worker. The worker replays the orchestration and then saves the state in a new extended session (state is only saved if the orchestration is not completed)
- Upon completion of an orchestration, the extended session state is evicted from the cache
- If the worker ends the extended session before the host does (this should be very rare but can happen if e.g. there is a network delay when sending the orchestration request to the worker, or something along those lines), then it will throw a `SessionAbortedException`. The work item will then be retried again.
- Other OOProc scenarios still work as expected (they do not have extended sessions enabled so the orchestration is replayed upon every request. If they attempt to enable extended sessions an error occurs upon startup indicating that the extended session feature does not work for these other languages).

### Performance testing
Two scenarios were run in in-process with extended sessions enabled/disabled, and in isolated with extended sessions enabled/disabled. Multiple trials were run, and the time it took to complete the orchestration was recorded as well as the number of times the history was loaded in each of these settings (with `extendedSessionIdleTimeoutInSeconds` set to 30 seconds). The two scenarios tested were

- Fan out to a large number of tasks (30,000) and then fan-in. For both isolated and in-process, without extended sessions this took about 35 minutes and with extended sessions around 25 minutes. In the isolated model there were consistently two history loads with ES enabled, with the second corresponding to a `SessionAbortedException` being thrown (so the worker ends the extended session before the host, and as expected, throws the exception which leads to a retry of the work item, this time with a history attached). In in-process, there was just one history load. Without ES, there were about 100 history loads in each. 
- Run a large number of sequential tasks (1000) with large payloads (10,000 character-length strings). For both isolated and in-process, without extended sessions this took about 30 minutes and with extended sessions enabled around 5 minutes. Without ES there were around 100 history loads, and with ES just 1. Interestingly, in isolated this scenario actually ran somewhat faster both with/without ES as opposed to in-process. 

resolves #2905 

### Pull request checklist


src/Worker.Extensions.DurableTask/DurableTaskFunctionsMiddleware.cs
```diff
@@ -7,20 +7,23 @@
 using Microsoft.Azure.Functions.Worker.Extensions.DurableTask.Exceptions;
 using Microsoft.Azure.Functions.Worker.Middleware;
 using Microsoft.DurableTask.Worker.Grpc;
+using Microsoft.Extensions.Caching.Memory;
```
andystaples: Why is this using statement failing?

src/Worker.Extensions.DurableTask/FunctionsWorkerApplicationBuilderExtensions.cs
```diff
@@ -59,6 +59,10 @@ public static IFunctionsWorkerApplicationBuilder ConfigureDurableExtension(this
         {
             builder.UseMiddleware<DurableTaskFunctionsMiddleware>();
         }
+        builder.Services.AddMemoryCache(options =>
+        {
+            options.ExpirationScanFrequency = TimeSpan.FromSeconds(5);
```
andystaples: Any specific reason why we picked 5 seconds? Not blocking, just curious

src/Worker.Extensions.DurableTask/DurableTaskFunctionsMiddleware.cs
sophiatev: Ah sorry I did not add the Directory.props change to this PR. Let me fix that

src/Worker.Extensions.DurableTask/FunctionsWorkerApplicationBuilderExtensions.cs
sophiatev: This is basically just a placeholder. Really what I want to do is in that "Open questions" section of the PR description:

"Is there a way to pass information from `DurableTaskOptions` to the worker upon startup? Specifically, the [`FunctionsWorkerApplicationBuilderExtensions`](https://github.com/Azure/azure-functions-durable-extension/pull/3154/files#diff-511b4afa2b8fca921550167126b4179908c8e31f2e928039d42f05bb248bf26e) class. Ideally, we would only make the cache for extended sessions if they are enabled (`DurableTaskOptions.ExtendedSessionsEnabled` is true), and set the cache `ExpirationScanFrequency` to be some function of the user-specified `DurableTaskOptions.ExtendedSessionIdleTimeoutInSeconds` (maybe something along the lines of `ExtendedSessionIdleTimeoutInSeconds` / 5)."

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -139,6 +144,11 @@ await this.LifeCycleNotificationHelper.OrchestratorStartingAsync(
                     byte[] triggerReturnValueBytes = Convert.FromBase64String(triggerReturnValue);
                     P.OrchestratorResponse response = P.OrchestratorResponse.Parser.ParseFrom(triggerReturnValueBytes);
 
+                    if (response.RequiresHistory)
+                    {
+                        throw new SessionAbortedException("The worker has since ended the extended session and needs an orchestration history to execute the orchestration request.");
```
andystaples: Some questions about this `SessionAbortedException` - 
1. Is `SessionAbortedException` retried indefinitely by Core or is there a limit?
2. Is it possible that the retry in Core happens so fast that the cache still hasn't been evicted from WebJobs, meaning that the exception will be thrown multiple times for the same task?
3. Have we observed this exception being thrown and retried in practice?
4. Any other gotchas we should document?

sophiatev: 1. Very good question, I'm not sure actually. We ultimately abandon the work item in DT.Core [here ](https://github.com/Azure/durabletask/blob/750d78f12d660f518f8a8b412ba6abf22314d203/src/DurableTask.Core/TaskOrchestrationDispatcher.cs#L301) but I'm not sure if there is any limit on how many times it will retry the item. 
2. So this exception actually happens when the worker evicts the extended session _earlier_ than the host does, so this is not a concern.
3. Yes, everything looks good to me.
4. Not that I can think of for now, but I think we should definitely test/discuss this more if we ultimately check this design into the main branch.

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -198,6 +206,11 @@ await this.LifeCycleNotificationHelper.OrchestratorStartingAsync(
             OrchestratorExecutionResult orchestratorResult;
             if (functionResult.Succeeded)
             {
+                if (workerRequiresHistory)
+                {
+                    throw new SessionAbortedException("The worker has since ended the extended session and needs an orchestration history to execute the orchestration request.");
```
bachuv, line 210: Can we add tests for extended sessions in the .NET Isolated testing framework? We could add a test for the use case when this exception gets thrown.

bachuv, line 210: Does the In Process implementation of extended sessions also throw a `SessionAbortedException` in this scenario?

sophiatev: It's not an issue in in-proc since there's no worker so there's no situation where a worker could evict an extended session before the host has (it's much easier for in-proc in that sense)

src/WebJobs.Extensions.DurableTask/RemoteOrchestratorConfiguration.cs
```diff
@@ -12,5 +12,20 @@ public class RemoteOrchestratorConfiguration
...
+
+        /// <summary>
+        /// Gets or sets whether or not to include the past history events in the orchestration request.
+        /// </summary>
+        public bool IncludePastEvents { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets whether or not the orchestration request is within an extended session.
+        /// </summary>
+        public bool ExtendedSession { get; set; } = false;
```
cgillum: nit: it's considered best practice to name Boolean properties in the form of a question. For example, `IsExtendedSession` instead of `ExtendedSession`.

src/Worker.Extensions.DurableTask/Worker.Extensions.DurableTask.csproj
```diff
@@ -43,6 +43,7 @@
     <PackageReference Include="Microsoft.DurableTask.Worker.Grpc" />
     <PackageReference Include="Grpc.Net.Client" />
     <PackageReference Include="Microsoft.DurableTask.Analyzers" />
+	<PackageReference Include="Microsoft.Extensions.Caching.Memory" />
```
cgillum: nit: replace tab with spaces.

src/Worker.Extensions.DurableTask/FunctionsWorkerApplicationBuilderExtensions.cs
```diff
@@ -10,6 +10,7 @@
 using Microsoft.DurableTask.Client;
 using Microsoft.DurableTask.Converters;
 using Microsoft.DurableTask.Worker;
+using Microsoft.DurableTask.Worker.Grpc;
```
cgillum, line 13: This might not be the right PR for this question, but why does the extended session cache live in the gRPC namespace instead of the worker namespace? Doesn't the concept of extended sessions also make sense in non-gRPC scenarios (not that we have any right now, but we might in the future allow the Durable Task SDK to talk directly to Azure Storage).

cgillum (approved):

cgillum (approved):

src/WebJobs.Extensions.DurableTask/Grpc/Protos/orchestrator_service.proto
```diff
@@ -751,6 +765,10 @@ service TaskHubSidecarService {
 
     // Abandon an entity work item
     rpc AbandonTaskEntityWorkItem(AbandonEntityTaskRequest) returns (AbandonEntityTaskResponse);
+
+    // "Skip" graceful termination of orchestrations by immediately changing their status in storage to "terminated".
+    // Note that a maximum of 500 orchestrations can be terminated at a time using this method.
+    rpc SkipGracefulOrchestrationTerminations(SkipGracefulOrchestrationTerminationsRequest) returns (SkipGracefulOrchestrationTerminationsResponse);
```
cgillum, line 771: Were these protobuf changes already merged as part of a different PR?

cgillum (approved):

src/WebJobs.Extensions.DurableTask/Grpc/Protos/orchestrator_service.proto
sophiatev: These are actually for a new feature I added to DTS, and they're showing up here just cause I pulled from the main protobuf branch (so sort of, yes)

nytian (approved):

PR #3160, opened 2025-07-21
Add Node as a supported E2E test language

andystaples:
Adds a Node.js app to the E2E testing suite, runs the E2E tests against it in the pipeline. 

### Pull request checklist


test/e2e/Apps/BasicNode/host.json
```diff
@@ -0,0 +1,23 @@
...
+        "isEnabled": true,
+        "excludedTypes": "Request"
+      }
+    }
+  },
+  "extensions": {
+    "durableTask": {
+      "hubName": "NodeE2ETaskHub",
+        "tracing": {
+            "DistributedTracingEnabled": true,
```
bachuv, line 15: Can we dynamically update this value so we can make tests pass when this setting is false? I'm just thinking of this now since we had the issue with Durable Entities + DT V2.

test/e2e/Tests/Localizers/NodeTestLanguageLocalizer.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
```
nytian, line 3: Can we add a comment here explaining what this class is used for? It’ll help in the future if we update the error message, so we’ll know where to update the corresponding tests.

nytian (approved):
Thanks for the PR! Just left a non-blocking comment

test/e2e/Apps/BasicNode/host.json
andystaples: I was thinking about this too - it would require a host restart, which is not yet supported in the E2E framework. I'm happy to add this, probably in a separate TestCollection (we'd need to experiment with how much additional time this takes in every run) but let's do it in a separate PR.

test/e2e/Tests/Localizers/NodeTestLanguageLocalizer.cs
andystaples: Yes. To prevent blocking this PR, I will add this in a separate PR, either the Java one or the documentation and usage PR I am planning already

PR #3165, opened 2025-07-23
Fixing Null Reference Bug in TaskEntityShim

### LINKED ISSUE #3167: NullReference exception in StartCallEntityAndEntityInvocationActivities
## Impact description
When using recent versions of Durable Functions with Durable entities, it is possible to get the following NullReferenceException from the extension: 

```
Function '<entity name> (Entity)' failed. TraceFlags: <flags>. Details: System.NullReferenceException: Object reference not set to an instance of an object.
at Microsoft.Azure.WebJobs.Extensions.DurableTask.TaskEntityShim.StartCallEntityAndEntityInvocationActivities(RequestMessage request, DateTimeOffset startTime) in /_/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs:line 759
at Microsoft.Azure.WebJobs.Extensions.DurableTask.TaskEntityShim.ProcessOperationRequestAsync(RequestMessage request) in /_/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs:line 458
at Microsoft.Azure.WebJobs.Extensions.DurableTask.TaskEntityShim.<>c__DisplayClass47_0.<<ExecuteBatch>g__ExecuteOperationsInBatch|0>d.MoveNext() in /_/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs:line 381
  --- End of stack trace from previous location ---
  at Microsoft.Azure.WebJobs.Extensions.DurableTask.TaskEntityShim.ExecuteBatch(CancellationToken onHostStopping) in /_/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs:line 392
  at Microsoft.Azure.WebJobs.Extensions.DurableTask.DurableTaskExtension.<>c__DisplayClass83_0.<<EntityMiddleware>b__0>d.MoveNext() in /_/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs:line 939.
```

This error may result in function execution failures and/or process recycles.

## Who is impacted
Apps may be impacted if they meet the following criteria:

1. Distributed Tracing is _not_ enabled in host.json (see [Setting up distributed tracing](https://learn.microsoft.com/azure/azure-functions/durable/durable-functions-diagnostics?tabs=csharp-inproc#setting-up-distributed-tracing))
2. The app is using [Durable Entities](https://learn.microsoft.com/azure/azure-functions/durable/durable-functions-entities), and  [calls](https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-entities?tabs=function-based%2Cin-process%2Cpython-v2&pivots=csharp#example-orchestration-signals-and-calls-an-entity) an entity from an orchestration (apps that only signal entities should not be impacted)
3. The app is using .NET **in-process** (.NET Isolated is **not** impacted), Node.js, or Python
4. The app has one of the following dependencies:
    - Microsoft.Azure.WebJobs.Extensions.DurableTask **v3.2.0** or **v3.3.0**
    - Extension bundles **4.25.0**

## Known workarounds

The current recommended workaround is to enable distributed tracing v2, which was first introduced in v3.2.0. This can be done in host.json as follows:
```json
{
   "extensions": {
     "durableTask": {
         "tracing": {
            "DistributedTracingEnabled": true,
            "Version": "V2"
         }
     }
   }
 }
```

sophiatev:
There was a bug in my implementation of distributed tracing for entities leading to a null reference exception in the case that a `callEntityActivity` trace activity was not successfully created in `TaskEntityShim.StartCallEntityAndEntityInvocationActivities`. The null reference exception is thrown if `callEntityActivity` is null but we attempt to access its `Context` anyway. This PR fixes that bug by essentially changing this to a `callEntityActivity?.Context` call. 

Resolves #3167 

### Pull request checklist

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
@@ -741,13 +741,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             Activity processEntityInvocationActivity = null;
             Activity callEntityActivity = null;
-            ActivityContext? parentTraceContext;
 
             // We only want to create a trace activity for processing the entity invocation in the case that we can successfully parse the trace context of the request that led to this entity invocation.
             // Otherwise, we will create an unlinked trace activity with no parent.
             if (ActivityContext.TryParse(request.ParentTraceContext?.TraceParent, request.ParentTraceContext?.TraceState, out ActivityContext parentTraceContextFromRequest))
             {
-                parentTraceContext = parentTraceContextFromRequest;
+                ActivityContext? parentTraceContext = parentTraceContextFromRequest;
                 if (!request.IsSignal)
                 {
                     callEntityActivity = TraceHelper.StartActivityForCallingOrSignalingEntity(
--- a/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
@@ -741,11 +741,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             Activity processEntityInvocationActivity = null;
             Activity callEntityActivity = null;
+            ActivityContext? parentTraceContext;
 
             // We only want to create a trace activity for processing the entity invocation in the case that we can successfully parse the trace context of the request that led to this entity invocation.
             // Otherwise, we will create an unlinked trace activity with no parent.
-            if (ActivityContext.TryParse(request.ParentTraceContext?.TraceParent, request.ParentTraceContext?.TraceState, out ActivityContext parentTraceContext))
+            if (ActivityContext.TryParse(request.ParentTraceContext?.TraceParent, request.ParentTraceContext?.TraceState, out ActivityContext parentTraceContextFromRequest))
             {
+                parentTraceContext = parentTraceContextFromRequest;
                 if (!request.IsSignal)
                 {
                     callEntityActivity = TraceHelper.StartActivityForCallingOrSignalingEntity(
@@ -756,7 +758,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                         request.ScheduledTime,
                         parentTraceContext,
                         request.RequestTime);
-                    parentTraceContext = callEntityActivity.Context;
+                    parentTraceContext = callEntityActivity?.Context;
                 }
 
                 processEntityInvocationActivity = TraceHelper.StartActivityForProcessingEntityInvocation(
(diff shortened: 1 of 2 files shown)


src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -741,11 +741,13 @@ private async Task ExecuteOutOfProcBatch()
         {
             Activity processEntityInvocationActivity = null;
             Activity callEntityActivity = null;
+            ActivityContext? parentTraceContext;
```
cgillum: I'm not quite understanding what we're changing in this file and need a little help. It looks like we're declaring two variables instead of one, but we're still only using one of those variables. Why was this change necessary?

sophiatev: Yeah sorry, this is confusing. Basically, the problem is that I want this `parentTraceContext` to now be nullable so that I can assign to it from the `callEntityActivity?.Context` call. Previously, since I didn't need it to be nullable, I could just make it the `out` variable of the `ActivityContext.TryParse` call. But now, since it's nullable, I can't have it be the `out` variable of that call, so I have two variables - one is the `out` variable of the call and the other is this newly-nullable `parentTraceContext`. If the `TryParse` succeeds I just assign its result to `parentTraceContext`. Let me know if there's a better way to do this

(I also moved this `parentTraceContext` to be inside the `if` statement in the most recent commit, there was no reason it should have been outside of it)

andystaples: I'm a little lost on this one too TBH - the code compiles without any changes to TaskEntityShim, right?. Passing a non-nullable object (parentTraceContext in this case) to a nullable parameter is valid in .NET, but not the other way around. Unless you were running into some error at runtime

sophiatev: Ah yeah that's not the issue, the issue is that I'm trying to _assign_ a nullable value (`callEntityActivity?.Context`) to a non-nullable parameter `parentTraceContext`. Previously, the call was instead `callEntityActivity.Context`, which had a non-nullable result, and could be assigned to the non-nullable `parentTraceContext`. (But this is precisely what is causing the null reference exception, since `callEntityActivity` can in fact be null). So I had to change `parentTraceContext` to be nullable

andystaples: Right yeah - I actually just enabled nullability for this solution and we have _a lot_ of dangerous null handling. 

I think you could do something like this here? Let me know if this makes sense - I'm not sure it's any more readable than what you have, though:
```csharp
parentTraceContext = callEntityActivity is not null ? callEntityActivity.Context : parentTraceContext;
```

andystaples: Or better, 
```csharp
parentTraceContext = callEntityActivity?.Context ?? parentTraceContext;
```

sophiatev: The problem is that I actually want `parentTraceContext` to hold a null value in the case that `callEntityActivity` is null. (This is because I want to then pass a null value to `StartActivityForProcessingEntityInvocation` for the parent trace context in that case). So unfortunately I can't use your suggestion (although I agree it's more readable).

andystaples (approved):

PR #3183, opened 2025-08-15
Support managed identity for Event Grid

### LINKED ISSUE #2924: Support managed identity for Event Grid notifications
# Issue Transfer

This issue has been transferred from the Azure SDK for .NET repository, [#46394](https://github.com/Azure/azure-sdk-for-net/issues/46394).

### Type of issue

Missing information

### Description

We see option to configure access key for the EventGrid using KeySettingName but how do we configure EventGrid that use Managed Identity based auth.

### Page URL

https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.webjobs.extensions.durabletask.eventgridnotificationoptions?view=azure-dotnet

### Content source URL

https://github.com/Azure/azure-docs-sdk-dotnet/blob/master/xml/Microsoft.Azure.WebJobs.Extensions.DurableTask/EventGridNotificationOptions.xml

### Document Version Independent Id

ec66102e-568e-9087-b442-6c9399e23ff1

### Article author

@azure-sdk

### Metadata

* ID: e7b049c9-bbe7-d26d-6e98-a026d93a6610 
* Service: **azure-functions**

bachuv:
This PR adds support for managed identity for publishing events to Event Grid.

I tested this by publishing Durable Functions events to Event Grid and then sent those events to an Azure Storage Queue.

For context, customers currently specify the event grid topic key through host.json and local.settings.json (https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-event-publishing?tabs=csharp-script):

- host.json
``` json
{
  "version": "2.0",
  "extensions": {
    "durableTask": {
      "notifications": {
        "eventGrid": {
          "topicEndpoint": "https://<topic_name>.westus2-1.eventgrid.azure.net/api/events",
          "keySettingName": "EventGridKey"
        }
      }
    }
  }
}
```

- local.settings.json
``` json
{
    "IsEncrypted": false,
    "Values": {
        "AzureWebJobsStorage": "UseDevelopmentStorage=true",
        "EventGridKey": "<topic_key>"
    }
}
```

This PR adds the following configuration support.

### System Assigned Managed identity configuration:
App Settings
- Add `EventGrid__topicEndpoint` app setting with the value as the Event Grid topic endpoint

Configuration
- Turn on system assigned identity for the function app
- In the event grid topic resource, give the function app the EventGrid Data Sender role.

### User Assigned Managed identity configuration:
App Settings
- Add `EventGrid__topicEndpoint` app setting with the value as the Event Grid topic endpoint
- Add `EventGrid__credential` app setting with the value `managedidentity`
- Add `EventGrid__clientId` app setting with the value of the user assigned managed identity client ID.

Configuration
- Create a user assigned managed identity (uami)
- Attach the uami to the function app resource (in the Identity section)
- Attach the uami to the event grid topic resource (in the Identity section)
- Create an event grid subscription and enable managed identity with the uami
- In the event grid topic resource, assign the EventGrid Data Sender role to the uami
- (If publishing events from the event grid to an Azure Storage queue) In the Azure Storage resource, assign the Storage Queue Data Message Sender role to the uami

Resolves #2924 

### Pull request checklist


src/WebJobs.Extensions.DurableTask/ManagedIdentityOptions.cs
```diff
@@ -16,10 +16,11 @@ public class ManagedIdentityOptions
         /// </summary>
         /// <param name="authorityHost">The host of the Azure Active Directory authority.</param>
         /// <param name="tenantId">The tenant id of the user to authenticate.</param>
-        public ManagedIdentityOptions(Uri authorityHost = null, string tenantId = null)
+        public ManagedIdentityOptions(Uri authorityHost = null, string tenantId = null, string clientId = null)
```
sophiatev: Can you add a parameter comment for this new parameter as well?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -508,8 +508,11 @@ private void TraceConfigurationSettings()
         private ILifeCycleNotificationHelper CreateLifeCycleNotificationHelper()
         {
             // First: EventGrid
-            if (this.Options.Notifications.EventGrid != null
-                && (!string.IsNullOrEmpty(this.Options.Notifications.EventGrid.TopicEndpoint) || !string.IsNullOrEmpty(this.Options.Notifications.EventGrid.KeySettingName)))
+
+            bool topicKeySettingsConfigured = this.Options.Notifications.EventGrid != null && !string.IsNullOrEmpty(this.Options.Notifications.EventGrid.TopicEndpoint);
```
sophiatev: If I understand correctly, it seems like before the check would pass if just the keySettingName was specified and not the topic endpoint. Was this existing check actually incorrect? I would imagine that both had to be specified before, no?  And if my understanding is correct, should we check here that if `usingManagedIdentity` is false then `KeySettingName` is not null/empty?

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
```diff
@@ -42,91 +42,123 @@ public EventGridLifeCycleNotificationHelper(
...
-                        eventGridNotificationsConfig.PublishRetryInterval,
-                        retryStatusCode);
+                // Currently, we support Event Grid Custom Topic for notify the lifecycle event of an orchestrator.
+                // For more detail about the Event Grid, please refer this document.
+                // Post to custom topic for Azure Event Grid
+                // https://docs.microsoft.com/en-us/azure/event-grid/post-to-custom-topic
+                this.HttpMessageHandler = options.NotificationHandler ?? new HttpRetryMessageHandler(
+                    new HttpClientHandler(),
+                    eventGridNotificationsConfig != null ? eventGridNotificationsConfig.PublishRetryCount : 0,
+                    eventGridNotificationsConfig != null ? eventGridNotificationsConfig.PublishRetryInterval : TimeSpan.FromMinutes(5),
```
sophiatev, line 172: Where did we get this 5 minute value from?

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
```diff
@@ -42,91 +42,123 @@ public EventGridLifeCycleNotificationHelper(
...
+            }
+
+            bool isTopicEndpointKeyConfigured = !string.IsNullOrEmpty(this.eventGridTopicEndpoint) && eventGridNotificationsConfig != null;
 
-            if (!string.IsNullOrEmpty(this.eventGridTopicEndpoint))
-            {
-                if (!string.IsNullOrEmpty(eventGridNotificationsConfig.KeySettingName))
-                {
-                    this.useTrace = true;
+            if (this.UseManagedIdentity || isTopicEndpointKeyConfigured)
```
sophiatev: Again should we have a check here for if `this.eventGridKeyValue` is non-null too? Like `this.UseManagedIdentity || (isTopicEndpointKeyConfigured && !string.IsNullOrEmpty(this.eventGridKeyValue)` or something along those lines (this is assuming that when not using managed identity the event grid key needs to be specified)

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
```diff
@@ -42,91 +42,123 @@ public EventGridLifeCycleNotificationHelper(
...
+
+                this.eventGridTopicEndpoint = eventGridNotificationsConfig.TopicEndpoint;
+
+                if (nameResolver.TryResolveWholeString(eventGridNotificationsConfig.TopicEndpoint, out var endpoint))
+                {
+                    this.eventGridTopicEndpoint = endpoint;
+                }
+            }
+
+            bool isTopicEndpointKeyConfigured = !string.IsNullOrEmpty(this.eventGridTopicEndpoint) && eventGridNotificationsConfig != null;
```
sophiatev: Do we need to check for `eventGridNotificationsConfig` being null here if we seem to throw an `ArgumentNullException` if it is?

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
```diff
@@ -42,91 +42,123 @@ public EventGridLifeCycleNotificationHelper(
...
+                    eventGridNotificationsConfig != null ? eventGridNotificationsConfig.PublishRetryInterval : TimeSpan.FromMinutes(5),
+                    retryStatusCode);
 
-                    if (string.IsNullOrEmpty(this.eventGridKeyValue))
-                    {
-                        throw new ArgumentException($"Failed to start lifecycle notification feature. Please check the configuration values for {eventGridNotificationsConfig.KeySettingName} on AppSettings.");
-                    }
-                }
-                else
+                if (eventGridNotificationsConfig != null && !string.IsNullOrEmpty(eventGridNotificationsConfig.KeySettingName) && string.IsNullOrEmpty(this.eventGridKeyValue) && !this.UseManagedIdentity)
```
sophiatev: Do we need the `eventGridNotificationsConfig != null` check here if we check for `!this.UseManagedIdentity`? I think if you move the `!this.UseManagedIdentity` to be the first check, then if that's true (i.e. that we're not using managed identities), `eventGridNotificationsConfig != null` is also guaranteed to be true since we throw the `ArgumentNullException` above if it is right?

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
```diff
@@ -42,91 +42,123 @@ public EventGridLifeCycleNotificationHelper(
...
                 {
-                    throw new ArgumentException($"Failed to start lifecycle notification feature. Please check the configuration values for {eventGridNotificationsConfig.TopicEndpoint} and {eventGridNotificationsConfig.KeySettingName}.");
+                    throw new ArgumentException($"Failed to start lifecycle notification feature. Please check the configuration values for {eventGridNotificationsConfig.KeySettingName} on AppSettings.");
                 }
-            }
-        }
+
+                if (eventGridNotificationsConfig != null && (string.IsNullOrEmpty(eventGridNotificationsConfig.KeySettingName) || string.IsNullOrEmpty(eventGridNotificationsConfig.TopicEndpoint)) && !this.UseManagedIdentity)
```
sophiatev: Same as the above comment for the earlier if check

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
bachuv: I think the check was like this because EventGridLifeCycleNotificationHelper handles the cases where there is missing configuration and throws exceptions. I updated it to just check for topic endpoint here, but I'll add back the check for `KeySettingName`.

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
bachuv: The `ArgumentNullException` is in the `else` case above which might not get checked if the customer enabled managed identity.

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
bachuv: I tried to follow the current implementation to keep the structure the same so the exceptions throw at the end after checking for `KeySettingName` and `eventGridValue`. It's probably better to throw the exceptions at the beginning though. I can make that change.

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
bachuv: The `eventGridNotificationsConfig != null` check happens in an `else` case that is separate so we can't rely on that `ArgumentNullException` getting thrown.

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
bachuv: The eventGridNotificationsConfig != null check happens in an else case that is separate so we can't rely on that ArgumentNullException getting thrown.

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
bachuv: I added this because if the customer is using managed identity, then `eventGridNotificationsConfig` will be `null` and `eventGridNotificationsConfig.PublishRetryCount` and `eventGridNotificationsConfig.PublishRetryInterval` will throw null ref exceptions. I set it to 5 minutes because the default for `eventGridNotificationsConfig.PublishRetryInterval` is 5 minutes. I misspoke when we discussed this offline. This is something that I added.

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
sophiatev: If I understand correctly !this.UseManagedIdentity guarantees that we entered the else case above right?

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
sophiatev: Right but the way this gets used later (the `if (this.UseManagedIdentity || isTopicEndpointKeyConfigured)`) means that if the first check, `this.UseManagedIdentity` is false, and we are checking the second part of the `or` statement, then the null exception would have been thrown so we only need to check for the `this.eventGridTopicEndpoint` having been set correctly (I think). But this might ultimately just a be a nit change

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
bachuv: Yes it does.

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
bachuv: Yes, what you explained makes sense. As part of a different comment, I refactored the code to move the exceptions earlier and removed any `this.UseManagedIdentity` checks for the exceptions. Can you take another look now?

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
```diff
@@ -42,91 +42,129 @@ public EventGridLifeCycleNotificationHelper(
...
+                if (nameResolver.TryResolveWholeString(eventGridNotificationsConfig.TopicEndpoint, out var endpoint))
+                {
+                    this.eventGridTopicEndpoint = endpoint;
+                }
+            }
+
+            // Check if we have the minimum required settings to enable Event Grid notifications with key based authentication.
+            bool eventGridNotificationSettingsConfigured = false;
+
+            if (eventGridNotificationsConfig != null)
```
sophiatev, line 99: Could we have a situation where eventGridNotificationsConfig is non-null but the user is also using managed identities? My worry is that in that case maybe one of these exceptions will get thrown (like the second one)

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
```diff
@@ -42,91 +42,129 @@ public EventGridLifeCycleNotificationHelper(
...
+                this.useTrace = true;
 
-                    var retryStatusCode = eventGridNotificationsConfig.PublishRetryHttpStatus?
-                                              .Where(x => Enum.IsDefined(typeof(HttpStatusCode), x))
-                                              .Select(x => (HttpStatusCode)x)
-                                              .ToArray()
-                                          ?? Array.Empty<HttpStatusCode>();
-
-                    if (eventGridNotificationsConfig.PublishEventTypes == null || eventGridNotificationsConfig.PublishEventTypes.Length == 0)
+                var retryStatusCode = eventGridNotificationsConfig != null ?
```
sophiatev: It looks like the previous logic was setting the `retryStatusCode` to `Array.Empty` in the case that `eventGridNotificationsConfig.PublishRetryHttpStatus` is null, but I think the current logic will instead set the `retryStatusCode` to null in this situation (and set it to `Array.Empty` only if the entire `eventGridNotificationsConfig` is null). I think it should be 
```suggestion
                var retryStatusCode = eventGridNotificationsConfig?.PublishRetryHttpStatus != null ?
```
That way, if `eventGridNotificationsConfig` itself is null, we will set `retryStatusCode` to `Array.Empty`, and will also do the same if the `PublishRetryHttpStatus` field is.

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
bachuv: Yes, I'll add a warning for this scenario and ask customers to either add the settings in host.json for key based authentication or app settings for managed identity.

sophiatev: Looking at what you added - couldn't we still trigger the second exception if `UseManagedIdentity` is true, `options.Notifications.EventGrid` is non-null, but both `options.Notifications.EventGrid.TopicEndpoint` and `options.Notifications.EventGrid.KeySettingName` are null? I'm actually not even sure if what I just described is possible, but I just wanted to double check.

Also what exactly does `this.traceHelper.ExtensionWarningEvent`, like does it exit the code or just surface a warning to the user and we still end up throwing exceptions

bachuv: If managed identity is enabled, then we don't ever initialize `eventGridNotificationsConfig` which means the exceptions will never get thrown.

The warning will just surface for the user.

sophiatev: Oh, of course! Sorry and thanks for the explanation

sophiatev (approved):

src/WebJobs.Extensions.DurableTask/ManagedIdentityOptions.cs
```diff
@@ -16,10 +16,12 @@ public class ManagedIdentityOptions
         /// </summary>
         /// <param name="authorityHost">The host of the Azure Active Directory authority.</param>
         /// <param name="tenantId">The tenant id of the user to authenticate.</param>
-        public ManagedIdentityOptions(Uri authorityHost = null, string tenantId = null)
+        /// <param name="clientId">The client id of the user assigned managed identity.</param>
+        public ManagedIdentityOptions(Uri authorityHost = null, string tenantId = null, string clientId = null)
```
nytian: Maybe nit, updating a ctor directly is breadking change. I would like another ctor and this one just called that one. eg

``` C#
public ManagedIdentityOptions(Uri authorityHost = null, string tenantId = null)
        : this(authorityHost, tenantId, null)
    {
    }

public ManagedIdentityOptions(Uri authorityHost = null, string tenantId = null, string clientId = null)
```

nytian (commented):
Thanks for the PR! Left a comment. Also, can you remove those alerts about the PR? And have we done any e2e testing on Azure  about msi?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -508,10 +508,15 @@ private void TraceConfigurationSettings()
         private ILifeCycleNotificationHelper CreateLifeCycleNotificationHelper()
         {
             // First: EventGrid
-            if (this.Options.Notifications.EventGrid != null
-                && (!string.IsNullOrEmpty(this.Options.Notifications.EventGrid.TopicEndpoint) || !string.IsNullOrEmpty(this.Options.Notifications.EventGrid.KeySettingName)))
+
+            bool topicKeySettingOrKeySettingNameConfigured = this.Options.Notifications.EventGrid != null && (!string.IsNullOrEmpty(this.Options.Notifications.EventGrid.TopicEndpoint) || !string.IsNullOrEmpty(this.Options.Notifications.EventGrid.KeySettingName));
```
cgillum: This line is too long to fit on most monitor screens without scrolling, so let's break it up into multiple lines. I suggest saving the result of `this.Options.Notifications.EventGrid` into a local variable and reusing that local variable in the next few lines.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -508,10 +508,15 @@ private void TraceConfigurationSettings()
         private ILifeCycleNotificationHelper CreateLifeCycleNotificationHelper()
         {
             // First: EventGrid
-            if (this.Options.Notifications.EventGrid != null
-                && (!string.IsNullOrEmpty(this.Options.Notifications.EventGrid.TopicEndpoint) || !string.IsNullOrEmpty(this.Options.Notifications.EventGrid.KeySettingName)))
+
+            bool topicKeySettingOrKeySettingNameConfigured = this.Options.Notifications.EventGrid != null && (!string.IsNullOrEmpty(this.Options.Notifications.EventGrid.TopicEndpoint) || !string.IsNullOrEmpty(this.Options.Notifications.EventGrid.KeySettingName));
+            bool usingManagedIdentity = !string.IsNullOrEmpty(this.nameResolver.Resolve("EventGrid:topicEndpoint"));
```
cgillum: Should we be using the `EventGridLifeCycleNotificationHelper.TopicEndpointKey` constant here?

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
```diff
@@ -37,96 +42,140 @@ public EventGridLifeCycleNotificationHelper(
...
-            this.eventGridKeyValue = nameResolver.Resolve(eventGridNotificationsConfig.KeySettingName);
-            this.eventGridTopicEndpoint = eventGridNotificationsConfig.TopicEndpoint;
+            this.UseManagedIdentity = false;
 
-            if (nameResolver.TryResolveWholeString(eventGridNotificationsConfig.TopicEndpoint, out var endpoint))
-            {
-                this.eventGridTopicEndpoint = endpoint;
-            }
+            // Check to see if we have a topic name app setting defined. If so, we will use managed identity to authenticate.
+            if (!string.IsNullOrEmpty(nameResolver.Resolve(TopicEndpointKey)))
```
cgillum: Instead of resolving the same value twice, resolve it once and save the result into a local variable.

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
```diff
@@ -37,96 +42,140 @@ public EventGridLifeCycleNotificationHelper(
...
+            {
+                this.UseManagedIdentity = true;
+                this.eventGridTopicEndpoint = nameResolver.Resolve(TopicEndpointKey);
+
+                if (string.Equals(nameResolver.Resolve(CredentialKey), ManagedIdentityValue) &&
+                    !string.IsNullOrEmpty(nameResolver.Resolve(ClientIdKey)))
+                {
+                    // Use user assigned managed identity
+                    string clientId = nameResolver.Resolve(ClientIdKey);
```
cgillum: Same feedback about using a local variable to cache the result of a .Resolve call.

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
```diff
@@ -37,96 +42,140 @@ public EventGridLifeCycleNotificationHelper(
...
+            }
+
+            // Log warning if both managed identity and key based authentication are configured.
+            if (this.UseManagedIdentity && options.Notifications != null && options.Notifications.EventGrid != null && (!string.IsNullOrEmpty(options.Notifications.EventGrid.TopicEndpoint) || !string.IsNullOrEmpty(options.Notifications.EventGrid.KeySettingName)))
+            {
+                this.traceHelper.ExtensionWarningEvent(
+                        hubName: this.options.HubName,
+                        functionName: "",
+                        instanceId: "",
+                        "Both managed identity and key based authentication are configured for Event Grid notifications. This may cause conflicts. Please configure either managed identity or key based authentication.");
```
cgillum: "This may cause conflicts" is unclear.  What's the expected behavior? If the actual behavior is that we use one authentication type and not another, then we should clearly say this.

src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs
```diff
@@ -139,11 +188,33 @@ public HttpMessageHandler HttpMessageHandler
...
+                httpClient = new HttpClient(httpMessageHandler);
             }
         }
 
+        internal void SetUpAuthentication()
+        {
+            if (this.UseManagedIdentity)
+            {
+                // Use Bearer token for Managed Identity
+                this.RefreshAccessTokenAsync().GetAwaiter().GetResult();
```
cgillum: It's not good to block on an async method call like this. Is it possible to instead make this method async?

src/WebJobs.Extensions.DurableTask/ManagedIdentityOptions.cs
```diff
@@ -22,6 +27,15 @@ public ManagedIdentityOptions(Uri authorityHost = null, string tenantId = null)
             this.TenantId = tenantId;
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ManagedIdentityOptions"/> class.
+        /// </summary>
+        /// <param name="clientId">The client id of the user assigned managed identity.</param>
+        public ManagedIdentityOptions(string clientId)
```
cgillum: Rather than creating a new constructor with just this argument, it's better to create a new constructor with all three arguments. All other constructors should then be changed to call into this new constructor. This should also help ensure that JSON deserialization works correctly.

bachuv: I added this change by making the following update, but I'm getting the same serialization error when the tests run:

Code:
``` csharp
public ManagedIdentityOptions(Uri authorityHost = null, string tenantId = null)
: this(authorityHost, tenantId, null)
{
}

public ManagedIdentityOptions(Uri authorityHost, string tenantId, string clientId)
{
  this.AuthorityHost = authorityHost;
  this.TenantId = tenantId;
  this.ClientId = clientId;
}
```

Error:
``` 
Newtonsoft.Json.JsonSerializationException : Error converting value "Orchestrator function 'CallHttpAsyncOrchestrator' failed: Unable to find a constructor to use for type Microsoft.Azure.WebJobs.Extensions.DurableTask.ManagedIdentityOptions. A class should either have a default constructor, one constructor with arguments or a constructor marked with the JsonConstructor attribute.`
````

Is it fine to add the `[JsonConstructor]` attribute to the constructor with 3 arguments? Would there be any unintended consequences?

File: src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs (hunk 16 above)
Comment (bachuv) [reply src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs]: If I update this method to make it async then I need to update all of the methods up the chain that call `SetUpAuthentication()` to async, all the way up to the `DurableTaskExtension` constructor where `CreateLifeCycleNotificationHelper()` is called.

Is there a clean workaround to this? I found that maybe we could use lazy initialization when creating the `LifeCycleNotificationHelper`. Is there an approach that you recommend?

Comment (cgillum) [reply src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs]: Yeah, lazy initialization is what we're doing in other places. I think that makes sense for this feature too.

File: src/WebJobs.Extensions.DurableTask/ManagedIdentityOptions.cs (hunk 17 above)
Comment (cgillum) [reply src/WebJobs.Extensions.DurableTask/ManagedIdentityOptions.cs]: Yes, let's add `[JsonConstructor]` to the constructor with 3 arguments. I don't expect there to be any unintended consequences.

**Review by cgillum — CHANGES_REQUESTED** [review]
I'm concerned about the large code churn in the latest iteration. I'm also concerned about the potential complexities and dangers of caching tasks, which have caused major outages for customers in other contexts.

Can we explore a simpler solution, such as keeping initialization as it was before, but calling `SetUpAuthenticationAsync()` inside an existing async method that we always run, such as inside the `StartTaskHubWorkerIfNotStartedAsync()` method?

File: src/WebJobs.Extensions.DurableTask/EventGridLifeCycleNotificationHelper.cs [hunk 18]
```diff
@@ -139,11 +189,33 @@ public HttpMessageHandler HttpMessageHandler
...
+                httpClient = new HttpClient(httpMessageHandler);
             }
         }
 
+        internal async Task SetUpAuthenticationAsync()
+        {
+            if (this.UseManagedIdentity)
+            {
+                // Use Bearer token for Managed Identity
+                await this.RefreshAccessTokenAsync();
```
cgillum, line 201: As far as I can tell, this is the only place where we refresh the access token, and it only happens once when the worker starts up. Suppose the token expires in 1 hour. How do we refresh the token to avoid this expiration?

bachuv: I added another call to `RefreshAccessTokenAsync()` before the POST request in `SendNotificationAsync()`. I checked the [DefaultAzureCredential Class (Azure.Identity)](https://learn.microsoft.com/en-us/dotnet/api/azure.identity.defaultazurecredential?view=azure-dotnet) docs and it says that `GetTokenAsync()` supports caching and refreshing so I didn't add any other code to handle that. Let me know if you recommend handling caching and refreshing another way.

src/WebJobs.Extensions.DurableTask/ManagedIdentityTokenSource.cs
```diff
@@ -70,6 +70,11 @@ public async Task<string> GetTokenAsync()
                 defaultAzureCredentialOptions.InteractiveBrowserTenantId = this.Options.TenantId;
             }
 
+            if (!string.IsNullOrEmpty(this.Options?.ClientId))
+            {
+                defaultAzureCredentialOptions.ManagedIdentityClientId = this.Options.ClientId;
+            }
+
             defaultCredential = this.Options == null ? new DefaultAzureCredential() : new DefaultAzureCredential(defaultAzureCredentialOptions); // CodeQL [SM05137] Use DefaultAzureCredential explicitly for local development and is decided by the user
```
cgillum, line 78: One change we should consider in the future is the ability to choose a credential type other than `DefaultAzureCredential`. This credential type is convenient but can cause problems or unexpected behavior in some setups because it may not always choose to use managed identity, even if managed identity is available.

cgillum (approved):

PR #3215, opened 2025-10-07
Include Exception Properties at FailureDetails via Custom Provider

nytian:

Add support for including exception properties at TaskFailureDetails via custom provider

Note: This PR won't work unless https://github.com/microsoft/durabletask-dotnet/pull/474 is merged. 

### Pull request checklist


src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -739,6 +776,69 @@ private static bool TryExtractSerializedFailureDetailsFromException(string excep
             details = null;
             return false;
         }
+
+        public static IDictionary<string, object?> TryExtractPropertiesFromExceptionJson(string json)
```
andystaples: Generally methods beginning with "Try" should return a bool representing if the operation was successful, and the parsed result as an "out" parameter. 
I'm also a bit concerned about exception handling here - if "json" is not a valid json string, JsonNode.Parse() will throw. Can we add try/catch around the contents of this method and just return false if an exception is thrown?

andystaples (commented):
Looking mostly good - just one thing to discuss

test/e2e/Apps/BasicDotNetIsolated/CustomExceptionPropertiesOrchestration.cs
```diff
@@ -0,0 +1,95 @@
...
+
+public class CustomExceptionPropertiesOrchestration
+{
+    [Function(nameof(OrchestrationWithCustomException))]
+    public async Task<TaskFailureDetails?> OrchestrationWithCustomException([OrchestrationTrigger] TaskOrchestrationContext context)
+    {
+        // Call the activity that will throw an exception
+        try
+        {
+            await context.CallActivityAsync(nameof(BusinessActivity));
```
andystaples, line 20: Can we correct the warning here? Should just need to pass an input or change the return type of BuisnessActivity

test/e2e/Apps/BasicDotNetIsolated/CustomExceptionPropertiesOrchestration.cs
```diff
@@ -0,0 +1,95 @@
...
+
+public class CustomExceptionPropertiesOrchestration
+{
+    [Function(nameof(OrchestrationWithCustomException))]
+    public async Task<TaskFailureDetails?> OrchestrationWithCustomException([OrchestrationTrigger] TaskOrchestrationContext context)
+    {
+        // Call the activity that will throw an exception
+        try
+        {
+            await context.CallActivityAsync(nameof(BusinessActivity), input:null, options: null);
```
andystaples: I looked more into this, and this may be a bug in the analyzer, but currently for activities with the signature containing only TaskActivityContext
```csharp
    public void BusinessActivity([ActivityTrigger] TaskActivityContext context)
```
this analyzer treats TaskActivityContext as the type of the expected input, and will still warn when passing null as the input value. Additionally, our analyzer still complains about passing null to nullable input types. For now, to fix this, you can either add an ingnored input: 
```csharp
            await context.CallActivityAsync(nameof(BusinessActivity), input:string.Empty, options: null);

    public void BusinessActivity([ActivityTrigger] string ignored, TaskActivityContext context)
```

or suppress this analyzer:
```csharp
#pragma warning disable DURABLE2001 // Incorrect input type
            await context.CallActivityAsync(nameof(BusinessActivity));
#pragma warning enable DURABLE2001
```

test/e2e/Apps/BasicJava/extensions.csproj
```diff
@@ -11,7 +11,7 @@
   <ItemGroup>
     <!-- This reference will be dynamically updated by build-e2e-tests.ps1 as part of test setup. -->
     <!-- Do not commit changes to this line unless necessary -->
-    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="3.4.1" />
```
andystaples, line 14: Please do not commit these changes, as it causes issues with the pipelines - can you revert all the extensions.csproj files from the e2e suite?

test/e2e/Tests/Tests/ErrorHandlingTests.cs
```diff
@@ -223,4 +223,72 @@ public async Task OrchestratorWithCustomRetriedActivityException_ShouldSucceed()
         Assert.Contains(this.fixture.TestLogs.CoreToolsLogs, x => x.Contains(nameof(OverflowException)) &&
                                                               x.Contains("Inner exception message"));
     }
+
+    [Fact]
+    [Trait("PowerShell", "Skip")] // FailureDetails is a dotnet-isolated implementation detail
+    [Trait("Python", "Skip")] // FailureDetails is a dotnet-isolated implementation detail
+    [Trait("Node", "Skip")] // FailureDetails is a dotnet-isolated implementation detail
+    [Trait("Java", "Skip")] // Include exception properties at Failure Details for Java is not supported yet.
+    [Trait("DTS", "Skip")] // DTS hasn't suppor this feature yet.
```
andystaples: nit: doesn't support

andystaples (approved):

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -740,6 +821,178 @@ private static bool TryExtractSerializedFailureDetailsFromException(string excep
             return false;
         }
 
+        // Reconstruct a FailureDetails instance from the worker's JSON payload,
+        // recursively converting Properties to native .NET types.
+        private static FailureDetails? BuildFailureDetailsFromJson(JObject obj)
+        {
+            string errorType = obj["ErrorType"]?.Value<string>() ?? string.Empty;
```
cgillum: Do we need to do manual parsing like this or can we just deserialize from the JObject directly? For example:
```csharp
return obj.ToObject<FailureDetails>();
```

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -740,6 +821,178 @@ private static bool TryExtractSerializedFailureDetailsFromException(string excep
...
+            }
+            catch (Exception)
+            {
+                return null;
+            }
+
+            return result;
+        }
+
+        // Convert a JSON representation of Google.Protobuf.WellKnownTypes.Value into a native .NET value.
```
cgillum: Where and why are we converting protobuf types to JSON? This is not something I would have expected.

src/WebJobs.Extensions.DurableTask/ProtobufUtils.cs
```diff
@@ -621,46 +636,128 @@ internal static P.PurgeInstancesResponse CreatePurgeInstancesResponse(PurgeResul
...
+                null => Value.ForNull(),
+                string str => Value.ForString(str),
+                bool b => Value.ForBool(b),
+                int i => Value.ForNumber(i),
+                long l => Value.ForNumber(l),
+                float f => Value.ForNumber(f),
+                double d => Value.ForNumber(d),
+                decimal dec => Value.ForNumber((double)dec),
+
+                // Handle Newtonsoft.Json types (JValue, JArray, JObject) from deserialized protobuf
```
cgillum, line 682: Same question (I think) - why do we have JSON types in the protobuf?

cgillum (commented):
I'm a bit confused about why we mix JSON and protobuf together in the same payloads. I don't remember us doing this before, so maybe I'm missing some context.

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
nytian: The confusion here is that .NET Isolated returns a JSON-serialized string, whereas Java returns a plain string.

Additionally, if I only perform the conversion between the Protobuf Value and the .NET object, the result is incorrect — it returns a serialized Protobuf struct instead of the actual value. To fix this, I added a manual JSON parsing step to convert the JSON-Protobuf struct into a normal .NET object.

Maybe I should just re-construct the code to make it more clear.

nytian: oh nvm my previous comment, i updated to use protobuf serialization and now we don't need to manually parse it.

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -639,15 +639,13 @@ private static FailureDetails GetFailureDetails(Exception e, out bool fromSerial
...
                     }
 
-                    // Try to extract properties from the serialized exception JSON
-                    IDictionary<string, object?>? properties = ExtractPropertiesFromExceptionJson(exception);
-
+                    // For non-.NET language, properties at FailureDetails is not supported yet.
                     if (TrySplitExceptionTypeFromMessage(exception, out string? exceptionType, out string? exceptionMessage))
                     {
-                        return new FailureDetails(exceptionType, exceptionMessage, stackTrace, innerFailure: null, isNonRetriable: false, properties: properties);
+                        return new FailureDetails(exceptionType, exceptionMessage, stackTrace, innerFailure: null, isNonRetriable: false, properties: null);
```
andystaples, line 643: What are the implications for Java with passing null to FailureDetails constructor here? Will we need to change this once Java SDK adds support for exception properties?

test/e2e/Apps/BasicDotNetIsolated/CustomExceptionPropertiesOrchestration.cs
```diff
@@ -0,0 +1,97 @@
...
+    public static async Task<HttpResponseData> CustomExceptionProperties_HttpStart(
+    [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post")] HttpRequestData req,
+    [DurableClient] DurableTaskClient client,
+    FunctionContext executionContext)
+    {
+        string instanceId = await client.ScheduleNewOrchestrationInstanceAsync(
+            nameof(OrchestrationWithCustomException));
+
+        return await client.CreateCheckStatusResponseAsync(req, instanceId);
+    }
```
andystaples: Is CustomExceptionProperties_HttpStart necessary, or can we use StartOrchestration instead?

src/WebJobs.Extensions.DurableTask/ProtobufUtils.cs
```diff
@@ -621,46 +636,128 @@ internal static P.PurgeInstancesResponse CreatePurgeInstancesResponse(PurgeResul
...
+                JValue jv => ConvertJValueToValue(jv),
+                JArray ja => Value.ForList(ja.Select(ConvertObjectToValue).ToArray()),
+                JObject jo => Value.ForStruct(new Struct
+                {
+                    Fields = { jo.Properties().ToDictionary(p => p.Name, p => ConvertObjectToValue(p.Value)) },
+                }),
```
andystaples, line 688: Couple of questions about this

1. Why Newtonsoft objects specifically, and not other JSON library objects? 
2. I notice we are not doing these Newtonsoft conversions in TaskFailureDetailsConverter.cs - is this deliberate? 
3. Also, this may be generally unintuitive for customers who expect their JValue properties to be returned as JValue (or the string equivalent) and instead get some defined object.

src/WebJobs.Extensions.DurableTask/ProtobufUtils.cs
```diff
@@ -621,46 +636,128 @@ internal static P.PurgeInstancesResponse CreatePurgeInstancesResponse(PurgeResul
...
+                    }
+
+                    // If the value starts with the 'dto:' prefix, it may represent a DateTime value — attempt to parse it.
+                    if (stringValue.StartsWith("dto:", StringComparison.Ordinal))
+                    {
+                        if (DateTimeOffset.TryParse(stringValue[4..], CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind, out DateTimeOffset date))
+                        {
+                            return date;
+                        }
+                    }
```
andystaples: I understand the intent with encoding and decoding DateTime and DateTimeOffset, but this implementation introduces some (probably) unintended behavior

For instance, just slightly tweaking the activity from the unit tests you've added:

```csharp
    [Function(nameof(BusinessActivityTypeInjection))]
    public void BusinessActivityTypeInjection([ActivityTrigger] TaskActivityContext context)
    {
        // Throw an exception with custom properties that should be captured
        throw new BusinessValidationException(
                message: "Business logic validation failed",
                stringProperty: $"dt:{DateTime.Now:O}",
                intProperty: 100,
                longProperty: 999999999L,
                dateTimeProperty: new DateTime(2025, 10, 15, 14, 30, 0, DateTimeKind.Utc),
                dictionaryProperty: new Dictionary<string, object?>
                {
                    ["error_code"] = "VALIDATION_FAILED",
                    ["retry_count"] = 3,
                    ["is_critical"] = true
                },
                listProperty: new List<object?> { "error1", "error2", 500, null },
                nullProperty: null);
    }
```

The stringProperty returned at FailureDetails is of type DateTime, even though the intent was to pass the literal string beginning with "dt:"

To properly guarantee that every string will never be accidentally treated as DateTime, we would need to replace the "dt:" and "dto:" delimiters with special characters that could never appear in a string provided by the customer, probably by first encoding the raw string to replace the delimiter. 

At this point we are probably getting into Json serializer territory anyway. I see a few paths forward here:
1. Only encode types natively supported by Google.Protobuf.WellKnownTypes - every other type calls ToString() and the customer is expected to deal with the stringified version
2. Use the same object-json serializer that we use for activity/entity/orchestration inputs and outputs to preserve all types
3. Allow only string values as custom exception properties - the customer's Exception property provider must handle the conversions for other types

I prefer option 1, but open to more discussion

src/WebJobs.Extensions.DurableTask/ProtobufUtils.cs
```diff
@@ -621,46 +636,128 @@ internal static P.PurgeInstancesResponse CreatePurgeInstancesResponse(PurgeResul
...
+                }),
+
+                // For DateTime and DateTimeOffset, add prefix to distinguish from normal string.
+                DateTime dt => Value.ForString($"dt:{dt.ToString("O")}"),
+                DateTimeOffset dto => Value.ForString($"dto:{dto.ToString("O")}"),
+                IDictionary<string, object?> dict => Value.ForStruct(new Struct
+                {
+                    Fields = { dict.ToDictionary(kvp => kvp.Key, kvp => ConvertObjectToValue(kvp.Value)) },
+                }),
+                IEnumerable e => Value.ForList(e.Cast<object?>().Select(ConvertObjectToValue).ToArray()),
```
andystaples: See the comment on ConvertValueToObject about DateTime serialization

andystaples (commented):
Comments

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
nytian: Yeah it needs to be changed once java sdk support properties.

src/WebJobs.Extensions.DurableTask/ProtobufUtils.cs
nytian: I should delete it.. sorry for the confusion here!

src/WebJobs.Extensions.DurableTask/ProtobufUtils.cs
nytian: Good point — yeah, it’s a bit tricky. I kept it this way because it’s consistent with how we handle it in durabletask.dotnet. I’m okay with option 1, though I’m just a bit concerned about alignment.

I will leave it as-it is for other folks opinion on that.

src/WebJobs.Extensions.DurableTask/ProtobufUtils.cs
nytian: I added these because, in the test case, some nested objects such as lists and dictionaries are serialized and deserialized as JSON values when passed from Worker.Extensions to WebJobs.Extensions. Therefore, we need to include this JSON block here.

For `TaskFailureDetailsConverter`, there’s no serialization or deserialization involved — it directly handles the object itself, so a JSON block isn’t needed there.

Yeah, if they specifically expect a JValue, we might be violating that here — but are there really many such cases? It seems that handling nested structures is more important overall. I’m fine with either approach, though. leave it as it-is for other folks to join the discussion..

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
nytian: yeah once we support Properties at FailureDetails, we need to go back here to update these as currently it will always be null.

Directory.Packages.props
```diff
@@ -27,8 +27,9 @@
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="3.9.0" />
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.9.0" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="3.9.0" />
-    <PackageVersion Include="Microsoft.DurableTask.Client.Grpc" Version="1.15.0" />
-    <PackageVersion Include="Microsoft.DurableTask.Worker.Grpc" Version="1.15.0" />
+    <PackageVersion Include="Microsoft.DurableTask.Client.Grpc" Version="1.16.1" />
+    <PackageVersion Include="Microsoft.DurableTask.Worker.Grpc" Version="1.16.1" />
+	<PackageVersion Include="Microsoft.DurableTask.Abstractions" Version="1.16.1" />
```
cgillum: nit: need to use spaces instead of tabs here.

src/WebJobs.Extensions.DurableTask/ProtobufUtils.cs
cgillum: This reminds me a bit of the problem JSON serializers in .NET have. JSON similarly doesn't define a datetime type. Newtonsoft.Json and System.Text.Json both serialize dates as an ISO-8601 string. I wonder if we should do the same since that might be closer to what customers expect? We still have the problem of not knowing whether to deserialize as a string or as a DateTime, but maybe we just don't support *deserialization* of DateTime, and require customers to do that themselves from the string value?

(I say this knowing that it was originally my idea to do the prefix, and now I'm changing my mind after reading this discussion and thinking about it more... 🥲)

cgillum (approved):

src/WebJobs.Extensions.DurableTask/ProtobufUtils.cs
AnatoliB: I will definitely vote for _not_ deserializing these strings. In the worst case, this "parsing by guessing" will lead to irreversible data loss (like getting `1/1/0001 12:00:00 AM` instead of `dt:12345`). When this data is used in the context of an exception, so someone relies on it for potentially challenging troubleshooting, this loss will be especially annoying. I'm sure it will be very rare, but, when it does happen, it may hurt. In this context, keeping DateTime values as strings is much better.

nytian: Removed prefix

andystaples (approved):

AnatoliB (approved):

cgillum (approved):

PR #3224, opened 2025-10-14
Code + warnings cleanup

andystaples:
Runs "Code cleanup" on solution - also several nullability-related warnings fixes


andystaples:
This is unmanageably large, closing and will split up over time

Issue #3193, opened 2025-09-04
Lightweight Durable Functions Packages for Scale Controller Only


nytian:
## Issue Description:
We need a lightweight version of the Durable Functions package that is tailored specifically for the Function Scale Controller. This package will be used only for Scale Controller scenarios and should exclude unnecessary dependencies included in the full DF extension and storage backend sdks.

## Background:
Recently added OpenTelemetry dependencies have caused the Scale Controller to fail to run. Also, current DF packages and all backend SDKs continue to accumulate dependencies over time, creating deployment and runtime burdens.

## Goals:

- Reduce package size and dependencies to minimize deployment and runtime overhead.
- Ensure compatibility with Scale Controller operations.

### Notes:

This package is not intended for general-purpose Durable Functions usage. Focus on functionality strictly required by Scale Controller.

PR #3240, opened 2025-10-29
Introduce a Lightweight Scaling Package for Multi Durable Backends

nytian:
Solve #3193 

This PR introduces a new lightweight package for Azure Functions scaling. The motivation is to reduce the footprint of scaling dependencies, since the current WebJobs.Extensions.DurableTask package has grown significantly in size and now pulls in additional dependencies such as OpenTelemetry.

~~**Changes in `WebJobs.Extensions.DurableTask`**~~
~~- Remove existing scale support from the durability provider.~~
~~- Register the scalability provider factory from WebJobs.Extensions.DurableTask.Scale in~~
~~`DurableTaskExtension` and `DurableTaskJobHostConfigurationExtensions` to enable **runtime scaling**.~~
~~- Update `DurableTaskListener` to delegate scaling-related operations to the new scalability package.~~

**Changes in `WebJobs.Extensions.DurableTask.Scale`**

- Add scaling support for four backend services, usable by **Scale Controller**.
- Related Scale Controller PR:
- https://msazure.visualstudio.com/One/_git/AAPT-Antares-ScaleController/pullrequest/14389357

**Tests**
- Add `ScaleTests` for `WebJobs.Extensions.DurableTask.Scale`

### Pull request checklist


src/WebJobs.Extensions.DurableTask.Scale/DurableTaskScaleExtension.cs
```diff
@@ -0,0 +1,75 @@
...
+		private readonly ILogger logger;
+        private readonly IEnumerable<IDurabilityProviderFactory> durabilityProviderFactories;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DurableTaskScaleExtension"/> class.
+        /// </summary>
+        /// <param name="options"></param>
+        /// <param name="logger"></param>
+        /// <param name="durabilityProviderFactories"></param>
+        public DurableTaskScaleExtension(
```
nytian: I am hoping to make a new extension here, so that SC will load this one instead of DurableTaskExtension.

src/WebJobs.Extensions.DurableTask.Scale/DurableTaskScaleConfigurationExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            builder
+                .AddExtension<DurableTaskScaleExtension>()
+                .BindOptions<DurableTaskOptions>();
+
+            IServiceCollection serviceCollection = builder.Services;
```
nytian, line 43: Here is a main question:

So this class/method is called by SC at registration and it registers many services. But it seems to me only `IconnectionInfoResolver` , `IStorageServiceClientProviderFactory`, and `IDurabilityProviderFactory` is needed for Scale Controller? I wonder can I delete all the other services? Just worry if I miss anything.

src/WebJobs.Extensions.DurableTask.Scale/DurableTaskOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Scale
+{
+    /// <summary>
+    /// Minimal options class for Scale package - only contains what's needed for scaling decisions.
+    /// </summary>
+    public class DurableTaskOptions
```
nytian: Question: I wonder if my thought is right..

`DurabilityProvide`r , `DurabilityProviderFactory`, and `DurableClientAttribute` need DurableTaskOptions. But it looks to me only parts of config is used by scaling. So my idea here is just to use a separate durabletaskoptions here ( not like the one defined under webjobs.extensions.durabletask), and only keeps the config that used by scaling.

src/WebJobs.Extensions.DurableTask.Scale/AzureStorage/AzureStorageOptions.cs
```diff
@@ -0,0 +1,343 @@
...
+using DurableTask.AzureStorage;
+
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Scale.AzureStorage
+{
+    /// <summary>
+    /// Configuration options for the Azure Storage storage provider.
+    /// </summary>
+    public class AzureStorageOptions
```
nytian: I think this is basically same as thought about `DurableTaskOptions`. Only parts of config are used by scaling and I want to only keep those parts. However, I worry if I filtered some parts, the OrchestraitonServiceSettings created will not point to the same `AzureStorageOrchestrationService`, which is used by collecting the metrics. But I guess as long as the `taskhubname` and authentication is same, then we should point to the same service?

src/WebJobs.Extensions.DurableTask.Scale/DurabilityProvider.cs
```diff
@@ -0,0 +1,131 @@
...
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Scale
+{
+    /// <summary>
+    /// The backend storage provider that provides the actual durability of Durable Functions.
+    /// This is functionally a superset of <see cref="IOrchestrationService"/> and <see cref="IOrchestrationServiceClient"/>.
+    /// If the storage provider does not any of the Durable Functions specific operations, they can use this class
+    /// directly with the expectation that only those interfaces will be implemented. All of the Durable Functions specific
+    /// methods/operations are virtual and can be overwritten by creating a subclass.
+    /// </summary>
+    public class DurabilityProvider
```
nytian: Normally the `DuraibilityProvider` will inherit interface like `IOrchestrationService`, `IOrchestrationClient` etc, while I think it's not used for scaling specifically? So what I am  doing there is to make the `DurabilityProvider` and `DurabilityProviderFactory` under this namespace is only for scaling... Hope that make sense.

src/WebJobs.Extensions.DurableTask.Scale/FunctionName.cs
```diff
@@ -0,0 +1,97 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Scale
+{
+    /// <summary>
+    /// The name of a durable function.
+    /// </summary>
+    internal struct FunctionName : IEquatable<FunctionName>
```
nytian: I have to make a copy of this class under this folder because when we create `DurabilityProvider` we need this.

src/WebJobs.Extensions.DurableTask.Scale/AzureStorage/AzureStorageDurabilityProviderFactory.cs
```diff
@@ -0,0 +1,270 @@
...
+            this.inConsumption = platformInfo.IsInConsumptionPlan();
+
+            // The consumption plan has different performance characteristics so we provide
+            // different defaults for key configuration values.
+            int maxConcurrentOrchestratorsDefault = this.inConsumption ? 5 : 10 * Environment.ProcessorCount;
+            int maxConcurrentActivitiesDefault = this.inConsumption ? 10 : 10 * Environment.ProcessorCount;
+            int maxConcurrentEntitiesDefault = this.inConsumption ? 10 : 10 * Environment.ProcessorCount;
+            int maxEntityOperationBatchSizeDefault = this.inConsumption ? 50 : 5000;
+
+            if (this.inConsumption)
```
nytian: I wonder if we should keep the logic from line 72 to line 92. Cause when I checked the TriggerMetadata json payload from SC I didn't find info about runtime. Also I think this basically is not related with scaling?

src/WebJobs.Extensions.DurableTask.Scale/IStorageServiceClientProviderFactory.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// <returns>The corresponding <see cref="IStorageServiceClientProvider{BlobServiceClient, BlobClientOptions}"/>.</returns>
+        IStorageServiceClientProvider<BlobServiceClient, BlobClientOptions> GetBlobClientProvider(string connectionName);
+
+        /// <summary>
+        /// Gets the <see cref="IStorageServiceClientProvider{QueueServiceClient, QueueClientOptions}"/> used
+        /// for accessing the Azure Storage Queue Service associated with the <paramref name="connectionName"/>.
+        /// </summary>
+        /// <param name="connectionName">The name associated with the connection information.</param>
+        /// <returns>The corresponding <see cref="IStorageServiceClientProvider{QueueServiceClient, QueueClientOptions}"/>.</returns>
+        IStorageServiceClientProvider<QueueServiceClient, QueueClientOptions> GetQueueClientProvider(string connectionName);
```
nytian: I think we might only need to keep QueueClient here, as we only need to calculate the queue length? No need for blob and tables

cgillum (commented):
I see that this is all new code and much of it is duplicated, which means that if we need to make a fix to it, we need to make the same fix in the other copy of the code. Is that right? Have you instead considered whether we should refactor the extension code so that we have one scale implementation that's used by both the extension and the scale controller?

src/WebJobs.Extensions.DurableTask.Scale/DurabilityProvider.cs
cgillum: Do we even need the durability provider at all? It's purpose is to connect runtime operations to different backend implementations. Trying to copy this entire abstraction seems like it will bring a lot of other stuff we don't need.

src/WebJobs.Extensions.DurableTask.Scale/DurableTaskScaleConfigurationExtensions.cs
cgillum: Yes, I don't think the other interface services being registered here are needed.

src/WebJobs.Extensions.DurableTask.Scale/DurableTaskOptions.cs
cgillum: If the idea is to have a lightweight, minimal dependency library for scaling, I suggest we create an entirely new class with a different name that has only the specific configurations needed for scaling, without all the other abstractions.

src/WebJobs.Extensions.DurableTask.Scale/DurableTaskScaleExtension.cs
cgillum: That sounds like the right approach, but I also suggest we reduce the number of services that this new extension depends on.

src/WebJobs.Extensions.DurableTask.Scale/FunctionName.cs
cgillum: I suggest we avoid using DurabilityProvider. It carries with it too many unnecessary dependencies.

src/WebJobs.Extensions.DurableTask.Scale/IStorageServiceClientProviderFactory.cs
cgillum: We need queue client for getting queue lengths, but I think we also need blob client for getting the partition count in PMv2 and table client for getting partitions in PMv3.

cgillum (commented):
Added some comments. I wonder if we should try to make this package even smaller, with fewer dependencies copied over. I think the current approach will be hard to maintain.

src/WebJobs.Extensions.DurableTask.Scale/ScalabilityProvider.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// <param name="functionName">Function name.</param>
+        /// <param name="hubName">Task hub name.</param>
+        /// <param name="connectionName">The name of the storage-specific connection settings.</param>
+        /// <param name="scaleMonitor">The scale monitor.</param>
+        /// <returns>True if autoscaling is supported, false otherwise.</returns>
+        public virtual bool TryGetScaleMonitor(
+            string functionId,
+            string functionName,
+            string hubName,
+            string connectionName,
```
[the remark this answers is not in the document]
cgillum: @nytian assuming we need both of these connection names, which `connectionName` takes precendence? We should change the names of both these parameters and the instance member so that we know what their purposes are. For example, one can be `defaultConnectionName` and the other can be `connectionNameOverride`.

src/WebJobs.Extensions.DurableTask.Scale/AzureStorage/AzureStorageScalabilityProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+            string functionName,
+            string hubName,
+            string connectionName,
+            out ITargetScaler targetScaler)
+        {
+            lock (this.initLock)
+            {
+                if (this.singletonDurableTaskMetricsProvider == null)
+                {
+                    // This is only called by the ScaleController, it doesn't run in the Functions Host process.
```
cgillum: This comment is true for all code in this new assembly, right? If so, we should delete this comment.

src/WebJobs.Extensions.DurableTask.Scale/AzureStorage/AzureStorageScalabilityProviderFactory.cs
```diff
@@ -0,0 +1,223 @@
...
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Scale.AzureStorage
+{
+    /// <summary>
+    /// Factory class responsible for creating <see cref="AzureStorageScalabilityProvider"/> instances.
+    /// </summary>
+    public class AzureStorageScalabilityProviderFactory : IScalabilityProviderFactory
+    {
+        private const string LoggerName = "Host.Triggers.DurableTask.AzureStorage";
```
cgillum: Why do we use the Host.Triggers prefix? Is this a scale controller convention? If so, we should add a comment explaining this because the we're not actually running in a "Host" context, so this choice of category name could be misleading.

src/WebJobs.Extensions.DurableTask.Scale/AzureStorage/AzureStorageScalabilityProviderFactory.cs
```diff
@@ -0,0 +1,223 @@
...
+        {
+            this.clientProviderFactory = clientProviderFactory ?? throw new ArgumentNullException(nameof(clientProviderFactory));
+            this.nameResolver = nameResolver ?? throw new ArgumentNullException(nameof(nameResolver));
+            this.loggerFactory = loggerFactory ?? throw new ArgumentNullException(nameof(loggerFactory));
+
+            var optionsValue = options?.Value ?? throw new ArgumentNullException(nameof(options));
+
+            // In Scale Controller context, optionsValue will be a default/empty object (can't read host.json)
+            // The real configuration comes from triggerMetadata in GetScalabilityProvider()
+            this.options = optionsValue;
```
cgillum: If this is default/empty as the comment says, why do we accept it as a parameter and save it?

src/WebJobs.Extensions.DurableTask.Scale/AzureStorage/AzureStorageScalabilityProviderFactory.cs
```diff
@@ -0,0 +1,223 @@
...
+        /// Creates and caches a default <see cref="ScalabilityProvider"/> instanceusing Azure Storage as the backend.
+        /// </summary>
+        /// <returns>
+        /// A singleton instance of <see cref="AzureStorageScalabilityProvider"/>.
+        /// </returns>
+        public virtual ScalabilityProvider GetScalabilityProvider()
+        {
+            if (this.defaultStorageProvider == null)
+            {
+                ILogger logger = this.loggerFactory.CreateLogger(LoggerName);
```
cgillum: Should we just create the logger once in the constructor and reuse it?

src/WebJobs.Extensions.DurableTask.Scale/AzureStorage/AzureStorageScalabilityProviderFactory.cs
```diff
@@ -0,0 +1,223 @@
...
+
+                // Validate Azure Storage specific options
+                this.ValidateAzureStorageOptions();
+
+                // Create StorageAccountClientProvider without credential (connection string)
+                var storageAccountClientProvider = this.clientProviderFactory.GetClientProvider(
+                    this.DefaultConnectionName,
+                    tokenCredential: null);
```
cgillum, line 75: Why can't we use a credential here?

src/WebJobs.Extensions.DurableTask.Scale/IScalabilityProviderFactory.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <summary>
+        /// Gets the default connection name for this backend provider.
+        /// </summary>
+        string DefaultConnectionName { get; }
+
+        /// <summary>
+        /// Creates or retrieves a scalability provider to be used throughout the extension.
+        /// </summary>
+        /// <returns>A scalability provider to be used by the Durable Task Extension.</returns>
+        ScalabilityProvider GetScalabilityProvider();
```
cgillum: What is the purpose of this method? The description says it's to be used by the durable extension, but is that true? I thought this package was only meant for the scale controller?

cgillum (commented):
A few initial questions to help me better understand the intent of the design.

src/WebJobs.Extensions.DurableTask.Scale/IScalabilityProviderFactory.cs
nytian: removed

src/WebJobs.Extensions.DurableTask.Scale/AzureStorage/AzureStorageScalabilityProviderFactory.cs
nytian: removed, sry I forgot to delete it..

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -1,1431 +1,1447 @@
...
+        /// <summary>
+        /// Gets the registered scalability provider factories for runtime scaling.
+        /// </summary>
+        internal IEnumerable<IScalabilityProviderFactory> GetScalabilityProviderFactories()
+        {
+            return this.scalabilityProviderFactories;
+        }
+
+        /// <summary>
+        /// Gets a logger for scale-related logging.
```
cgillum: Why do we say it's for scale-related logging? The returned logger's category seems to be the same one we use for everything else.

src/WebJobs.Extensions.DurableTask/DurableTaskJobHostConfigurationExtensions.cs
```diff
@@ -99,24 +104,52 @@ public static IWebJobsBuilder AddDurableTask(this IWebJobsBuilder builder)
...
             });
 
-            // Commenting out incremental scale model for hotfix release 3.0.0-rc.4, SC uses TBS by default
-            // builder.Services.AddSingleton<IScaleMonitorProvider>(serviceProvider => serviceProvider.GetServices<DurableTaskTriggersScaleProvider>().Single(x => x == provider));
-            builder.Services.AddSingleton<ITargetScalerProvider>(serviceProvider => serviceProvider.GetServices<DurableTaskTriggersScaleProvider>().Single(x => x == provider));
-            return builder;
+            // Register all scalability provider factories from Scale package
+            serviceCollection.AddSingleton<IScalabilityProviderFactory>(serviceProvider =>
+            {
+                return new AzureStorageScalabilityProviderFactory(
```
cgillum: Why do we create these manually? The conventional way to use the dependency injection pattern is to simply do this:
```csharp
services.AddSingleton<IScalabilityProviderFactory, AzureStorageScalabilityProviderFactory>();
services.AddSingleton<IScalabilityProviderFactory, AzureManagedScalabilityProviderFactory>();
services.AddSingleton<IScalabilityProviderFactory, SqlServerScalabilityProviderFactory>();
services.AddSingleton<IScalabilityProviderFactory, NetheriteScalabilityProviderFactory>();
```

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -52,6 +52,10 @@
     <Protobuf Include="Grpc/Protos/**/*.proto" GrpcServices="Both" Access="Internal" />
   </ItemGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="..\WebJobs.Extensions.DurableTask.Scale\WebJobs.Extensions.DurableTask.Scale.csproj" />
```
cgillum: Adding this new project as a dependency seems like a big potential problem because of the dependency bloat it creates, particularly for .NET customers. It effectively means that every .NET function app that uses Durable now depends on the Netherite and MSSQL libraries, which wasn't the case before. Things like security scans will be much more problematic because an outdated dependency in something like Netherite will impact all .NET customers instead of just all Netherite customers.

Is there a different way we can make this work without creating the dependency problem?

cgillum (changes requested):
I haven't reviewed everything, but ran into one blocking issue regarding how this changes the set of dependencies for the durable extension.

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
nytian: Thanks for the feedback! I think we have a few options:

1. Move Azure Storage scale logic into `WebJobs.Extensions.DurableTask`
Since it already depends on DurableTask.AzureStorage, we can register Azure Storage scaling there. If customers use other backends, they would add the new  scale package to their app, and the corresponding ScalabilityProvider would be picked upby reflection. 
Con: If a customer wants to enable runtime scaling, they may need to redeploy to include the scale package if they haven't. This sounds not convenient.

2. Reuse the existing durability path for runtime scaling
Leverage the current durability support path (so when runtime scaling is enabled, it continues using the existing flow), and only have the scale controller use the new scale package.
Pros: Fewer changes and faster to ship.
Con: Our code / design is less clean, since we could have overlapping logic between the scale package and individual durability providers.

3. Split the scale package by backend (scale.AzureStorage, scale.Sql, scale.Netherite, etc.)
Have `WebJobs.Extensions.DurableTask` only register `scale.AzureStorage`. The tricky part is that we likely need to move the other three backend scale packages/logic out of this extensions repo, to their own backend repo instead. So when a customer pulls in a backend Functions package, the matching backend scale package is loaded atomically.
Then the scale controller can register all backend-specific scale packages (instead of a single combined one), and we can avoid relying on WebJobs.Extensions.DurableTask for non–Azure Storage backends.
Pros: Cleanest architecture and resolves the dependency issue properly.
Con: Requires more code movement and larger refactoring.

What do you think? I prefer option 2 or 3 ( both is fine to me).

nytian: btw I updated this PR to be in option2 for now. As this resolves the main concern :1. provide a light pkg for sc to use 2. no extra dependency for df app and runtime scaling will still use the existing pattern . Let me know if there is any concern or other plan is preferred or any other thoughts. thanks!

cgillum: My initial preference was for option 3 since that's the cleanest. But option 2 is definitely more practical. In fact, it's closer to the original Scale Controller architecture. :) I'm fine with that approach. We should consider adding a GitHub Copilot rule and/or some code comments to make sure that we keep the two duplicate code paths in sync.

src/WebJobs.Extensions.DurableTask.Scale/Netherite/NetheriteScalabilityProviderFactory.cs
```diff
@@ -0,0 +1,274 @@
...
+                    if (componentFactory != null || eventHubsCredentialFunc != null)
+                    {
+                        effectiveServiceProvider = new NetheriteScaleControllerServiceProvider(
+                            componentFactory,
+                            eventHubsCredentialFunc,
+                            eventHubsConnectionName,
+                            this.logger);
+                    }
+                }
+
```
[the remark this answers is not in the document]
AnatoliB: @nytian This comment looks fair. Are you planning to address it?

src/WebJobs.Extensions.DurableTask.Scale/TriggerMetadataExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                return null;
+            }
+
+            try
+            {
+                // Parse the JSON metadata to extract configuration values
+                return triggerMetadata.Metadata.ToObject<DurableTaskMetadata>();
+            }
+            catch
```
AnatoliB, line 32: Nit: Do we really need to silently swallow all exceptions here? In general, my worry is that this will mask exceptions and make real issues difficult to diagnose.

If this method is supposed to be used in tests only (is this correct?), perhaps we should just remove this try/catch (so, if any test setup leads to this exception, this test will catch an exception that should be expected)? Alternatively, catch only the specific exceptions that actually happen on these tests. Yet another alternative is to move this code to a place that makes it very obvious that this is just a test helper, so we don't inadvertently invoke it in production code.

src/WebJobs.Extensions.DurableTask.Scale/DurableTaskMetadata.cs
```diff
@@ -0,0 +1,82 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Creates a DurableTaskMetadata instance from DurableTaskOptions for runtime-driven scaling.
+        /// Extracts only the configuration needed for scaling decisions.
+        /// </summary>
+        /// <param name="options">The Durable Task options from host.json.</param>
+        /// <returns>A DurableTaskMetadata instance with configuration for scaling.</returns>
+        public static DurableTaskMetadata FromOptions(object options)
```
AnatoliB, line 64: What is expected to invoke this method? If SC or Host are supposed to do it, please explain this in a comment

nytian: oh I should remove it as this used for runtime-scaling support. But now since we decide to not include it here.

PR #3256, opened 2025-11-13
Extended sessions for entities in .NET isolated

sophiatev:
This PR introduces the extended sessions feature for entities in .NET isolated. It essentially copies the existing logic for extended sessions for orchestrations and makes certain properties/classes more generic for shared logic between the two.

- `EntityTriggerAttributeBindingProvider`: Passes the `RemoteEntityContext` configuration to the proto converter so that the entity state is only attached to the proto request if necessary (extended sessions are not enabled, or this is the first execution within an extended session)
- `OrchestrationTriggerAttributeBindingProvider`: Changed to reference the more generic property `IncludeState` (previously `IncludePastEvents`)
- `RemoteEntityContext`: Introduced a new `Configurations` field to pass information about extended sessions to the worker
- `RemoteOrchestratorContext`: Changed to reference the more generic `RemoteInstanceConfiguration` and `IncludeState` objects/properties
- `OutOfProcMiddleware`: Existing orchestration logic changed to reference the more generic `WorkItemMetadata.IncludeState` field, and logic introduces for entities to extract information from the `WorkItemMetadata` object and attach it to the `RemoteEntityContext`, as well as to throw a `SessionAbortedException` if the worker needed an entity state and none was attached to the request
- `ProtobufUtils`: Changed to only include entity state in the request if necessary, and to serialize a map of values containing information about extended sessions for the worker process
- `RemoteInstanceConfiguration`: Renamed version of `RemoteOrchestratorConfiguration` to be used by both orchestrations and entities
- `DurableFunctionExecutor.Entity`: Changed to pass the extended sessions cache to the `TaskEntityDispatcher`
- `TaskEntityDispatcher`: Changed to pass the cache to `GrpcEntityRunner`

### Pull request checklist

* [] My changes **do not** need to be backported to a previous version


src/WebJobs.Extensions.DurableTask/RemoteInstanceConfiguration.cs
```diff
@@ -6,18 +6,18 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
     /// <summary>
     /// Configuration settings for RemoteOrchestratorContext in out-of-process mode, transmitted via gRPC.
     /// </summary>
-    public class RemoteOrchestratorConfiguration
+    public class RemoteInstanceConfiguration
```
cgillum: This is another case of a breaking API change. Is it safe to make?

src/WebJobs.Extensions.DurableTask/RemoteInstanceConfiguration.cs
```diff
@@ -6,18 +6,18 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
...
         /// </summary>
         public int HttpDefaultAsyncRequestSleepTimeMilliseconds { get; set; } = 30000;
 
         /// <summary>
-        /// Gets or sets whether or not to include the past history events in the orchestration request.
+        /// Gets or sets whether or not to include the instance state in the instance request.
         /// True by default.
         /// </summary>
-        public bool IncludePastEvents { get; set; } = true;
+        public bool IncludeState { get; set; } = true;
```
cgillum: I'm wondering whether this breaking change is safe. Does this property get used by anything, either directly or indirectly? An example of indirect usage would be something like it gets set by deserializing configuration in host.json (I'm not sure if it does, but it looks like something that potentially could be).

sophiatev: This property was introduced by me in the PR for extended sessions for orchestrations, so it shouldn't be getting used by anything outside of this repo. I think it should be a safe change to make.

Which does beg the question why I made it a public property... probably something I need to change in this PR.

src/WebJobs.Extensions.DurableTask/RemoteInstanceConfiguration.cs
sophiatev: Checked with @nytian, this class is actually not used outside of this repo so this should be a safe change. I'll actually go ahead and make it an `internal` class in this PR too.

src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteEntityContext.cs
```diff
@@ -11,13 +11,25 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
...
+                };
+            }
         }
 
         [JsonProperty("request")]
-        internal EntityBatchRequest Request { get; private set; }
+        internal EntityBatchRequest Request { get; private set; }
+
+        [JsonProperty("configurations")]
+        public RemoteInstanceConfiguration? Configurations { get; private set; }
```
cgillum: nit: remove the `private set;`, you don't need it if you're only assigning the property in the constructor. This is useful to communicate to the reader that this is effectively a read-only property.

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -449,6 +460,11 @@ void SetErrorResult(FailureDetails failureDetails)
                 }
 
                 return;
+            }
+
+            if (workerRequiresEntityState)
+            {
+                throw new SessionAbortedException("The worker has since ended the extended session and needs an entity state to execute the request.");
```
cgillum: Just curious: why do we wait until we get all the way down here to throw this exception rather than throwing it right when we get the response?

cgillum (approved):

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
sophiatev: If we throw earlier, it gets registered as a "host runtime exception" [here](https://github.com/Azure/azure-functions-durable-extension/blob/1f5384d7b238060ded5e2a7f1f375a1767f05c91/src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs#L405). I could have altered this exception-catching logic but I figured it'd be easier to just throw the exception later.

sophiatev: Actually I can at least move it before the `(!functionResult.Succeeded)` check. I don't think it could be the case that if we need an entity state, the function could have failed, since we don't attempt to invoke anything in that case. But wouldn't hurt

YunchuWang (approved):

PR #3260, opened 2025-11-21
Open gRPC server for Python if requested

andystaples:
Adds a new environment variable that allows changing the worker communication protocol to gRPC manually. Will be used for preview builds of the durabletask-azurefunctions package before support is added to Python worker + Host to allow WebJobs extensions to access worker capabilities

### Pull request checklist


src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -175,19 +175,19 @@ public DurableTaskExtension(
...
             {
-                this.OutOfProcProtocol = OutOfProcOrchestrationProtocol.MiddlewarePassthrough;
-                this.localGrpcListener = LocalGrpcListener.Create(this, this.Options.GrpcListenerMode);
-                this.HostLifetimeService.OnStopped.Register(this.StopLocalGrpcServer);
+                this.ConfigureForGrpcProtocol();
             }
             else
             {
-                this.OutOfProcProtocol = OutOfProcOrchestrationProtocol.OrchestratorShim;
+                this.ConfigureForHttpProtocol();
```
cgillum, line 183: During startup initially we'll call `ConfigureForHttpProtocol()` but will later call `ConfigureForGrpcProtocol()` if we see the appropriate metadata on the bindings?

andystaples: Correct - ConfigureForHttpProtocol() only sets the OutOfProcOrchestrationProtocol setting in the extension, which will be changed by the call to ConfigureForGrpcProtocol() during parsing of metadata in Python before that property is ever accessed

src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
```diff
@@ -51,6 +51,11 @@ public ActivityTriggerAttributeBindingProvider(
 
             // The activity name defaults to the method name.
             var activityName = new FunctionName(name);
+            if (trigger.DurableRequiresGrpc)
```
cgillum, line 54: Let's add some comments explaining the scenario and what we're doing. It will be good to have these kinds of comments for future reviewers and for any AI that needs to understand the code logic (for debugging or for maintanence purposes).

src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
```diff
@@ -23,6 +23,11 @@ public BindingHelper(DurableTaskExtension config)
         public IAsyncCollector<StartOrchestrationArgs> CreateAsyncCollector(DurableClientAttribute clientAttribute)
         {
             IDurableOrchestrationClient client = this.config.GetClient(clientAttribute);
+            if (clientAttribute.DurableRequiresGrpc)
```
cgillum: I'm trying to remember when `CreateAsyncCollector` gets called, and whether it gets called early enough for us to correctly configure the gRPC protocol. I'm also wondering if this gets called at all for out-of-proc, or whether it's for in-proc only. Can you confirm these questions, and if there are no concerns, explain via code comments?

src/WebJobs.Extensions.DurableTask/TriggerAttributes/ActivityTriggerAttribute.cs
```diff
@@ -27,5 +27,18 @@ public sealed class ActivityTriggerAttribute : Attribute
         [AutoResolve]
 #pragma warning restore CS0618 // Type or member is obsolete
         public string Activity { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether this orchestration requires gRPC communication. Default false.
+        /// </summary>
+        /// <remarks>
+        /// If set to true, the Durable extension will start a gRPC server to handle communication for this orchestration,
```
cgillum:
```suggestion
        /// If set to true, the Durable extension will start a gRPC server to handle communication for this activity,
```

src/WebJobs.Extensions.DurableTask/TriggerAttributes/EntityTriggerAttribute.cs
```diff
@@ -30,5 +30,18 @@ public sealed class EntityTriggerAttribute : Attribute
         [AutoResolve]
 #pragma warning restore CS0618 // Type or member is obsolete
         public string EntityName { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether this orchestration requires gRPC communication. Default false.
+        /// </summary>
+        /// <remarks>
+        /// If set to true, the Durable extension will start a gRPC server to handle communication for this orchestration,
```
cgillum:
```suggestion
        /// If set to true, the Durable extension will start a gRPC server to handle communication for this entity,
```

src/WebJobs.Extensions.DurableTask/DurableClientAttribute.cs
```diff
@@ -61,6 +61,19 @@ public DurableClientAttribute(DurableClientOptions durableClientOptions)
...
+        /// </summary>
+        /// <remarks>
+        /// If set to true, the Durable extension will start a gRPC server to handle communication for this orchestration,
+        /// regardless of the detected worker runtime. This is used for languages migrating from HTTP to gRPC to allow the
+        /// worker to communicate the desired connection type back to the host.
+        /// </remarks>
+        /// <value>
+        /// True if gRPC is required; otherwise, false.
+        /// </value>
+        public bool DurableRequiresGrpc { get; set; } = false;
```
cgillum: nit: no need to assign properties to their default values (for example, we don't do this for `ExternalClient` above).

src/WebJobs.Extensions.DurableTask/DurableClientAttribute.cs
```diff
@@ -61,6 +61,19 @@ public DurableClientAttribute(DurableClientOptions durableClientOptions)
         /// </summary>
         public bool ExternalClient { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether this orchestration requires gRPC communication. Default false.
```
cgillum:
```suggestion
        /// Gets or sets a value indicating whether this client requires gRPC communication. Default false.
```

src/WebJobs.Extensions.DurableTask/DurableClientAttribute.cs
```diff
@@ -61,6 +61,19 @@ public DurableClientAttribute(DurableClientOptions durableClientOptions)
...
+        /// </summary>
+        /// <remarks>
+        /// If set to true, the Durable extension will start a gRPC server to handle communication for this orchestration,
```
cgillum:
```suggestion
        /// If set to true, the Durable extension will start a gRPC server to handle communication for this client,
```

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
cgillum: Let's add a comment about this in the code for the benefit of other reviewers and maintainers since it's a non-obvious behavior.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -201,14 +200,22 @@ public DurableTaskExtension(
 
         internal TypedCodeProvider TypedCodeProvider { get; private set; }
 
+        /// <summary>
+        /// A global static accessible to DurabilityProviders to know whether the app has requested gRPC via worker metadata.
+        /// </summary>
+        /// <value>
+        /// True if the functionapp requested gRPC via function metadata, otherwise false.
+        /// </value>
+        public static bool DurableRequiresGrpc { get; set; } = false;
```
cgillum: Why does this need to be static? All triggers and bindings should be associated with a single instance of the `DurableTaskExtension` class, so I'd think it'd be more appropriate for this to be an instance property.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -358,44 +363,71 @@ void IExtensionConfigProvider.Initialize(ExtensionConfigContext context)
             context.AddBindingRule<EntityTriggerAttribute>()
                 .BindToTrigger(new EntityTriggerAttributeBindingProvider(this, connectionName));
 
-            this.taskHubWorker = new TaskHubWorker(this.defaultDurabilityProvider, this, this, loggerFactory: this.loggerFactory, versioningSettings: new VersioningSettings
+            this.getTaskHubWorker = () =>
             {
```
cgillum: I wonder if we should keep some state indicating whether we've run this already. I worry about what could happen if this method gets run twice and we create multiple `TaskHubWorker` objects, creating hard-to-diagnose problems.

test/Common/TestHelpers.cs
```diff
@@ -353,9 +353,19 @@ public static async Task WaitUntilTrue(Func<bool> predicate, string conditionDes
...
-                if (predicate())
+                try
                 {
-                    return;
+                    if (predicate())
+                    {
+                        return;
+                    }
+                }
+                catch (Exception ex)
```
cgillum, line 363: With this change we now swallow exceptions and let the test continue running, assuming they are transient. Is this the intended behavior? If so, can you add comments explaining what kind of transient errors this is meant to handle?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
andystaples: I'd love to have an instance property here, but we need it in AzureStorageDurabilityProviderFactory.GetAzureStorageOrchestrationServiceSettings which does not have access to the extension instance. If you can think of another way to communicate this value through to that method, I'm happy to use that approach.

test/Common/TestHelpers.cs
andystaples: This change was needed to prevent a newly-exposed race condition in RemovesNewlinesFromExceptions from failing the tests when the LinuxAppServiceLogger was still buffering the file output when the test went to read the file, leading to an equivalent to FileNotFound in our complex read-safe log file reader. 
I can add a comment about this, but I think that tests where the predicate fails due to a determinate exception will continue to fail, so the effect will be
1. These tests will start failing due to timeout errors, with the "real" exception details in the logs
2. Non-determinate/race-condition failures will be swallowed. 

I'll add a comment explaining the same.

src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
andystaples: You're right, it isn't called in this code path, even though I could have sworn I hit the breakpoint during testing. I moved the function metadata parsing for the client to GetClient(), which is later than the other trigger types but still early enough that we can defer TaskHubWorker init until after it is needed.

There is still a possibility for unexpected behavior if (somehow) a user app defines an app with two or more client functions that don't all define DurableRequiresGrpc, where the app might throw inconsistent error types depending on which function executes first, but this is already an unsupported scenario, and we warn against it explicitly in the Functions Python SDK.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
andystaples: Modified the lambda to re-assign getTaskHubWorker at the beginning of the call to a new lambda that will warn, attempt to return the existing task hub worker, and throws if it is null. Let me know if this behavior is acceptable.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -1287,6 +1348,13 @@ internal string GetInvalidEntityFunctionMessage(string name)
             return message;
         }
 
+        internal string GetTaskHubWorkerDuplicateMessage(string hubName)
+        {
+            return $"A Task Hub Worker is already started for the task hub '{hubName}' but the extension called getTaskHubWorker() again. " +
+                "Please report this at https://github.com/Azure/azure-functions-durable-extension/issues \n" +
+                "At: " + new StackTrace().ToString();
```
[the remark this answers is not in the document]
andystaples: Resolved

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
andystaples: Turns out you were right, this being static caused issues in the tests. 
I changed this to use an instanced property and have AzureStorageDurabiltyProviderFactory implement a new interface IExtensionAwareDurabilityProviderFactory, which should be extensible by the other DurabilityProviderFactories like MSSQL/DTS. Can you take another look and confirm this new approach makes sense?

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
```diff
@@ -5,8 +5,8 @@
     <AssemblyName>Microsoft.Azure.WebJobs.Extensions.DurableTask</AssemblyName>
     <RootNamespace>Microsoft.Azure.WebJobs.Extensions.DurableTask</RootNamespace>
     <MajorVersion>3</MajorVersion>
-    <MinorVersion>8</MinorVersion>
-    <PatchVersion>1</PatchVersion>
+    <MinorVersion>9</MinorVersion>
+    <PatchVersion>0</PatchVersion>
```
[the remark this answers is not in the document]
andystaples: Reviewers - these changes really only apply to the WebJobs stuff, but I understand why we might want to bump and release the worker side as well. Would you prefer I bump both versions, leave both versions unchanged until we are ready to release, or keep this change as-is?

src/WebJobs.Extensions.DurableTask/TriggerAttributes/OrchestrationTriggerAttribute.cs
```diff
@@ -30,5 +30,18 @@ public sealed class OrchestrationTriggerAttribute : Attribute
...
+        /// </summary>
+        /// <remarks>
+        /// If set to true, the Durable extension will start a gRPC server to handle communication for this orchestration,
+        /// regardless of the detected worker runtime. This is used for languages migrating from HTTP to gRPC to allow the
+        /// worker to communicate the desired connection type back to the host.
+        /// </remarks>
+        /// <value>
+        /// True if gRPC is required; otherwise, false.
+        /// </value>
+        public bool DurableRequiresGrpc { get; set; }
```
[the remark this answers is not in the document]
andystaples: Tests will be added that cover these code paths in the E2E for the new Python SDK soon after this.

test/FunctionsV2/EntityIdParsingTests.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        [Fact]
+        public void EntityIdParsing_ThrowsOnInvalidEntityId()
+        {
+            Assert.Throws<ArgumentException>(() => EntityId.FromString(null));
+            Assert.Throws<ArgumentException>(() => EntityId.FromString(""));
+            Assert.Throws<ArgumentException>(() => EntityId.FromString("invalidEntityId"));
+            Assert.Throws<ArgumentException>(() => EntityId.FromString("invalid@symbolplacement"));
+            Assert.Throws<ArgumentException>(() => EntityId.FromString("invalid@symbol@placement"));
+        }
```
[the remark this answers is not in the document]
andystaples: The test case is not valid, it does not begin with @

src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj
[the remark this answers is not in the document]
cgillum: My preference is to bump versions ahead of time so that we don't need to open another PR at release time, but I'm flexible on this.

test/Common/TestHelpers.cs
```diff
@@ -353,9 +353,25 @@ public static async Task WaitUntilTrue(Func<bool> predicate, string conditionDes
...
+                    // This is to prevent tests from failing due to transient errors in the predicate evaluation,
+                    // specifically a newly-exposed race condition in RemovesNewlinesFromExceptions where the LinuxAppServiceLogger
+                    // was still buffering the log output when the predicate tried to read it, causing the equivalent of a
+                    // FileNotFound error.
+                    // Predicates that fail consistently will still fail tests, though the thrown exception type becomes TimeoutException.
+                    if (output is not null)
+                    {
+                        output.WriteLine($"Exception thrown while evaluating predicate: {ex}");
+                    }
                 }
```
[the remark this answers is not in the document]
andystaples: I feel debuggability is OK with this setup, the "skipped" exceptions are still logged into the test output so should be very visible, but open to the idea of storing the last skipped exception and re-throwing it as the inner.

src/WebJobs.Extensions.DurableTask/ProtobufUtils.cs
```diff
@@ -314,6 +313,82 @@ public static OrchestratorAction ToOrchestratorAction(P.OrchestratorAction a)
...
+                    string? targetInstance = null;
+                    DateTime? scheduledTime = null;
+                    switch (a.SendEntityMessage.EntityMessageTypeCase)
+                    {
+                        case P.SendEntityMessageAction.EntityMessageTypeOneofCase.EntityLockRequested:
+                            entityMessage = new RequestMessage()
+                            {
+                                Operation = null,
+                                Id = Guid.Parse(a.SendEntityMessage.EntityLockRequested.CriticalSectionId),
+                                LockSet = a.SendEntityMessage.EntityLockRequested.LockSet.Select(s => EntityId.FromString(s)).ToArray(),
```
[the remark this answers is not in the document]
andystaples: I think debuggability is OK here, FromString would include either the information that the entity ID was null or empty, or the actual text of the ID in any other case when ArgumentException is thrown, both of which (probably) provide enough information for the customer/us to understand what went wrong. Open to feedback, though

src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
```diff
@@ -51,6 +51,15 @@ public ActivityTriggerAttributeBindingProvider(
 
             // The activity name defaults to the method name.
             var activityName = new FunctionName(name);
+            if (trigger.DurableRequiresGrpc)
+            {
+                // The function's metadata includes the DurableRequiresGrpc attribute, indicating that a language typically
+                // configured for HTTP wants to use gRPC instead. Currently only Python uses this flag for the durabletask-based
+                // Python SDK.
```
cgillum, line 58: Can I assume that we'll eventually use this for the JavaScript SDK as well?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -208,7 +206,77 @@ internal TimeSpan MessageReorderWindow
...
+            // and this is the latest point in the pipeline available to us.
+            if (this.OutOfProcProtocol == OutOfProcOrchestrationProtocol.MiddlewarePassthrough)
+            {
+                this.localGrpcListener?.StartAsync(default).GetAwaiter().GetResult();
+            }
+
+            if (this.OutOfProcProtocol == OutOfProcOrchestrationProtocol.OrchestratorShim)
+            {
+                bool? shouldEnable = this.Options.LocalRpcEndpointEnabled;
+                if (!shouldEnable.HasValue)
```
cgillum: Not related to your change (I see you simply moved this code from one place to another), but I got stuck on this line of code for a minute, wondering what it meant and whether it was correct. To make the intent simpler, consider turning it into a simple null check.
```suggestion
                if (shouldEnable == null)
```

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -357,45 +440,41 @@ void IExtensionConfigProvider.Initialize(ExtensionConfigContext context)
...
+        internal void ConfigureForHttpProtocol()
+        {
+            if (this.OutOfProcProtocol != OutOfProcOrchestrationProtocol.OrchestratorShim)
             {
-                this.StartLocalGrpcServer();
+                this.OutOfProcProtocol = OutOfProcOrchestrationProtocol.OrchestratorShim;
+                try
+                {
+                    this.durabilityProviderFactory.SetUseSeparateQueueForEntityWorkItems(false);
+                    this.DefaultDurabilityProvider.SetUseSeparateQueueForEntityWorkItems(false);
```
cgillum: There's something a little strange about the fact that we're setting a property for both a factory and an object presumably created by that factory. I haven't looked at the code closely enough to understand why this is needed, but something to maybe reflect on in case it indicates something fragile in the design.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -357,45 +440,41 @@ void IExtensionConfigProvider.Initialize(ExtensionConfigContext context)
...
+                }
+                catch (NotImplementedException ex)
+                {
+                    this.TraceHelper.ExtensionWarningEvent(this.Options.HubName, string.Empty, string.Empty, $"Could not set UseSeparateQueueForEntityWorkItems: {ex}");
```
cgillum: Consider adding a comment explaining in what situations we expect this as well as why it's safe for us to swallow this exception.

src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
andystaples: Yes that seems likely

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
andystaples: Good callout - the reason I'm doing both here is that the DefaultDurabilityProvider instance will be created using this factory before this method is called/is relevant, but we also need to ensure that all future durability providers created using the factory use the correct setting. I'm not sure this is fragile, per-se, maybe just over-eager as the factory keeps track of the default provider as well, meaning we could call a method on just the factory, but that would be assuming that the extension's default provider and the factory's default provider are always the same, which I think is risky

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -112,7 +112,7 @@ await this.LifeCycleNotificationHelper.OrchestratorStartingAsync(
 
             WorkItemMetadata workItemMetadata = dispatchContext.GetProperty<WorkItemMetadata>();
             bool isExtendedSession = workItemMetadata.IsExtendedSession;
-            bool includePastEvents = workItemMetadata.IncludePastEvents;
+            bool includePastEvents = workItemMetadata.IncludeState;
 
             var context = new RemoteOrchestratorContext(runtimeState, entityParameters, this.extension.Options, isExtendedSession, includePastEvents);
```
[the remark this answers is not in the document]
andystaples: This will be fixed with rebase

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -439,7 +503,7 @@ private void InitializeLinuxLogging()
             // Since our logging payload can be quite large, linux telemetry by default
             // disables verbose-level telemetry to avoid a performance hit.
             bool enableVerbose = this.Options.Tracing.AllowVerboseLinuxTelemetry;
-            this.eventSourceListener = new EventSourceListener(linuxLogger, enableVerbose, this.TraceHelper, this.defaultDurabilityProvider.EventSourceName, this.extensionGuid);
+            this.eventSourceListener = new EventSourceListener(linuxLogger, enableVerbose, this.TraceHelper, this.DefaultDurabilityProvider.EventSourceName, this.extensionGuid);
```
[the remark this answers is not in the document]
andystaples: Why, copilot, why couldn't you have found this earlier. Fixed

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -357,45 +441,25 @@ void IExtensionConfigProvider.Initialize(ExtensionConfigContext context)
...
+        {
+            if (this.OutOfProcProtocol != OutOfProcOrchestrationProtocol.MiddlewarePassthrough)
             {
-                this.StartLocalHttpServer();
+                this.OutOfProcProtocol = OutOfProcOrchestrationProtocol.MiddlewarePassthrough;
+                this.localGrpcListener = LocalGrpcListener.Create(this, this.Options.GrpcListenerMode);
+                this.HostLifetimeService.OnStopped.Register(this.StopLocalGrpcServer);
+                this.DurableRequiresGrpc = true;
             }
         }
```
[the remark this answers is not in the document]
andystaples: Reviewers - thoughts?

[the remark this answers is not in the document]
cgillum: Agreed that if this can be called by multiple threads concurrently, we should use a lock to protect it.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
cgillum: I think it would be good to mention some of this context as code comments so that future reviewers/agents understand the rationale behind the approach.

cgillum (approved):

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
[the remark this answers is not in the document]
andystaples: Added a lock object around this logic and also a check ensuring that the localGrpcListener is not null, in the (currently impossible but just-to-be-safe) case that the call order is ConfigureForGrpcProtocol(), ConfigureForHttpProtocol(), ConfigureForGrpcProtocol(), IntializeTaskHubWorker()

src/WebJobs.Extensions.DurableTask/Bindings/OrchestrationTriggerAttributeBindingProvider.cs
```diff
@@ -70,6 +70,15 @@ public OrchestrationTriggerAttributeBindingProvider(
...
 
             this.config.RegisterOrchestrator(orchestratorName, null);
+            if (trigger.DurableRequiresGrpc)
+            {
+                // The function's metadata sets the DurableRequiresGrpc attribute property, indicating that a language typically
+                // configured for HTTP wants to use gRPC instead. Currently only Python uses this property for the durabletask-based
+                // Python SDK.
+                // Calling this method will cause the extension to use gRPC instead of HTTP when starting the task hub.
+                this.config.ConfigureForGrpcProtocol();
+            }
```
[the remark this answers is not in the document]
andystaples: This will be covered implicitly when we add E2E for new Python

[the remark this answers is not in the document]
andystaples: Is implicitly covered by E2E tests

src/WebJobs.Extensions.DurableTask/ProtobufUtils.cs
```diff
@@ -326,6 +325,82 @@ public static OrchestratorAction ToOrchestratorAction(P.OrchestratorAction a)
...
+                    return new SendEventOrchestratorAction
+                    {
+                        Id = a.Id,
+                        Instance = new OrchestrationInstance
+                        {
+                            InstanceId = targetInstance,
+                        },
+                        EventName = eventName,
+                        EventData = JsonConvert.SerializeObject(entityMessage, new JsonSerializerSettings() { TypeNameHandling = TypeNameHandling.None }),
+                    };
```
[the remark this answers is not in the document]
andystaples: Will be implicitly covered by E2E for new python

src/WebJobs.Extensions.DurableTask/ProtobufUtils.cs
```diff
@@ -326,6 +325,82 @@ public static OrchestratorAction ToOrchestratorAction(P.OrchestratorAction a)
...
+                        case P.SendEntityMessageAction.EntityMessageTypeOneofCase.EntityLockRequested:
+                            entityMessage = new RequestMessage()
+                            {
+                                Operation = null,
+                                Id = Guid.Parse(a.SendEntityMessage.EntityLockRequested.CriticalSectionId),
+                                LockSet = a.SendEntityMessage.EntityLockRequested.LockSet.Select(s => EntityId.FromString(s)).ToArray(),
+                                Position = a.SendEntityMessage.EntityLockRequested.Position,
+                                ParentInstanceId = a.SendEntityMessage.EntityLockRequested.ParentInstanceId,
+                            };
+                            targetInstance = a.SendEntityMessage.EntityLockRequested.LockSet.ElementAt(a.SendEntityMessage.EntityLockRequested.Position);
```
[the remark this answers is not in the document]
andystaples: See previous Copilot comment - this is OK, as we don't expect to be able to parse messages with invalid entity IDs and the exceptions thrown provide enough information to figure out what went wrong

cgillum (approved):

PR #3316, opened 2026-01-17
Add support for dedupe statuses in isolated

sophiatev:
# Summary
## What changed?
This PR:
1. Adds support for honoring dedupe statuses specified in a create orchestration request in isolated
2. Also adds support for terminating a non-terminal orchestration if a creation request is issued for the same instance ID before creating the new orchestration (assuming this orchestration's status is not in the list of dedupe statuses).
3. Changes the restart orchestration functionality to first terminate any non-terminal orchestrations, if restarting with the same instance ID, rather than throwing an exception in this case.

What is ultimately used to determine the dedupe statuses is a combination of those specified in the request and those obtains from the user's app-wide [`OverridableExistingInstanceStates`](https://github.com/Azure/azure-functions-durable-extension/blob/ef23ad97746d56b05a420faddd46f77839e4fd8b/src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs#L209) setting. These two are "unioned" to obtain the strictest dedupe statuses.
 

## Why is this change needed?
1. Currently we are ignoring the dedupe statuses passed proto of the create orchestration request for isolated apps
2. And, if `OverridableExistingInstanceStates` was set to `AnyState`, we would allow just creating a new orchestration with the same instance ID without terminating any existing running orchestration, which is unsafe.

---

# Project checklist
  - [ ] Otherwise: Notes added to `release_notes.md`
  - [ ] Otherwise: Ensure EventIds are within the supported range in the existing Windows infrastructure (validate via deployed telemetry). If needed, extend the range via a PR such as https://msazure.visualstudio.com/One/_git/AAPT-Antares-Websites/pullrequest/7463263?_a=files
  - [ ] Otherwise: This change applies exclusively to `WebJobs.Extensions.DurableTask` v3.x and will be retained only in the `dev` and `main` branches
- [ ] Breaking change?
  - [ ] If yes:
    - Impact:
    - Migration guidance:
---

- [ ] No
- [x] Yes, AI helped write parts of this PR (e.g., GitHub Copilot)
- [ ] Yes, an AI agent generated most of this PR

If AI was used:
- Tool(s): GitHub Copilot
- AI-assisted areas/files: Some of the logic in the `DedupeStatusesTest`
- What you changed after AI output: Updated the test AI wrote for me to make more sense

# Notes for reviewers
1. What do we want to do if a user's dedupe statuses allows reusing running instances, but does not allow reusing a terminal instance? Then the creation request will fail after we terminate the first orchestration. Should we alter the user-passed dedupe statuses, or throw an invalid argument exception?
2. Currently, I take an `OrchestrationIdReusePolicy` of "null" to mean all statuses can be reused (equivalent to no dedupe statuses). Does this make sense? I do this because none of the other languages (that use gRPC, so Java and Python) set this field currently. In that case I just default to relying on the `OverridableExistingInstanceStates` setting.


src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
```diff
@@ -407,9 +415,45 @@ public Task CreateTaskOrchestrationAsync(TaskMessage creationMessage)
...
+        /// before a new orchestration is created.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>A task that completes when the creation message for the task orchestration instance is enqueued.</returns>
+        /// <exception cref="OrchestrationAlreadyExistsException">Thrown if an orchestration with the same instance ID already exists and its status
+        /// is in <paramref name="dedupeStatuses"/>.</exception>
+        /// <exception cref="OperationCanceledException">Thrown if the operation is cancelled via <paramref name="cancellationToken"/>.</exception>
+        public async Task CreateTaskOrchestrationAsync(TaskMessage creationMessage, OrchestrationStatus[] dedupeStatuses, CancellationToken cancellationToken)
+        {
+            using var timeoutCts = new CancellationTokenSource(TimeSpan.FromSeconds(this.OrchestrationCreationRequestTimeoutInSeconds));
+            using var linkedCts =
```
cgillum: It's redundant to have both a timeout and a cancellation token parameter and I worry that this could lead to bugs and/or unexpected behavior. I suggest removing the use of `OrchestrationCreationRequestTimeoutInSeconds` in this method.

src/WebJobs.Extensions.DurableTask/Options/DurableTaskOptions.cs
```diff
@@ -270,7 +271,22 @@ public string HubName
...
+        /// <remarks>
+        /// This setting is applicable when <see cref="OverridableExistingInstanceStates"/> is set to <see cref="OverridableStates.AnyState"/>.
+        /// If an orchestration in a non-terminal state already exists with the instance ID passed to the creation request, then this
+        /// orchestration will be terminated before the new orchestration is created. This setting controls how long the extension will wait
+        /// for the orchestration to reach a status of <see cref="OrchestrationStatus.Terminated"/> before failing the creation request.
+        /// </remarks>
+        /// <value>
+        /// The number of seconds before a creation request for an orchestration times out.
+        /// </value>
+        public int OrchestrationCreationRequestTimeoutInSeconds { get; set; } = 180;
```
cgillum: Is this needed, or can we instead rely on the caller to provide the timeout (via a user-provided cancellationToken or an HTTP request timeout)?

src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
```diff
@@ -407,9 +414,40 @@ public Task CreateTaskOrchestrationAsync(TaskMessage creationMessage)
         }
 
         /// <inheritdoc />
```
sophiatev: Should this method comment be updated to mention terminating existing instances as well? And should we change the method that doesn't exist any dedupe statuses to also terminate running orchestrations, since it effectively treats all statuses as reusable?

src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
```diff
@@ -614,10 +652,72 @@ public virtual Task<IAsyncEnumerable<HistoryEvent>> StreamOrchestrationHistoryAs
...
+                            $"{orchestrationState.OrchestrationStatus}. Since the dedupe statuses of the creation request, " +
+                            $"{string.Join(", ", dedupeStatuses)}, do not contain the orchestration's status, the orchestration has been " +
+                            $"terminated and a new instance with the same instance ID will be created.");
+
+                        while (orchestrationState != null && orchestrationState.OrchestrationStatus != OrchestrationStatus.Terminated)
+                        {
+                            cancellationToken.ThrowIfCancellationRequested();
+                            await Task.Delay(TimeSpan.FromSeconds(1), cancellationToken);
+                            orchestrationState = await this.GetOrchestrationStateAsync(instanceId, executionId: null);
+                        }
```
sophiatev: What should we do here if dedupe statuses contains terminated?  The creation call afterwards will fail. So, should we manually remove this dedupe status if it exists, or should we throw an invalid argument exception if the dedupe statuses contains terminated but also allows for reuse of a running status?

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -933,13 +935,21 @@ private async Task<HttpResponseMessage> HandleStartOrchestratorRequestAsync(
...
+                            {
+                                Event = executionStartedEvent,
+                                OrchestrationInstance = instance,
+                            },
+                            this.config.Options.OverridableExistingInstanceStates.ToDedupeStatuses(),
+                            cancellationToken);
+                    }
+                    catch (OperationCanceledException)
+                    {
+                        return request.CreateErrorResponse(HttpStatusCode.RequestTimeout, $"Create instance request cancelled for instance ID {instanceId}");
```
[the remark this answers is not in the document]
cgillum: BadRequest seems like a worse option to me. I think RequestTimeout is fine.

cgillum: Is it even worth catching and throwing here? If the request was cancelled, it doesn't do us much good to return anything.

src/WebJobs.Extensions.DurableTask/TaskHubGrpcServer.cs
```diff
@@ -54,6 +54,28 @@ public override Task<Empty> Hello(Empty request, ServerCallContext context)
         {
             try
             {
+                var allStatuses = new List<OrchestrationStatus>()
```
cgillum: It's probably best to not hardcode this since you'll have to remember to update it if we ever add a new status (like Canceled, which is not included in this list currently). There are other ways to get all values of an enum, though I don't have the specific API on hand.

src/WebJobs.Extensions.DurableTask/TaskHubGrpcServer.cs
```diff
@@ -100,6 +123,10 @@ await this.GetDurabilityProvider(context).CreateTaskOrchestrationAsync(
             catch (InvalidOperationException ex) when (ex.Message.EndsWith("already exists.")) // for older versions of DTF.AS and DTFx.Netherite
             {
                 throw new RpcException(new Status(StatusCode.AlreadyExists, $"An Orchestration instance with the ID {request.InstanceId} already exists."));
+            }
+            catch (OperationCanceledException)
+            {
+                throw new RpcException(new Status(StatusCode.Cancelled, $"Create instance request cancelled for instance ID {request.InstanceId}"));
```
cgillum: Same comment about whether it really makes sense to add this catch/throw combo.

test/Common/DurableTaskEndToEndTests.cs
```diff
@@ -5278,19 +5279,39 @@ await Assert.ThrowsAsync<FunctionInvocationException>(async () =>
             }
         }
 
+        public static IEnumerable<object[]> GetBooleanAndFullFeaturedStorageProviderOptionsWithOverridableStatesAndSuspend()
+        {
+            foreach (object[] data in GetBooleanAndFullFeaturedStorageProviderOptionsWithOverridableStates())
+            {
+                yield return new object[] { data[0], data[1], data[2], true };
```
cgillum, line 5320: Please add comments explaining what each entry in the array represents.

src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
cgillum: My recommendation:

- Throw an `ArgumentException` early (at the top of `TerminateTaskOrchestrationWithReusableRunningStatusAndWaitAsync`) if the dedupe statuses contain both a non-running status that would trigger termination AND `OrchestrationStatus.Terminated`. This is a contradictory configuration and should fail fast with a clear error message rather than silently failing later. Something like:

```csharp
  if (runningStatuses.Any(
      s => !dedupeStatuses.Contains(s)) && dedupeStatuses.Contains(OrchestrationStatus.Terminated))
    {
        throw new ArgumentException(
            "Invalid dedupe statuses: cannot include 'Terminated' while also allowing reuse of running instances, " +
            "because the running instance would be terminated and then immediately conflict with the dedupe check.");
    }
```
- Don't silently strip `Terminated` from the array — that would change the caller's intent without their knowledge, which is surprising behavior. Failing explicitly is safer and more debuggable.

src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
cgillum: Yes, let's update the documentation comments here.

My gut says we should leave the other method alone to minimize the chance of unexpected breaking behavior for customers. I don't know whether it's implicitly understood that it "treats all statuses as reusable".

src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
```diff
@@ -406,10 +409,62 @@ public Task CreateTaskOrchestrationAsync(TaskMessage creationMessage)
...
+        /// <exception cref="ArgumentException">
+        /// Thrown if <paramref name="dedupeStatuses"/> contains <see cref="OrchestrationStatus.Terminated"/> but also allows at least one running status
+        /// to be reusable. In this case, an existing orchestration with that running status would be terminated, but the creation of the new orchestration
+        /// would immediately fail due to the existing orchestration now having status <see cref="OrchestrationStatus.Terminated"/>.
+        /// </exception>
         public Task CreateTaskOrchestrationAsync(TaskMessage creationMessage, OrchestrationStatus[] dedupeStatuses)
         {
-            return this.GetOrchestrationServiceClient().CreateTaskOrchestrationAsync(creationMessage, dedupeStatuses);
+            using var timeoutCts = new CancellationTokenSource(TimeSpan.FromSeconds(this.orchestrationCreationRequestTimeoutInSeconds));
+            return this.CreateTaskOrchestrationAsync(creationMessage, dedupeStatuses, timeoutCts.Token);
```
cgillum: There's a subtle bug here where we dispose `timeoutCts` too soon. Need to change this method to be `async` and replace `return` with `await` to ensure the `timeoutCts` gets disposed only after the call to `this.CreateTaskOrchestrationAsync(...)` completes.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -1068,6 +1079,10 @@ private async Task<HttpResponseMessage> HandleRestartInstanceRequestAsync(
             {
                 return request.CreateErrorResponse(HttpStatusCode.BadRequest, "InstanceId does not match a valid orchestration instance.", e);
             }
+            catch (OrchestrationAlreadyExistsException e)
+            {
+                return request.CreateErrorResponse(HttpStatusCode.BadRequest, "A non-terminal instance with this instance ID already exists.", e);
```
cgillum: In the `StartOrchestration` call you're returning HTTP 409 (Conflict) but here you're returning HTTP 400 (Bad Request). I assume we should be using `HttpStatusCode.Conflict` here too?

cgillum (commented):
Just a couple of small things to fix, including build warnings, and I'm good with this PR.

There is a breaking behavioral change, though, so we need to include this in the release notes. Specifically, the old RestartInstance code explicitly blocked restarting running instances regardless of `OverridableExistingInstanceStates`, while the new code delegates to the dedupe mechanism, meaning `OverridableExistingInstanceStates = AnyState` could allow silently terminating and restarting a running instance. That's a change in semantics, and we need to document this clearly.

cgillum (approved):

PR #3331, opened 2026-02-04
Support Tags when RestartInstance

nytian:
This PR adds tags at RestartInstane by using the tags at `OrchestrationState`

At `CreateGetInstanceResponse` include `tags` when creating OrchestrationState at `TaskHubGrpcServer`

E2E test has been udpated -- testing tags should be included.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -174,7 +174,21 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         }
 
         /// <inheritdoc />
-        async Task<string> IDurableOrchestrationClient.StartNewAsync<T>(string orchestratorFunctionName, string instanceId, T input)
+        Task<string> IDurableOrchestrationClient.StartNewAsync<T>(string orchestratorFunctionName, string instanceId, T input)
+        {
+            return this.StartNewAsync<T>(orchestratorFunctionName, instanceId, input, tags: null);
+        }
+
+        /// <summary>
+        /// Starts a new instance of the specified orchestrator function with optional tags.
+        /// </summary>
+        /// <param name="orchestratorFunctionName">The name of the orchestrator function to start.</param>
+        /// <param name="instanceId">The ID to use for the new orchestration instance.</param>
+        /// <param name="input">JSON-serializable input value for the orchestrator function.</param>
+        /// <param name="tags">A dictionary of key-value pairs to associate with the orchestration instance.</param>
+        /// <typeparam name="T">The type of the input value for the orchestrator function.</typeparam>
+        /// <returns>A task that completes when the orchestration is started.</returns>
+        public async Task<string> StartNewAsync<T>(string orchestratorFunctionName, string instanceId, T input, IDictionary<string, string> tags)
         {
             if (this.ClientReferencesCurrentApp(this))
             {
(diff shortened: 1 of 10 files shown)


src/WebJobs.Extensions.DurableTask/TaskHubGrpcServer.cs
sophiatev, line 1: It looks like there's a line ending difference issue in this file, can you fix it?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -1150,25 +1150,51 @@ Task<string> IDurableOrchestrationClient.StartNewAsync<T>(string orchestratorFun
...
+                instanceId: restartWithNewInstanceId ? newInstanceId : instanceId,
+                input: input,
+                tags: state.Tags,
+                dedupeStatuses: dedupeStatuses);
+
+            this.traceHelper.FunctionScheduled(
+                this.TaskHubName,
+                state.Name,
+                instanceId,
+                reason: "NewInstance",
```
sophiatev: Would it make sense to make the reason "RestartedInstance" or something like that?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -1150,25 +1150,51 @@ Task<string> IDurableOrchestrationClient.StartNewAsync<T>(string orchestratorFun
 
         async Task<string> IDurableOrchestrationClient.RestartAsync(string instanceId, bool restartWithNewInstanceId)
         {
-            DurableOrchestrationStatus status = await ((IDurableOrchestrationClient)this).GetStatusAsync(instanceId, showHistory: false, showHistoryOutput: false, showInput: true);
+            OrchestrationState state = await this.GetOrchestrationInstanceStateAsync(instanceId);
```
bachuv, line 1141: Why do we need to make this update?

nytian: because the previous status doesn't have tags property. if we need to include tags, then OrchetrationState is one of the easiest options here.

test/e2e/Tests/Tests/RestartOrchestrationTests.cs
```diff
@@ -80,6 +81,14 @@ public async Task RestartOrchestration_CreatedTimeAndOutputChange(bool restartWi
...
             Assert.Equal(instanceId, restartInstanceId);
         }
+
+        HttpResponseMessage result = await HttpHelpers.InvokeHttpTrigger("RestartOrchestrator_Query_Tags", $"?id={restartInstanceId}");
+
+        // Verify that restarted orchestration instance contains tags.
+        Assert.Equal(HttpStatusCode.OK, result.StatusCode);
+        var content = await result.Content.ReadAsStringAsync();
+        Assert.Contains("output", content);
+        Assert.Contains("testtag", content);
```
bachuv, line 92: Can you add a comment that shows what we should expect `content` to look like?

test/Common/DurableTaskEndToEndTests.cs
bachuv, line 1: Should we add any unit or integration tests?

src/WebJobs.Extensions.DurableTask/TaskHubGrpcServer.cs
nytian: I think that's introduced by me VS for different encoding or what, but it should not impact anything.

test/Common/DurableTaskEndToEndTests.cs
nytian: what is the test you mentioned? If its about this PR change, I just updated the `RestartOrchestration` to include tags and  existing test `RestartOrchestration_CreatedTimeAndOutputChange` to test that tags are included when instance is restarted. So that we don't need to add other tests

bachuv (approved):

PR #3339, opened 2026-02-12
Fix ArgumentNullException when ControlQueueLengths is null in Durable…

alrod:
# Summary

## What changed?

Added a null/empty guard in `DurableTaskTargetScaler.GetScaleResultAsync()` before passing `ControlQueueLengths` to `JsonConvert.DeserializeObject()`. When `ControlQueueLengths` is null or empty, control queue metrics now default to zero instead of throwing an `ArgumentNullException`.

**Before:**
```csharp
var controlQueueLengths = JsonConvert.DeserializeObject<IReadOnlyList<int>>(serializedControlQueueLengths);
var controlQueueMessages = controlQueueLengths!.Sum();
var activeControlQueues = controlQueueLengths!.Count(x => x > 0);
```

**After:**
```csharp
IReadOnlyList<int>? controlQueueLengths = null;
if (!string.IsNullOrEmpty(serializedControlQueueLengths))
{
    controlQueueLengths = JsonConvert.DeserializeObject<IReadOnlyList<int>>(serializedControlQueueLengths);
}
var controlQueueMessages = controlQueueLengths?.Sum() ?? 0;
var activeControlQueues = controlQueueLengths?.Count(x => x > 0) ?? 0;
```

## Why is this change needed?

This bug generates **~640 million error events per day** in `ScaleControllerEvents` across **12,330+ Durable Functions apps**.

**Root cause flow:**
1. `DurableTaskMetricsProvider.GetMetricsAsync()` returns a `DurableTaskTriggerMetrics` where `ControlQueueLengths` is `null` — this happens when `PulseAsync()` throws (e.g., storage connectivity issues) or hasn't executed yet, leaving `heartbeat` as null.
2. `DurableTaskTargetScaler.GetScaleResultAsync()` passes this null string to `JsonConvert.DeserializeObject()`, whose first parameter (`value`) cannot be null → throws `ArgumentNullException: Value cannot be null. (Parameter 'value')`.
3. `ScaleManager.GetTargetScalersResultAsync()` catches and logs the exception as an error event per function per polling interval → ~640M events/day.

**Kusto evidence (ScaleControllerEvents):**

| Date | Error Count |
|------|-------------|
| 2025-08-16 | 611,058,054 |
| 2025-08-17 | 639,546,126 |
| 2025-08-18 | 641,153,320 |
| 2025-08-19 | 644,527,250 |
| 2025-08-20 | 640,640,212 |
| 2025-08-21 | 638,558,769 |

**Stack trace (from all 640M+ events):**
```
System.ArgumentNullException: Value cannot be null. (Parameter 'value')
   at Newtonsoft.Json.JsonConvert.DeserializeObject(String value, Type type, JsonSerializerSettings settings)
   at DurableTaskTargetScaler.GetScaleResultAsync(TargetScalerContext context)
   at ScaleManager.GetTargetScalersResultAsync(ScaleStatusContext context, IEnumerable`1 targetScalersToProcess)
```

**Behavioral impact:** When `ControlQueueLengths` is null, the target scaler throws instead of returning a valid vote. The `ScaleManager` catches this and skips the scaler, meaning Durable Functions apps with null heartbeats effectively have **no scaling vote** for that polling interval. With this fix, the scaler returns a vote based on activity queue length alone (orchestrator metrics treated as zero), which is the correct degraded behavior when heartbeat data is unavailable.

## Issues / work items

- Related: [ADO Work Item 34697783](https://msazure.visualstudio.com/Antares/_workitems/edit/34697783)

---

# Project checklist

- [x] This change should be added to the `v2.x` branch

---

- [x] Yes, AI helped write parts of this PR (e.g., GitHub Copilot)

**If AI was used:**

- **Tool(s):** GitHub Copilot (Claude Opus 4.6)
- **AI-assisted areas/files:** `DurableTaskTargetScaler.cs` (fix implementation), `DurableTaskTargetScalerTests.cs` (new test cases), Kusto investigation queries
- **What you changed after AI output:** Reviewed root cause analysis, validated fix logic and test coverage

**AI verification (required if AI was used):**

---

# Testing

## Automated tests

- **Result:** Passed (10/10 — 7 existing + 3 new)
- New `[InlineData]` test cases added to `DurableTaskTargetScalerTests.TestTargetScaler`:
  - `controlQueueLengths: null, workItemQueueLength: 0` → expects 0 workers
  - `controlQueueLengths: "", workItemQueueLength: 0` → expects 0 workers
  - `controlQueueLengths: null, workItemQueueLength: 5` → expects 5 workers (activity-only scaling works)

N/A — logic-only change with full unit test coverage.

---

# Notes for reviewers

- The fix is intentionally minimal — a simple null guard at the point of failure.
- When `ControlQueueLengths` is null, orchestrator worker count defaults to 0 and the vote is driven entirely by `workItemQueueLength` (activity queue). This is better than the current behavior of throwing and producing no vote at all.
- The 640M daily error volume also creates significant noise in Scale Controller telemetry, making it harder to investigate real scaling issues.


src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs
```diff
@@ -53,10 +53,14 @@ public async Task<TargetScalerResult> GetScaleResultAsync(TargetScalerContext co
                 double activityWorkers = Math.Ceiling(workItemQueueLength / (double)this.MaxConcurrentActivities);
 
                 var serializedControlQueueLengths = metrics.ControlQueueLengths;
-                var controlQueueLengths = JsonConvert.DeserializeObject<IReadOnlyList<int>>(serializedControlQueueLengths);
+                IReadOnlyList<int>? controlQueueLengths = null;
```
nytian, line 56: I think the exception is thrown because metrics/ metrics.ControlQueueLengths can be null, and passing null into JsonConvert.DeserializeObject causes an ArgumentNullException.

Thus the fix should be  checking null or empty input before calling DeserializeObject.
```C#
if (!string.IsNullOrEmpty(serializedControlQueueLengths))
{
    controlQueueLengths =
        JsonConvert.DeserializeObject<IReadOnlyList<int>>(serializedControlQueueLengths)
}
```

nytian (approved):

PR #3357, opened 2026-02-23
Lightweight Scale Controller package for Azure Storage backend

nytian:
This PR is a partial reorganization related to PR https://github.com/Azure/azure-functions-durable-extension/pull/3240

It introduces a new lightweight package `Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale` to support scaling Durable Functions within the Functions Scale Controller.

This PR also defines a basic interface to connect Durable Functions backends with the Scale Controller and adds backend service support for Azure Storage.


WebJobs.Extensions.DurableTask.sln
```diff
@@ -82,65 +82,209 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "E2E", "E2E", "{02EA681E-C7D
...
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "app", "test\e2e\Apps\BasicDotNetIsolated\app.csproj", "{DBD5C544-70AC-5928-5DD1-C02918927439}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale", "src\Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale\Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale.csproj", "{CDDAD51A-3516-401E-B0A4-7688E2787AB7}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale.Tests", "test\ScaleTests\Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale.Tests.csproj", "{F47A638F-634F-48CA-A1FC-83694327FBAE}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
+		Debug|x64 = Debug|x64
```
AnatoliB: Is this intentional? Do we need it?

test/ScaleTests/DurableTaskScaleConfigurationExtensionsTests.cs
```diff
@@ -0,0 +1,210 @@
...
+        /// Ensures WebJobs framework can discover and configure the scale extension.
+        /// </summary>
+        [Fact]
+        public void AddDurableTask_RegistersDurableTaskScaleExtension()
+        {
+            // Arrange
+            // Use TestWebJobsBuilder directly (no HostBuilder needed) - this matches how Scale Controller uses it
+            var services = new ServiceCollection();
+            services.AddSingleton<INameResolver>(new SimpleNameResolver());
+            services.AddSingleton<ILoggerFactory>(new LoggerFactory());
```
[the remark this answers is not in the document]
nytian: udpated

WebJobs.Extensions.DurableTask.sln
nytian: I already updated. I should remove it before/

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/AzureStorage/AzureStorageScalabilityProviderFactory.cs
```diff
@@ -0,0 +1,183 @@
...
+                        if (credential is global::Azure.Core.TokenCredential tokenCredential)
+                        {
+                            return tokenCredential;
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        logger?.LogWarning(ex, "Failed to extract TokenCredential from AzureComponentFactory. Using null credential instead.");
+                        return null;
+                    }
```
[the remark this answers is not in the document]
nytian: Do we really need this?

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/DurableTaskScaleConfigurationExtensions.cs
```diff
@@ -0,0 +1,141 @@
...
+            builder.AddExtension<DurableTaskScaleExtension>();
+
+            IServiceCollection serviceCollection = builder.Services;
+
+            // Register StorageServiceClientProviderFactory using factory function to ensure proper construction
+            serviceCollection.TryAddSingleton<IStorageServiceClientProviderFactory>(serviceProvider =>
+            {
+                return new StorageServiceClientProviderFactory(
+                    serviceProvider.GetRequiredService<IConfiguration>(),
+                    serviceProvider.GetRequiredService<ILoggerFactory>());
```
cgillum: nit: are you using Claude Opus for this? I'm not sure why it always uses DI this way. The simpler approach is to simply do:

```csharp
serviceCollection.TryAddSingleton<IStorageServiceClientProviderFactory, StorageServiceClientProviderFactory>()
```

No need for manually constructing the object using `serviceProvider`. It should be equivalent.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/DurableTaskTriggersScaleProvider.cs
```diff
@@ -0,0 +1,121 @@
...
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale
+{
+    /// <summary>
+    /// Provides scale monitoring and target scaling for Durable Task triggers
+    /// by delegating to backend-specific scalability providers.
+    /// </summary>
+    public class DurableTaskTriggersScaleProvider : IScaleMonitorProvider, ITargetScalerProvider
+    {
+        private const string DefaultConnectionName = "connectionName";
+        private const string ConnectionNameOverride = "connectionStringName";
```
AnatoliB: These two constants are not used?

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/AzureStorage/AzureStorageScalabilityProviderFactory.cs
```diff
@@ -0,0 +1,183 @@
...
+        /// </summary>
+        public virtual string Name => ProviderName;
+
+        /// <summary>
+        /// Gets default connection name of this provider service.
+        /// </summary>
+        public string DefaultConnectionName { get; }
+
+        /// <summary>
+        /// Creates and caches a default <see cref="ScalabilityProvider"/> instance using Azure Storage as the backend using
```
AnatoliB: I don't see it caching anything: it always returns a newly constructed provider instance, right?

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/DurableTaskMetadata.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// <summary>
+        /// Resolves app settings in <see cref="DurableTaskMetadata"/> using the provided <see cref="INameResolver"/>.
+        /// This allows configuration values such as connection strings to be expanded from environment variables or host settings.
+        /// </summary>
+        /// <param name="metadata">The scale options instance containing configuration values to resolve.</param>
+        /// <param name="nameResolver">The name resolver used to resolve app setting placeholders.</param>
+        public static void ResolveAppSettingOptions(DurableTaskMetadata metadata, INameResolver nameResolver)
+        {
+            if (metadata.StorageProvider != null &&
+                metadata.StorageProvider.TryGetValue("connectionName", out object? connectionNameObj) &&
```
AnatoliB: Copilot noticed this:

> `ResolveAppSettingOptions` only resolves `connectionName`, not `connectionStringName`.
>
> `TriggerMetadataExtensions.ResolveConnectionName` checks both `connectionName `and `connectionStringName` keys for compatibility, but `ResolveAppSettingOptions` only resolves placeholder values (e.g. `%MyConn%`) for `connectionName`. If a Scale Controller sends `connectionStringName` with an app-setting placeholder, it won't be expanded. Consider resolving both keys for consistency.

I don't know if there is a chance to actually get `connectionStringName`, but I think it would be safer to handle both, just in case. Unless you have a reason to be absolutely certain that `connectionStringName` will never happen.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/DurableTaskScaleConfigurationExtensions.cs
```diff
@@ -0,0 +1,141 @@
...
+                    var logger = loggerFactory?.CreateLogger(typeof(DurableTaskScaleConfigurationExtensions));
+                    logger?.LogError(
+                        ex,
+                        "Failed to create DurableTaskTriggersScaleProvider for function {FunctionName}. This may prevent scaling from working correctly.",
+                        triggerMetadata.FunctionName);
+                    throw;
+                }
+            });
+
+            // builder.Services.AddSingleton<IScaleMonitorProvider>(serviceProvider => serviceProvider.GetServices<DurableTaskTriggersScaleProvider>().Single(x => x == provider));
```
cgillum: What is this commented-out line of code for? Can it be removed? Are we intentionally not supporting IScaleMonitorProvider?

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/AzureStorage/AzureStorageScalabilityProviderFactory.cs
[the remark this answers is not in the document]
AnatoliB: The only reason I could say "no" would be if this is just a duplicate of code already in use in the same scenario. Otherwise, limiting exception types is a good thing here, so we don't accidentally swallow bad things like OOM exceptions.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/AzureStorage/AzureStorageScalabilityProviderFactory.cs
```diff
@@ -0,0 +1,183 @@
...
+
+            // Check if metadata contains an AzureComponentFactory wrapper
+            // ScaleController passes it as: metadata.Properties[nameof(AzureComponentFactory)] = new AzureComponentFactoryWrapper(...)
+            if (triggerMetadata.Properties.TryGetValue("AzureComponentFactory", out object? componentFactoryObj) && componentFactoryObj != null)
+            {
+                // The AzureComponentFactoryWrapper has CreateTokenCredential method
+                // Call it using reflection to get the TokenCredential
+                var factoryType = componentFactoryObj.GetType();
+                var method = factoryType.GetMethod("CreateTokenCredential");
+                if (method != null)
```
cgillum: Is it expected that `method` will ever be null here? Do we need to throw or at least log a warning when this happens?

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/AzureStorage/DurableTaskTargetScaler.cs
```diff
@@ -0,0 +1,135 @@
...
+                var controlQueueMessages = controlQueueLengths?.Sum() ?? 0;
+                var activeControlQueues = controlQueueLengths?.Count(x => x > 0) ?? 0;
+
+                // compute orchestratorWorkers: the number of workers we need to process all orchestrator messages.
+                // We bound this result to be no larger than the partition count
+                var upperBoundControlWorkers = Math.Ceiling(controlQueueMessages / (double)this.MaxConcurrentOrchestrators);
+                var orchestratorWorkers = Math.Min(activeControlQueues, upperBoundControlWorkers);
+
+                int numWorkersToRequest = (int)Math.Max(activityWorkers, orchestratorWorkers);
+                this.scaleResult.TargetWorkerCount = numWorkersToRequest;
```
cgillum: Here we're mutating shared state. Is it possible that multiple threads might call `GetScaleResultAsync` concurrently? If not, we should call that out explicitly in the method comments so that future reviewers know that there isn't risk of a race condition here.

But it looks like this is the only place we do anything with this object. Could we instead simply `return TargetScaleResult { TargetWorkerCount = numWorkersToRequest };` at the end of this method and remove the `scaleResult` field? That would make this method thread safe.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/AzureStorage/AzureStorageScalabilityProviderFactory.cs
nytian: Yes, it won’t cache anything. I will update the comments here. 

 I had it because previously I had another method here that would cache a default backend as a fallback when nothing was provided — that is, when there was no metadata from the scale controller. But after testing, that scenario seems impossible. If nothing is configured, the host will automatically choose a default backend, so triggerMetadata should never be null.

So, coming back to the question: no, it does not cache any provider. Do you think we actually need caching here?

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/AzureStorage/AzureStorageScalabilityProviderFactory.cs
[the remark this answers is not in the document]
nytian: added

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/AzureStorage/AzureStorageScalabilityProviderFactory.cs
nytian: This method refers to [this one ](https://msazure.visualstudio.com/One/_git/AAPT-Antares-ScaleController?path=/src/ScaleController/Models/AzureComponentFactoryWrapper.cs&version=GBnytian/memory-leak&line=18&lineEnd=19&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) which passed from [here ](https://msazure.visualstudio.com/One/_git/AAPT-Antares-ScaleController?path=/src/ScaleController/BuilderExtensions/RegisteredExtensionHelper.cs&version=GBnytian/memory-leak&line=120&lineEnd=121&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents).

yeah it shouldn’t be null, if so then it would mean the passed `AzureComponentFactory` is empty or broken. I’ll add a warning log here.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/DurableTaskMetadata.cs
```diff
@@ -0,0 +1,58 @@
...
+                connectionNameObj is string connectionName)
+            {
+                metadata.StorageProvider["connectionName"] = nameResolver.Resolve(connectionName) ?? string.Empty;
```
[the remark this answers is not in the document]
nytian: this doesn't need to be handled. SC wil handle the % % wrapping for us. [here ](https://msazure.visualstudio.com/One/_git/AAPT-Antares-ScaleController?path=/src/ScaleController/BuilderExtensions/ScaleControllerBuilderExtensions.cs&version=GBmain&_a=contents&line=263&lineStyle=plain&lineEnd=264&lineStartColumn=1&lineEndColumn=1)

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/DurableTaskScaleConfigurationExtensions.cs
nytian: Oh, I was trying to remove the whole scale monitor implementation at first because they previously mentioned that Scale Monitor is no longer used in SC v3, so in theory we could remove it. But I’m not totally sure if we should tbh — in case something goes wrong and we need to revert back later.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/AzureStorage/DurableTaskTargetScaler.cs
```diff
@@ -0,0 +1,132 @@
...
+                return new TargetScalerResult { TargetWorkerCount = numWorkersToRequest };
+            }
+            catch (Exception ex)
+            {
+                // We want to augment the exception with metrics information for investigation purposes
+                var metricsLog = $"Metrics: workItemQueueLength={metrics?.WorkItemQueueLength}. controlQueueLengths={metrics?.ControlQueueLengths}. " +
+                    $"maxConcurrentOrchestrators={this.MaxConcurrentOrchestrators}. maxConcurrentActivities={this.MaxConcurrentActivities}";
+                var errorLog = $"Error: target worker count for '{this.scaler}' resulted in exception. " + metricsLog;
+                throw new Exception(errorLog, ex);
+            }
```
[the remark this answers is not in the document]
nytian: udpated.

cgillum (approved):
It looks like the CI is failing, but I'm otherwise good with these changes.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/AzureStorage/AzureStorageScalabilityProviderFactory.cs
AnatoliB: @nytan No, I don't think we need caching, I just wanted to make sure the comment agrees with the actual implementation.

AnatoliB (approved):

bachuv (approved):

PR #3373, opened 2026-03-02
Dispose DurableTaskTriggersScaleProvider  to prevent Azure Managed provider leaks

nytian:
This change makes `DurableTaskTriggersScaleProvider` disposable so backend resources are released when scale hosts are rebuilt.

For Azure Managed backends, this ensures `EventSourceListener` instances are disposed , preventing memory/thread growth across repeated scale-controller reinitializations.


AnatoliB (approved):
I agree with Copilot's comment (https://github.com/Azure/azure-functions-durable-extension/pull/3373/changes#r2875087442) but will not block on this.

AnatoliB (approved):

PR #3393, opened 2026-03-17
Fix gRPC listener resilience for transient sidecar failures

berndverst:
## Summary

When the local gRPC sidecar is temporarily unavailable, `BindingHelper.DurableOrchestrationClientToString` threw `InvalidOperationException("The local RPC address has not been configured!")`. The Azure Functions host treats this as a **non-retryable application error**, which is incorrect for what is fundamentally a transient infrastructure condition.

For **queue-triggered functions** with `[DurableClient]` input bindings, this is especially damaging: the rapid, non-retryable failures cause messages to exhaust their dequeue count (default: 5) within seconds and move to **poison queues**, resulting in data loss that requires manual reprocessing.

The failure occurs in the host-side input binding pipeline — **before user function code executes** — so no application-level retry strategy (e.g., Polly) can work around it.

### Related issue
- Azure/durabletask#1306

## Root Cause Analysis

The error is thrown at [`BindingHelper.cs:37`](src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs) when `GetLocalRpcAddress()` returns `null`. Through investigation, I identified the following contributing scenarios:

1. **gRPC server fails to start** (e.g., port conflict in `GetFreeTcpPort` TOCTOU race) — `ListenAddress` is never set
2. **gRPC server stops unexpectedly** — `StopAsync()` did not reset `host` or `ListenAddress`, so `StartAsync()` fast-path would skip re-initialization, permanently leaving a dead server
3. **`ListenAddress` set to `null` after Kestrel starts** — if `IServerAddressesFeature` returns empty addresses, the server appears "started" but `ListenAddress` is `null`, and subsequent calls hit the fast-path and never retry

In all cases, the `InvalidOperationException` exception type made recovery impossible at every layer.

### Fix 1: Platform-level exception + wait-with-timeout

**`GrpcChannelTemporarilyUnavailableException.cs`** (new) — A dedicated `[Serializable]` exception type for transient gRPC sidecar unavailability. This is recognized as a platform-level error by the durable middleware, causing orchestrations/activities to be safely aborted and retried by the backend rather than permanently failed.

**`BindingHelper.cs`** — Replaced `InvalidOperationException` with `GrpcChannelTemporarilyUnavailableException`. The message clearly describes the transient sidecar unavailability condition.

**`OutOfProcMiddleware.cs`** — Added `GrpcChannelTemporarilyUnavailableException` to `IsPlatformLevelException()`, following the same pattern as PR #3355. When this exception is the `InnerException` of a function invocation failure, the middleware aborts the dispatch and triggers a durable retry instead of marking the orchestration as failed.

**`DurableTaskExtension.cs` → `GetLocalRpcAddress()`** — Instead of returning `null` immediately when `ListenAddress` is not set, the method now:
1. Returns immediately if the address is already available (fast path)
2. Attempts to restart the gRPC listener via `EnsureStartedAsync()`
3. Waits up to 30 seconds for the listen address to become available via `WaitForListenAddressAsync()`
4. Only returns `null` (triggering the exception) after exhausting the timeout

### Fix 2: Health checking + self-healing restart for the gRPC listener

**`ILocalGrpcListener.cs`** — Added three new interface members:
- `bool IsHealthy` — checks if the gRPC server is alive and responsive
- `Task EnsureStartedAsync()` — verifies health and restarts if needed
- `Task<string?> WaitForListenAddressAsync()` — waits with timeout for the address to become available

**`AspNetCoreLocalGrpcListener.cs`** — Major resilience improvements:
- **`IsHealthy` property**: Checks if the Kestrel `IServer` is still reporting addresses
- **`StartAsync()` now checks health**: If the host exists but is unhealthy, it stops and restarts (previously skipped via fast-path)
- **`StopAsync()` resets state**: Sets `host = null` and `ListenAddress = null`, and re-creates the `TaskCompletionSource` so subsequent `StartAsync()` calls properly reinitialize and waiters don't receive stale addresses
- **`StopInternalAsync()` threads `CancellationToken`**: The cancellation token is forwarded to the inner `host.StopAsync()` to avoid shutdown hangs
- **`EnsureStartedAsync()`**: Logs a warning and delegates to `StartAsync()` for health-aware restart
- **`WaitForListenAddressAsync()`**: Uses a `TaskCompletionSource<string>` to efficiently wait for the address without polling; the TCS is signaled upon successful startup
- **Error-resilient `StopInternalAsync()`**: Gracefully handles errors during stop so cleanup always completes

### Behavior comparison

| Scenario | Before | After |
|---|---|---|
| gRPC server not started | `InvalidOperationException` (instant, non-retryable) | Attempts restart, waits up to 30s, then `GrpcChannelTemporarilyUnavailableException` (platform-level, retriable) |
| gRPC server crashed | Fast-path skip (stale host ref), `ListenAddress` null forever | Health check detects dead server, restarts it |
| `StopAsync()` then `StartAsync()` | Skipped (host ref was non-null) | Properly resets state, allows clean restart |
| Queue-triggered function binding failure | Poison queue after ~5 rapid failures | Wait-with-retry gives server time to recover; platform-level exception triggers durable retry |
| Orchestration during gRPC unavailability | Orchestration marked as permanently failed | `IsPlatformLevelException` aborts dispatch → durable backend retries safely |

## Tests

Added 13 new tests to `LocalGrpcListenerTests.cs` and 1 to `OutOfProcMiddlewareTests.cs` (all pass on both net8.0 and net10.0):

**gRPC Listener tests:**
| Test | What it verifies |
|---|---|
| `IsHealthy_FalseBeforeStart` | `IsHealthy` is false before `StartAsync` |
| `IsHealthy_TrueAfterStart_FalseAfterStop` | `IsHealthy` tracks lifecycle correctly |
| `CanRestartAfterStop` | `StartAsync` works after `StopAsync` (new behavior) |
| `EnsureStarted_NoOpWhenHealthy` | No-op when already healthy |
| `EnsureStarted_RestartsWhenUnhealthy` | Restarts after stopped state |
| `WaitForAddress_ReturnsImmediatelyWhenReady` | Instant return when address available |
| `WaitForAddress_ReturnsWhenStartedConcurrently` | Waiter unblocks on concurrent start |
| `WaitForAddress_ReturnsNullOnTimeout` | Returns null on timeout (doesn't hang) |
| `WaitForAddress_ReturnsNullOnCancellation` | Respects cancellation token |
| `ConcurrentStartAsync_IsSafe` | 5 concurrent starts succeed safely |
| `StartAsync_IdempotentWhenHealthy` | Repeated starts return same address |
| `GetLocalRpcAddress_ReturnsAddressWhenStarted` | End-to-end via `DurableTaskExtension` |
| `BindingHelper_ThrowsGrpcChannelTemporarilyUnavailableException` | Confirms correct exception type when gRPC address unavailable |

**OutOfProcMiddleware test:**
| Test | What it verifies |
|---|---|
| `PlatformLevelExceptions` (new theory case) | `GrpcChannelTemporarilyUnavailableException` as InnerException triggers `SessionAbortedException` (safe abort + retry) |

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -527,11 +527,15 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 {
                     try
                     {
-                        this.localGrpcListener.EnsureStartedAsync(default).GetAwaiter().GetResult();
+                        // Use the host stopping token so the wait is canceled promptly during shutdown
+                        // rather than blocking for up to 30 seconds with a default token.
+                        CancellationToken stoppingToken = this.HostLifetimeService.OnStopping;
+
+                        this.localGrpcListener.EnsureStartedAsync(stoppingToken).GetAwaiter().GetResult();
 
                         // Wait up to 30 seconds for the listen address to become available.
                         address = this.localGrpcListener.WaitForListenAddressAsync(
-                            TimeSpan.FromSeconds(30), default).GetAwaiter().GetResult();
+                            TimeSpan.FromSeconds(30), stoppingToken).GetAwaiter().GetResult();
                     }
                     catch (Exception ex)
                     {
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -511,7 +511,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         }
 
 #nullable enable
-        internal string GetLocalRpcAddress()
+        internal string? GetLocalRpcAddress()
         {
             if (this.OutOfProcProtocol == OutOfProcOrchestrationProtocol.MiddlewarePassthrough)
             {
@@ -543,7 +543,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     }
                 }
 
-                return address!;
+                return address;
             }
 
             return this.HttpApiHandler.GetBaseUrl();
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -510,15 +510,45 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             }
         }
 
+#nullable enable
         internal string GetLocalRpcAddress()
         {
             if (this.OutOfProcProtocol == OutOfProcOrchestrationProtocol.MiddlewarePassthrough)
             {
-                return this.localGrpcListener.ListenAddress;
+                string? address = this.localGrpcListener?.ListenAddress;
+                if (address != null)
+                {
+                    return address;
+                }
+
+                // The address is not yet available. This can happen if the gRPC server failed to start,
+                // stopped unexpectedly, or hasn't finished initializing. Try to (re)start it.
+                if (this.localGrpcListener != null)
+                {
+                    try
+                    {
+                        this.localGrpcListener.EnsureStartedAsync(default).GetAwaiter().GetResult();
+
+                        // Wait up to 30 seconds for the listen address to become available.
+                        address = this.localGrpcListener.WaitForListenAddressAsync(
+                            TimeSpan.FromSeconds(30), default).GetAwaiter().GetResult();
+                    }
+                    catch (Exception ex)
+                    {
+                        this.TraceHelper.ExtensionWarningEvent(
+                            this.Options.HubName,
+                            instanceId: string.Empty,
+                            functionName: string.Empty,
+                            message: $"Failed to start/restart local gRPC listener: {ex.Message}");
+                    }
+                }
+
+                return address!;
             }
 
             return this.HttpApiHandler.GetBaseUrl();
         }
+#nullable restore
 
         internal DurabilityProvider GetDurabilityProvider(DurableClientAttribute attribute)
         {
--- a/src/WebJobs.Extensions.DurableTask/Grpc/AspNetCoreLocalGrpcListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/Grpc/AspNetCoreLocalGrpcListener.cs
@@ -108,22 +108,19 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Grpc
                 return this.ListenAddress;
             }
 
-            // Wait for the TaskCompletionSource to be signaled or timeout
+            // Wait for the TaskCompletionSource to be signaled, with a timeout.
+            // Task.WhenAny does not throw when the delay is canceled, so we check
+            // which task completed to distinguish success from timeout.
             using var timeoutCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
             timeoutCts.CancelAfter(timeout);
 
-            try
+            Task<string> addressTask = this.addressReady.Task;
+            Task delayTask = Task.Delay(Timeout.Infinite, timeoutCts.Token);
+            Task completedTask = await Task.WhenAny(addressTask, delayTask);
+
+            if (completedTask == addressTask)
             {
-                Task<string> addressTask = this.addressReady.Task;
-                Task completedTask = await Task.WhenAny(addressTask, Task.Delay(Timeout.Infinite, timeoutCts.Token));
-                if (completedTask == addressTask)
-                {
-                    return await addressTask;
-                }
-            }
-            catch (OperationCanceledException) when (!cancellationToken.IsCancellationRequested)
-            {
-                // Timeout expired, not caller cancellation
+                return await addressTask;
             }
 
             return this.ListenAddress;
@@ -195,9 +192,24 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Grpc
             this.host = newHost;
 
             // Signal any waiters that the address is now available.
+            // If ListenAddress is null (e.g. IServerAddressesFeature returned no addresses),
+            // stop the host to avoid a partially-initialized state that would cause
+            // repeated stop/restart loops from the IsHealthy check.
             if (this.ListenAddress != null)
             {
                 this.addressReady.TrySetResult(this.ListenAddress);
+            }
+            else
+            {
+                this.extension.TraceHelper.ExtensionWarningEvent(
+                    this.extension.Options.HubName,
+                    instanceId: string.Empty,
+                    functionName: string.Empty,
+                    message: "gRPC listener started but ListenAddress is null. Stopping to allow retry.");
+
+                // Roll back: stop the host so the next StartAsync call can try again
+                // rather than leaving the listener in a broken state.
+                await this.StopInternalAsync();
             }
         }
 
--- a/src/WebJobs.Extensions.DurableTask/Grpc/AspNetCoreLocalGrpcListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/Grpc/AspNetCoreLocalGrpcListener.cs
@@ -23,6 +23,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Grpc
         private const string HostName = "127.0.0.1";
         private readonly DurableTaskExtension extension;
         private readonly SemaphoreSlim startLock = new SemaphoreSlim(1, 1);
+        private readonly TaskCompletionSource<string> addressReady = new TaskCompletionSource<string>(TaskCreationOptions.RunContinuationsAsynchronously);
         private IHost? host;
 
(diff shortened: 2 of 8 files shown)


src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -662,6 +662,7 @@ private static bool IsPlatformLevelException(Exception? exception)
             return exception is Host.FunctionTimeoutException
                 || exception is Host.FunctionTimeoutAbortException
                 || exception?.InnerException is SessionAbortedException // see RemoteOrchestrationContext.TrySetResultInternal for details on OOM-handling
+                || exception?.InnerException is GrpcChannelTemporarilyUnavailableException
```
andystaples, line 666: I think this should be `exception is ...` not `exception?.InnerException is ...`, similar to the other Host exceptions which are thrown in-process.  Pretty sure the latter syntax is for exceptions thrown by the worker. Can you run a local test that makes DurableOrchestrationClientToString always throw the new error, and verify this code path is hit?

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -405,12 +405,12 @@ void SetErrorResult(FailureDetails failureDetails)
...
-                    // In this case we want to make sure to retry this entity's execution rather than marking it as failed.
-                    if (functionResult.Exception is Host.FunctionTimeoutAbortException)
-                    {
-                        throw functionResult.Exception;
+                {
+                    // This exception is thrown when another Function on the worker exceeded the Function timeout.
+                    // In this case we want to make sure to retry this entity's execution rather than marking it as failed.
+                    if (functionResult.Exception is Host.FunctionTimeoutAbortException)
+                    {
+                        throw functionResult.Exception;
```
andystaples, line 413: What changed here?

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -561,12 +561,12 @@ public async Task CallActivityAsync(DispatchMiddlewareContext dispatchContext, F
...
-                    // In this case we want to make sure to retry this Activity's execution rather than marking it as failed.
-                    if (result.Exception is Host.FunctionTimeoutAbortException)
-                    {
-                        throw result.Exception;
+                {
+                    // This exception is thrown when another Function on the worker exceeded the Function timeout.
+                    // In this case we want to make sure to retry this Activity's execution rather than marking it as failed.
+                    if (result.Exception is Host.FunctionTimeoutAbortException)
+                    {
+                        throw result.Exception;
```
andystaples, line 569: Same here - can you double-check we don't need to modify these methods with any GrpcChannelTemporarilyUnavailableException logic?

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
berndverst: I have no idea why it is formatted like this - don't blame me for GitHub doing this :)

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
berndverst: Good catch! Updated to check **both** paths:
- `exception is GrpcChannelTemporarilyUnavailableException` — for the in-process case where the binding pipeline throws it directly as the top-level exception
- `exception?.InnerException is GrpcChannelTemporarilyUnavailableException` — for cases where the host wraps it (e.g. `FunctionInvocationException` wrapping the binding error)

Also added a test case for the top-level exception in `OutOfProcMiddlewareTests.PlatformLevelExceptions`.

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
berndverst: Nothing changed here — this is a GitHub diff rendering artifact. The file was modified elsewhere (the `IsPlatformLevelException` method and the `CallActivityAsync` method), and GitHub is showing additional context lines around those changes.

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
berndverst: The activity path doesn't need this change. The orchestrator path (where `IsPlatformLevelException` is called) is the right place since a QueueTrigger function with `[DurableClient]` that starts an orchestration is where the binding error occurs, and the orchestrator middleware handles the resulting invocation failure.

src/WebJobs.Extensions.DurableTask/Grpc/AspNetCoreLocalGrpcListener.cs
```diff
@@ -32,10 +33,34 @@ public AspNetCoreLocalGrpcListener(DurableTaskExtension extension)
 
         public string? ListenAddress { get; private set; }
 
+        public bool IsHealthy
+        {
+            get
+            {
+                if (this.host is not { } currentHost)
```
nytian, line 40: what is the `currentHost` here?

berndverst: `this.host is not { } currentHost` is a C# pattern match — it's equivalent to `if (this.host == null) return false; var currentHost = this.host;`. The `is not { }` pattern checks for null, and the `currentHost` capture variable gives us a local copy to work with safely (avoids the field being set to null between the null check and the `Services.GetService` call).

src/WebJobs.Extensions.DurableTask/Grpc/AspNetCoreLocalGrpcListener.cs
```diff
@@ -55,6 +81,51 @@ public async Task StartAsync(CancellationToken cancellationToken)
...
+            Task<string> addressTask = this.addressReady.Task;
+            Task delayTask = Task.Delay(Timeout.Infinite, timeoutCts.Token);
+            Task completedTask = await Task.WhenAny(addressTask, delayTask);
+
+            if (completedTask == addressTask)
+            {
+                return await addressTask;
+            }
+
+            return this.ListenAddress;
```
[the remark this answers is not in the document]
berndverst: Fixed — the timeout/cancellation path now returns `null` explicitly instead of reading `this.ListenAddress`, which avoids the race where a concurrent start could set the address between the `Task.WhenAny` check and the return.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -510,15 +510,45 @@ internal void ConfigureForGrpcProtocol()
...
+                // stopped unexpectedly, or hasn't finished initializing. Try to (re)start it.
+                if (this.localGrpcListener != null)
+                {
+                    try
+                    {
+                        this.localGrpcListener.EnsureStartedAsync(default).GetAwaiter().GetResult();
+
+                        // Wait up to 30 seconds for the listen address to become available.
+                        address = this.localGrpcListener.WaitForListenAddressAsync(
+                            TimeSpan.FromSeconds(30), default).GetAwaiter().GetResult();
```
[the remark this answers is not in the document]
berndverst: Fixed — now uses `this.HostLifetimeService.OnStopping` instead of `default`, so the 30s wait is canceled promptly during host shutdown rather than needlessly blocking.

src/WebJobs.Extensions.DurableTask/Grpc/AspNetCoreLocalGrpcListener.cs
```diff
@@ -119,16 +190,66 @@ private async Task StartInternalAsync(CancellationToken cancellationToken)
...
+                }
 
-            return Task.CompletedTask;
+                // Reset state so StartAsync can re-initialize.
+                // Re-create the TCS so that WaitForListenAddressAsync callers
+                // don't receive a stale address from a previous start cycle.
+                this.host = null;
+                this.ListenAddress = null;
+                this.addressReady = new TaskCompletionSource<string>(TaskCreationOptions.RunContinuationsAsynchronously);
+            }
```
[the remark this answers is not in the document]
berndverst: Fixed — `StopInternalAsync` now disposes the previous `IHost` in a `finally` block after stopping it, so server, sockets, and DI container resources are always released even if `StopAsync` throws.

src/WebJobs.Extensions.DurableTask/Grpc/AspNetCoreLocalGrpcListener.cs
berndverst: This was copilot replying for me by the way :)

andystaples (approved):

nytian (approved):

berndverst:
I pushed this PR also to a branch `beverst/grpcSidecarResiliency`. I triggered the MSSQL smoketest on that branch:

https://github.com/Azure/azure-functions-durable-extension/actions/runs/23266893289

PR #3395, opened 2026-03-18
Lightweight scale controller package for Mssql backend

nytian:
As titled. This PR adds support for MSSQL backend at the lightweight scale controller package


src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/Sql/SqlServerScaleMonitor.cs
```diff
@@ -0,0 +1,89 @@
...
+    /// </summary>
+    public class SqlServerScaleMonitor : IScaleMonitor<SqlServerScaleMetric>
+    {
+        private static readonly ScaleStatus ScaleInVote = new ScaleStatus { Vote = ScaleVote.ScaleIn };
+        private static readonly ScaleStatus NoScaleVote = new ScaleStatus { Vote = ScaleVote.None };
+        private static readonly ScaleStatus ScaleOutVote = new ScaleStatus { Vote = ScaleVote.ScaleOut };
+
+        private readonly SqlServerMetricsProvider metricsProvider;
+        private int? previousWorkerCount = -1;
+
```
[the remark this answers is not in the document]
nytian: I don;t know why it's set to -1 but this is what we do at sql repo before https://github.com/microsoft/durabletask-mssql/blob/main/src/DurableTask.SqlServer.AzureFunctions/SqlScaleMonitor.cs#L24

.github/workflows/validate-build-scale.yml
```diff
@@ -62,5 +63,71 @@ jobs:
...
         done
 
-        # Run tests
-        dotnet test ./test/ScaleTests/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale.Tests.csproj -c $config --no-build --verbosity normal
\ No newline at end of file
+        # Run tests (exclude AzureManaged and SqlServer tests)
+        dotnet test ./test/ScaleTests/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale.Tests.csproj -c $config --no-build --verbosity normal --filter "FullyQualifiedName!~AzureManaged&FullyQualifiedName!~SqlServer"
+
+  mssql:
+    name: Functions Scale Tests - MSSQL
```
cgillum, line 160: Are these workflow updates related to the scale controller package? Just trying to make sure I understand what to look for.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/Sql/SqlServerMetricsProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// <summary>
+    /// Metrics provider for SQL Server backend scaling.
+    /// Provides recommended replica count based on SQL Server orchestration service metrics.
+    /// Note: This class should be kept in sync with SqlMetricsProvider in DurableTask.SqlServer.AzureFunctions.
+    /// </summary>
+    public class SqlServerMetricsProvider
+    {
+        private readonly SqlOrchestrationService service;
+        private DateTime metricsTimeStamp = DateTime.MinValue;
+        private SqlServerScaleMetric metrics;
```
cgillum: I suggest we renamed this to `cachedMetrics` to make it clear that this is a cached value. Let's also update the timestamp variable name to better reflect how it's used, e.g. "cachedMetricsLastRefreshTime" or something like that.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/Sql/SqlServerMetricsProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                // GetRecommendedReplicaCountAsync will write a trace if the recommendation results
+                // in a worker count that is different from the worker count we pass in as an argument.
+                int recommendedReplicaCount = await this.service.GetRecommendedReplicaCountAsync(
+                    previousWorkerCount,
+                    CancellationToken.None);
+
+                this.metricsTimeStamp = DateTime.UtcNow;
+                this.metrics = new SqlServerScaleMetric { RecommendedReplicaCount = recommendedReplicaCount };
+            }
```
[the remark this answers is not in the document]
cgillum: I don't know if there's a serious data integrity race here, but I agree we should use an async lock (SemaphoreSlim, etc.) to at least ensure we're doing doing multiple concurrent calls to `GetRecommendedReplicaCountAsync`, which could be quite expensive.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/Sql/SqlServerScalabilityProviderFactory.cs
```diff
@@ -0,0 +1,149 @@
...
+            // Create SQL Server orchestration service settings - following durabletask-mssql pattern
+            // Connection string should include authentication method (e.g., Authentication=Active Directory Default)
+            var settings = new SqlOrchestrationServiceSettings(
+                connectionString,
+                taskHubName)
+            {
+                // Set concurrency limits from trigger metadata (from Scale Controller payload)
+                // Default: 10 times the number of processors on the current machine
+                MaxActiveOrchestrations = metadata?.MaxConcurrentOrchestratorFunctions ?? (Environment.ProcessorCount * 10),
+                MaxConcurrentActivities = metadata?.MaxConcurrentActivityFunctions ?? (Environment.ProcessorCount * 10),
```
cgillum: Do we have a setting for MaxConcurrentEntities? I'm not sure if we do but wanted to check.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/Sql/SqlServerScalabilityProviderFactory.cs
```diff
@@ -0,0 +1,149 @@
...
+
+            // Create SQL Server orchestration service settings - following durabletask-mssql pattern
+            // Connection string should include authentication method (e.g., Authentication=Active Directory Default)
+            var settings = new SqlOrchestrationServiceSettings(
+                connectionString,
+                taskHubName)
+            {
+                // Set concurrency limits from trigger metadata (from Scale Controller payload)
+                // Default: 10 times the number of processors on the current machine
+                MaxActiveOrchestrations = metadata?.MaxConcurrentOrchestratorFunctions ?? (Environment.ProcessorCount * 10),
```
cgillum: This code runs on the scale controller, right? If so, doesn't `Environment.ProcessorCount` return the number of CPUs on the frontend role (where the Scale Controller is hosted)? Seems we shouldn't be falling back to that value, if so.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/Sql/SqlServerScalabilityProviderFactory.cs
```diff
@@ -0,0 +1,149 @@
...
+                MaxConcurrentActivities = metadata?.MaxConcurrentActivityFunctions ?? (Environment.ProcessorCount * 10),
+            };
+
+            // Note: When connection string includes "Authentication=Active Directory Default" or
+            // "Authentication=Active Directory Managed Identity", SQL Server will automatically use
+            // the appropriate Azure identity (managed identity in Azure, or DefaultAzureCredential locally).
+            // So we don't need to extract token credential here from site metadata.
```
cgillum, line 119: Regarding this comment, is it still true in the case where this code runs on the scale controller? We should probably update these comments to talk about behavior on the scale controller since that's the most important architectural consideration of this code.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/Sql/SqlServerScalabilityProviderFactory.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// <summary>
+        /// Validates SQL Server specific metadata.
+        /// </summary>
+        private void ValidateSqlServerMetadata(DurableTaskMetadata metadata)
+        {
+            // Validate hub name (SQL Server has less strict requirements than Azure Storage)
+            if (string.IsNullOrWhiteSpace(metadata.TaskHubName))
+            {
+                // Hub name defaults to "default" for SQL Server, so this is acceptable
+                return;
```
cgillum: Why do we return here? I'm a bit confused about what this if-statement is intended for. The comment says that we're trying to validate the task hub name, but we don't seem to be doing that at all. But even if the task hub name is unspecified, I would expect we'd still need to do the other validation steps below to verify the concurrency configuration values.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/Sql/SqlServerScaleMonitor.cs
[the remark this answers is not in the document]
cgillum: Looking at the `SqlServerMetricsProvider.GetMetrics()` code, it seems initializing this value to `null` is the correct/expected behavior. It's possible that the existing MSSQL code is buggy in this way.

.github/workflows/validate-build-scale.yml
nytian: This is just to run the tests I added under test/ScaleTests, which include scalability provider setup unit tests and target scaler tests.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/Sql/SqlServerScalabilityProviderFactory.cs
nytian: yeah I remember I added it mainly becasue our current pkg( all backend durability provider)  contains this default settings. Code [here](https://github.com/microsoft/durabletask-mssql/blob/ccd104375a11fadd1113a76cf4f1668a997c964c/src/DurableTask.SqlServer/SqlOrchestrationServiceSettings.cs#L84).  So I should change to a different default value?

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/Sql/SqlServerScalabilityProviderFactory.cs
nytian: It seems we don't have this for [sql](https://github.com/microsoft/durabletask-mssql/blob/ccd104375a11fadd1113a76cf4f1668a997c964c/src/DurableTask.SqlServer/SqlOrchestrationServiceSettings.cs#L84).  Do we need to consider entities for scaling? I didn't add entities but only consider activity functions and orchestrations for all backends now.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/Sql/SqlServerScalabilityProviderFactory.cs
nytian: Yeah, that’s the tricky part, and what I’m confused about it. Thanks for catching that. I could not find any MSI authentication handling in tout current  durabletask sql package, so I am not sure how this case was handled. But I tested a real Durable Functions app with MSSQL using MSI, and it worked. That is why I did not add any specific MSI handling here.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/Sql/SqlServerScalabilityProviderFactory.cs
nytian: Sorry  the code here is not very elegant. I think I wrote it like this is because if the hub name is null or not set, we can still back to the default hub name, `default`. Basically I was trying to follow the logic [here](https://github.com/microsoft/durabletask-mssql/blob/ccd104375a11fadd1113a76cf4f1668a997c964c/src/DurableTask.SqlServer.AzureFunctions/SqlDurabilityOptions.cs#L62). And I was mainly not sure when the customer doesn’t set a hub name, whether the scale controller would pass an empty string or the default name defined in the SQL package. That's also why I didn't throw exception here. I will try test this in private stamp.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/Sql/SqlServerScalabilityProviderFactory.cs
cgillum: Maybe not for SQL since SQL doesn't support entities for out-of-proc languages, but we'll need it for DTS.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/Sql/SqlServerScalabilityProviderFactory.cs
cgillum: `Environment.ProcessorCount` only makes sense if we're running on the worker because it's a hint about how much concurrent processing the worker can handle. If the scale logic runs on the scale controller, we don't have this hint. `Environment.ProcessorCount` returns the *scale controller's* CPUs and not the worker's CPUs, making it useless in this package. We should instead hardcode the default value to something like 10 so that we don't mislead people into misunderstanding how this setting works.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/Sql/SqlServerScalabilityProviderFactory.cs
cgillum: The defaulting behavior makes sense to me, but my concerns is that we're not actually validating in the non-null case. My other concern is that by returning early for the null/whitespace case, we're skipping the other validations that happen further down.

cgillum: Looks like Copilot provided the same feedback as me in its most recent round of feedback.

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/Sql/SqlServerScalabilityProviderFactory.cs
nytian: yes DTS ounts entities. https://github.com/Azure/azure-functions-durable-extension/pull/3387/changes#diff-4b3f260ce6808614e702952db963c75556f27210c7ea6e3931f36f2f84eef687R56

.github/workflows/validate-build-scale.yml
```diff
@@ -78,5 +79,82 @@ jobs:
...
+        run: |
+          echo "Waiting for SQL Server to be ready..."
+          for i in {1..60}; do
+            if nc -z localhost 1433; then
+              echo "SQL Server is ready"
+              break
+            fi
+            echo "SQL Server is not ready, waiting ($i/30)"
+            sleep 2
+          done
```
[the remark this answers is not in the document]
nytian: nit

src/Microsoft.Azure.WebJobs.Extensions.DurableTask.FunctionsScale/Sql/SqlServerScalabilityProviderFactory.cs
cgillum: @nytian Just to be sure, I asked Copilot to take a look, and here's what it said:

> SqlServerScalabilityProviderFactory still assumes SQL managed-identity auth will “just work” from the scale controller by passing only a connection string into SqlOrchestrationService
(...SqlServerScalabilityProviderFactory.cs:91-120, especially 116-119). Unlike Azure Storage, it never extracts the app/site TokenCredential from triggerMetadata.Properties (
AzureStorageScalabilityProviderFactory.cs:70-71,95-123). So Authentication=Active Directory Default/Managed Identity would use the scale-controller process identity, not the customer
app identity. That means MI-backed SQL apps likely still won’t scale correctly.

Can you expand a bit more on how you tested?

cgillum (approved):

PR #3419, opened 2026-04-08
Add dependency version validation pipeline

YunchuWang:
## Summary

Adds a GitHub Actions workflow (dep-version-validation.yml) that validates dependency version bumps in the extension csproj files do not break the Durable Functions extension packages at build or runtime.

## What this pipeline does

1. **Packs both extension packages** (WebJobs + Worker) from source into a local NuGet feed
2. **Builds a .NET isolated Function App** using \PackageReference\ to the local Worker extension package, with \packageSourceMapping\ to ensure the extension comes from the local feed
3. **Starts Azurite** (Azure Storage emulator)
4. **Runs the Function App** with \unc host start\ and triggers a HelloCities orchestration
5. **Validates** orchestration output contains expected greetings
6. **Verifies loaded extension assembly versions** match the locally built version

## Trigger paths

- \src/Worker.Extensions.DurableTask/Worker.Extensions.DurableTask.csproj\
- \src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj\
- \	est/SmokeTests/DepValidation/**\

## Motivation

The existing smoke tests (e.g., \smoketest-dotnet-isolated-v4.yml\) use \ProjectReference\ for the Worker extension. While existing smoke tests trigger on \src/**\ changes (which covers dep changes since deps are inline in csproj files), they don't validate the NuGet package resolution flow that real users experience.

This pipeline specifically targets dependency version changes and validates the full NuGet package → Function App → orchestration flow. Mirrors the [SDK-PR-Validation pipeline](https://msazure.visualstudio.com/DefaultCollection/One/_git/AAPT-DTMB?path=/.pipelines/CI/SDK-PR-Validation.yaml) pattern from AAPT-DTMB.


test/SmokeTests/DepValidation/Isolated/DepValidationApp.csproj
```diff
@@ -0,0 +1,42 @@
...
+    <FunctionsEnableWorkerIndexing>false</FunctionsEnableWorkerIndexing>
+    <ManagePackageVersionsCentrally>false</ManagePackageVersionsCentrally>
+    <!-- Default version for standalone restore (e.g. Dependency Submission workflow).
+         The CI pipeline bumps the patch version which only exists in local-packages. -->
+    <SmokeTestExtVersion Condition="'$(SmokeTestExtVersion)' == ''">1.16.3</SmokeTestExtVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!-- Azure Functions host packages (from nuget.org) -->
+    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="2.51.0" />
```
bachuv: Can we use central package management here instead of setting a specific version so it matches the version that other smoke tests are using? Example: https://github.com/Azure/azure-functions-durable-extension/blob/9d2bcb4f183ea20a2fc9af5e781b536099646d34/test/SmokeTests/OOProcSmokeTests/DotNetIsolated/DotNetIsolated.csproj#L12

.github/workflows/dep-version-validation.yml
```diff
@@ -0,0 +1,245 @@
+name: Dependency Version Validation
+
+on:
+  push:
+    branches: [ main, dev ]
+    paths:
+      - 'src/Worker.Extensions.DurableTask/Worker.Extensions.DurableTask.csproj'
+      - 'src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj'
+      - 'Directory.Packages.props'
+      - 'test/SmokeTests/DepValidation/**'
```
bachuv, line 14: Should we add https://github.com/Azure/azure-functions-durable-extension/blob/dev/test/Directory.Packages.props to the list?

.github/workflows/dep-version-validation.yml
```diff
@@ -0,0 +1,245 @@
+name: Dependency Version Validation
+
```
bachuv, line 6: Can you add a short description in a comment explaining what this yaml file runs and tests?

test/SmokeTests/DepValidation/Isolated/DepValidationApp.csproj
YunchuWang: Done — switched to CPM for Worker/Http/Sdk packages (matching OOProcSmokeTests/DotNetIsolated pattern). Only the DurableTask extension uses `VersionOverride` since its version is injected at build time.

.github/workflows/dep-version-validation.yml
YunchuWang: Done — added a 4-line description comment at the top of the workflow file.

bachuv: Looks like the formatting isn't correct for the description. `name: Dependency Version Validation` isn't showing up correctly either.

test/SmokeTests/DepValidation/Isolated/DepValidationApp.csproj
YunchuWang: Actually reverted — this project needs `ManagePackageVersionsCentrally=false` because the DurableTask extension version is dynamically injected via `-p:SmokeTestExtVersion` and cannot go through CPM. The other packages (Worker/Http/Sdk) have explicit versions matching `Directory.Packages.props` for clarity.

.github/workflows/dep-version-validation.yml
```diff
@@ -0,0 +1,245 @@
...
+      - name: Verify package version in assets
+        run: |
+          set -e
+          KEY="Microsoft.Azure.WebJobs.Extensions.DurableTask/${{ steps.version.outputs.ext_version }}"
+          grep -qi "$KEY" test/SmokeTests/DepValidation/InProcess/obj/project.assets.json && echo "OK: $KEY" || { echo "FAIL: $KEY not found"; exit 1; }
+
+      - name: Start Azurite
+        run: |
+          npm install -g azurite
+          nohup azurite --silent --skipApiVersionCheck --blobPort 10000 --queuePort 10001 --tablePort 10002 > /tmp/azurite.log 2>&1 &
```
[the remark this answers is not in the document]
YunchuWang: no need to cleanup

.github/workflows/dep-version-validation.yml
YunchuWang: Good call — but I think `test/Directory.Packages.props` only manages test-time dependency versions (e.g. xunit, Moq) that don't flow into the shipped extension packages. This pipeline specifically validates that the **published Worker and WebJobs extension NuGet packages** work correctly with Azure Functions at runtime. A version bump in test deps wouldn't affect that compatibility. Happy to add it if you feel differently though!

.github/workflows/dep-version-validation.yml
YunchuWang: updated

bachuv (approved):

PR #3420, opened 2026-04-09
Fix CI workflows: .NET 6 SDK, perf scenario TFM, flaky E2E test

YunchuWang:
## Fixes

1. **Add .NET 6.0 SDK installation** to all 3 Validate Build workflows — \windows-latest\ runners no longer have it pre-installed
2. **Fix DFPerfScenariosV4.csproj** — TFM typo \
et60\ → \
et6.0\, upgrade \Microsoft.NET.Sdk.Functions\ from \4.0.*\ (resolves to 4.0.1 which needs .NET 6 RC) to \4.2.0\
3. **Exclude \DurableEntity_AsyncIO\** from E2E tests — it calls external URLs (microsoft.com, bing.com) which return HTTP 403 from WAF on CI runners


YunchuWang:
Changes merged into #3419 to unblock CI.

PR #3430, opened 2026-04-27
Add SetRegisteredFunctions to IDurabilityProviderFactory for work item filter support

bachuv:
# Summary
## What changed?
- Added `SetRegisteredFunctions(IReadOnlyCollection<string>, IReadOnlyCollection<string>, IReadOnlyCollection<string>)` default interface method to `IDurabilityProviderFactory` with a no-op default implementation
- Added a call to `SetRegisteredFunctions` in `DurableTaskExtension.StartTaskHubWorkerIfNotStartedAsync()`, before `EnsureTaskHubWorker().StartAsync()`, passing orchestrator/activity/entity names from `knownOrchestrators`/`knownActivities`/`knownEntities` (projected via `FunctionName.Name`). The durability provider is created before function indexing runs, so it doesn't know what orchestrators, activities, and entities the app has registered. SetRegisteredFunctions is called after indexing but before the worker starts, giving the provider the function names it needs to build work item filters for selective dispatch.
- Added a sample under samples/workitem-filters

## Why is this change needed?
- When multiple Function apps share the same Durable Task Scheduler (DTS) task hub, work items are dispatched to any connected worker causing failures when a worker receives a function it doesn't have (e.g., "The function 'X' doesn't exist")
- The DTS backend already supports `WorkItemFilters` on the `GetWorkItems` gRPC stream (from https://github.com/microsoft/durabletask-dotnet/pull/616). The AzureManaged durability provider needs the registered function names to build these filters. This PR provides the hook, the AzureManaged provider overrides `SetRegisteredFunctions` to build and send the proto filters.
- Non-AzureManaged backends are unaffected (the default implementation is a no-op)

## Issues / work items
- Related: https://github.com/microsoft/durabletask-dotnet/pull/616

---

# Project checklist
- [ ] Documentation changes are not required
  - [x] Documentation will be needed when the full feature ships (host.json `workItemFilteringEnabled` setting), tracked separately
  - [x] This change applies exclusively to v3.x (default interface methods require C# 8+)
- [x] Not a breaking change — adds a new default interface method with no-op implementation

---

- [ ] No

If AI was used:
- Tool(s): GitHub Copilot (Claude Opus 4.6)
- AI-assisted areas/files:
  - `IDurabilityProviderFactory.cs` (new `SetRegisteredFunctions` method)
  - `DurableTaskExtension.cs` (call site in `StartTaskHubWorkerIfNotStartedAsync`)
- What you changed after AI output: Reviewed code, verified lifecycle timing, validated against live DTS emulator E2E

---

# Testing
## Automated tests
- Result: Passed
- - Added `DefaultInterfaceMethod_IsNoOp` test

- Environment: Windows 11, .NET 8/10, DTS emulator (in-memory), Azurite (Docker), func 4.9.0
- Steps + observed results:
  1. Built local NuGet packages with this change + companion AzureManaged provider changes
  2. Ran .NET isolated Function app with `workItemFilteringEnabled: true` against DTS emulator
  3. Host log confirmed: `Work item filtering enabled. Registered N orchestrators, N activities, N entities.`
  4. Ran 8 E2E tests: matching orchestrations completed; non-matching stayed Pending (filter isolation confirmed)

---

# Notes for reviewers
- This is the extension-side half of work item filter support. The AzureManaged durability provider (separate package) implements the `SetRegisteredFunctions` override.
- `SetRegisteredFunctions` is called from `StartTaskHubWorkerIfNotStartedAsync` — guaranteed to run after function indexing but before `TaskHubWorker.StartAsync()`. This is a new call point; no existing factory method is called here today.
- The no-op default means all non-AzureManaged backends (Azure Storage, MSSQL, Netherite) are completely unaffected.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -1171,6 +1171,26 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             return info;
         }
 
+        internal (
+            IReadOnlyCollection<string> orchestratorNames,
+            IReadOnlyCollection<string> activityNames,
+            IReadOnlyCollection<string> entityNames) GetActiveRegisteredFunctionNames()
+        {
+            return (
+                orchestratorNames: this.knownOrchestrators
+                    .Where(kvp => !kvp.Value.IsDeregistered)
+                    .Select(kvp => kvp.Key.Name)
+                    .ToList(),
+                activityNames: this.knownActivities
+                    .Where(kvp => !kvp.Value.IsDeregistered)
+                    .Select(kvp => kvp.Key.Name)
+                    .ToList(),
+                entityNames: this.knownEntities
+                    .Where(kvp => !kvp.Value.IsDeregistered)
+                    .Select(kvp => kvp.Key.Name)
+                    .ToList());
+        }
+
         // This is temporary until script loading
         private static void ConfigureLoaderHooks()
         {
@@ -1431,10 +1451,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                         // Pass registered function names to the factory so backends that support
                         // work-item filtering (e.g., DTS/AzureManaged) can build WorkItemFilters
                         // before the task hub worker opens its GetWorkItems stream.
+                        var activeFunctions = this.GetActiveRegisteredFunctionNames();
                         this.durabilityProviderFactory.SetRegisteredFunctions(
-                            orchestratorNames: this.knownOrchestrators.Keys.Select(fn => fn.Name).ToList(),
-                            activityNames: this.knownActivities.Keys.Select(fn => fn.Name).ToList(),
-                            entityNames: this.knownEntities.Keys.Select(fn => fn.Name).ToList());
+                            orchestratorNames: activeFunctions.orchestratorNames,
+                            activityNames: activeFunctions.activityNames,
+                            entityNames: activeFunctions.entityNames);
 
                         await this.EnsureTaskHubWorker().StartAsync();
 
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -1427,6 +1427,15 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
                         Stopwatch sw = Stopwatch.StartNew();
                         await this.defaultDurabilityProvider.CreateIfNotExistsAsync();
+
+                        // Pass registered function names to the factory so backends that support
+                        // work-item filtering (e.g., DTS/AzureManaged) can build WorkItemFilters
+                        // before the task hub worker opens its GetWorkItems stream.
+                        this.durabilityProviderFactory.SetRegisteredFunctions(
+                            orchestratorNames: this.knownOrchestrators.Keys.Select(fn => fn.Name).ToList(),
+                            activityNames: this.knownActivities.Keys.Select(fn => fn.Name).ToList(),
+                            entityNames: this.knownEntities.Keys.Select(fn => fn.Name).ToList());
+
                         await this.EnsureTaskHubWorker().StartAsync();
 
                         this.GetTaskHubWorkerOrThrow().TaskOrchestrationDispatcher.EntitiesEnabled = true;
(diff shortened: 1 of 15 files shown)


nytian (approved):
Do we need add a e2e test on DTS backend? But anyways I am not blocked on that.

YunchuWang (commented):
A few inline comments on filter wiring and the new e2e test setup.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -1427,6 +1427,15 @@ internal async Task<bool> StartTaskHubWorkerIfNotStartedAsync()
...
                         Stopwatch sw = Stopwatch.StartNew();
                         await this.defaultDurabilityProvider.CreateIfNotExistsAsync();
+
+                        // Pass registered function names to the factory so backends that support
+                        // work-item filtering (e.g., DTS/AzureManaged) can build WorkItemFilters
+                        // before the task hub worker opens its GetWorkItems stream.
+                        this.durabilityProviderFactory.SetRegisteredFunctions(
+                            orchestratorNames: this.knownOrchestrators.Keys.Select(fn => fn.Name).ToList(),
+                            activityNames: this.knownActivities.Keys.Select(fn => fn.Name).ToList(),
+                            entityNames: this.knownEntities.Keys.Select(fn => fn.Name).ToList());
```
YunchuWang: **Medium — `Keys.Select(...)` does not exclude deregistered functions.**

`knownOrchestrators` / `knownActivities` / `knownEntities` retain entries with `IsDeregistered = true`; other call sites in this file consistently filter them out (see lines 1253, 1289, 1327: `&& !existing.IsDeregistered`).

As written, a function disabled via `[Disable]` will still be advertised to DTS, so the backend will dispatch its work items to this worker, which is exactly the failure mode work-item filtering is supposed to eliminate.

Suggested fix:
```csharp
orchestratorNames: this.knownOrchestrators
    .Where(kvp => !kvp.Value.IsDeregistered)
    .Select(kvp => kvp.Key.Name)
    .ToList(),
// same for activityNames / entityNames
```

Would also be worth a unit test that registers a function, marks it deregistered, and asserts it is not in the list passed to `SetRegisteredFunctions`.

test/e2e/Tests/Tests/WorkItemFilterTests.cs
```diff
@@ -0,0 +1,124 @@
...
+        using HttpResponseMessage response = await HttpHelpers.InvokeHttpTrigger(
+            "StartOrchestration",
+            "?orchestrationName=HelloCities");
+
+        Assert.Equal(HttpStatusCode.Accepted, response.StatusCode);
+
+        string statusQueryGetUri = await DurableHelpers.ParseStatusQueryGetUriAsync(response);
+
+        await DurableHelpers.WaitForOrchestrationStateAsync(statusQueryGetUri, "Completed", 30);
+
```
YunchuWang, line 49: **Low/Medium — fixed `Task.Delay(10s)` makes this test ambiguous and flaky.**

A 10-second sleep cannot distinguish "filter is working (item is held by DTS)" from "the worker just hasn't picked it up yet". On a slow CI run the assertion can pass for the wrong reason; on a fast run it might race the dispatcher.

A more deterministic pattern: start a known orchestration and the unknown one together, wait for the **known** one to reach `Completed` (which proves the worker is actively dispatching), then assert the unknown one is still `Pending`.

```csharp
var known   = await StartHelloCities();
var unknown = await StartUnknown();
await DurableHelpers.WaitForOrchestrationStateAsync(known, "Completed", 30);
var details = await DurableHelpers.GetRunningOrchestrationDetailsAsync(unknown);
Assert.Equal("Pending", details.RuntimeStatus);
```

This removes the dependency on wall-clock time and makes the assertion meaningful.

test/e2e/Tests/Fixtures/FixtureHelpers.cs
```diff
@@ -141,6 +141,7 @@ internal static void AddDurableBackendEnvironmentVariables(Process funcProcess,
                 funcProcess.StartInfo.EnvironmentVariables["AzureFunctionsJobHost__extensions__durableTask__hubName"] = "default";
                 funcProcess.StartInfo.EnvironmentVariables["AzureFunctionsJobHost__extensions__durableTask__storageProvider__type"] = "azureManaged";
                 funcProcess.StartInfo.EnvironmentVariables["AzureFunctionsJobHost__extensions__durableTask__storageProvider__connectionStringName"] = "DURABLE_TASK_SCHEDULER_CONNECTION_STRING";
+                funcProcess.StartInfo.EnvironmentVariables["AzureFunctionsJobHost__extensions__durableTask__storageProvider__workItemFilteringEnabled"] = "true";
```
YunchuWang: **Low — `workItemFilteringEnabled=true` is set globally for every test using this fixture, not just `WorkItemFilterTests`.**

Today this is benign because the AzureManaged backend package that consumes the setting hasn't shipped yet, so it's a no-op env var. Once the companion package is released, every DTS e2e test in this fixture will run with filtering enabled.

That is probably fine for the happy path, but any test that depends on receiving messages for a function the worker hasn't registered (e.g., raise-event scenarios, sub-orchestrations across apps, future cross-app tests) could become flaky or start hanging in `Pending`.

Suggest scoping this to a class fixture or test-class–level setup used only by `WorkItemFilterTests`, or at minimum tracking it with a follow-up issue so it doesn't silently broaden test surface coverage when the AzureManaged package upgrades.

YunchuWang (approved):

PR #3432, opened 2026-04-28
Bump patch versions and dependencies

sophiatev:
Bump .NET SDK and DT.Core dependencies and the patch version


sophiatev:
@copilot the dependency version validation pipeline is failing because it is not considering local-packages for the WebJobs version, the problem is this version doesn't exist on nuget yet because it is being updated by this PR. Please fix the pipeline, probably the nuget.config needs to be changed to check local-packages

YunchuWang (approved):

PR #3439, opened 2026-05-06
Add Function retry when the worker has not been fully initialized

sophiatev:
# Summary
## What changed?
This PR adds a new case for Activity/orchestration/entity execution failures that detects errors related to a worker not being fully initialized, and throws a `SessionAbortedException` in these cases which will ensure the Function execution is retried rather than failed.

## Why is this change needed?
Previously these Function invocations were marked as failed, but they should be retried for platform-related errors.

## Issues / work items
- Resolves #3438 

---

# Project checklist
- [ ] Documentation changes are not required
  - [ ] Otherwise: Notes added to `release_notes.md`
- [ ] Backport is not required
- [ ] All required tests have been added/updated (unit tests, E2E tests)
- [ ] No extra work is required to be leveraged by OutOfProc SDKs
- [ ] No change to the version of the `WebJobs.Extensions.DurableTask` package
- [ ] No EventIds were added to `EventSource` logs
  - [ ] Otherwise: Ensure EventIds are within the supported range in the existing Windows infrastructure (validate via deployed telemetry). If needed, extend the range via a PR such as https://msazure.visualstudio.com/One/_git/AAPT-Antares-Websites/pullrequest/7463263?_a=files
  - [ ] Otherwise: This change applies exclusively to `WebJobs.Extensions.DurableTask` v3.x and will be retained only in the `dev` and `main` branches
- [ ] Breaking change?
  - [ ] If yes:
    - Impact:
    - Migration guidance:
---

- [ ] No
- [ ] Yes, an AI agent generated most of this PR

If AI was used:
- Tool(s):
- AI-assisted areas/files:
- What you changed after AI output:

- [ ] I understand the code and can explain it
- [ ] I verified referenced APIs/types exist and are correct
- [ ] I reviewed edge cases/failure paths (timeouts, retries, cancellation, exceptions)
- [ ] I reviewed concurrency/async behavior
- [ ] I checked for unintended breaking or behavior changes

---

# Testing
## Automated tests
- Result: Passed / Failed (link logs if failed)

- Environment (OS, .NET version, components):
- Steps + observed results:
  1.
  2.
  3.
- Evidence (optional):

---

# Notes for reviewers
- N/A

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
+++ b/src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
@@ -409,9 +409,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
                 if (!functionResult.Succeeded)
                 {
-                    // This exception is thrown when another Function on the worker exceeded the Function timeout.
-                    // In this case we want to make sure to retry this entity's execution rather than marking it as failed.
-                    if (functionResult.Exception is Host.FunctionTimeoutAbortException)
+                    // These exception are thrown when either:
+                    // 1. Another Function on the worker exceeded the Function timeout.
+                    // 2. The worker the Activity was sent to has not yet been fully initialized and is not ready to process the Activity execution.
+                    // In these cases we want to make sure to retry this Activity's execution rather than marking it as failed.
+                    if (functionResult.Exception is Host.FunctionTimeoutAbortException || IsWorkerNotFullyInitializedException(functionResult.Exception))
                     {
                         throw functionResult.Exception;
                     }
--- a/src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
+++ b/src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using DurableTask.Core;
@@ -21,6 +22,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
     internal class OutOfProcMiddleware
     {
         private const string NoProcessAssociatedMessage = "No process is associated";
+        private const string NoWorkerInitializedMessage = "Did not find any initialized language workers";
+        private const string AssemblyNotLoadedMessage = "Could not load file or assembly";
 
         private readonly DurableTaskExtension extension;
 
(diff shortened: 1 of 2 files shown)


sophiatev:
@copilot Update `OutOfProcMiddlewareTests` to add tests for the new cases in this PR

sophiatev:
@copilot Add a test for the entity case too now

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -666,7 +673,14 @@ private static bool IsPlatformLevelException(Exception? exception)
...
                 || (exception?.InnerException is InvalidOperationException ioe
-                    && ioe.Message.Contains(NoProcessAssociatedMessage, StringComparison.Ordinal));
+                    && ioe.Message.Contains(NoProcessAssociatedMessage, StringComparison.Ordinal))
+                || IsWorkerNotFullyInitializedException(exception);
+        }
+
+        private static bool IsWorkerNotFullyInitializedException(Exception? exception)
+        {
+            return (exception?.InnerException is InvalidOperationException ioe && ioe.Message.Contains(NoWorkerInitializedMessage, StringComparison.Ordinal))
+                || (exception?.InnerException is FileNotFoundException fnfe && fnfe.Message.Contains(AssemblyNotLoadedMessage, StringComparison.Ordinal));
```
AnatoliB, line 683: I'm concerned about the "assembly not loaded" exception being potentially thrown by customer code for reasons unrelated to the issue we are fixing. Is there anything else that we can use to distinguish between these two cases? Anything different on the stack trace?

sophiatev: This is a good point - I think the outer exception would be of type `FunctionInvocationException` in this case, although that may be the case even for customer-code-errors... let me see if I can find anything useful for this

sophiatev: I had copilot do a small test for me where it threw a `FileNotFoundException` from user code and this is what it found:

"Your inner-type checking is the right discriminator — `FileNotFoundException` and `InvalidOperationException` instances only appear there in the host-dispatch path; user-code exceptions of those types would arrive as an RpcException with "Result: ..." text and would not match."

In other words, when the exception originates from user code, the inner exception type is actually `RpcException`. This is already something we rely on in our code in the same class [here](https://github.com/Azure/azure-functions-durable-extension/blob/f0c960b01e57e893b014fed22cdd431ad4d2d82a/src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs#L672)  when attempting to extract failure details related to user exceptions. So I think the check we have now should distinguish between the two cases already.

AnatoliB: Ok, this what I was hoping for, thank you for checking!

AnatoliB (approved):

PR #3451, opened 2026-06-01
Support Custom FailureDetails at durable functions javascript 

nytian:
This pull request enhances the propagation of structured failure details from out-of-process (out-of-proc) workers, such as the durable-functions JS SDK, through the Durable Task Framework. The changes ensure that when an activity or sub-orchestration fails with structured error information (including custom properties and inner failures), this information is preserved and made available to parent orchestrations, improving error transparency and debuggability across language boundaries.

Related PR https://github.com/Azure/azure-functions-durable-js/pull/705

**Structured Failure Details Propagation:**

* Added the `OrchestrationFailureDetails` property to `DurableOrchestrationContext` to store structured failure details from out-of-proc workers, allowing these details to be attached to orchestration completion actions and propagated up the orchestration chain.
* In `OutOfProcOrchestrationShim`, now detects and stores structured failure details from the worker's result, converting them into `FailureDetails` objects and attaching them to the orchestration context. [[1]](diffhunk://#diff-b78a9b57bdd33bfec3f85b6762bb78d0c45339928637e1d805201f40b4b14ff0R91-R103) [[2]](diffhunk://#diff-b78a9b57bdd33bfec3f85b6762bb78d0c45339928637e1d805201f40b4b14ff0R342-R368) [[3]](diffhunk://#diff-b78a9b57bdd33bfec3f85b6762bb78d0c45339928637e1d805201f40b4b14ff0R386-R420)
* The orchestration middleware in `DurableTaskExtension` now attaches any structured failure details present in the context to the orchestration completion action, ensuring they are persisted and available to parent orchestrations.

**Activity Failure Handling:**

* `TaskActivityShim` now extracts and stores structured failure details from out-of-proc worker exceptions, making them available for middleware to override the resulting `TaskFailedEvent` with detailed `FailureDetails`. [[1]](diffhunk://#diff-fa6cd73d3750116a7ae3d86c7cec2650ae6c734c5e69a246bb5db88d47bb65f3R49-R58) [[2]](diffhunk://#diff-fa6cd73d3750116a7ae3d86c7cec2650ae6c734c5e69a246bb5db88d47bb65f3R132-R133)
* The activity middleware in `DurableTaskExtension` is updated to catch failures with structured details and override the dispatch result, ensuring these details are persisted similarly to orchestration failures.

**Utility and Parsing Improvements:**

* Added a utility method in `OutOfProcMiddleware` to parse and extract structured `FailureDetails` from exception payloads, supporting the new error propagation flow.


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -114,6 +114,16 @@ internal int DefaultHttpAsyncRequestSleepTimeMillseconds
...
 
+        /// <summary>
+        /// Structured failure details supplied by an out-of-proc worker (e.g. the durable-functions
+        /// JS SDK) when an uncaught sub-orchestration or activity failure propagates out of the
+        /// orchestrator. Carries the full <c>InnerFailure</c> chain (including any custom
+        /// <c>Properties</c>) so <see cref="DurableTaskExtension.OrchestrationMiddleware"/> can attach
+        /// it to the orchestration completion action, allowing a calling parent orchestration to
+        /// reconstruct the failure. <c>null</c> for all other workers/paths, leaving behavior unchanged.
+        /// </summary>
+        internal FailureDetails OrchestrationFailureDetails { get; set; }
```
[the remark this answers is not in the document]
nytian: this file doens't support nullable, adding this will cause build error

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -325,6 +339,33 @@ private async Task ProcessAsyncActionsV1(AsyncAction[][] actions, SchemaVersion
...
+            if (dto == null)
+            {
+                return null;
+            }
+
+            IDictionary<string, object> properties = null;
+            if (dto.Properties != null && dto.Properties.Count > 0)
+            {
+                properties = new Dictionary<string, object>(dto.Properties);
+            }
```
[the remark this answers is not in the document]
nytian: Same no nullable at this file

src/WebJobs.Extensions.DurableTask/Listener/TaskActivityShim.cs
```diff
@@ -119,6 +129,8 @@ public override async Task<string> RunAsync(TaskContext context, string rawInput
                         isReplay: false,
                         taskEventId: this.taskEventId);
 
+                    this.StructuredFailureDetails = OutOfProcMiddleware.TryGetStructuredFailureDetails(exceptionToReport!);
+
                     throw new TaskFailureException(
                             $"Activity function '{this.activityName}' failed: {exceptionToReport!.Message}",
                             Utils.SerializeCause(exceptionToReport, this.config.ErrorDataConverter));
```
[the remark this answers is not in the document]
nytian: I don't think this is correct, a new TakActivityShim will be created for every execution in DurablTaskExtension, so StructedFailureDetails always starts null. Shims won't be reused accross invocations so nothing stale can leak actually

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -710,16 +711,39 @@ TaskActivity INameVersionObjectManager<TaskActivity>.GetObject(string name, stri
...
+                FailureDetails details = shim.StructuredFailureDetails;
+                dispatchContext.SetProperty(new ActivityExecutionResult
+                {
+                    ResponseEvent = new TaskFailedEvent(
+                        eventId: -1,
+                        taskScheduledId: scheduledEvent?.EventId ?? -1,
+                        reason: $"Activity function '{shim.ActivityName}' failed with an unhandled exception.",
+                        details: null,
+                        details),
+                });
```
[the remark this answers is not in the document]
nytian: done

test/e2e/Apps/BasicNode/package-lock.json
```diff
@@ -425,6 +465,7 @@
       "version": "10.5.0",
       "resolved": "https://registry.npmjs.org/glob/-/glob-10.5.0.tgz",
       "integrity": "sha512-DfXN8DfhJ7NH3Oe7cFmu3NCu1wKbkReJ8TorzSAFbSKrlNaQSKfIzqYqVY8zlbs2NLBbWpRiU52GX2PbaBVNkg==",
+      "deprecated": "Old versions of glob are not supported, and contain widely publicized security vulnerabilities, which have been fixed in the current version. Please update. Support for old versions may be purchased (at exorbitant rates) by contacting i@izs.me",
       "dev": true,
```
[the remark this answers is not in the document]
nytian: removed

test/e2e/Tests/Tests/ErrorHandlingTests.cs
```diff
@@ -224,7 +224,7 @@ await this.fixture.TestLogs.AssertLogExistsAsync(
     [Fact]
     [Trait("PowerShell", "Skip")] // FailureDetails is a dotnet-isolated implementation detail
     [Trait("Python", "Skip")] // FailureDetails is a dotnet-isolated implementation detail
-    [Trait("Node", "Skip")] // FailureDetails is a dotnet-isolated implementation detail
+    [Trait("Node", "Skip")] // Custom exception properties are not yet supported by the durable-functions JS SDK.
```
bachuv, line 227: Can we add unit tests?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -710,16 +711,39 @@ TaskActivity INameVersionObjectManager<TaskActivity>.GetObject(string name, stri
...
+            try
+            {
+                // Move to the next stage of the DTFx pipeline to trigger the activity shim.
+                await next();
+            }
+            catch (TaskFailureException) when (shim?.StructuredFailureDetails != null)
+            {
+                // The shim extracted a structured TaskFailureDetails payload from the out-of-proc
+                // worker's exception (e.g. produced by the durable-functions JS SDK's
+                // ExceptionPropertiesProvider). Override the dispatch result so DTFx persists the
```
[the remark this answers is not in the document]
nytian: we will add e2e test later. As right now we didn't release js pkg yet and we need to use some classes there

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableOrchestrationContext.cs
```diff
@@ -114,6 +114,16 @@ internal int DefaultHttpAsyncRequestSleepTimeMillseconds
 
         internal ExceptionDispatchInfo OrchestrationException { get; set; }
 
+        /// <summary>
+        /// Structured failure details supplied by an out-of-proc worker (e.g. the durable-functions
+        /// JS SDK) when an uncaught sub-orchestration or activity failure propagates out of the
```
bachuv: Let's remove this mention of the durable-functions JS SDK from the comment.


```suggestion
        /// Structured failure details supplied by an out-of-proc worker when an uncaught sub-orchestration or activity failure propagates out of the
```

src/WebJobs.Extensions.DurableTask/Listener/TaskActivityShim.cs
```diff
@@ -46,6 +46,16 @@ public TaskActivityShim(
             this.activityName = activityName;
         }
 
+        /// <summary>
+        /// Structured failure details extracted from an out-of-proc worker's serialized exception payload
+        /// (e.g. produced by the durable-functions JS SDK's ExceptionPropertiesProvider). The shim stashes
```
bachuv: Can we remove this mention of the durable-functions JS SDK too?

```suggestion
        /// The shim stashes
```

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
bachuv:
```suggestion
                // worker's exception. Override the dispatch result so DTFx persists the
```

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -710,16 +711,39 @@ TaskActivity INameVersionObjectManager<TaskActivity>.GetObject(string name, stri
...
+                // FailureDetails on the TaskFailedEvent, mirroring what OutOfProcMiddleware does
+                // for the new-protocol .NET-isolated/Java/Custom workers.
```
bachuv:
```suggestion
                // for the new-protocol.
```

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -342,12 +383,41 @@ private class OutOfProcOrchestratorState
...
             [JsonProperty("schemaVersion", DefaultValueHandling = DefaultValueHandling.Populate)]
             internal SchemaVersion SchemaVersion { get; set; }
         }
 
+        /// <summary>
+        /// Wire-format DTO matching the JSON shape the durable-functions worker SDKs serialize for
+        /// <see cref="DurableTask.Core.FailureDetails"/> (PascalCase). Mirrors the
+        /// <c>FailureDetailsDto</c> the host extension already sends down on history events.
+        /// </summary>
+        private class FailureDetailsDto
```
bachuv: non-blocking - is `FailureDetailsPayload` a better name?

bachuv (approved):

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -325,6 +339,33 @@ private async Task ProcessAsyncActionsV1(AsyncAction[][] actions, SchemaVersion
             }
         }
 
+        /// <summary>
+        /// Recursively converts a worker-supplied <see cref="FailureDetailsPayload"/> into a
+        /// <see cref="DurableTask.Core.FailureDetails"/>, preserving the <c>InnerFailure</c> chain
+        /// and any custom <c>Properties</c>.
+        /// </summary>
+        private static FailureDetails ConvertToFailureDetails(FailureDetailsPayload dto)
```
andystaples, line 347: **Test coverage — orchestration path (the headline feature).**

This recursive conversion (the `InnerFailure` chain + custom `Properties`) and the `OrchestrationMiddleware` attachment that consumes it have no unit coverage. The only added tests exercise `TryGetStructuredFailureDetails` (the *activity* parsing path), and the Node `CustomExceptionPropertiesInFailureDetails` e2e test stays `[Trait("Node", "Skip")]`, so there's effectively no automated test for the propagate-to-parent behavior this PR is adding.

Consider making `ConvertToFailureDetails` `internal` (as was done for `TryGetStructuredFailureDetails`) and adding tests covering the nested `InnerFailure` chain, property preservation, and the `null` payload path.

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
```diff
@@ -325,6 +339,33 @@ private async Task ProcessAsyncActionsV1(AsyncAction[][] actions, SchemaVersion
...
+        {
+            if (dto == null)
+            {
+                return null;
+            }
+
+            IDictionary<string, object> properties = null;
+            if (dto.Properties != null && dto.Properties.Count > 0)
+            {
+                properties = new Dictionary<string, object>(dto.Properties);
```
andystaples, line 357: **Consideration: property CLR-type divergence between the two channels.**

Here the `Properties` values come straight from Newtonsoft deserialization (`int`→`long`, nested object→`JObject`/`JArray`, ISO-ish string→`DateTime`), whereas the activity path builds them via `ProtobufUtils.ConvertValueToObject` where JSON numbers surface as `double` (see the `IntProperty` assertion in the new unit test). So the *same* custom property can reach a parent with a different runtime type depending on whether it arrived via a `TaskFailedEvent` or via a propagated orchestrator failure. Impact is low (re-serialization normalizes), but it's worth a note — or normalizing — so a parent orchestration that inspects property types isn't surprised.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -821,6 +844,28 @@ private async Task OrchestrationMiddleware(DispatchMiddlewareContext dispatchCon
...
             }
 
+            // If an out-of-proc worker supplied structured failure details for an uncaught
+            // sub-orchestration/activity failure that propagated out of the orchestrator, attach
+            // them to the orchestration completion action so DTFx persists them on the resulting
+            // ExecutionCompleted / SubOrchestrationInstanceFailed event. This lets a calling parent
+            // orchestration reconstruct the full InnerFailure chain (including custom Properties).
+            // The completion action's FailureDetails is publicly settable and is read by the DTFx
+            // dispatcher after this middleware returns, so mutating it here is safe and works
+            // regardless of the configured ErrorPropagationMode. No-op for all other workers/paths.
```
andystaples, line 854: The `Where(... && a.FailureDetails == null)` guard means that under `ErrorPropagationMode.UseFailureDetails` — where DTFx already populates `FailureDetails` from the `OrchestrationFailureException` — these structured details would be **skipped**, not attached. It works today only because the extension runs in the default `SerializeExceptions` mode (the completion action's `FailureDetails` is null there). Suggest rewording “works regardless of the configured ErrorPropagationMode” to something scoped to the current mode, so a future switch to `UseFailureDetails` doesn't silently drop the richer details.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -710,16 +711,38 @@ TaskActivity INameVersionObjectManager<TaskActivity>.GetObject(string name, stri
...
             }
 
-            // Move to the next stage of the DTFx pipeline to trigger the activity shim.
-            return next();
+            try
+            {
+                // Move to the next stage of the DTFx pipeline to trigger the activity shim.
+                await next();
+            }
+            catch (TaskFailureException) when (shim?.StructuredFailureDetails != null)
```
andystaples, line 729: **Consideration: is this override reachable for any worker routed through this (old) middleware today?**

It only fires when `shim.StructuredFailureDetails != null`, which requires the worker to embed a protobuf `TaskFailureDetails` JSON in the RPC `Result:` message. The worker that emits that shape (.NET-isolated) uses the *new* `OutOfProcMiddleware` path, while the languages routed here are Node/Python/PowerShell/in-proc .NET. So this looks like forward-looking plumbing that's dormant until a worker SDK starts emitting the payload. Not a problem — just confirming it's intentionally staged rather than expected to be exercised now (which also explains why the Node e2e test remains skipped).

src/WebJobs.Extensions.DurableTask/OutOfProcMiddleware.cs
```diff
@@ -677,10 +677,38 @@ private static bool IsPlatformLevelException(Exception? exception)
...
+        /// structured payload is present so callers can fall back to legacy behavior.
+        /// </summary>
+        internal static FailureDetails? TryGetStructuredFailureDetails(Exception e)
+        {
+            string? candidate = null;
+            if (e.InnerException != null && e.InnerException.Message.StartsWith("Result:", StringComparison.Ordinal))
+            {
+                candidate = e.InnerException.Message;
+            }
+            else if (e.Message != null && e.Message.StartsWith("Result:", StringComparison.Ordinal))
```
andystaples: Nit: `Exception.Message` never returns `null` (it falls back to a default message), so the `e.Message != null` guard here is redundant.

andystaples (commented):
## Review summary

I traced the structured-failure-details propagation end-to-end against DTFx **v3.9.0** (the pinned `Microsoft.Azure.DurableTask.Core` version) and the companion JS SDK PR [Azure/azure-functions-durable-js#705](https://github.com/Azure/azure-functions-durable-js/pull/705). **The core mechanism is correct.** The dispatcher sets `OrchestratorExecutionResult` in the core delegate before the middleware's post-`next()` code runs and reads it back afterward, so mutating the completion action's `FailureDetails` is persisted; the activity `TaskFailureException` propagates through the middleware and the overridden `ActivityExecutionResult` is picked up; a fresh `TaskActivityShim` per work item makes `StructuredFailureDetails` race-free; and the PascalCase wire DTO matches the JS SDK. The changes are guarded to be no-ops for in-proc .NET and for failures without a structured payload, so existing behavior is preserved.

The inline comments below are **recommendations / considerations only — nothing blocking**. The main one is unit-test coverage for the orchestration path, which is the headline feature of this PR.

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
nytian: will add a test later once the js pkg is released

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
nytian: reword comment

src/WebJobs.Extensions.DurableTask/Listener/OutOfProcOrchestrationShim.cs
nytian: yes expected. I think the impact should be low since re-serialization normalizes

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
nytian: yeah actually the js worker now starts emitting the payload. When an activity throws, the new exception property provider and the wrap activityhandler will reshape the error into the taskfailuredetails json. Node now route through this middleware so the catch block will fire. I have a e2e test for this for its skipped for now as i need the js pkg to be released first.

andystaples (approved):

PR #3472, opened 2026-07-22
Fix disabled activity/entity functions poison-looping instead of failing gracefully

### LINKED ISSUE #3471: Disabled (but still deployed) activity functions poison-loop with ArgumentNullException('executor') instead of failing gracefully

When an activity function is **disabled but still deployed**, the extension throws `ArgumentNullException('executor')` every time an in-flight orchestration dispatches that activity, and the Durable Task backend treats the throw as a *transient* error — so the work item is abandoned and retried indefinitely (a permanent poison loop). Affected orchestrations never make progress and never fail; they just spin forever.

Root cause: disabled functions are still **indexed** (their trigger binding is created), but no **listener** is started for them, which leaves the activity registered with a `null` executor:

- During indexing, `ActivityTriggerAttributeBindingProvider.TryCreateAsync` runs and calls `RegisterActivity(name, null)`, adding the activity to `knownActivities` with `Executor == null`.
- The WebJobs `HostListenerFactory.CreateAsync` **skips listener creation for disabled functions** (`if (functionDefinition.Descriptor.IsDisabled) { log "Function '...' is disabled"; continue; }`), so `ActivityTriggerBinding.CreateListenerAsync` — which would call `RegisterActivity(name, realExecutor)` — never runs. The executor stays `null`.

When DTFx later dispatches the activity, `DurableTaskExtension`'s `INameVersionObjectManager<TaskActivity>.GetObject` finds the entry (so it does **not** fall back to `TaskNonexistentActivityShim`) and constructs `new TaskActivityShim(this, info.Executor /* null */, ...)`, which throws at the null-guard in the shim constructor:

```
System.ArgumentNullException: Value cannot be null. (Parameter 'executor')
   at ...DurableTask.TaskActivityShim..ctor(DurableTaskExtension config, ITriggeredFunctionExecutor executor, ...) in .../Listener/TaskActivityShim.cs:line 38
   at ...DurableTask.DurableTaskExtension.DurableTask.Core.INameVersionObjectManager<DurableTask.Core.TaskActivity>.GetObject(String name, String version) in .../DurableTaskExtension.cs
   at DurableTask.Core.TaskActivityDispatcher.OnProcessWorkItemAsync(TaskActivityWorkItem workItem)
   at DurableTask.Core.WorkItemDispatcher`1.ProcessWorkItemAsync(WorkItemDispatcherContext context, Object workItemObj)

Backing off for 10 seconds ...
```

Because the exception is thrown during **object construction** (not from `RunAsync`), it surfaces as a work-item *processing* failure, so the message is abandoned and retried forever instead of failing the activity/orchestration.

This is easy to hit with a very common migration pattern: introduce `MyActivity_V2`, disable the old `MyActivity`, but keep it deployed. Any orchestration that was in-flight when the swap happened (its history references the old activity name) then poison-loops permanently after the next restart/recycle. It has been observed in production (Node.js / out-of-proc, host 4.1052.x, extension 3.x, Azure Storage backend) with thousands of instances abandoning/retrying every ~10 s for days across worker recycles.

This is a gap in the graceful handling added by #1901 ("Improve handling of deleted/disabled/renamed orchestrator and activity functions"), which introduced `TaskNonexistentActivityShim`. That change only covers the case where the function is **absent** from `knownActivities` (deleted/renamed). A **disabled-but-still-indexed** function is *present* in `knownActivities` with a `null` executor, so it bypasses that guard and hits the throwing path. (Related but distinct: #3449 fixed a separate null-for-disabled-functions `NullReferenceException` in `GetActiveRegisteredFunctionNames`.)

In `GetObject`, treat a *registered-but-inactive* activity (null executor / not an active listener) the same as a nonexistent one, so the orchestration receives a deterministic, non-transient failure instead of an infinite poison loop:

```csharp
RegisteredFunctionInfo info;
if (!this.knownActivities.TryGetValue(activityFunction, out info) || info?.Executor == null)
{
    return new TaskNonexistentActivityShim(this, name);
}

return new TaskActivityShim(this, info.Executor, this.HostLifetimeService, name);
```

`RegisteredFunctionInfo` already exposes `HasActiveListener => Executor != null && !IsDeregistered`, which could be used directly to also cover the deregistered case. `TaskNonexistentActivityShim.Run` already throws a `TaskFailureException` wrapping `FunctionFailedException("Activity function 'X' does not exist.")`, giving the orchestration a clear, catchable failure. Ideally the message would distinguish "disabled" from "does not exist" for better diagnosability, and emit a warning trace when a disabled activity is dispatched.

The orchestrator and entity dispatch paths should be reviewed for the same registered-with-null pattern.

- Throwing a clearer exception from `GetObject`/the shim constructor — rejected: any exception thrown there stays transient and keeps poison-looping. The failure must be converted into a deterministic activity failure (i.e., surfaced via a shim's `RunAsync`).
- Not indexing disabled functions / removing them from `knownActivities` — a larger, riskier change to the indexing pipeline than the localized `GetObject` guard.

**Additional context**

- Repro (deterministic, language-agnostic):
  1. Orchestrator calls `context.CallActivityAsync("MyActivity", ...)`.
  2. Start an instance so `MyActivity` is scheduled in history.
  3. Disable `MyActivity` (e.g., `AzureWebJobs.MyActivity.Disabled = true`) and restart the app.
  4. The instance resumes and poison-loops on `ArgumentNullException('executor')` indefinitely.
- Suggested labels: `Enhancement`, `Reliability`.

andystaples:
## Summary

Fixes #3471.

When a function is **disabled but still deployed**, it is still **indexed** (its trigger binding is created, so it lands in `knownActivities`/`knownEntities`) but no **listener** is started, so its `RegisteredFunctionInfo.Executor` stays `null`. When DTFx later dispatches such a function for an in-flight orchestration, the extension dereferenced that null executor. Because the failure happened during object construction / before user code, DTFx treated it as a **transient** work-item failure and abandoned + retried it forever — a permanent poison loop (`ArgumentNullException('executor')` every ~10s), with the orchestration never progressing and never failing.

This is a common migration footgun: introduce `MyActivity_V2`, disable the old `MyActivity` but keep it deployed, and any orchestration that was in-flight when the swap happened poison-loops after the next restart.

## Root cause & dispatch-path review

`null` executor already means "registered but inactive" in the codebase (`RegisteredFunctionInfo.HasActiveListener`). I reviewed every dispatch path for the same registered-with-null pattern:

| Path | Classic (in-proc) middleware | Passthrough (`OutOfProcMiddleware`) |
|---|---|---|
| **Activity** | `GetObject` built `TaskActivityShim(null)` → **ArgumentNullException** ❌ | `CallActivityAsync` dereferenced `function.Executor` → **NRE → SessionAborted** ❌ |
| **Orchestrator** | `OrchestrationMiddleware` handles `info == null` ✅ | `CallOrchestratorAsync` handles `function == null` ✅ |
| **Entity** | `EntityMiddleware` dereferenced `GetFunctionInfo().Executor` → **NRE** ❌ | `CallEntityAsync` handles `functionInfo == null` ✅ |

`GetObject` (the activity object manager) is invoked by DTFx for **both** middleware flavors, so the activity fix is required for every language.

## Changes

1. **`TaskNonexistentActivityShim`** — carry an `isDisabled` flag so the failure message distinguishes *"is disabled"* from *"does not exist"*, and emit a warning trace when a disabled activity is dispatched.
2. **`DurableTaskExtension` activity `GetObject`** — when the activity is found but `Executor == null`, return the graceful shim instead of constructing `TaskActivityShim(null)`.
3. **`OutOfProcMiddleware.CallActivityAsync`** — treat `function?.Executor == null` like a missing function and fail with a deterministic `TaskFailedEvent`.
4. **`DurableTaskExtension.EntityMiddleware` + `TaskEntityShim.ExecuteBatch`** — for the classic entity path, fail every operation in the batch deterministically (via a throwing invocation callback that reuses the existing per-operation application-error machinery) instead of dereferencing null / aborting + retrying. Regression-safe: the new branch is gated on `functionUnavailable`, so registered entities take the exact same path as before. Host-shutdown stays transient (gated on `!OnStopping`).
5. **Orchestrator** paths already fail disabled functions gracefully — no change.

In all cases the orchestration now receives a deterministic, catchable failure (`FunctionFailedException`) instead of an infinite poison loop.

## Tests

- **Unit** (`DisabledFunctionDispatchTests`): `GetObject` returns the disabled shim (with an *"is disabled"* failure) for a null-executor activity, the nonexistent shim (*"does not exist"*) otherwise, and the real `TaskActivityShim` when an executor is present.
- **Unit** (`OutOfProcMiddlewareTests`): `CallActivityAsync` with a null-executor activity sets a `TaskFailedEvent` and does **not** throw `SessionAbortedException`.
- **E2E** (`DisabledOrchestrationTests`, .NET isolated): new orchestrations schedule the disabled activity/entity; the tests assert the orchestration reaches `Failed` promptly (a poison loop would leave it `Running`).

### Validation performed
- Extension + all test projects build with 0 warnings/errors.
- 4 new unit tests pass.
- In-proc regression run against Azurite: 14 entity tests pass — including all 8 `DurableEntity_CallFaultyEntity` variants, which exercise the exact per-operation application-error machinery the entity fix reuses — plus the nonexistent activity/orchestrator tests.

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -24,7 +24,7 @@
 
 ### Bug Fixes
 
-- Fixed a poison loop where dispatching a disabled-but-still-deployed activity or entity function caused in-flight orchestrations to retry indefinitely (throwing `ArgumentNullException('executor')`) instead of failing gracefully. Such registered-but-inactive functions are now treated as unavailable and fail deterministically. (#3471)
+- Fixed a poison loop where dispatching a disabled-but-still-deployed activity or entity function caused in-flight orchestrations to retry indefinitely (e.g. throwing `ArgumentNullException('executor')` on the activity dispatch path) instead of failing gracefully. Such registered-but-inactive functions are now treated as unavailable and fail deterministically. (#3471)
 
 ### Breaking Changes
 
--- a/release_notes.md
+++ b/release_notes.md
@@ -24,6 +24,8 @@
 
 ### Bug Fixes
 
+- Fixed a poison loop where dispatching a disabled-but-still-deployed activity or entity function caused in-flight orchestrations to retry indefinitely (throwing `ArgumentNullException('executor')`) instead of failing gracefully. Such registered-but-inactive functions are now treated as unavailable and fail deterministically. (#3471)
+
 ### Breaking Changes
 
 ### Dependency Updates
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -708,7 +708,22 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             RegisteredFunctionInfo info;
             if (!this.knownActivities.TryGetValue(activityFunction, out info))
             {
+                // The activity function is not registered at all (e.g. it was deleted or renamed).
                 return new TaskNonexistentActivityShim(this, name);
+            }
+
+            if (info.Executor == null)
+            {
+                // The activity function was indexed (so it is present in knownActivities) but no
+                // listener was ever started for it. This happens when the function is disabled but
+                // still deployed: the binding provider registers the name with a null executor
+                // during indexing, and the disabled function's listener never replaces it. Returning
+                // a shim that fails deterministically — instead of constructing a TaskActivityShim
+                // with a null executor, which throws ArgumentNullException during object construction
+                // and causes the work item to be abandoned and retried forever (a poison loop) — lets
+                // the orchestration receive a catchable failure.
+                // See https://github.com/Azure/azure-functions-durable-extension/issues/3471.
+                return new TaskNonexistentActivityShim(this, name, isDisabled: true);
             }
 
             return new TaskActivityShim(this, info.Executor, this.HostLifetimeService, name);
@@ -1077,11 +1092,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
             WrappedFunctionResult result;
 
-            if (entityShim.OperationBatch.Count > 0 && !this.HostLifetimeService.OnStopping.IsCancellationRequested)
+            RegisteredFunctionInfo entityFunctionInfo = entityShim.GetFunctionInfo();
+            bool functionUnavailable = entityFunctionInfo?.Executor == null;
+
+            if (entityShim.OperationBatch.Count > 0
+                && !this.HostLifetimeService.OnStopping.IsCancellationRequested
+                && !functionUnavailable)
             {
                 // 3a. (function execution) Start the functions invocation pipeline (billing, logging, bindings, and timeout tracking).
                 result = await FunctionExecutionHelper.ExecuteFunctionInOrchestrationMiddleware(
-                    entityShim.GetFunctionInfo().Executor,
+                    entityFunctionInfo.Executor,
                     new TriggeredFunctionData
                     {
                         TriggerValue = entityShim.Context,
@@ -1174,6 +1194,42 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     throw new SessionAbortedException(
                         $"An internal error occurred while attempting to execute '{entityContext.FunctionName}'.",
                         result.Exception);
+                }
+            }
+            else if (functionUnavailable
+                && entityShim.OperationBatch.Count > 0
+                && !this.HostLifetimeService.OnStopping.IsCancellationRequested)
+            {
+                // The entity function is registered/indexed but has no active listener — e.g. it is
+                // disabled but still deployed, or was deleted/renamed. Rather than dereferencing a
+                // null executor (which would surface as a transient work-item failure and poison-loop
+                // the batch forever), fail every operation in the batch deterministically. Each
+                // operation gets an exception response and is recorded as an application (not internal)
+                // error, so the batch is not aborted/retried. This mirrors the graceful handling in
+                // OutOfProcMiddleware.CallEntityAsync and the activity GetObject path.
+                // See https://github.com/Azure/azure-functions-durable-extension/issues/3471.
+                entityShim.AddTraceFlag(EntityTraceFlags.FunctionUnavailable);
+
+                this.TraceHelper.ExtensionWarningEvent(
+                    this.Options.HubName,
+                    entityContext.Name,
+                    entityContext.InstanceId,
+                    $"The entity function '{entityContext.Name}' is disabled or does not exist. Failing {entityShim.OperationBatch.Count} operation(s) deterministically instead of retrying indefinitely.");
+
+                var failureMessage = this.GetInvalidEntityFunctionMessage(entityContext.Name);
+                entityShim.SetFunctionInvocationCallback(() => throw new FunctionFailedException(failureMessage));
+
+                if (entityContext.InternalError == null)
+                {
+                    try
+                    {
+                        await entityShim.ExecuteBatch(this.HostLifetimeService.OnStopping);
+                        await next();
+                    }
+                    catch (Exception e)
+                    {
+                        entityContext.CaptureInternalError(e, entityShim);
+                    }
                 }
             }
             else
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -1533,7 +1533,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         internal string GetInvalidEntityFunctionMessage(string name)
         {
             string message = $"The function '{name}' doesn't exist, is disabled, or is not an entity function. Additional info: ";
-            if (this.knownOrchestrators.Keys.Count > 0)
+            if (this.knownEntities.Keys.Count > 0)
             {
                 message += $"The following are the known entity functions: '{string.Join("', '", this.knownEntities.Keys)}'.";
             }
--- a/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
@@ -1214,7 +1214,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     this.Options.HubName,
                     entityContext.Name,
                     entityContext.InstanceId,
-                    $"The entity function '{entityContext.Name}' is disabled or does not exist. Failing {entityShim.OperationBatch.Count} operation(s) deterministically instead of retrying indefinitely.");
+                    $"The entity function '{entityContext.Name}' is disabled or does not exist. Failing {entityShim.OperationBatch.Count} operation(s).");
 
                 var failureMessage = this.GetInvalidEntityFunctionMessage(entityContext.Name);
                 entityShim.SetFunctionInvocationCallback(() => throw new FunctionFailedException(failureMessage));
--- a/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
@@ -316,7 +316,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             async Task ExecuteOperationsInBatch()
             {
-                if (this.GetFunctionInfo().IsOutOfProc)
+                // GetFunctionInfo() can be null when the entity function is registered/indexed but has
+                // no active listener (e.g. disabled but still deployed). In that case the caller has
+                // set a throwing invocation callback, so we take the per-operation (in-proc) path and
+                // let each operation fail deterministically rather than dereferencing a null executor.
+                if (this.GetFunctionInfo()?.IsOutOfProc == true)
                 {
                     // process all operations in the batch using a single function call
                     await this.ExecuteOutOfProcBatch();
--- a/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
@@ -316,16 +316,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             async Task ExecuteOperationsInBatch()
             {
-                // Only use the single-call out-of-proc batch path for a runnable out-of-proc entity.
-                // For an unavailable entity (disabled but still deployed, or deleted/renamed)
-                // GetFunctionInfo() is null, so functionInfo?.Executor is null and we fall through to
-                // the per-operation in-proc path. That matters because for the unavailable case
-                // EntityMiddleware sets an invocation callback that throws synchronously: the in-proc
-                // path invokes it inside a per-operation try/catch and records a deterministic
-                // application error, whereas ExecuteOutOfProcBatch would let the synchronous throw
-                // bubble out of the batch and be recorded as an internal error (aborting and retrying
-                // the work item — the poison loop this guards against). The Executor null-check also
-                // defends against a registered-but-null-executor RegisteredFunctionInfo.
+                // A disabled/deleted entity is registered with a null executor. In that case, fall
+                // through to the per-operation loop below, which runs EntityMiddleware's throwing
+                // callback inside a per-operation try/catch and records a deterministic failure.
+                // ExecuteOutOfProcBatch would instead let that throw abort + retry the batch (the
+                // poison loop this guards against). See https://github.com/Azure/azure-functions-durable-extension/issues/3471.
                 RegisteredFunctionInfo functionInfo = this.GetFunctionInfo();
                 if (functionInfo?.Executor != null && functionInfo.IsOutOfProc)
                 {
--- a/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
+++ b/src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
@@ -316,13 +316,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         {
             async Task ExecuteOperationsInBatch()
             {
-                // Only take the single-call out-of-proc batch path for a genuinely runnable out-of-proc
-                // entity (non-null executor). When the executor is null — the entity is registered/indexed
-                // but has no active listener (e.g. disabled but still deployed) — the caller has set a
-                // throwing invocation callback; that callback throws synchronously, so routing it through
-                // ExecuteOutOfProcBatch would bubble the exception out of the batch and reintroduce the
-                // transient failure / poison loop. Instead we always take the per-operation (in-proc) path
-                // for an unavailable entity, so each operation fails deterministically.
+                // Only use the single-call out-of-proc batch path for a runnable out-of-proc entity.
+                // For an unavailable entity (disabled but still deployed, or deleted/renamed)
+                // GetFunctionInfo() is null, so functionInfo?.Executor is null and we fall through to
+                // the per-operation in-proc path. That matters because for the unavailable case
+                // EntityMiddleware sets an invocation callback that throws synchronously: the in-proc
+                // path invokes it inside a per-operation try/catch and records a deterministic
+                // application error, whereas ExecuteOutOfProcBatch would let the synchronous throw
+                // bubble out of the batch and be recorded as an internal error (aborting and retrying
+                // the work item — the poison loop this guards against). The Executor null-check also
+                // defends against a registered-but-null-executor RegisteredFunctionInfo.
                 RegisteredFunctionInfo functionInfo = this.GetFunctionInfo();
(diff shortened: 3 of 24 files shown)


src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -1176,6 +1196,42 @@ private async Task EntityMiddleware(DispatchMiddlewareContext dispatchContext, F
...
+
+                this.TraceHelper.ExtensionWarningEvent(
+                    this.Options.HubName,
+                    entityContext.Name,
+                    entityContext.InstanceId,
+                    $"The entity function '{entityContext.Name}' is disabled or does not exist. Failing {entityShim.OperationBatch.Count} operation(s) deterministically instead of retrying indefinitely.");
+
+                var failureMessage = this.GetInvalidEntityFunctionMessage(entityContext.Name);
+                entityShim.SetFunctionInvocationCallback(() => throw new FunctionFailedException(failureMessage));
+
```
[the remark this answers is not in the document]
andystaples: Good catch — fixed in ab1a185c. GetInvalidEntityFunctionMessage now checks knownEntities.Keys.Count (instead of knownOrchestrators.Keys.Count) before listing the registered entity functions, so the disabled-entity failure message reports the real entities. Added a unit test (GetInvalidEntityFunctionMessage_ListsRegisteredEntities_WhenNoOrchestratorsExist) to guard it.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -1176,6 +1196,42 @@ private async Task EntityMiddleware(DispatchMiddlewareContext dispatchContext, F
...
+                {
+                    try
+                    {
+                        await entityShim.ExecuteBatch(this.HostLifetimeService.OnStopping);
+                        await next();
+                    }
+                    catch (Exception e)
+                    {
+                        entityContext.CaptureInternalError(e, entityShim);
+                    }
```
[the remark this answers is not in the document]
andystaples: Intentionally keeping the broad `catch (Exception e)` here. This block is byte-for-byte identical to the adjacent direct-execution branch, and the broad catch is the intended entity mechanism: a captured internal error is converted into a `SessionAbortedException` by `AbortOnInternalError` a few lines below, which is exactly the graceful abort-and-backoff we want on host shutdown / cancellation. Adding `when (!(e is OperationCanceledException))` would let cancellation escape uncaught and bypass that backoff, and would diverge from the existing pattern. So no change here.

release_notes.md
```diff
@@ -24,6 +24,8 @@
 
 ### Bug Fixes
 
+- Fixed a poison loop where dispatching a disabled-but-still-deployed activity or entity function caused in-flight orchestrations to retry indefinitely (throwing `ArgumentNullException('executor')`) instead of failing gracefully. Such registered-but-inactive functions are now treated as unavailable and fail deterministically. (#3471)
```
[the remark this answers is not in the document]
andystaples: Fixed in bff8473b. You're right — \ArgumentNullException('executor')\ was specific to the activity path (the TaskActivityShim constructor null-guard); the entity path dereferenced \GetFunctionInfo().Executor\ and surfaced as a NullReferenceException. Reworded the note to scope the exception example to the activity dispatch path.

test/e2e/Apps/BasicJava/src/main/java/com/function/DisabledActivity.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Activity function DisabledActivity. Disabled at runtime via the
+     * AzureWebJobs.DisabledActivity.Disabled app setting in local.settings.json, so no listener is
+     * started for it even though it is still deployed/indexed.
+     */
+    @FunctionName("DisabledActivity")
+    public String disabledActivity(
+            @DurableActivityTrigger(name = "input") String input,
+            final ExecutionContext context) {
```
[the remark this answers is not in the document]
andystaples: Fixed in bff8473b — removed the unused \ExecutionContext context\ parameter from \disabledActivity\.

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -316,7 +316,11 @@ public async Task ExecuteBatch(CancellationToken onHostStopping)
         {
             async Task ExecuteOperationsInBatch()
             {
-                if (this.GetFunctionInfo().IsOutOfProc)
+                // GetFunctionInfo() can be null when the entity function is registered/indexed but has
+                // no active listener (e.g. disabled but still deployed). In that case the caller has
+                // set a throwing invocation callback, so we take the per-operation (in-proc) path and
+                // let each operation fail deterministically rather than dereferencing a null executor.
+                if (this.GetFunctionInfo()?.IsOutOfProc == true)
                 {
```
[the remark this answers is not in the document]
andystaples: Good catch — fixed in bff8473b. You're right that the throwing invocation callback throws synchronously at 	his.FunctionInvocationCallback() in ExecuteOutOfProcBatch, so if that path were ever reached for an unavailable entity the exception would bubble out of the batch and reintroduce the poison loop. ExecuteBatch now gates the out-of-proc batch path on \unctionInfo?.Executor != null && functionInfo.IsOutOfProc\, so a null executor always takes the per-operation in-proc path (which catches the throwing callback per operation and produces deterministic failures). Verified no regression across the entity E2E suite (SignalAndCall, BatchedSignals, all CallFaultyEntity variants, Nonexistent).

src/WebJobs.Extensions.DurableTask/Listener/TaskNonexistentActivityShim.cs
```diff
@@ -9,22 +9,51 @@
...
-            string message = $"Activity function '{this.activityName}' does not exist.";
+            if (this.isDisabled)
+            {
+                // Emit a warning so operators can diagnose why an in-flight orchestration's activity
+                // started failing after a deploy that disabled (but kept) the activity function.
+                this.config.TraceHelper.ExtensionWarningEvent(
+                    hubName: this.config.Options.HubName,
+                    functionName: this.activityName,
+                    instanceId: context.OrchestrationInstance?.InstanceId ?? string.Empty,
+                    message: $"Activity function '{this.activityName}' was scheduled but is disabled (registered without an active listener). Failing the activity deterministically instead of retrying indefinitely.");
```
sophiatev: Can we remove the "instead of retrying indefinitely" here, feels like an implementation detail

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
```diff
@@ -316,7 +316,15 @@ public async Task ExecuteBatch(CancellationToken onHostStopping)
         {
             async Task ExecuteOperationsInBatch()
             {
-                if (this.GetFunctionInfo().IsOutOfProc)
+                // Only take the single-call out-of-proc batch path for a genuinely runnable out-of-proc
```
sophiatev: Nit-picking as always but did you actually confirm the logic in this comment is correct?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -1176,6 +1196,42 @@ private async Task EntityMiddleware(DispatchMiddlewareContext dispatchContext, F
...
+                // error, so the batch is not aborted/retried. This mirrors the graceful handling in
+                // OutOfProcMiddleware.CallEntityAsync and the activity GetObject path.
+                // See https://github.com/Azure/azure-functions-durable-extension/issues/3471.
+                entityShim.AddTraceFlag(EntityTraceFlags.FunctionUnavailable);
+
+                this.TraceHelper.ExtensionWarningEvent(
+                    this.Options.HubName,
+                    entityContext.Name,
+                    entityContext.InstanceId,
+                    $"The entity function '{entityContext.Name}' is disabled or does not exist. Failing {entityShim.OperationBatch.Count} operation(s) deterministically instead of retrying indefinitely.");
```
sophiatev: Again can we remove the "retrying indefinitely"?

test/e2e/Tests/Tests/DisabledOrchestrationTests.cs
```diff
@@ -37,4 +38,67 @@ public async Task ActiveOrchestration_StartsWhenDisabledFunctionsAreRegistered()
...
+        Assert.Contains("DisabledActivity", details.Output);
+    }
+
+    // Companion to the activity test above, for the entity dispatch path: calling an operation on a
+    // disabled-but-still-deployed entity must fail the orchestration deterministically rather than
+    // poison-looping the entity work item forever.
+    //
+    // Runs on the languages whose durable SDK supports entities (dotnet-isolated, Node, Python).
+    // PowerShell and Java are skipped because durable entities are not implemented in those SDKs, and
+    // MSSQL/DTS are skipped because durable entities are not supported on those backends — the same
```
sophiatev: DTS supports entities so this comment should be refined and/or we can actually enable the test for DTS? Or is the issue the error surfacing stuff

test/FunctionsV2/Tests/Unit/DisabledFunctionDispatchTests.cs
sophiatev, line 1: Can we also add tests for a disabled entity?

test/FunctionsV2/Tests/Unit/OutOfProcMiddlewareTests.cs
sophiatev, line 1: Same question about tests for a disabled entity

sophiatev:
Do we need to add the same logic for disabled orchestrations? Is that even a thing?

src/WebJobs.Extensions.DurableTask/Listener/TaskNonexistentActivityShim.cs
andystaples: Done in 9f0839e3 — dropped the "instead of retrying indefinitely" clause. The warning is now just: Activity function 'X' was scheduled but is disabled. Failing the activity.

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
andystaples: Done in 9f0839e3 — removed "deterministically instead of retrying indefinitely"; the entity warning is now: The entity function 'X' is disabled or does not exist. Failing N operation(s).

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
andystaples: Yes — I traced it through and confirmed it, and refined the comment to be precise. For a disabled entity the binding provider registers it during indexing with a null RegisteredFunctionInfo, so GetFunctionInfo() returns null and functionInfo?.Executor is null — we fall through to the per-operation in-proc path. EntityMiddleware sets an invocation callback that throws synchronously for the unavailable case; the in-proc path invokes it inside a per-operation try/catch (ProcessOperationRequestAsync) and records a deterministic application error, whereas ExecuteOutOfProcBatch calls the callback directly at Task.WhenAny(FunctionInvocationCallback(), ...) so the synchronous throw would bubble out of the batch and be captured as an internal error -> SessionAborted -> retry (the poison loop). The Executor null-check is also defense-in-depth against a non-null info with a null executor. Behavior is exercised by the Node/Python entity E2E tests (classic path) and the in-proc entity regression suite, all green.

test/e2e/Tests/Tests/DisabledOrchestrationTests.cs
andystaples: Good point — DTS does support entities, so I refined the comment. The real reason for the DTS skip is the same as the activity test (and ErrorHandlingTests.OrchestratorWithUncaughtEntityException_ShouldFail): this test asserts on the failure being surfaced in the orchestration status output, which DTS doesn't do reliably until the referenced bug (31779638) is fixed. MSSQL stays skipped because entities aren't supported there. Updated the comment/trait to say exactly that rather than "entities not supported on those backends".

test/FunctionsV2/Tests/Unit/DisabledFunctionDispatchTests.cs
andystaples: Added in 9f0839e3: DisabledFunctionDispatchTests.DisabledEntity_IsTreatedAsUnavailableByClassicDispatch. It covers the classic-path invariant (GetEntityInfo(name)?.Executor is null / HasActiveListener false) for both a null RegisteredFunctionInfo and the defense-in-depth null-executor info, which is exactly what routes EntityMiddleware into the deterministic branch. The full classic dispatch is also exercised by the Node/Python entity E2E tests.

test/FunctionsV2/Tests/Unit/OutOfProcMiddlewareTests.cs
andystaples: Added in 9f0839e3: OutOfProcMiddlewareTests.CallEntityAsync_DisabledEntity_FailsDeterministicallyWithoutAbort. It registers a disabled (null-executor) entity and asserts CallEntityAsync sets a non-retriable FailureDetails EntityBatchResult without throwing. I also hardened CallEntityAsync's guard from functionInfo == null to functionInfo?.Executor == null so the passthrough entity path is robust to the null-executor case, matching the activity fix.

test/e2e/Tests/Tests/DisabledOrchestrationTests.cs
andystaples: The DTS bug  _may_ have already been fixed (I haven't checked), but re-enabling all those tests for DTS and string-matching DTS's style is outside the scope of this PR

test/FunctionsV2/Tests/Unit/DisabledFunctionDispatchTests.cs
```diff
@@ -0,0 +1,176 @@
...
+
+            // A normally registered entity is available.
+            var mockExecutor = new Mock<ITriggeredFunctionExecutor>();
+            extension.RegisterEntity(new FunctionName("ActiveEntity"), new RegisteredFunctionInfo(mockExecutor.Object, isOutOfProc: true));
+
+            // The unavailable entities have no active executor (EntityMiddleware's functionUnavailable check).
+            Assert.Null(extension.GetEntityInfo(new FunctionName("DisabledEntityNullInfo"))?.Executor);
+
+            RegisteredFunctionInfo nullExecutorInfo = extension.GetEntityInfo(new FunctionName("DisabledEntityNullExecutor"));
+            Assert.NotNull(nullExecutorInfo);
```
sophiatev, line 123: Can we actually try to run it here like we do the activity or no?

src/WebJobs.Extensions.DurableTask/Listener/TaskEntityShim.cs
sophiatev: I'm struggling to understand why the logic added for a null executor in OutOfProcMiddleware is not sufficient

andystaples: The `OutOfProcMiddleware.CallEntityAsync` fix only covers the `MiddlewarePassthrough` protocol (newer runtimes / .NET isolated). In-proc .NET **and the legacy out-of-proc languages — Node.js, Python, and PowerShell — dispatch entities through the classic `EntityMiddleware` → `TaskEntityShim` path instead**, which never touches `OutOfProcMiddleware`. You can see the split in `InitializeTaskHubWorker`: `AddEntityDispatcherMiddleware(ooprocMiddleware.CallEntityAsync)` is only wired up when `OutOfProcProtocol == MiddlewarePassthrough`; otherwise we register `EntityMiddleware`.

So for a disabled Node/Python/PowerShell (or in-proc .NET) entity, the null-executor guard in `OutOfProcMiddleware` does nothing — those never reach it. The `TaskEntityShim` gate (`functionInfo?.Executor != null && functionInfo.IsOutOfProc`) is what routes a null-executor entity to the deterministic per-operation in-proc path instead of `ExecuteOutOfProcBatch` (whose synchronous throw would abort + retry the work item = the poison loop). The two fixes cover disjoint runtimes, so both are needed.

I've also trimmed the comment down to just the essential "why" (10 lines → 5).

test/FunctionsV2/Tests/Unit/DisabledFunctionDispatchTests.cs
andystaples: The activity shim is self-contained (`GetObject(...).Run(...)`), so running it inline is trivial. Driving the classic entity dispatch is much heavier — it needs a rehydrated `DurableEntityContext`, a populated operation batch, and the throwing callback fed through `TaskEntityShim.ExecuteBatch` — so I kept this unit test focused on the state signal that actually routes the dispatch (`GetEntityInfo(name)?.Executor is null` / `HasActiveListener == false`, which is what `EntityMiddleware`'s `functionUnavailable` check keys off of).

The full run is covered end-to-end by the Node/Python entity E2E tests (the classic dispatch path), plus the `.NET isolated` `DisabledOrchestrationTests` for the passthrough path — those assert the orchestration actually reaches `Failed` promptly rather than poison-looping. So the executed behavior is exercised; this unit test just guards the routing invariant cheaply.

sophiatev (approved):

Issue #3027, opened 2025-02-04
Unclear exception when calling TerminateInstanceAsync on invalid instance


andystaples:
### Description
When calling TerminateInstanceAsync() in dotnet-isolated on an orchestration ID that is already completed, terminated, or otherwise not eligable for termination, you get an RpcException with the following message: 
Status(StatusCode=\"Unknown\", Detail=\"Exception was thrown by handler.\")
This is unhelpful and we are losing the detail provided by the exception thrown here:
https://github.com/Azure/azure-functions-durable-extension/blob/3bfd14970291ecc4b7e5bf7210a27e7ac6251cad/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs#L399

### Expected behavior
> Terminating an invalid orchestration should throw an exception with a helpful stack trace and error message. 

### Actual behavior
> Terminating a scheduled orchestration throws a generic RpcException. 

### Relevant source code snippets
```csharp
        [Function("TerminateInstance")]
        public static async Task<HttpResponseData> Run(
            [HttpTrigger(AuthorizationLevel.Anonymous, "get", "post")] HttpRequestData req,
            [DurableClient] DurableTaskClient client,
            string instanceId)
        {
            string reason = "Long-running orchestration was terminated early.";
            try 
            {
                await client.TerminateInstanceAsync(instanceId, reason);
                return req.CreateResponse(HttpStatusCode.OK);
            }
            catch (RpcException ex) 
            {
                var response = req.CreateResponse(HttpStatusCode.BadRequest);
                response.Headers.Add("Content-Type", "text/plain");
                await response.WriteStringAsync(ex.Message);
                return response;
            }
        }
```

### Known workarounds
None

### App Details

Latest Durable worker extension **1.2.2/dotnet-isolated/FunctionsV4

PR #3478, opened 2026-07-23
fix(grpc): map wrong-state Suspend/Resume to FAILED_PRECONDITION instead of opaque UNKNOWN

YunchuWang:
## Summary

Server-side fix for the gRPC error-mapping gap behind microsoft/durabletask-js#315.

A Durable Functions gRPC client (e.g. the `durable-functions` v4 / `durabletask-js` compat package) that calls a control-plane operation on an instance in a **terminal (or otherwise invalid) state** received an opaque `2 UNKNOWN: Exception was thrown by handler` instead of a meaningful status.

## Root cause

In `src/WebJobs.Extensions.DurableTask/TaskHubGrpcServer.cs`, the `TerminateInstance` and `RaiseEvent` handlers already wrap their body in a try/catch that maps:

- `InvalidOperationException` → `FAILED_PRECONDITION` (with detail)
- `ArgumentException` → `NOT_FOUND`
- `ArgumentNullException` → `INVALID_ARGUMENT`

But `SuspendInstance` and `ResumeInstance` had **no try/catch at all**. So the `InvalidOperationException` thrown by `DurableClient.SuspendAsync`/`ResumeAsync` (`Cannot suspend/resume the orchestration instance {id} because instance is in the {State} state.`) bubbled to the gRPC framework's default handler and was surfaced as an opaque `StatusCode.Unknown`.

The shared `GetOrchestrationInstanceStateAsync` in `DurableClient` throws the same exception types that `TerminateInstance` already handles (`ArgumentNullException` for null/empty id, `ArgumentException` for a missing instance, `InvalidOperationException` for wrong/terminal state), so mirroring its catch blocks into Suspend/Resume is a like-for-like fix.

## Fix

Add the same try/catch mapping used by `TerminateInstance` to `SuspendInstance` and `ResumeInstance`. Wrong-state Suspend/Resume now returns `FAILED_PRECONDITION` (with the detail message) instead of opaque `UNKNOWN`; missing instances return `NOT_FOUND`; null ids return `INVALID_ARGUMENT`.

Because the fix lives in the shared gRPC server, it benefits **all language SDKs** (.NET isolated, Python, Node, Java) that talk to the Durable extension over gRPC.

## Testing

- `dotnet build src/WebJobs.Extensions.DurableTask/WebJobs.Extensions.DurableTask.csproj -c Release` → **succeeds** (0 warnings, 0 errors).
- **No unit test added.** The existing gRPC error-mapping coverage (`FailedPrecondition` for Terminate / RaiseEvent) lives only in the e2e suite (`test/e2e/Tests/Tests/TerminateOrchestratorTests.cs`, `ExternalEventTests.cs`), which uses `FunctionAppFixture` + `HttpHelpers.InvokeHttpTrigger` and requires a live Functions host. There is no unit harness that directly instantiates `TaskHubGrpcServer`, so rather than invent a heavy new harness this change relies on mirroring the already-tested `TerminateInstance` path.

## Related

Server-side portion of microsoft/durabletask-js#315 (plain reference — intentionally **no** auto-close keyword, since it is a cross-repo issue).


YunchuWang:
Addressed the review feedback in `b4080ca3` (append-only).

**1. Cancellation no longer masked as `Unknown`.** The generic `catch (Exception) => Unknown` now has a filter:

```csharp
catch (Exception ex) when (ex is not RpcException && ex is not OperationCanceledException)
{
    throw new RpcException(new Status(StatusCode.Unknown, ex.Message));
}
```

So a client cancel/deadline (`OperationCanceledException`) propagates and gRPC surfaces `Cancelled`, and an already-formed `RpcException` is never re-wrapped. Genuinely fatal exceptions (OOM, stack overflow, etc.) are non-catchable / should crash the process rather than be remapped, so they're intentionally not special-cased here.

**2. Duplication removed.** The identical mapping is now a single shared helper, `InvokeControlPlaneOperationAsync(Func<Task> operation)`, and the three control-plane handlers (`TerminateInstance`, `SuspendInstance`, `ResumeInstance`) route through it — so the mapping can't drift between them. `TerminateInstance`'s status codes and detail strings are byte-identical to before (its existing e2e `FAILED_PRECONDITION` assertion still holds); Terminate additionally picks up the cancellation improvement for free.

**Deliberately out of scope:** `RaiseEvent` is left as-is — its mapping is conditionally gated (`ThrowStatusExceptionsOnRaiseEvent`) and uses different messages, so it isn't a clean fit for the shared control-plane helper. `Rewind`/`Restart` are likewise untouched.

Local `dotnet build -c Release` is clean (0 warnings; the project has `TreatWarningsAsErrors=true` + StyleCop.Analyzers).

src/WebJobs.Extensions.DurableTask/TaskHubGrpcServer.cs
bachuv, line 1: Can we add some tests for this fix? Maybe mirror the tests for TerminateInstance?

YunchuWang:
Thanks @bachuv! Added e2e coverage mirroring the `TerminateInstance` tests, in two parts:

**1. Wrong-state suspend/resume — updated the existing tests (`af36b914`)**

`SuspendResumeTests.cs` already mirrored `TerminateOrchestratorTests.cs` for the wrong-state cases, but the expected strings were pinned to the old opaque `UNKNOWN` gRPC status (with a `// ... will update this code when that bug is fixed` note pointing at #3027). This PR *is* that fix, so I updated the Isolated/Java localizer expectations to the new `FAILED_PRECONDITION` mapping and removed the stale comment. They now assert the fixed behavior directly, e.g.:

`InvalidOperationException: Cannot suspend the orchestration instance {id} because instance is in the Suspended state.`

(covers suspend-suspended, resume-running, and suspend/resume-completed). These are the tests that were red on the earlier commits — they were encoding the old bug.

**2. Not-found suspend/resume — new tests (`70e753f5`)**

Added `SuspendNonExistentOrchestration_ShouldFail` + `ResumeNonExistentOrchestration_ShouldFail`, mirroring `TerminateNonExistantOrchestration_ShouldFail`, to cover the other path this fix improves — `ArgumentException` now maps to `NotFound` instead of the old `Unknown`:

- isolated asserts `Status(StatusCode="NotFound", ...)`
- java asserts `NOT_FOUND: ArgumentException: No instance ...`

I gated these two new not-found tests to the gRPC workers (`dotnet-isolated` + `java`) via `[Trait(..., "Skip")]` for node/python/powershell, because suspend/resume of a non-existent instance goes through the HTTP webhook path for those workers, which this gRPC-server change doesn't touch — I didn't want to assert behavior this PR doesn't change. Happy to extend to full-language coverage if you'd prefer.

bachuv (approved):

Issue #1343, opened 2020-05-02
Binding to System.Object in activity function


cgillum:
### Description
Here is an example orchestrator function:

```csharp
[FunctionName(nameof(AnyToString))]
public static string AnyToString([ActivityTrigger] object input) => input.ToString();
```
And here is the orchestrator code that calls it, passing in a `5` as the parameter.

```csharp
ctx.CallActivityAsync<string>(nameof(AnyToString), 5)
```

### Expected behavior
I expected the runtime type of the `input` parameter to be `int` and the result of the function to be `"5"`.

### Actual behavior
The runtime type of `input` is `Microsoft.Azure.WebJobs.Extensions.DurableTask.DurableActivityContext` and the result is always `"Microsoft.Azure.WebJobs.Extensions.DurableTask.DurableActivityContext""`

This is unintuitive. The fix will require changing the activity trigger binding logic. This is potentially breaking if anyone depended on this behavior, but seems unlikely to actually impact any real apps or even tests.

### Known workarounds
Use explicit types instead of `object`. However, this can be inconvenient if the activity function needs to work with arbitrary types, like in my trivial example.

### App Details

- **Durable Functions extension version**: v2.2.1
- **Azure Functions runtime version**: 3.0
- **Programming language used**: C# 8

YunchuWang:
not planned, please see chris' comment in https://github.com/Azure/azure-functions-durable-extension/pull/3484

PR #3484, opened 2026-07-24
Add opt-in [ActivityTrigger(BindToInput = true)] for object/dynamic parameters (#1343)

YunchuWang:
## Summary

Part of #1343.

An `[ActivityTrigger]` parameter typed as `System.Object` (or `dynamic`, whose reflected `ParameterType` is also `typeof(object)`) receives the `DurableActivityContext` instead of the activity input, unlike every other parameter type.

This PR does **not** change that default. It adds a per-function opt-in so applications can get the intuitive behavior today, and leaves flipping the default to the next major version.

### Root cause

`ActivityTriggerAttributeBindingProvider.BindAsync` special-cases `typeof(object)` and assigns the **raw trigger value** — the `DurableActivityContext` on the normal in-proc path, or the raw serialized string on the admin/portal path — instead of binding the input:

```csharp
if (destinationType == typeof(object))
{
    convertedValue = value; // the DurableActivityContext, not the input
}
```

That branch was introduced in c9936623 (2017) with the comment `// Straight assignment`.

### Change: opt in per activity function

A new `BindToInput` property on the in-proc `ActivityTriggerAttribute`:

```csharp
[ActivityTrigger(BindToInput = true)] object input   // NEW: the activity input
[ActivityTrigger] object input                       // UNCHANGED: DurableActivityContext
```

- **The default is unchanged.** `BindToInput` defaults to `false`, so existing apps that do `((IDurableActivityContext)input).GetInput<T>()` or `dynamic input; input.GetInput<T>()` keep working exactly as before.
- **This is purely additive and non-breaking**, and therefore safe to ship in a 3.x minor release.
- **It is opt-in per activity function**, so an app can migrate one function at a time.
- The flag only affects `object`/`dynamic` parameters. It is ignored for every other parameter type, because all other types already bind to the input.
- **Intent for v4:** flip the default to `true` and mark `BindToInput` obsolete. At that point, declaring the parameter as `IDurableActivityContext` is the way to receive the context.

Because the unintuitive default remains until v4, this PR is deliberately worded as *Part of* #1343 rather than *Fixes*, so merging it does not auto-close the issue.

The isolated (.NET worker) model does not need a matching change: `ActivityInputConverter` always deserializes to the target type and has no `typeof(object)` special case, so it never had this bug.

### Testing

New E2E tests in `test/FunctionsV2` assert the **concrete CLR type** of the bound value (`value.GetType().FullName`) in addition to its value, so that a raw `"5"` string, a boxed `Int64`, and a `JValue` are all distinguishable — comparing `ToString()` alone cannot tell them apart.

| Test | Parameter | Input | Bound value |
| --- | --- | --- | --- |
| `ActivityTriggerAsObject_LegacyDefault_BindsToContext` | `[ActivityTrigger] object` | `5` | `DurableActivityContext` (and `GetInput<object>()` on it returns `System.Int64` `5`) |
| `ActivityTriggerAsObject_BindToInput_Scalar` | `[ActivityTrigger(BindToInput = true)] object` | `5` | `System.Int64` `5` |
| `ActivityTriggerAsObject_BindToInput_StructuredObject` | `[ActivityTrigger(BindToInput = true)] object` | `{ "Name": "foo", "Count": 3 }` | `Newtonsoft.Json.Linq.JObject` |
| `ActivityTriggerAsObject_BindToInput_Array` | `[ActivityTrigger(BindToInput = true)] object` | `[1, "two", 3.5]` | `Newtonsoft.Json.Linq.JArray` |
| `ActivityTriggerAsDynamic_BindToInput` | `[ActivityTrigger(BindToInput = true)] dynamic` | `{ "Name": "foo", "Count": 3 }` | `Newtonsoft.Json.Linq.JObject` |
| `ActivityTriggerAsObject_LegacyDefault_AdminInvocation` | `[ActivityTrigger] object` | raw serialized string | `System.String` (the raw JSON) |
| `ActivityTriggerAsObject_BindToInput_AdminInvocation` | `[ActivityTrigger(BindToInput = true)] object` | raw serialized string | `Newtonsoft.Json.Linq.JObject` |

The structured-object and array cases matter because they are not `JValue`s, so they take the serialize-and-reparse path in `DurableActivityContext.GetInput` rather than the cheap `JValue.ToObject` path. The last two rows cover the Azure Portal / admin API invocation path, where the trigger value arrives as a raw serialized string rather than a `DurableActivityContext`.

The `LegacyDefault` tests are the compatibility guard for the promise above: they fail if the default ever changes silently.

Verified locally against Azurite: all 16 tests in `BindingTests` pass. Inverting the `BindToInput` gate causes all 7 new tests to fail, confirming none of them is vacuous.


src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
bachuv, line 1: This would be a breaking change right? Can we add to the checks instead of replacing them?

src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
```diff
@@ -148,6 +144,9 @@ public Task<ITriggerData> BindAsync(object value, ValueBindingContext context)
                 }
                 else
                 {
+                    // Deserialize the activity's input to the requested type. This includes System.Object and dynamic
+                    // parameters, which previously received the raw trigger value (DurableActivityContext or serialized input)
```
berndverst: Could we preserve the legacy binding behind an opt-in and only change the default in a major release? For normal invocations, `[ActivityTrigger] object`/`dynamic` currently receives `DurableActivityContext`; for portal/admin invocations it receives the raw JSON string. After this change it receives a deserialized scalar or `JToken` instead. Existing casts to `IDurableActivityContext` and dynamic calls such as `value.GetInput<T>()` will fail after a v3 update, and there is no compatibility switch.

src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
```diff
@@ -148,6 +144,9 @@ public Task<ITriggerData> BindAsync(object value, ValueBindingContext context)
                 }
                 else
                 {
+                    // Deserialize the activity's input to the requested type. This includes System.Object and dynamic
```
berndverst: This also introduces an extra full JSON round trip for structured object inputs. `DurableActivityContext.GetInput` already has a parsed `JObject`/`JArray`, then calls `ToString(Formatting.None)` and `MessagePayloadDataConverter.Deserialize`, allocating a payload-sized string and reparsing it on every invocation. Could the default converter deserialize via `JToken.CreateReader()` and retain the text path only for custom converters whose reader semantics must be preserved?

test/FunctionsV2/Tests/E2E/BindingTests.cs
```diff
@@ -207,6 +207,34 @@ public async Task ActivityTriggerAsNumber(string storageProviderType)
...
+        public async Task ActivityTriggerAsObject(string storageProviderType)
+        {
+            using (ITestHost host = TestHelpers.GetJobHost(this.loggerProvider, nameof(this.ActivityTriggerAsObject), false, storageProviderType: storageProviderType))
+            {
+                await host.StartAsync();
+
+                // Using StartOrchestrationArgs to start an activity function because it's easier than creating a new type.
+                var startArgs = new StartOrchestrationArgs();
+                startArgs.FunctionName = nameof(TestActivities.BindToObject);
+                startArgs.Input = 5;
```
berndverst: Please add a structured object/array case and assert the concrete CLR types, including the portal/admin invocation path. `5` is a `JValue` and takes the cheap `ToObject` path; comparing only `ToString()` cannot expose the `JObject`/`JArray` serialize-and-reparse path or distinguish a raw `"5"` string from `Int64(5)` or `JValue(5)`.

berndverst (commented):
The object/dynamic binding change has a compatibility break and an avoidable structured-payload round trip; details are inline.

berndverst (changes requested):

src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
YunchuWang: You were right to push back — I've reworked the PR so the default is unchanged.

One note on "add to the checks instead of replacing them": that specific shape isn't implementable. `object` and `dynamic` **both** reflect as `typeof(object)` (I verified: `dynamic` differs only by a compiler-emitted `DynamicAttribute`), so the parameter type alone carries no signal to branch on. It's genuinely either/or — which is why I went with an explicit opt-in instead:

```csharp
[ActivityTrigger(BindToInput = true)] object input   // NEW: the activity input
[ActivityTrigger] object input                       // UNCHANGED: DurableActivityContext
```

`BindToInput` defaults to `false`, so this is purely additive and nothing existing changes behavior. There's now a dedicated `ActivityTriggerAsObject_LegacyDefault_BindsToContext` test that asserts the old behavior is still exactly what it was, so the default can't drift silently.

src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
YunchuWang: Done — the PR now preserves the legacy binding behind an opt-in, and defers the default flip to the next major.

```csharp
[ActivityTrigger(BindToInput = true)] object input   // NEW: the activity input
[ActivityTrigger] object input                       // UNCHANGED: DurableActivityContext
```

`BindToInput` defaults to `false`, so `((IDurableActivityContext)input).GetInput<T>()` and `dynamic input; input.GetInput<T>()` keep working untouched. Purely additive, safe for a 3.x minor. Intent for v4 is to flip the default and mark the property `[Obsolete]`, at which point `IDurableActivityContext` is the way to receive the context.

Your point about the failure mode is the reason I didn't just ship the flip: it's a *silent runtime* break — the cast still compiles and dies in production. That's not something to land in a minor.

Two bits of history that shaped the choice, in case they're useful:

- The branch was introduced in c9936623 (2017) with the comment `// Straight assignment` — a fallthrough, not a designed contract.
- It was then filed as #1343 **by @cgillum, labeled `bug`**, where he wrote: *"potentially breaking if anyone depended on this behavior, but seems unlikely to actually impact any real apps or even tests."* Confirming that today: no test in the repo binds `[ActivityTrigger] object`, and the behavior appears in no documentation — Learn documents `IDurableActivityContext` as the way to get the context.

That's what makes me want the default to eventually flip rather than live behind a flag forever, but I agree entirely that it can't happen here. I've added @cgillum to weigh in on the v4 timing.

test/FunctionsV2/Tests/E2E/BindingTests.cs
YunchuWang: Done, and you were right that `ToString()` alone was hiding things. The test activities now return `value.GetType().FullName` alongside the value, and I asserted the types I actually observed rather than the ones I expected:

| Test | Parameter | Input | Bound value |
| --- | --- | --- | --- |
| `..._LegacyDefault_BindsToContext` | `[ActivityTrigger] object` | `5` | `DurableActivityContext` (and `GetInput<object>()` on it returns `System.Int64` `5`) |
| `..._BindToInput_Scalar` | `BindToInput = true` `object` | `5` | `System.Int64` `5` |
| `..._BindToInput_StructuredObject` | `BindToInput = true` `object` | `{"Name":"foo","Count":3}` | `Newtonsoft.Json.Linq.JObject` |
| `..._BindToInput_Array` | `BindToInput = true` `object` | `[1,"two",3.5]` | `Newtonsoft.Json.Linq.JArray` |
| `ActivityTriggerAsDynamic_BindToInput` | `BindToInput = true` `dynamic` | `{"Name":"foo","Count":3}` | `JObject` (identical to `object`) |
| `..._LegacyDefault_AdminInvocation` | `[ActivityTrigger] object` | raw serialized string | `System.String` (raw JSON) |
| `..._BindToInput_AdminInvocation` | `BindToInput = true` `object` | raw serialized string | `JObject` |

Concretely proving your point: the scalar boxes as **`Int64`**, not `Int32` or `JValue`. A `ToString()` comparison would never have surfaced that.

The structured-object and array cases take the serialize-and-reparse path you flagged in the other thread, rather than the cheap `JValue.ToObject` path. The last two rows are the portal/admin invocation path — I reused the `[NoAutomaticTrigger]` + `outputWrapper` + `host.CallAsync` shape from `CallActivity_Like_From_Azure_Portal`, but put them in `BindingTests` because `DurableTaskEndToEndTests` requires ETW/admin rights to construct.

Verification: 16/16 `BindingTests` pass against Azurite. I also inverted the `BindToInput` gate as a mutation check — all 7 new tests fail, so none of them is vacuous. The legacy guard in particular fails with `System.Int64|5`, i.e. exactly what the previous version of this PR produced.

src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
YunchuWang: Good catch, and the round trip is real — I traced it: `DurableActivityContext.GetInput` takes the cheap `JValue.ToObject` path for scalars, but for `JObject`/`JArray` it does `ToString(Formatting.None)` and re-deserializes, allocating a payload-sized string and reparsing on every invocation.

I'd like to handle it separately rather than in this PR, for two reasons:

1. **It's pre-existing and not introduced here.** That path already runs today for every structured input bound to a typed parameter (`MyPoco`, `JObject`, etc.). This PR doesn't add a new caller — and now that the fix is opt-in, it doesn't even change which parameters reach it unless someone explicitly sets `BindToInput`.
2. **`JToken.CreateReader()` isn't behavior-neutral.** It changes Json.NET's date/`JValue` typing semantics versus a fresh text parse, and it would affect *all* callers of `GetInput`, not just `object` parameters. That's a materially riskier change than this one, and it deserves its own PR with its own tests rather than riding along on a binding fix.

Happy to file an issue for it and pick it up next — say the word and I'll open one referencing this thread. If you'd rather it block here, I'll do it, but I think the blast radius argues for separating them.

src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
```diff
@@ -148,6 +144,9 @@ public Task<ITriggerData> BindAsync(object value, ValueBindingContext context)
                 }
                 else
                 {
+                    // Deserialize the activity's input to the requested type. This includes System.Object and
+                    // dynamic parameters, which previously received the DurableActivityContext instead of the
+                    // input value (issue #1343).
```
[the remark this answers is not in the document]
YunchuWang: Fixed — and the comment that prompted this is gone entirely, since the branch is now restored (gated on an opt-in) rather than removed.

Both defects you flagged across the review rounds are addressed:

- **Wording.** I dropped "Deserialize the activity's input". You were right that it was inaccurate: `GetInput(Type)` returns the *raw serialized string* when `rawInput` is true (the modern OOProc path). The comment now describes what the branch actually does — the parameter "receives the trigger value" — without claiming deserialization always happens.
- **Stray carriage returns.** Confirmed there were exactly 3 `\r\r` sequences. Root cause worth recording: `ActivityTriggerAttributeBindingProvider.cs` is stored in git with **CRLF** while the rest of the repo is LF, so with `core.autocrlf=true` a normal `git add` renormalizes the whole file and produces a 499-line diff. I staged that file with `core.autocrlf=false` so the blob keeps CRLF and matches `dev` byte-for-byte outside the 4 lines I changed. Byte-scanned the committed blobs: `strayCR=0` across all touched files.

YunchuWang:
## Reworked: opt-in instead of a default change

Pushed cd7b020c. This PR no longer changes any existing behavior.

@berndverst's objection was that the previous version shipped a **silent runtime** break in a 3.x minor — `((IDurableActivityContext)input).GetInput<T>()` still compiles and fails in production. That's correct and not something to land in a minor, so the fix is now opt-in per activity function:

```csharp
[ActivityTrigger(BindToInput = true)] object input   // NEW: the activity input
[ActivityTrigger] object input                       // UNCHANGED: DurableActivityContext
```

`BindToInput` defaults to `false` → purely additive, nothing existing changes. All review threads are addressed inline (opt-in, test coverage w/ concrete CLR type assertions incl. the portal/admin path, comment wording, stray CRs). The one item I'm proposing to defer is the `GetInput` serialize-and-reparse round trip — it's pre-existing and affects all callers, so I'd rather do it in its own PR; details in that thread.

Retitled and reworded to **Part of #1343** rather than *Fixes*, so merging doesn't auto-close — the unintuitive default survives this PR.

---

### @cgillum — question for you, since you filed #1343

You opened this in 2020 and labeled it `bug`, writing:

> *"This is potentially breaking if anyone depended on this behavior, but seems unlikely to actually impact any real apps or even tests."*

Two data points on that prediction, six years on:

- **No test** in this repo binds `[ActivityTrigger] object`/`dynamic` — your "or even tests" held exactly.
- **No documentation** describes it. Learn documents `IDurableActivityContext` as the way to receive the context, and the branch originated in your c9936623 (2017) with the comment `// Straight assignment` — a fallthrough, not a designed contract.

So the behavior looks undocumented, untested, and accidental. But it's also had six more years for Hyrum's Law to accumulate, and the failure mode is the bad kind (silent, runtime, production).

**What I'd like your read on: is the plan to flip the default in v4 and mark `BindToInput` obsolete the right endgame, or should the corrected behavior stay opt-in indefinitely?**

That distinction matters now, because it decides whether this property is a temporary migration aid or permanent public API we support forever. My bias is the former — a flag whose only job is to faithfully reproduce a 2017 bug is a real long-term tax, and an opt-in that users must first discover doesn't actually fix the unintuitive default you originally reported.

If flipping in v4 is the plan, one thing that would de-risk it: emitting a one-time **indexing-time warning** in 3.x naming any function with an `object`-typed `[ActivityTrigger]`, pointing at `BindToInput` / `IDurableActivityContext`. That converts the silent break into a loud one well before v4. Happy to do it as a follow-up PR if you both think it's worth it.

src/WebJobs.Extensions.DurableTask/Bindings/ActivityTriggerAttributeBindingProvider.cs
berndverst: If we think it's only a theoretical breaking change that no real user would be impacted by then go ahead :)

berndverst (approved):

cgillum:
Hi @YunchuWang. Thanks for proactively looking into older issues that can be fixed.

I think we should be really careful about making changes to .NET in-proc binding behavior. This fix only benefits .NET in-proc customers, and .NET in-proc is going out of support later this year. All .NET customers should be migrating to the .NET out-of-proc experience for Functions, which has been available for a few years now. For this reason, I don't expect customers to start using this new feature. New customers will be starting with .NET isolated and existing customers will be just making minor fixes to existing in-proc apps or will otherwise be migrating away.

My recommendation would be to leave the code as-is to avoid any accidental regressions for the remaining in-proc customers.

PR #3491, opened 2026-07-27
Optimize typed interface discovery during startup

### LINKED ISSUE #3486: [Performance] Avoid enumerating every loaded type during Durable extension startup

`DurableTaskExtension` unconditionally constructs and initializes `TypedCodeProvider` during extension startup, including for applications that never use generated typed interfaces:

https://github.com/Azure/azure-functions-durable-extension/blob/4a52e5a802660bdb8fed0141d75b9d1e89463d2f/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs#L121-L150

`TypedCodeProvider.Initialize()` enumerates `DefinedTypes` from every loaded assembly, materializes all of them into a `List<TypeInfo>`, and then performs five separate linear searches for known full names:

https://github.com/Azure/azure-functions-durable-extension/blob/4a52e5a802660bdb8fed0141d75b9d1e89463d2f/src/WebJobs.Extensions.DurableTask/TypedCodeProvider.cs#L43-L69

The result is only used to conditionally register the optional typed-client bindings (and similarly to convert typed orchestration contexts):

https://github.com/Azure/azure-functions-durable-extension/blob/4a52e5a802660bdb8fed0141d75b9d1e89463d2f/src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs#L414-L449

### Performance impact

This work is directly on the cold-start path and scales with the total number of assemblies and types loaded by the Functions host, not with the number of Durable functions or typed interfaces in the app. It also creates a large temporary list and the backing arrays needed while that list grows.

A directional .NET 10.0.10 microbenchmark with 84 loaded assemblies and 22,288 types measured:

| Discovery strategy | Mean per scan | Allocated per scan |
|---|---:|---:|
| Current `DefinedTypes` + `ToList()` + five scans | 61.5 ms | 1.94 MB |
| Prototype `Assembly.GetType()` lookup for the five known names | 0.31 ms | 10.6 KB |

This is not an end-to-end Functions cold-start benchmark, and absolute results will vary with the host/app assembly graph. It does demonstrate that the current cost is avoidable and grows in exactly the wrong dimension for larger apps.

I did not find an exact open issue for `TypedCodeProvider` startup/type-enumeration performance. This proposal is deliberately limited to discovery internals; it does not change the typed binding feature or its public surface.

Replace full type enumeration with targeted, allocation-light lookup while preserving the current eager initialization point initially:

1. Store the five fully qualified generated type names once.
2. Iterate the already loaded assemblies and call `Assembly.GetType(fullName, throwOnError: false, ignoreCase: false)` for unresolved names.
3. Stop once all five types have been found; do not enumerate unrelated `DefinedTypes` or materialize a global list.
4. Set `IsInitialized` only when the same five types required today are present.
5. Retain equivalent handling for assemblies that cannot be inspected, but scope exception handling to the individual lookup/assembly rather than changing startup behavior.

This approach still permits the generated types to live in any loaded assembly, including separate assemblies, instead of assuming a particular generated assembly name.

- No changes to public APIs, binding names, generated type names, or host configuration.
- Preserve the current case-sensitive full-name match.
- Preserve all-or-nothing `IsInitialized` semantics and the existing typed-client/typed-context binding behavior.
- Preserve behavior when no generated typed interfaces are present.
- Do not move discovery later in the lifecycle in the first change unless tests prove binding registration timing remains identical.

- **Initialize lazily only when a typed binding is requested.** This could remove almost all cost for non-typed apps, but it has more risk because binding rules are conditionally registered during extension initialization. Targeted eager lookup captures most of the gain with less lifecycle risk.
- **Search only an assembly with a hard-coded name.** This is faster but could break apps whose generated types are emitted into a differently named assembly.
- **Cache the current complete type list.** This reduces repeated work only if initialization occurs more than once and still pays the dominant first-start cost and memory spike.

**Additional context**

- Add tests for zero, partial, and all five generated types, including types split across multiple loaded assemblies.
- Verify the same typed binding rules are registered and the same typed context/client instances are created before and after the change.
- Add a focused benchmark with a representative loaded-assembly graph and report elapsed time plus allocated bytes.
- Acceptance target: no enumeration/materialization of unrelated `TypeInfo` instances and a material reduction in cold-start CPU/allocation (for example, at least 90% lower allocations in the no-typed-interface benchmark).
- Run an end-to-end cold-start comparison for a non-typed app and a typed-interface app to ensure the optimization benefits the common case without regressing the optional feature.

berndverst:
## Summary
- replace global `DefinedTypes` enumeration and list materialization with targeted `Assembly.GetType` lookups for the five generated implementation types
- keep eager discovery during `DurableTaskExtension` construction and stop scanning as soon as all required types are resolved
- add focused regression coverage for absent, partial, complete, split-assembly, case-sensitive, early-stop, inspection-failure, and typed-instantiation behavior

## Performance rationale
The startup path now performs only case-sensitive lookups for unresolved generated type names and never enumerates unrelated types. This implements the lookup shape from the directional benchmark in #3486 (61.5 ms / 1.94 MB for full enumeration versus 0.31 ms / 10.6 KB for targeted lookup). This repository has no benchmark harness, so tests structurally assert that `DefinedTypes` is never touched and that assembly inspection stops after all five types are found.

## Backward compatibility
- no public API, binding name, configuration, or serialization changes
- preserves eager lifecycle timing, exact case-sensitive full names, first-match assembly order, and support for the five types being split across arbitrary loaded assemblies
- preserves all-or-nothing `IsInitialized` semantics and existing typed client/context construction
- preserves normal startup when typed interfaces are absent and typed binding registration when all required implementations are present
- skips expected assembly/type load failures per assembly while allowing unexpected failures to surface instead of broadly swallowing them

## Validation
Tests were added first; the initial targeted net8.0 run failed with CS1501 because the assembly-scoped initialization seam did not yet exist.

- `dotnet test test\FunctionsV2\WebJobs.Extensions.DurableTask.Tests.V2.csproj --filter "FullyQualifiedName~TypedCodeProviderTests" --no-restore --nologo`
  - net8.0: 9 passed, 0 failed
  - net10.0: 9 passed, 0 failed
- `dotnet build src\WebJobs.Extensions.DurableTask\WebJobs.Extensions.DurableTask.csproj --configuration Release --no-restore --nologo`
  - succeeded for net8.0 and net10.0 with 0 warnings and 0 errors

Fixes #3486

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/TypedCodeProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/TypedCodeProvider.cs
@@ -2,8 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
-using System.Collections.Generic;
-using System.Linq;
+using System.IO;
 using System.Reflection;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
@@ -27,6 +26,21 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         public const string TypedDurableOrchestrationStarter = "TypedDurableOrchestrationStarter";
         public const string ITypedDurableOrchestrationStarter = "ITypedDurableOrchestrationStarter";
 
+        private const string TypedDurableOrchestrationContextFullName =
+            Namespace + "." + TypedDurableOrchestrationContext;
+
+        private const string TypedDurableClientFullName =
+            Namespace + "." + TypedDurableClient;
+
+        private const string TypedDurableOrchestrationCallerFullName =
+            Namespace + "." + TypedDurableOrchestrationCaller;
+
+        private const string TypedDurableActivityCallerFullName =
+            Namespace + "." + TypedDurableActivityCaller;
+
+        private const string TypedDurableOrchestrationStarterFullName =
+            Namespace + "." + TypedDurableOrchestrationStarter;
+
         private Type typedDurableOrchestrationContextType;
         private Type typedDurableClientType;
 
@@ -42,26 +56,80 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         public void Initialize()
         {
-            var types = AppDomain.CurrentDomain.GetAssemblies().SelectMany(t =>
+            this.Initialize(AppDomain.CurrentDomain.GetAssemblies());
+        }
+
+        internal void Initialize(Assembly[] assemblies)
+        {
+            this.typedDurableOrchestrationContextType = null;
+            this.typedDurableClientType = null;
+            this.typedDurableOrchestrationCallerType = null;
+            this.typedDurableActivityCallerType = null;
+            this.typedDurableOrchestrationStarterType = null;
+            this.IsInitialized = false;
+
+            foreach (Assembly assembly in assemblies)
             {
+                Type typedDurableOrchestrationContextType = this.typedDurableOrchestrationContextType;
+                Type typedDurableClientType = this.typedDurableClientType;
+                Type typedDurableOrchestrationCallerType = this.typedDurableOrchestrationCallerType;
+                Type typedDurableActivityCallerType = this.typedDurableActivityCallerType;
+                Type typedDurableOrchestrationStarterType = this.typedDurableOrchestrationStarterType;
+
                 try
                 {
-                    return t.DefinedTypes;
+                    typedDurableOrchestrationContextType ??= assembly.GetType(
+                        TypedDurableOrchestrationContextFullName,
+                        throwOnError: false,
+                        ignoreCase: false);
+                    typedDurableClientType ??= assembly.GetType(
+                        TypedDurableClientFullName,
+                        throwOnError: false,
+                        ignoreCase: false);
+                    typedDurableOrchestrationCallerType ??= assembly.GetType(
+                        TypedDurableOrchestrationCallerFullName,
+                        throwOnError: false,
+                        ignoreCase: false);
+                    typedDurableActivityCallerType ??= assembly.GetType(
+                        TypedDurableActivityCallerFullName,
+                        throwOnError: false,
+                        ignoreCase: false);
+                    typedDurableOrchestrationStarterType ??= assembly.GetType(
+                        TypedDurableOrchestrationStarterFullName,
+                        throwOnError: false,
+                        ignoreCase: false);
                 }
-                catch
+                catch (Exception exception) when (IsAssemblyInspectionException(exception))
                 {
-                    return new List<TypeInfo>();
+                    continue;
                 }
-            }).ToList();
 
-            this.typedDurableOrchestrationContextType = types.FirstOrDefault(t => t.FullName == $"{Namespace}.{TypedDurableOrchestrationContext}");
-            this.typedDurableClientType = types.FirstOrDefault(t => t.FullName == $"{Namespace}.{TypedDurableClient}");
+                this.typedDurableOrchestrationContextType = typedDurableOrchestrationContextType;
+                this.typedDurableClientType = typedDurableClientType;
+                this.typedDurableOrchestrationCallerType = typedDurableOrchestrationCallerType;
+                this.typedDurableActivityCallerType = typedDurableActivityCallerType;
+                this.typedDurableOrchestrationStarterType = typedDurableOrchestrationStarterType;
 
-            this.typedDurableOrchestrationCallerType = types.FirstOrDefault(t => t.FullName == $"{Namespace}.{TypedDurableOrchestrationCaller}");
-            this.typedDurableActivityCallerType = types.FirstOrDefault(t => t.FullName == $"{Namespace}.{TypedDurableActivityCaller}");
-            this.typedDurableOrchestrationStarterType = types.FirstOrDefault(t => t.FullName == $"{Namespace}.{TypedDurableOrchestrationStarter}");
+                if (this.AreAllTypesResolved())
+                {
+                    this.IsInitialized = true;
+                    return;
+                }
+            }
+        }
 
-            this.IsInitialized = this.typedDurableOrchestrationContextType != null &&
+        private static bool IsAssemblyInspectionException(Exception exception)
+        {
+            return exception is FileNotFoundException ||
+                exception is FileLoadException ||
+                exception is BadImageFormatException ||
+                exception is TypeLoadException ||
+                exception is ReflectionTypeLoadException;
+        }
+
+        private bool AreAllTypesResolved()
+        {
+            return this.typedDurableOrchestrationContextType != null &&
                 this.typedDurableClientType != null &&
                 this.typedDurableOrchestrationCallerType != null &&
                 this.typedDurableActivityCallerType != null &&
--- a/test/FunctionsV2/Tests/Unit/TypedCodeProviderTests.cs
+++ b/test/FunctionsV2/Tests/Unit/TypedCodeProviderTests.cs
@@ -1,0 +1,289 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Reflection;
+using System.Reflection.Emit;
+using Moq;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
+{
+    public class TypedCodeProviderTests
+    {
+        private const string TypedDurableOrchestrationContextFullName =
+            TypedCodeProvider.Namespace + "." + TypedCodeProvider.TypedDurableOrchestrationContext;
+
+        private const string TypedDurableClientFullName =
+            TypedCodeProvider.Namespace + "." + TypedCodeProvider.TypedDurableClient;
+
+        private const string TypedDurableOrchestrationCallerFullName =
+            TypedCodeProvider.Namespace + "." + TypedCodeProvider.TypedDurableOrchestrationCaller;
+
+        private const string TypedDurableActivityCallerFullName =
+            TypedCodeProvider.Namespace + "." + TypedCodeProvider.TypedDurableActivityCaller;
+
+        private const string TypedDurableOrchestrationStarterFullName =
+            TypedCodeProvider.Namespace + "." + TypedCodeProvider.TypedDurableOrchestrationStarter;
+
+        private static readonly string[] RequiredTypeNames =
+        {
+            TypedDurableOrchestrationContextFullName,
+            TypedDurableClientFullName,
+            TypedDurableOrchestrationCallerFullName,
+            TypedDurableActivityCallerFullName,
+            TypedDurableOrchestrationStarterFullName,
+        };
+
+        [Fact]
+        [Trait("Category", PlatformSpecificHelpers.TestCategory)]
+        public void Initialize_NoGeneratedTypes_IsNotInitialized()
+        {
+            var provider = new TypedCodeProvider();
+
+            provider.Initialize(new[] { CreateAssembly() });
+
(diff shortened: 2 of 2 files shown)


src/WebJobs.Extensions.DurableTask/TypedCodeProvider.cs
```diff
@@ -42,26 +56,80 @@ internal class TypedCodeProvider
...
+                        TypedDurableActivityCallerFullName,
+                        throwOnError: false,
+                        ignoreCase: false);
+                    typedDurableOrchestrationStarterType ??= assembly.GetType(
+                        TypedDurableOrchestrationStarterFullName,
+                        throwOnError: false,
+                        ignoreCase: false);
                 }
-                catch
+                catch (Exception exception) when (IsAssemblyInspectionException(exception))
```
YunchuWang, line 102: **This narrows exception handling on the host startup path.**

`Initialize()` runs inside the `DurableTaskExtension` constructor, so anything that escapes here fails host startup. The previous `catch { }` swallowed everything; this filter now lets non-listed exceptions propagate.

Two concrete problems:

1. **`Assembly.GetType(string, bool, bool)` is `virtual`, and the base `Assembly` implementation is `{ throw NotImplemented.ByDesign; }`** (`NotImplementedException`) — see [Assembly.cs in dotnet/runtime](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Reflection/Assembly.cs). Any `Assembly` subclass that does not override it — profiler/instrumentation shims, AOP proxies, `PersistedAssemblyBuilder` (which throws `NotSupportedException`) — will now take down startup instead of being skipped. Neither `NotImplementedException` nor `NotSupportedException` is in the whitelist.

2. **`ReflectionTypeLoadException` in the whitelist is unreachable here.** It is thrown by `GetTypes()` / `DefinedTypes`, never by `GetType()`. Its presence suggests the list was derived from the old enumeration code path rather than from the contract of the new API.

Since this PR is scoped to performance and explicitly promises no behavior change, I would keep the original semantics:

```csharp
catch
{
    // An assembly that cannot be inspected must never fail host startup.
}
```

If surfacing these failures is desirable, that seems worth its own PR — and should probably log via `EndToEndTraceHelper` rather than throw from a constructor.

src/WebJobs.Extensions.DurableTask/TypedCodeProvider.cs
```diff
@@ -42,26 +56,80 @@ internal class TypedCodeProvider
...
+            this.typedDurableOrchestrationStarterType = null;
+            this.IsInitialized = false;
+
+            foreach (Assembly assembly in assemblies)
             {
+                Type typedDurableOrchestrationContextType = this.typedDurableOrchestrationContextType;
+                Type typedDurableClientType = this.typedDurableClientType;
+                Type typedDurableOrchestrationCallerType = this.typedDurableOrchestrationCallerType;
+                Type typedDurableActivityCallerType = this.typedDurableActivityCallerType;
+                Type typedDurableOrchestrationStarterType = this.typedDurableOrchestrationStarterType;
```
YunchuWang, line 77: Nit (readability): these five locals shadow the fields of the same name, so inside the loop `typedDurableClientType` and `this.typedDurableClientType` are two different variables one `this.` apart.

The intent is sound — commit to the fields only if all five lookups on this assembly succeeded, which preserves the all-or-nothing "an un-inspectable assembly contributes nothing" semantics that `DefinedTypes` gave you for free (it was a single call, so partial success was not representable). Worth keeping, but please make it readable:

```csharp
// Commit only if the entire assembly could be inspected, matching the
// previous all-or-nothing behavior of Assembly.DefinedTypes.
Type contextType  = this.typedDurableOrchestrationContextType;
Type clientType   = this.typedDurableClientType;
Type callerType   = this.typedDurableOrchestrationCallerType;
Type activityType = this.typedDurableActivityCallerType;
Type starterType  = this.typedDurableOrchestrationStarterType;
```

One tradeoff to make deliberate: a mid-assembly failure now discards types already resolved from that assembly, which can only turn a would-be-successful initialization into a failed one, never the reverse. Very unlikely in practice (the generator emits all five together), and moot if the `catch` below is restored — just flagging it.

src/WebJobs.Extensions.DurableTask/TypedCodeProvider.cs
```diff
@@ -42,26 +56,80 @@ internal class TypedCodeProvider
...
+        }
+
+        internal void Initialize(Assembly[] assemblies)
+        {
+            this.typedDurableOrchestrationContextType = null;
+            this.typedDurableClientType = null;
+            this.typedDurableOrchestrationCallerType = null;
+            this.typedDurableActivityCallerType = null;
+            this.typedDurableOrchestrationStarterType = null;
+            this.IsInitialized = false;
```
YunchuWang, line 69: Nit: this reset block exists only so tests can call `Initialize` repeatedly — production calls it exactly once, from the `DurableTaskExtension` constructor. It is also the one piece of new logic with no test coverage. Either add a case asserting that a second `Initialize` clears the results of the first, or drop it.

It also briefly flips `IsInitialized` from `true` to `false`, which the old code never did. Harmless at today's single-threaded call site, but a new (if theoretical) window.

Separately: `Initialize(Assembly[] assemblies)` will `NullReferenceException` on a null argument. An `ArgumentNullException` guard would be clearer for an internal seam that tests call directly.

test/FunctionsV2/Tests/Unit/TypedCodeProviderTests.cs
```diff
@@ -0,0 +1,289 @@
...
+                .Throws(new InvalidOperationException("DefinedTypes must not be enumerated."));
+            var provider = new TypedCodeProvider();
+
+            provider.Initialize(new[] { assembly.Object });
+
+            Assert.False(provider.IsInitialized);
+            assembly.VerifyGet(a => a.DefinedTypes, Times.Never);
+            assembly.Verify(
+                a => a.GetType(It.IsAny<string>(), false, false),
+                Times.Exactly(RequiredTypeNames.Length));
```
YunchuWang, line 130: Nit: `Times.Exactly(RequiredTypeNames.Length)` pins the implementation to "exactly five `GetType` calls per assembly." The invariant this test actually protects is the `VerifyGet(a => a.DefinedTypes, Times.Never)` on the line above, which is already asserted.

Relaxing this to `Times.AtLeastOnce()` (or dropping it) would keep the test from needing edits when a sixth generated type is added, or when the lookups are reordered or short-circuited.

YunchuWang (commented):
Nice targeted fix — replacing whole-assembly `DefinedTypes` materialization with five `Assembly.GetType` lookups is the right shape for this problem, and adding the first unit tests for `TypedCodeProvider` is a welcome bonus.

One blocking concern plus a few nits, details inline.

**Blocking:** the narrowed exception filter is a behavior change riding along in a perf PR, and it can now fail host startup. See the comment on `TypedCodeProvider.cs`.

**Non-blocking:**
- Shadowed locals in the scan loop hurt readability (the "commit only on success" intent is reasonable, the naming is what makes it hard to read).
- The reset block at the top of `Initialize(Assembly[])` is untested and exists only for test re-entrancy.
- `Initialize(Assembly[])` has no null guard.
- `Times.Exactly(RequiredTypeNames.Length)` over-asserts on call count.
- The five `*FullName` constants are re-concatenated in the test file. Making them `internal const` on `TypedCodeProvider` (already an `internal` class) would remove the duplicated concatenation logic.
- Consider a `release_notes.md` entry under `Microsoft.Azure.WebJobs.Extensions.DurableTask` — a cold-start improvement seems worth calling out.

FYI, not for this PR: `Assembly.GetType` resolves type forwarders while `DefinedTypes` does not. Not reachable for these five generated types, but a real semantic difference worth recording somewhere.

berndverst:
## Component microbenchmark: baseline vs. this PR

I ran a scratch-only benchmark against merge base `c07ebda7f91d90cc3baf4916fca76191a841b4fe` and current PR head `9ff788c51f85b86b11ba323257d5c74d6620cf64`. The harness copies the two discovery loops exactly; no benchmark code or artifacts are added to the PR.

### Results

Times are **median single-threaded wall time per discovery scan**, with the 25th-75th percentile range in parentheses.

| Case | Assembly/type graph and assumption | Baseline time | PR time | Throughput, baseline -> PR | PR relative speed | Allocated bytes, baseline -> PR |
|---|---|---:|---:|---:|---:|---:|
| **Average-case proxy** | Built FunctionsV2 test-host graph: 108 loaded assemblies, 19,928 inspectable types; typed implementations absent. This represents the common non-typed app path, not a statistical average of all Functions apps. | 15.344 ms (15.076-15.758) | 0.2247 ms (0.2243-0.2255) | 65.2 -> 4,451 scans/s | **68.3x faster** | 1,030,012 -> 159,864 B (**84.5% less**, 870,148 B saved) |
| **Best realistic bound** | Same host graph plus one normally loaded app `RuntimeAssembly` containing all five required types at the start of the scan. This exercises the PR's early exit. | 13.033 ms (12.634-13.245) | 1.865 us (1.846-1.886) | 76.7 -> 536,106 scans/s | **6,987x faster** | 649,861 -> 2,368 B (**99.6% less**) |
| **Worst case tested / adversarial** | 84 physical, type-light `RuntimeAssembly` instances with only 89 total types; the five required types are split one-per-assembly across the final five. This deliberately maximizes missing-name lookups while minimizing enumeration work. | 13.544 us (13.534-13.577) | 166.528 us (166.093-167.372) | 73,835 -> 6,005 scans/s | **12.3x slower** | 4,550 -> 121,416 B (**116,866 B more; +2,568%**) |

The crossover is real: `Assembly.GetType` missing-name lookups are not allocation-free on this runtime. A graph with roughly one type per assembly can therefore favor enumeration. The tested worst case is intentionally unlike the host proxy (89 vs. 19,928 types at a similar assembly count), and its absolute regression is about **0.153 ms and 114 KiB once per extension construction**. It is not a formal upper bound: more type-empty assemblies could extend that lookup cost.

### Methodology / reproduction

- Environment: Windows 10.0.26200 x64, .NET 10.0.10, AMD EPYC 7763, 32 logical processors available to the process, Workstation GC, Release build, no debugger.
- Host graph:

  ```powershell
  dotnet build test\FunctionsV2\WebJobs.Extensions.DurableTask.Tests.V2.csproj `
    --configuration Release --framework net10.0
  ```

  The harness loads every managed DLL from `test\FunctionsV2\bin\Release\net10.0` into the default `AssemblyLoadContext`, then captures `AppDomain.CurrentDomain.GetAssemblies()`. Fourteen of the 108 assemblies throw while enumerating `DefinedTypes`; the baseline skips them exactly as production does.
- Best/worst fixtures are physical DLLs emitted with `PersistedAssemblyBuilder` and reloaded as ordinary `RuntimeAssembly` objects, rather than benchmarking `AssemblyBuilder` behavior.
- Each operation clones the captured `Assembly[]` to model the common `AppDomain.GetAssemblies()` snapshot allocation, then runs either the merge-base `DefinedTypes`/`ToList()` plus five `FirstOrDefault` scans or the PR's unresolved-name `GetType(fullName, false, false)` loop and early exit.
- Per scenario and algorithm: 1 second warmup; iteration count independently calibrated to at least 250 ms/sample; 21 interleaved samples; forced GC before each sample. Allocations use `GC.GetAllocatedBytesForCurrentThread()`. Reported ranges are interquartile ranges, not confidence intervals.

### Cold-start interpretation

This measures only `TypedCodeProvider.Initialize()`, after JIT/reflection metadata warmup. It is **not an end-to-end Functions cold-start benchmark**. On the representative absent-types graph, the component delta is about **15.1 ms and 0.87 MB per extension construction**; on the early-hit graph it is about **13.0 ms and 0.65 MB**. Those are component-level measurements, not claims that total app cold start improves by the same amount or percentage. A process-level Functions benchmark would still be needed to quantify whole-app impact, and true first-call costs may differ from these deliberately warmed medians.

PR #3493, opened 2026-07-27
Gate suppressed replay traces before sanitization

### LINKED ISSUE #3489: [Performance] Gate suppressed replay traces before payload and exception sanitization

Replay tracing is disabled unless `TraceReplayEvents` is enabled:

https://github.com/Azure/azure-functions-durable-extension/blob/4a52e5a802660bdb8fed0141d75b9d1e89463d2f/src/WebJobs.Extensions.DurableTask/Options/TraceOptions.cs#L45-L52

`EndToEndTraceHelper.ShouldLogEvent()` enforces that setting, but multiple replay-aware methods sanitize/format payloads before checking it. `SanitizeString()` always constructs a redaction string, and `SanitizeException()` eagerly calls `Exception.ToString()` and constructs stack-trace text:

https://github.com/Azure/azure-functions-durable-extension/blob/4a52e5a802660bdb8fed0141d75b9d1e89463d2f/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs#L62-L108

https://github.com/Azure/azure-functions-durable-extension/blob/4a52e5a802660bdb8fed0141d75b9d1e89463d2f/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs#L1176-L1179

Examples where formatting happens before the replay gate include function start/completion/failure:

https://github.com/Azure/azure-functions-durable-extension/blob/4a52e5a802660bdb8fed0141d75b9d1e89463d2f/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs#L214-L243

https://github.com/Azure/azure-functions-durable-extension/blob/4a52e5a802660bdb8fed0141d75b9d1e89463d2f/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs#L301-L325

https://github.com/Azure/azure-functions-durable-extension/blob/4a52e5a802660bdb8fed0141d75b9d1e89463d2f/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs#L443-L466

Entity operation completion/failure can sanitize two values and stringify exceptions before the eventual gate:

https://github.com/Azure/azure-functions-durable-extension/blob/4a52e5a802660bdb8fed0141d75b9d1e89463d2f/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs#L511-L562

https://github.com/Azure/azure-functions-durable-extension/blob/4a52e5a802660bdb8fed0141d75b9d1e89463d2f/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs#L565-L594

External events and entity responses have the same ordering:

https://github.com/Azure/azure-functions-durable-extension/blob/4a52e5a802660bdb8fed0141d75b9d1e89463d2f/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs#L618-L640

https://github.com/Azure/azure-functions-durable-extension/blob/4a52e5a802660bdb8fed0141d75b9d1e89463d2f/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs#L736-L759

### Performance impact

Replay is a normal part of orchestration execution, and the number of replayed actions grows with history. When replay tracing is disabled, none of these formatted values are emitted, but each redaction string still contributes short-lived allocation; failure paths additionally materialize potentially large exception strings.

A directional .NET 10.0.10 microbenchmark of only the current redaction interpolation (`"(Redacted {length} characters)"`) measured 80 B/call and 56.7 ms for 500,000 calls (about 40 MB total). Some trace methods sanitize two fields, and exception formatting is substantially more expensive. This is not an end-to-end replay benchmark, but it isolates work that can be skipped completely because the subsequent gate discards it.

I did not find an exact open issue for eager sanitization of suppressed replay traces. Existing logging work is broader; this proposal preserves logging semantics and only removes work on a path that currently emits nothing.

Move the replay eligibility check to the beginning of each replay-aware public trace entry point, before any sanitization, `ToString()`, enum conversion, or other formatting:

```csharp
if (!this.ShouldLogEvent(isReplay))
{
    return;
}

this.SanitizeString(...);
// Existing ETW and ILogger emission remains unchanged.
```

For overload chains such as `FunctionFailed` and `OperationFailed`, gate in the earliest overload that receives the original payload/exception so an ignored exception is never stringified. Keep a defensive gate in a shared private emitter only if it can be done without duplicated checks on the enabled path.

Audit all `SanitizeString`/`SanitizeException` calls in `EndToEndTraceHelper`, not only the examples above, and apply the change only where `isReplay` currently controls both ETW and `ILogger` emission.

- When `TraceReplayEvents == true`, preserve every ETW event, `ILogger` event, payload redaction/raw-data choice, log level, template, property, and sequence number.
- For non-replay events, preserve output exactly.
- When replay tracing is disabled, continue emitting no replay event; the only change should be avoiding discarded work.
- Preserve the current `TraceInputsAndOutputs` privacy behavior on every emitted event.
- Do not remove trace calls from orchestration/entity call sites or change host.json settings.
- Ensure early return does not alter `sequenceNumber`; today it increments only inside the successful logging blocks.

- **Make sanitization lazy via delegates.** This centralizes gating but adds delegates/closures and complexity to a very hot path. An early branch is cheaper and clearer.
- **Check only `ILogger.IsEnabled`.** ETW emission is separate, and replay suppression is independent of configured log level; this does not address the known no-output condition.
- **Remove replay trace calls at their call sites.** That risks missing paths and could break the opt-in `TraceReplayEvents` feature.
- **Cache redaction strings by payload length.** Length cardinality and retention make this inferior to not formatting suppressed events at all; it also does not solve exception stringification.

**Additional context**

- Extend `EndToEndTraceHelperTests` to cover each affected method with `(isReplay, TraceReplayEvents)` combinations and assert emitted messages remain identical on enabled paths:

https://github.com/Azure/azure-functions-durable-extension/blob/4a52e5a802660bdb8fed0141d75b9d1e89463d2f/test/FunctionsV2/Tests/Unit/EndToEndTraceHelperTests.cs#L15-L108

- Use a test exception that overrides `ToString()` to prove it is not evaluated when replay logging is suppressed and is evaluated when enabled.
- Add an allocation benchmark covering function start/completion, entity operation completion, and exception failure paths.
- Acceptance target: no payload-redaction allocation and no exception stringification on suppressed replay paths, with byte-for-byte/log-property-equivalent enabled telemetry.
- Run an orchestration replay regression test with replay tracing both disabled and enabled to verify no behavioral or observability regression.

berndverst:
## Summary

- Gate replay-aware `EndToEndTraceHelper` entry points before payload redaction, exception stringification, and downstream enum formatting.
- Cover function start/completion/failure, entity operation completion/failure, external events, and entity responses across the full `(isReplay, TraceReplayEvents)` matrix and both raw/redacted ILogger modes.
- Keep shared failure emitters gate-free after public overloads validate eligibility, so enabled paths perform only one replay check.

## Allocation rationale

When `isReplay` is true and `TraceReplayEvents` is false, these events are discarded. The early gates now avoid constructing one or two `"(Redacted N characters)"` strings per suppressed event and avoid `Exception.ToString()` plus sanitized stack text on suppressed failures. This removes work entirely instead of introducing delegates, closures, or caching on hot replay paths.

## Observability compatibility

For replay events when tracing is enabled, and for every non-replay event, the ETW and ILogger emitters are unchanged: event ordering remains ETW before ILogger; log levels, templates, structured property names and values, raw-vs-redacted payload selection, replay flags, task event IDs, durations, exception arguments, and sequence-number increments are preserved. Suppressed replay events still emit neither ETW nor ILogger telemetry and still do not increment the sequence number.

## Tests

- `dotnet test test\FunctionsV2\WebJobs.Extensions.DurableTask.Tests.V2.csproj --no-restore --filter "FullyQualifiedName~EndToEndTraceHelperTests" --logger "console;verbosity=minimal"`
  - Passed: 93/93 on `net8.0`; 93/93 on `net10.0`.
- `dotnet test test\FunctionsV2\WebJobs.Extensions.DurableTask.Tests.V2.csproj -f net8.0 --no-restore --filter "FullyQualifiedName~OutOfProcTests.CallHttpActionOrchestration|FullyQualifiedName~OutOfProcTests.LockEntitiesAction" --logger "console;verbosity=minimal"`
  - Passed: 10/10 on `net8.0`.
- `dotnet test test\FunctionsV2\WebJobs.Extensions.DurableTask.Tests.V2.csproj -f net8.0 --no-restore --filter "FullyQualifiedName~HelloWorldOrchestration_ValidateReplayEventLogs|FullyQualifiedName~DurableEntity_SignalAndCallStringStore|FullyQualifiedName~DurableEntity_CallFaultyEntity" --logger "console;verbosity=minimal"`
  - Environment-blocked before test bodies: 12/12 cases failed in the shared fixture with `UnauthorizedAccessException: Error Starting ETW: Access Denied (Administrator rights required to start ETW)`.

Fixes #3489

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/EndToEndTraceHelper.cs
@@ -220,10 +220,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             bool isReplay,
             int taskEventId = -1)
         {
-            this.SanitizeString(input, out string loggerInput, out string sanitizedInput);
-
             if (this.ShouldLogEvent(isReplay))
             {
+                this.SanitizeString(input, out string loggerInput, out string sanitizedInput);
+
                 EtwEventSource.Instance.FunctionStarting(
                     hubName,
                     LocalAppName,
@@ -308,10 +308,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             bool isReplay,
             int taskEventId = -1)
         {
-            this.SanitizeString(output, out string loggerOutput, out string sanitizedOutput);
-
             if (this.ShouldLogEvent(isReplay))
             {
+                this.SanitizeString(output, out string loggerOutput, out string sanitizedOutput);
+
                 EtwEventSource.Instance.FunctionCompleted(
                     hubName,
                     LocalAppName,
@@ -449,8 +449,13 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             bool isReplay,
             int taskEventId = -1)
         {
+            if (!this.ShouldLogEvent(isReplay))
+            {
+                return;
+            }
+
             this.SanitizeException(exception, out string loggerReason, out string sanitizedReason);
-            this.FunctionFailed(hubName, functionName, instanceId, loggerReason, sanitizedReason, functionType, isReplay, taskEventId);
+            this.WriteFunctionFailed(hubName, functionName, instanceId, loggerReason, sanitizedReason, functionType, isReplay, taskEventId);
         }
 
         public void FunctionFailed(
@@ -463,25 +468,40 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             bool isReplay,
             int taskEventId = -1)
         {
-            if (this.ShouldLogEvent(isReplay))
+            if (!this.ShouldLogEvent(isReplay))
             {
-                EtwEventSource.Instance.FunctionFailed(
-                    hubName,
-                    LocalAppName,
-                    LocalSlotName,
-                    functionName,
-                    taskEventId,
-                    instanceId,
-                    sanitizedReason,
-                    functionType.ToString(),
-                    ExtensionVersion,
-                    isReplay);
-
-                this.logger.LogError(
-                    "{instanceId}: Function '{functionName} ({functionType})' failed with an error. Reason: {reason}. IsReplay: {isReplay}. State: {state}. RuntimeStatus: {runtimeStatus}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}. TaskEventId: {taskEventId}",
-                    instanceId, functionName, functionType, reason, isReplay, FunctionState.Failed, OrchestrationRuntimeStatus.Failed, hubName,
-                    LocalAppName, LocalSlotName, ExtensionVersion, this.sequenceNumber++, taskEventId);
+                return;
             }
+
+            this.WriteFunctionFailed(hubName, functionName, instanceId, reason, sanitizedReason, functionType, isReplay, taskEventId);
+        }
+
+        private void WriteFunctionFailed(
+            string hubName,
+            string functionName,
+            string instanceId,
+            string reason,
+            string sanitizedReason,
+            FunctionType functionType,
+            bool isReplay,
+            int taskEventId)
+        {
+            EtwEventSource.Instance.FunctionFailed(
+                hubName,
+                LocalAppName,
+                LocalSlotName,
+                functionName,
+                taskEventId,
+                instanceId,
+                sanitizedReason,
+                functionType.ToString(),
+                ExtensionVersion,
+                isReplay);
+
+            this.logger.LogError(
+                "{instanceId}: Function '{functionName} ({functionType})' failed with an error. Reason: {reason}. IsReplay: {isReplay}. State: {state}. RuntimeStatus: {runtimeStatus}. HubName: {hubName}. AppName: {appName}. SlotName: {slotName}. ExtensionVersion: {extensionVersion}. SequenceNumber: {sequenceNumber}. TaskEventId: {taskEventId}",
+                instanceId, functionName, functionType, reason, isReplay, FunctionState.Failed, OrchestrationRuntimeStatus.Failed, hubName,
+                LocalAppName, LocalSlotName, ExtensionVersion, this.sequenceNumber++, taskEventId);
         }
 
         public void FunctionAborted(
@@ -519,11 +539,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
            double duration,
            bool isReplay)
         {
-            this.SanitizeString(input, out string loggerInput, out string sanitizedInput);
-            this.SanitizeString(output, out string loggerOutput, out string sanitizedOutput);
-
             if (this.ShouldLogEvent(isReplay))
             {
+                this.SanitizeString(input, out string loggerInput, out string sanitizedInput);
+                this.SanitizeString(output, out string loggerOutput, out string sanitizedOutput);
+
                 EtwEventSource.Instance.OperationCompleted(
                     hubName,
                     LocalAppName,
@@ -557,6 +577,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
            double duration,
            bool isReplay)
         {
+            if (!this.ShouldLogEvent(isReplay))
+            {
+                return;
+            }
+
             this.SanitizeString(input, out string loggerInput, out string sanitizedInput);
             this.SanitizeException(exception, out string loggerException, out string sanitizedException);
             this.OperationFailed(hubName, functionName, instanceId, operationId, operationName, sanitizedInput, loggerInput, sanitizedException, loggerException, duration, isReplay);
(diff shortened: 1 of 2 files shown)


berndverst:
### Replay-trace microbenchmark: merge-base vs. PR head

I measured the real `EndToEndTraceHelper` assemblies built from merge-base `4a52e5a8` and PR head `290109ed` (source blob hashes verified). This is deliberately a **trace-helper microbenchmark**, not an end-to-end orchestration-latency or orchestration-throughput claim.

All values below are medians; brackets are the retained sample min-max. `~0` means below `0.00001 B/event` after fixed harness overhead.

| Scenario | Wall ns/event: base -> PR | Process CPU ns/event: base -> PR | B/event: base -> PR | M events/s: base -> PR | Throughput |
|---|---:|---:|---:|---:|---:|
| **Average: suppressed replay mix** | 59.89 [58.31-60.73] -> 4.72 [4.70-4.74] | 59.67 [53.71-63.19] -> 4.67 [4.41-4.91] | 112.72 -> ~0 | 16.70 -> 212.08 | **12.70x** |
| One 256-char payload, suppressed | 32.83 [31.83-33.91] -> 0.519 [0.518-0.540] | 33.01 [32.08-38.08] -> 0.469 [0.469-0.781] | 72.00 -> ~0 | 30.46 -> 1,924.95 | 63.21x |
| Two 256-char payloads, suppressed | 69.87 [67.49-71.78] -> 0.632 [0.629-0.694] | 70.46 [66.75-75.32] -> 0.625 [0.625-0.938] | 144.00 -> ~0 | 14.31 -> 1,582.52 | 110.58x |
| Input + 1,024-char serialized failure, suppressed | 65.88 [64.29-67.71] -> 0.634 [0.629-0.653] | 65.86 [62.39-72.79] -> 0.625 [0.625-1.250] | 152.00 -> ~0 | 15.18 -> 1,577.28 | 103.91x |
| Realistic nested `FunctionFailed(Exception)`, suppressed | 9,752 [9,575-10,014] -> 3.78 [3.77-3.81] | 9,980 [7,795-11,135] -> 3.75 [3.55-4.14] | 18,800.80 -> ~0 | 0.103 -> 264.32 | 2,577.81x |
| **Best: realistic nested entity `OperationFailed(Exception)`, suppressed** | 10,197 [9,937-10,390] -> 4.10 [4.09-4.14] | 10,268 [9,241-10,293] -> 4.12 [3.86-5.01] | 18,872.78 -> ~0 | 0.098 -> 243.61 | **2,484.12x** |
| Replay tracing enabled (same mix) | 257.63 [244.22-379.96] -> 252.98 [242.05-398.68] | 257.30 [224.95-496.93] -> 262.71 [234.27-528.78] | 555.97 -> 555.98 | 3.882 -> 3.953 | 1.018x |
| **Worst measured control: non-replay (same mix)** | 257.39 [248.79-274.56] -> 267.73 [254.96-330.39] | 256.36 [233.91-281.12] -> 265.94 [240.61-331.19] | 660.67 -> 660.67 | 3.885 -> 3.735 | **0.961x** |

#### Average / best / worst translations

- **Average (default `TraceReplayEvents=false`)**: I used an explicit illustrative replay mix, not fleet telemetry: 70% one-payload function traces, 25% two-payload entity completions, 4.9% serialized failures, and only 0.1% exception failures. The PR saved **55.17 ns wall, 55.00 ns process CPU, and 112.72 B per helper call**. At 100/1,000 suppressed calls, that is about **5.52/55.17 us wall**, **5.50/55.00 us CPU**, and **11.0/110.1 KiB** avoided.
- **Best (failure-heavy replay)**: the realistic nested entity exception path saved **10.19 us wall, 10.26 us CPU, and 18.87 KB per call** by avoiding `Exception.ToString()`, sanitized stack text, and input redaction. At 100/1,000 such calls, that is about **1.02/10.19 ms wall** and **1.80/18.00 MiB** avoided. Deeper exception chains or longer rendered stacks can save more; simpler exceptions save less.
- **Worst (emitting paths)**: no eligibility branch was added; it moved earlier, and eligible telemetry does the same work. The worst retained control median was the non-replay mix at **+10.34 ns wall (+4.0%), +9.58 ns CPU (+3.7%), -3.9% throughput, and no allocation increase**. That is about **+1.03 us per 100** or **+10.34 us per 1,000** calls. Base/head ranges overlap, alternating-process results moved in both directions, and the enabled-replay wall median was actually 1.8% faster, so I treat this small control delta as measurement/code-layout noise rather than a demonstrated regression.

#### Methodology and caveats

- Windows NT `10.0.26200`, x64 .NET runtime `10.0.10` / SDK `10.0.302`, Release build, AMD EPYC 7763 VM (16 cores / 32 logical processors), 128 GiB RAM.
- Scratch-only harness (not committed) directly project-referenced each real assembly; no reflection in the timed loop. Payloads were 256 chars; serialized failure text was 1,024 chars; the exception was thrown/caught with nested inner/outer stack frames.
- `NullLogger` and no active ETW consumer isolate helper-side sanitization/formatting and log-call construction. A real logger/ETW sink adds downstream cost equally to eligible base/head paths.
- Per process: adaptive calibration toward 300 ms/batch (100M-event cap), two warmup batches, then 15 samples. Versions ran in alternating order. The first 3 samples/process were discarded, leaving **24 samples/version/scenario** for suppressed paths and **48** for emitting controls. Wall time used `Stopwatch`; CPU used `Process.TotalProcessorTime`; allocation used `GC.GetAllocatedBytesForCurrentThread`.
- The very large suppressed-path throughput ratios mean the PR path became a branch-only no-op in this tight loop. They must **not** be interpreted as equivalent end-to-end orchestration throughput gains. Real savings scale with the number/type of replay trace-helper calls and are only one component of replay execution.

PR #3494, opened 2026-07-27
Make orchestration history projection linear

### LINKED ISSUE #3490: [Performance] Make orchestration history projection linear for large histories

When `showHistory=true`, `GetDurableOrchestrationStatusAsync()` parses the complete persisted history into a mutable `JArray`, walks and mutates every event, tracks source indices that should be hidden, sorts those indices, and then calls `JArray.RemoveAt()` once per hidden event:

https://github.com/Azure/azure-functions-durable-extension/blob/4a52e5a802660bdb8fed0141d75b9d1e89463d2f/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs#L934-L1037

Matched `TaskScheduled` and `SubOrchestrationInstanceCreated` records are added to the removal list after their data has been aggregated into completion/failure events:

https://github.com/Azure/azure-functions-durable-extension/blob/4a52e5a802660bdb8fed0141d75b9d1e89463d2f/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs#L1055-L1073

`OrchestratorStarted` and `OrchestratorCompleted` records are also removed:

https://github.com/Azure/azure-functions-durable-extension/blob/4a52e5a802660bdb8fed0141d75b9d1e89463d2f/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs#L1008-L1033

The management HTTP API exposes this projection through the `showHistory` query parameter:

https://github.com/Azure/azure-functions-durable-extension/blob/4a52e5a802660bdb8fed0141d75b9d1e89463d2f/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs#L645-L676

### Performance impact

`JArray` is list-backed, so removing an item shifts later elements. Repeating adjusted ascending `RemoveAt()` calls makes the filtering phase O(n × k), approaching O(n²) when the number of removed events grows with history size. Durable histories naturally contain many start/completion replay markers and matched schedule/completion pairs, so this is not a pathological event mix.

A directional list-compaction proxy with 50,000 events and 10,000 distributed removals measured 16.7 ms for the current adjusted-ascending `RemoveAt` shape versus 0.9 ms for a one-pass filter (about 19x). The proxy excludes JSON parsing, token mutation, storage, and response serialization; it isolates only the avoidable removal complexity. At larger histories the gap grows superlinearly, increasing management-API tail latency and CPU precisely for long-running instances where history inspection is most useful.

I did not find an exact open issue for the history projection/removal algorithm. This proposal is limited to the `showHistory` response projection and does not alter persisted history, replay, purge behavior, or storage-provider contracts.

Replace repeated mutation/removal with an O(n) projection while retaining the existing first-pass aggregation semantics. One low-risk shape is:

1. Keep the current event walk and transformations, but record removals in a `HashSet<int>` or compact boolean marker array rather than a sorted `List<int>`.
2. Preserve the current rule that a scheduled/created event is removed only when a matching completion/failure event was found; unmatched/in-flight scheduled events must remain.
3. After the transform pass, collect retained `JToken` references in source order.
4. Clear/detach the original `JArray` once and append retained tokens once, or return a new projected `JArray` without repeatedly shifting the old one.
5. Skip `OrchestratorStarted`/`OrchestratorCompleted` with the same semantics as today.

An alternative is a two-pass output builder that delays emission decisions until all matched scheduled-event indices are known. Either implementation should be linear, preserve order, and avoid deep-cloning every token if tokens can be safely detached and reused.

The projected JSON is a public management/API response and must remain structurally identical:

- Preserve event order, event-type string conversion, renamed/removed properties, timestamps, and aggregation behavior.
- Preserve `showInput` and `showHistoryOutput` behavior for successful and failed tasks/sub-orchestrations.
- Preserve unmatched/in-flight `TaskScheduled` and `SubOrchestrationInstanceCreated` events.
- Preserve removal of replay-only orchestrator start/completion markers.
- Preserve malformed/unknown event handling and current exception behavior.
- Do not change persisted history, provider APIs, HTTP query parameters, or response models.

- **Remove indices in descending order.** This is a small improvement and avoids counter adjustment, but repeated list shifts remain superlinear for distributed/low indices.
- **Return raw history.** This would be faster but is a breaking response-shape change and loses the existing aggregation/readability contract.
- **Paginate history.** Potentially valuable for very large instances, but it changes the public API and provider interaction. Linear projection is an internal, compatible improvement that should be done first.
- **Use a `LinkedList<JToken>`.** Removal becomes cheap, but conversion and node overhead add complexity; a one-pass array/list compaction is simpler and cache-friendlier.

**Additional context**

- Add golden/snapshot tests comparing old and new projected JSON for completed, failed, and in-flight activities and sub-orchestrations.
- Cover multiple replay segments, duplicate-looking names with distinct scheduled IDs, unknown event types, and all `showInput`/`showHistoryOutput` combinations.
- Include histories with zero removals, all-removable markers, and removals at the beginning/middle/end.
- Add scaling benchmarks (for example 1K, 10K, 50K, and 100K events) that report elapsed time and allocated bytes for projection after history parsing.
- Acceptance target: output-equivalent JSON with projection/filter cost that scales approximately linearly and no repeated `JArray.RemoveAt()` loop.
- Run existing end-to-end history tests to guard the public shape, including activity and sub-orchestration history scenarios:

https://github.com/Azure/azure-functions-durable-extension/blob/4a52e5a802660bdb8fed0141d75b9d1e89463d2f/test/FunctionsV2/Tests/E2E/HelloWorldActivityTests.cs#L346-L395

https://github.com/Azure/azure-functions-durable-extension/blob/4a52e5a802660bdb8fed0141d75b9d1e89463d2f/test/FunctionsV2/Tests/E2E/ExceptionHandlingTests.cs#L598-L666

berndverst:
## Summary

- replace sorted, repeated `JArray.RemoveAt` calls with stable linear history compaction
- retain the existing aggregation pass and matching rules for activities and sub-orchestrations
- add golden public-path coverage for completed, failed, and in-flight work, replay segments, edge removals, unknown/malformed events, and all `showInput`/`showHistoryOutput` combinations

## Performance rationale

The previous filter sorted hidden source indexes and removed each event from a list-backed `JArray`, shifting later elements on every removal and approaching O(n^2) when removed events scale with history size. The new path records removal multiplicity by source index, collects retained token references in source order, detaches the original array once, and reattaches retained tokens once. Valid histories are projected in O(n) time with O(n) auxiliary references and without deep-cloning retained `JToken`s. The compatibility path for malformed duplicate completion links also uses linear cursor movement rather than list shifts.

## API compatibility

The public projected JSON remains unchanged: source/event ordering, event-type strings, property renames/removals, scheduled/completion aggregation, timestamps, successful and failed result handling, and all `showInput`/`showHistoryOutput` behavior are covered by exact JSON assertions. Matched scheduled/created events are hidden only after a matching completion/failure; in-flight events remain. `OrchestratorStarted`/`OrchestratorCompleted` markers remain hidden. Unknown events and malformed-history exception behavior are preserved. Persisted history is not mutated; retained tokens are detached and reparented to the response array.

## Tests

- `dotnet test test\FunctionsV2\WebJobs.Extensions.DurableTask.Tests.V2.csproj --configuration Release --framework net8.0 --filter 'FullyQualifiedName~DurableClientHistoryTests|FullyQualifiedName~DurableClientBaseTests' --no-build --no-restore --logger 'console;verbosity=minimal'`
  - Passed: 45, Failed: 0, Skipped: 0
- `dotnet build test\FunctionsV2\WebJobs.Extensions.DurableTask.Tests.V2.csproj --configuration Release --verbosity minimal`
  - Succeeded for net8.0 and net10.0 with 0 warnings and 0 errors
- Linux container E2E run:
  - `docker run -d --rm --name durable-history-azurite mcr.microsoft.com/azure-storage/azurite:latest azurite --silent --skipApiVersionCheck --blobHost 0.0.0.0 --queueHost 0.0.0.0 --tableHost 0.0.0.0`
  - `docker run --rm --network container:durable-history-azurite --env 'AzureWebJobsStorage=UseDevelopmentStorage=true' --mount "type=bind,source=$PWD,target=/src" --workdir /src mcr.microsoft.com/dotnet/sdk:10.0 dotnet test test/FunctionsV2/WebJobs.Extensions.DurableTask.Tests.V2.csproj --framework net10.0 --filter 'FullyQualifiedName~HelloWorldOrchestration_Activity_History|FullyQualifiedName~HelloWorldOrchestration_ShowHistoryInputOutput|FullyQualifiedName~SubOrchestration_ComplexType_History' --logger 'console;verbosity=minimal'`
  - Passed: 6, Failed: 0, Skipped: 0

The history E2E tests were run in Linux because the existing Windows test logger requires an elevated process to create its ETW session.

Fixes #3490

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -942,7 +942,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     historyArray = MessagePayloadDataConverter.ConvertToJArray(history);
 
                     var eventMapper = new Dictionary<string, EventIndexDateMapping>();
-                    var indexList = new List<int>();
+                    var eventsToRemove = new int[historyArray.Count];
 
                     for (var i = 0; i < historyArray.Count; i++)
                     {
@@ -959,7 +959,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                                     break;
                                 case EventType.TaskCompleted:
                                 case EventType.TaskFailed:
-                                    AddScheduledEventDataAndAggregate(ref eventMapper, "TaskScheduled", historyItem, indexList, showInput);
+                                    AddScheduledEventDataAndAggregate(ref eventMapper, "TaskScheduled", historyItem, eventsToRemove, showInput);
                                     historyItem["TaskScheduledId"]?.Parent.Remove();
                                     if (!showHistoryOutput && eventType == EventType.TaskCompleted)
                                     {
@@ -974,7 +974,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                                     break;
                                 case EventType.SubOrchestrationInstanceCompleted:
                                 case EventType.SubOrchestrationInstanceFailed:
-                                    AddScheduledEventDataAndAggregate(ref eventMapper, "SubOrchestrationInstanceCreated", historyItem, indexList, showInput);
+                                    AddScheduledEventDataAndAggregate(ref eventMapper, "SubOrchestrationInstanceCreated", historyItem, eventsToRemove, showInput);
                                     historyItem.Remove("TaskScheduledId");
                                     if (!showHistoryOutput && eventType == EventType.SubOrchestrationInstanceCompleted)
                                     {
@@ -1010,7 +1010,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                                     break;
                                 case EventType.OrchestratorStarted:
                                 case EventType.OrchestratorCompleted:
-                                    indexList.Add(i);
+                                    eventsToRemove[i]++;
                                     break;
                             }
 
@@ -1024,13 +1024,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                         }
                     }
 
-                    var counter = 0;
-                    indexList.Sort();
-                    foreach (var indexValue in indexList)
-                    {
-                        historyArray.RemoveAt(indexValue - counter);
-                        counter++;
-                    }
+                    CompactHistory(historyArray, eventsToRemove);
                 }
             }
 
@@ -1057,7 +1051,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             eventMapper.Add($"{eventType}_{historyItem["EventId"]}", new EventIndexDateMapping { Index = index, Name = (string)historyItem["Name"], Date = (DateTime)historyItem["Timestamp"], Input = (string)historyItem["Input"] });
         }
 
-        private static void AddScheduledEventDataAndAggregate(ref Dictionary<string, EventIndexDateMapping> eventMapper, string prefix, JToken historyItem, List<int> indexList, bool showInput)
+        private static void AddScheduledEventDataAndAggregate(ref Dictionary<string, EventIndexDateMapping> eventMapper, string prefix, JToken historyItem, int[] eventsToRemove, bool showInput)
         {
             if (eventMapper.TryGetValue($"{prefix}_{historyItem["TaskScheduledId"]}", out EventIndexDateMapping taskScheduledData))
             {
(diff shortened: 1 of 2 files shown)


cgillum (approved):
I'm good with this change assuming it's 100% correct. The caveat is that this API is called very infrequently so the real-world improvement may not be noticeable.

PR #3498, opened 2026-07-28
Log unexpected local gRPC failures

### LINKED ISSUE #3042: Add error logging for GetInstance (and other methods) in LocalGrpcListener
In out-of-process languages, for many methods in LocalGrpcListener.cs, if the durability provider throws an exception it is not caught by the WebJobs extension but is thrown back to the worker as a generic grpc exception type, which may or may not include details about what went wrong on the WebJobs side. 

None known

We should generate error logs using this.extension.TraceHelper.ExtensionWarningEvent() or similar to ensure that exception details in these scenarios are accessible. 

Debugging issues with the durability providers in out-of-process requires attaching a debugger to the WebJobs extension. 

**Additional context**
Customer exception from an issue with getInstanceMetadata() in java, note that only the java and host stack traces are preserved, but nothing from the extension. 

Microsoft.Azure.WebJobs.Host.FunctionInvocationException : Exception while executing function: Functions.HttpTriggerStartSAPPayments ---> Microsoft.Azure.WebJobs.Script.Workers.Rpc.RpcException : Result: Failure
Exception: StatusRuntimeException: UNKNOWN: Exception was thrown by handler.
Stack: java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.microsoft.azure.functions.worker.broker.JavaMethodInvokeInfo.invoke(JavaMethodInvokeInfo.java:22)
	at com.microsoft.azure.functions.worker.broker.JavaMethodExecutorImpl.execute(JavaMethodExecutorImpl.java:22)
	at com.microsoft.azure.functions.worker.chain.FunctionExecutionMiddleware.invoke(FunctionExecutionMiddleware.java:19)
	at com.microsoft.azure.functions.worker.chain.InvocationChain.doNext(InvocationChain.java:21)
	at com.microsoft.azure.functions.worker.broker.JavaFunctionBroker.invokeMethod(JavaFunctionBroker.java:125)
	at com.microsoft.azure.functions.worker.handler.InvocationRequestHandler.execute(InvocationRequestHandler.java:34)
	at com.microsoft.azure.functions.worker.handler.InvocationRequestHandler.execute(InvocationRequestHandler.java:10)
	at com.microsoft.azure.functions.worker.handler.MessageHandler.handle(MessageHandler.java:44)
	at com.microsoft.azure.functions.worker.JavaWorkerClient$StreamingMessagePeer.lambda$onNext$0(JavaWorkerClient.java:94)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:750)
Caused by: io.grpc.StatusRuntimeException: UNKNOWN: Exception was thrown by handler.
	at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:271)
	at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:252)
	at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:165)
	at com.microsoft.durabletask.implementation.protobuf.TaskHubSidecarServiceGrpc$TaskHubSidecarServiceBlockingStub.getInstance(TaskHubSidecarServiceGrpc.java:929)
	at com.microsoft.durabletask.DurableTaskGrpcClient.getInstanceMetadata(DurableTaskGrpcClient.java:140)
	at com.tmna.mar.ui.functions.SAPPaymentsFun.HttpTriggerStartSAPPayments(SAPPaymentsFun.java:65)
	... 18 more

   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.Azure.WebJobs.Script.Description.WorkerFunctionInvoker.InvokeCore(Object[] parameters,FunctionInvocationContext context) at /_/src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs : 101
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.Azure.WebJobs.Script.Description.FunctionInvokerBase.Invoke(Object[] parameters) at /_/src/WebJobs.Script/Description/FunctionInvokerBase.cs : 82
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.Azure.WebJobs.Script.Description.FunctionGenerator.Coerce[T](Task`1 src) at /_/src/WebJobs.Script/Description/FunctionGenerator.cs : 225
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.Azure.WebJobs.Host.Executors.FunctionInvoker`2.InvokeAsync[TReflected,TReturnValue](Object instance,Object[] arguments) at D:\a\_work\1\s\src\Microsoft.Azure.WebJobs.Host\Executors\FunctionInvoker.cs : 53
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.Azure.WebJobs.Host.Executors.FunctionExecutor.InvokeWithTimeoutAsync(IFunctionInvoker invoker,ParameterHelper parameterHelper,CancellationTokenSource timeoutTokenSource,CancellationTokenSource functionCancellationTokenSource,Boolean throwOnTimeout,TimeSpan timerInterval,IFunctionInstance instance) at D:\a\_work\1\s\src\Microsoft.Azure.WebJobs.Host\Executors\FunctionExecutor.cs : 581
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.Azure.WebJobs.Host.Executors.FunctionExecutor.ExecuteWithWatchersAsync(IFunctionInstanceEx instance,ParameterHelper parameterHelper,ILogger logger,CancellationTokenSource functionCancellationTokenSource) at D:\a\_work\1\s\src\Microsoft.Azure.WebJobs.Host\Executors\FunctionExecutor.cs : 523
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.Azure.WebJobs.Host.Executors.FunctionExecutor.ExecuteWithLoggingAsync(IFunctionInstanceEx instance,FunctionStartedMessage message,FunctionInstanceLogEntry instanceLogEntry,ParameterHelper parameterHelper,ILogger logger,CancellationToken cancellationToken) at D:\a\_work\1\s\src\Microsoft.Azure.WebJobs.Host\Executors\FunctionExecutor.cs : 306
   End of inner exception
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.Azure.WebJobs.Host.Executors.FunctionExecutor.ExecuteWithLoggingAsync(IFunctionInstanceEx instance,FunctionStartedMessage message,FunctionInstanceLogEntry instanceLogEntry,ParameterHelper parameterHelper,ILogger logger,CancellationToken cancellationToken) at D:\a\_work\1\s\src\Microsoft.Azure.WebJobs.Host\Executors\FunctionExecutor.cs : 352
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.Azure.WebJobs.Host.Executors.FunctionExecutor.TryExecuteAsync(IFunctionInstance functionInstance,CancellationToken cancellationToken) at D:\a\_work\1\s\src\Microsoft.Azure.WebJobs.Host\Executors\FunctionExecutor.cs : 108

YunchuWang:
## Summary
- add a centralized ASP.NET Core gRPC interceptor that records unexpected provider exceptions through `ExtensionWarningEvent`
- preserve existing client-facing gRPC status codes while retaining the original exception stack for host-side diagnostics
- suppress warnings for deliberate `RpcException` mappings and client-triggered cancellation
- await task-hub create/delete operations so asynchronous provider failures reach the exception boundary

## Validation
- `LocalGrpcListenerTests`: 30/30 passed on .NET 8
- `LocalGrpcListenerTests`: 30/30 passed on .NET 10
- Release build passed for both target frameworks with 0 warnings and 0 errors

Closes #3042

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Grpc/AspNetCoreLocalGrpcListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/Grpc/AspNetCoreLocalGrpcListener.cs
@@ -144,6 +144,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Grpc
                         services.AddSingleton(this.extension);
                         services.AddGrpc(options =>
                         {
+                            options.Interceptors.Add<TaskHubGrpcExceptionInterceptor>();
                             options.MaxReceiveMessageSize = int.MaxValue;
                             options.MaxSendMessageSize = int.MaxValue;
                         });
--- a/src/WebJobs.Extensions.DurableTask/Grpc/AspNetCoreLocalGrpcListener.cs
+++ b/src/WebJobs.Extensions.DurableTask/Grpc/AspNetCoreLocalGrpcListener.cs
@@ -142,6 +142,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Grpc
                     builder.ConfigureServices(services =>
                     {
                         services.AddSingleton(this.extension);
+
+                        // Register the interceptor in DI so DefaultGrpcInterceptorActivator resolves
+                        // this singleton instead of constructing a new instance for every RPC.
+                        services.AddSingleton<TaskHubGrpcExceptionInterceptor>();
                         services.AddGrpc(options =>
                         {
                             options.Interceptors.Add<TaskHubGrpcExceptionInterceptor>();
--- a/src/WebJobs.Extensions.DurableTask/Grpc/TaskHubGrpcExceptionInterceptor.cs
+++ b/src/WebJobs.Extensions.DurableTask/Grpc/TaskHubGrpcExceptionInterceptor.cs
@@ -1,0 +1,82 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#nullable enable
+using System;
+using System.Threading.Tasks;
+using Grpc.Core;
+using Grpc.Core.Interceptors;
+
+namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Grpc
+{
+    internal sealed class TaskHubGrpcExceptionInterceptor : Interceptor
+    {
+        private readonly DurableTaskExtension extension;
+
+        public TaskHubGrpcExceptionInterceptor(DurableTaskExtension extension)
+        {
+            this.extension = extension ?? throw new ArgumentNullException(nameof(extension));
+        }
+
+        public override async Task<TResponse> UnaryServerHandler<TRequest, TResponse>(
+            TRequest request,
+            ServerCallContext context,
+            UnaryServerMethod<TRequest, TResponse> continuation)
+        {
+            try
+            {
+                return await continuation(request, context);
+            }
+            catch (Exception exception) when (ShouldLog(exception, context))
+            {
+                this.LogException(context, exception);
+                throw;
+            }
+        }
+
+        public override async Task ServerStreamingServerHandler<TRequest, TResponse>(
+            TRequest request,
+            IServerStreamWriter<TResponse> responseStream,
+            ServerCallContext context,
+            ServerStreamingServerMethod<TRequest, TResponse> continuation)
+        {
+            try
+            {
+                await continuation(request, responseStream, context);
+            }
+            catch (Exception exception) when (ShouldLog(exception, context))
+            {
+                this.LogException(context, exception);
+                throw;
+            }
+        }
+
+        private static bool ShouldLog(Exception exception, ServerCallContext context)
+        {
+            if (context.CancellationToken.IsCancellationRequested)
+            {
+                return false;
+            }
+
+            if (exception is TaskHubRpcException)
+            {
+                return true;
+            }
+
+            return exception is not RpcException;
+        }
+
+        private void LogException(ServerCallContext context, Exception exception)
+        {
+            Exception exceptionToLog = exception is TaskHubRpcException rpcException
+                ? rpcException.Cause
+                : exception;
+
+            this.extension.TraceHelper.ExtensionWarningEvent(
+                this.extension.Options.HubName,
+                instanceId: string.Empty,
+                functionName: string.Empty,
+                message: $"Unhandled exception in local gRPC call '{context.Method}': {exceptionToLog}");
+        }
+    }
+}
--- a/src/WebJobs.Extensions.DurableTask/Grpc/TaskHubGrpcExceptionInterceptor.cs
+++ b/src/WebJobs.Extensions.DurableTask/Grpc/TaskHubGrpcExceptionInterceptor.cs
@@ -11,6 +11,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Grpc
 {
     internal sealed class TaskHubGrpcExceptionInterceptor : Interceptor
     {
+        // gRPC metadata key used by clients to target a task hub other than the host's default.
+        // Kept in sync with TaskHubGrpcServer.GetAttribute, which uses it to select the provider.
+        private const string TaskHubMetadataKey = "Durable-TaskHub";
+
         private readonly DurableTaskExtension extension;
 
         public TaskHubGrpcExceptionInterceptor(DurableTaskExtension extension)
@@ -73,10 +77,20 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Grpc
                 : exception;
 
             this.extension.TraceHelper.ExtensionWarningEvent(
-                this.extension.Options.HubName,
+                this.GetHubName(context),
                 instanceId: string.Empty,
                 functionName: string.Empty,
                 message: $"Unhandled exception in local gRPC call '{context.Method}': {exceptionToLog}");
+        }
+
+        /// <summary>
+        /// Resolves the task hub the failing call targeted. Clients can address a non-default hub
+        /// through request metadata, so the host's configured hub is only a fallback.
+        /// </summary>
+        private string GetHubName(ServerCallContext context)
+        {
+            string? taskHub = context.RequestHeaders.GetValue(TaskHubMetadataKey);
+            return string.IsNullOrWhiteSpace(taskHub) ? this.extension.Options.HubName : taskHub;
         }
     }
 }
--- a/test/FunctionsV2/Tests/Unit/LocalGrpcListenerTests.cs
+++ b/test/FunctionsV2/Tests/Unit/LocalGrpcListenerTests.cs
@@ -2,10 +2,16 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Net;
 using System.Net.Sockets;
+using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
+using DurableTask.Core;
+using DurableTask.Core.History;
+using Grpc.Core;
+using Grpc.Net.Client;
 using Microsoft.Azure.WebJobs.Extensions.DurableTask.Grpc;
 using Microsoft.Azure.WebJobs.Host.TestCommon;
 using Microsoft.Extensions.Logging;
@@ -14,6 +20,7 @@ using Microsoft.Extensions.Options;
 using Moq;
 using Xunit;
 using Xunit.Abstractions;
+using P = Microsoft.DurableTask.Protobuf;
 
 namespace Microsoft.Azure.WebJobs.Extensions.DurableTask.Tests
 {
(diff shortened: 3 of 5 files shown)


src/WebJobs.Extensions.DurableTask/Grpc/AspNetCoreLocalGrpcListener.cs
```diff
@@ -144,6 +144,7 @@ private async Task StartInternalAsync(CancellationToken cancellationToken)
                         services.AddSingleton(this.extension);
                         services.AddGrpc(options =>
                         {
+                            options.Interceptors.Add<TaskHubGrpcExceptionInterceptor>();
```
berndverst, line 151: ASP.NET Core gRPC server interceptors are per-request unless the interceptor type is registered with DI. With the resolved `Grpc.AspNetCore.Server` 2.49.0, this registration alone causes `DefaultGrpcInterceptorActivator<T>` to construct and release a new `TaskHubGrpcExceptionInterceptor` for every RPC. That adds avoidable activation and object-allocation overhead to high-rate calls such as `CompleteActivityTask`, `CompleteOrchestratorTask`, and `CompleteEntityTask`.

This interceptor is stateless and only holds the singleton `DurableTaskExtension`, so please register it once before `AddGrpc`:

```suggestion
                            options.Interceptors.Add<TaskHubGrpcExceptionInterceptor>();
```

and add `services.AddSingleton<TaskHubGrpcExceptionInterceptor>();` alongside `services.AddSingleton(this.extension);`. This preserves the interceptor pipeline while removing the per-RPC interceptor allocation.

berndverst (commented):
Reviewed all five changed files with a focus on runtime cost and compatibility.

I found one actionable performance issue, left inline. Otherwise, there are no public API, protobuf/wire-format, persisted-history, replay/determinism, or binary compatibility breaks.

`CreateTaskHub` and `DeleteTaskHub` now wait for provider completion and surface failures instead of returning success immediately. That is externally observable for these infrequent management RPCs, but it is a sound contract correction and does not affect orchestration hot paths. The cancellation changes apply after the request is already cancelled or deadline-expired, so they are not a practical client compatibility break.

One non-blocking test nuance: under Grpc.AspNetCore.Server 2.49.0, a raw server-side `OperationCanceledException` is handled as an unrecognized exception; the new cancellation test observes cancellation of the client call, not proof of a server-generated `Cancelled` status. If an explicit server-side `Cancelled` status is part of the intended contract, throw `RpcException(StatusCode.Cancelled, ...)` directly and assert that path.

All current CI checks are green. I recommend addressing the inline singleton-lifetime issue before approval.

berndverst (dismissed):

src/WebJobs.Extensions.DurableTask/Grpc/AspNetCoreLocalGrpcListener.cs
YunchuWang: Good catch — fixed in 3a5ab48.

`TaskHubGrpcExceptionInterceptor` is now registered in DI before `AddGrpc`, while `options.Interceptors.Add<TaskHubGrpcExceptionInterceptor>()` stays so the pipeline is unchanged:

```csharp
services.AddSingleton(this.extension);

// Register the interceptor in DI so DefaultGrpcInterceptorActivator resolves
// this singleton instead of constructing a new instance for every RPC.
services.AddSingleton<TaskHubGrpcExceptionInterceptor>();
services.AddGrpc(options =>
{
    options.Interceptors.Add<TaskHubGrpcExceptionInterceptor>();
    ...
});
```

Singleton lifetime is safe here: the interceptor has a single constructor whose only parameter is `DurableTaskExtension`, which is already registered as a singleton instance on the same container, so there is no captive dependency. All of its fields are `readonly` and the handlers only delegate plus log, so there is no per-call state to share.

I verified the activation behaviour empirically against the resolved `Grpc.AspNetCore.Server` 2.49.0 rather than assuming it — a throwaway probe resolving `IGrpcInterceptorActivator<T>` and calling `Create` twice for the same `InterceptorRegistration`:

```
WITHOUT services.AddSingleton<T>() -> sameInstance=False createdFlag=True  ctorInvocations=2
WITH    services.AddSingleton<T>() -> sameInstance=True  createdFlag=False ctorInvocations=1
```

So it was indeed constructing (and releasing, given `created: true`) a fresh interceptor per RPC, and the registration removes that allocation on `CompleteActivityTask` / `CompleteOrchestratorTask` / `CompleteEntityTask`.

Validation:
- `dotnet build src/WebJobs.Extensions.DurableTask -c Release` — **0 warnings, 0 errors** across `net8.0` and `net10.0`.
- `dotnet test test/FunctionsV2 -c Release --filter FullyQualifiedName~LocalGrpcListenerTests` — **30/30 passed on `net8.0` and 30/30 on `net10.0`**, including the restart-after-stop cases that build a fresh host/container per listener start.

berndverst (commented):
One remaining merge-blocking diagnostic correctness issue is noted inline.

src/WebJobs.Extensions.DurableTask/Grpc/TaskHubGrpcExceptionInterceptor.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        private void LogException(ServerCallContext context, Exception exception)
+        {
+            Exception exceptionToLog = exception is TaskHubRpcException rpcException
+                ? rpcException.Cause
+                : exception;
+
+            this.extension.TraceHelper.ExtensionWarningEvent(
+                this.extension.Options.HubName,
```
berndverst: Warnings for non-default task hubs are attributed to the host's default hub here. `FunctionsDurableClientProvider` sends the selected hub in the `Durable-TaskHub` metadata header, and `TaskHubGrpcServer.GetAttribute` uses that header to select the actual provider. If a call targeting hub B fails, this currently records the warning against hub A (`Options.HubName`), which makes the new diagnostic misleading in supported multi-hub scenarios.

Please resolve the hub from the request metadata on this failure-only path, with the existing default as fallback:

```csharp
string hubName = context.RequestHeaders.GetValue("Durable-TaskHub")
    ?? this.extension.Options.HubName;
```

Pass `hubName` to `ExtensionWarningEvent` and add a failing-RPC test with a non-default `Durable-TaskHub` header that verifies the warning's hub field.

YunchuWang: Fixed in c10295c.

`LogException` now resolves the hub from the request metadata, matching how `TaskHubGrpcServer.GetAttribute` selects the provider, with the configured hub as fallback:

```csharp
private string GetHubName(ServerCallContext context)
{
    string? taskHub = context.RequestHeaders.GetValue(TaskHubMetadataKey);
    return string.IsNullOrWhiteSpace(taskHub) ? this.extension.Options.HubName : taskHub;
}
```

Both `UnaryServerHandler` and `ServerStreamingServerHandler` route through `LogException`, so this covers both. The `TaskHubRpcException` unwrapping and message format are unchanged, and this stays on the failure-only path, so nothing is cached or precomputed.

On the guard: I used `IsNullOrWhiteSpace` rather than a bare `??`, and verified why rather than assuming. With `IsNullOrEmpty`, a `Durable-TaskHub: "   "` header reaches the server verbatim and gets logged as the hub name:

```
header=[   ]        loggedHub=[   ]          // IsNullOrEmpty
header=[   ]        loggedHub=[DefaultHub]   // IsNullOrWhiteSpace
```

So blank values are not trimmed anywhere in the pipeline and would otherwise produce a blank hub field. The worker only sets the header when the name is non-empty (`FunctionsDurableClientProvider.GetHeaders`), so this only guards against non-standard callers.

On the metadata key: there is no existing shared constant for `Durable-TaskHub` (`TaskHubGrpcServer` only defines `FunctionInvocationIdMetadataKey`), so I added a local `private const` rather than refactoring `TaskHubGrpcServer` to share one.

New test: `TestGrpcListener_AttributesWarningToTaskHubFromRequestHeader`, a `[Theory]` over four cases — non-default hub, absent header, empty header, whitespace header. It issues a failing `GetInstance` RPC and asserts the structured `hubName` field on the emitted `ExtensionWarningEvent`. I confirmed it actually catches the regression by reverting the fix:

```
Expected: "AttributionOtherHub"
Actual:   "AttributionDefaultHub"
```

Validation:
- `dotnet build src/WebJobs.Extensions.DurableTask -c Release` — **0 warnings, 0 errors** on `net8.0` and `net10.0` (unchanged from baseline).
- `dotnet test test/FunctionsV2 -c Release --filter FullyQualifiedName~LocalGrpcListenerTests` — **34/34 passed on `net8.0` and 34/34 on `net10.0`** (30 existing, unchanged, plus the 4 new cases).

test/FunctionsV2/Tests/Unit/LocalGrpcListenerTests.cs
```diff
@@ -61,6 +75,323 @@ public async Task TestGrpcListener_ShouldStartAndStopSuccessfully(TestGrpcListen
...
+        {
+            const string InstanceId = "test-instance";
+            const string ErrorMessage = "The durability provider failed.";
+
+            var providerException = new InvalidOperationException(ErrorMessage);
+            DurabilityProvider durabilityProvider = CreateFailingGetStateProvider(InstanceId, providerException);
+
+            Metadata headers = requestTaskHub is null
+                ? null
+                : new Metadata { { TaskHubMetadataKey, requestTaskHub } };
```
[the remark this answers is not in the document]
YunchuWang: This is not correct — `Grpc.Core` normalizes mixed-case metadata keys rather than rejecting them, so there is nothing to fix here.

`Metadata.Entry` runs keys through `NormalizeKey`, whose validator explicitly accepts `A-Z` and then lowercases the key. Only genuinely invalid characters are rejected. Verified against the resolved gRPC version rather than assumed:

```
--- does a mixed-case key throw? ---
  NO THROW. stored key = [durable-taskhub]  value = [HubB]

--- lookup with the mixed-case name (as TaskHubGrpcServer does) ---
  GetValue("Durable-TaskHub") = [HubB]
  GetValue("durable-taskhub") = [HubB]

--- which keys are actually rejected? ---
  [Durable-TaskHub  ] -> OK, normalized to [durable-taskhub]
  [durable-taskhub  ] -> OK, normalized to [durable-taskhub]
  [Durable_TaskHub  ] -> OK, normalized to [durable_taskhub]
  [Durable.TaskHub  ] -> OK, normalized to [durable.taskhub]
  [Durable TaskHub  ] -> ArgumentException: Metadata entry key not valid. ...
  [Durable:TaskHub  ] -> ArgumentException: Metadata entry key not valid. ...
```

So both parts of the comment are inaccurate: construction does not throw, and the header does flow. The second block is the decisive case — the entry is stored lowercase (exactly as HTTP/2 delivers it) yet a mixed-case `GetValue` still matches, which is precisely the server-side path this test exercises.

Two further points:

- This is pre-existing, shipped behaviour, not something the test introduces. `TaskHubGrpcServer.GetAttribute` already calls `context.RequestHeaders.GetValue("Durable-TaskHub")` with the same casing to select the durability provider. If mixed-case lookup did not match, multi-hub routing would already be broken in production. The interceptor was deliberately made consistent with it.
- `TestGrpcListener_AttributesWarningToTaskHubFromRequestHeader` passes on both `net8.0` and `net10.0`, including the non-default-hub case that asserts the logged hub is `AttributionOtherHub`. A runtime throw on construction would fail that test, and a header that did not flow would fall back to the default hub and fail the assertion. The passing test is end-to-end evidence against both claims.

The likely source of the confusion is the validator's error message, which reads "lowercase alphanumeric characters" while the implementation accepts uppercase and normalizes it.

Keeping the constant mixed-case, matching the existing `TaskHubGrpcServer` usage and the header name the worker sets in `FunctionsDurableClientProvider.GetHeaders`. Resolving as not applicable.

berndverst (approved):
Re-reviewed current head c10295c. The interceptor lifetime and multi-hub attribution feedback are correctly addressed; no remaining correctness, performance, or compatibility findings.

PR #3501, opened 2026-07-29
Fix purge requests with Python minimum dates

### LINKED ISSUE #2531: Purge function returning 400 response status
### Description
I've created a timer-triggered Azure Function (code below) to purge orchestration data in my storage account older than 14 days. I'm running it locally using Azure Function Core Tools and a local Azurite storage emulator. When running it locally, the timer trigger works OK, but the `purge_instance_history_by()` method returns the following exception:

> The operation failed with an unexpected status code 400 

 I’ve investigated a couple of things, listed out below. See relevant code snippets in the next section.

- Investigated the ‘starter’ parameter passed into the timer-triggered function
    - This returned a string with contents including the below. All seem OK to me.
        - taskHubName: “TestHubName”
        - connectionName: null,
        - baseUrl: “[http://localhost:7071/runtime/webhooks/durabletask”](http://localhost:7071/runtime/webhooks/durabletask%E2%80%9D)
        - rpcBaseUrl “[http://127.0.0.1:17071/durabletask/”](http://127.0.0.1:17071/durabletask/%E2%80%9D)
- Investigated the `DurableOrchestrationClient` instance generated to perform the purge (named client in my code below):
    - The orchestration bindings (`client._orchestration_bindings`) seem to contain all the things I expect when comparing against the input ‘starter’ argument.
- Investigated the request url produced by the `RpcManagementOptions` instance method `to_url()` which is passed onto the `delete_async_request` method.
    - Url produced: http://127.0.0.1:17071/durabletask/instances/?createdTimeFrom=1-01-01T00:00:00.000000Z&createdTimeTo=2023-07-17T17:51:00.008019Z&runtimeStatus=Completed,Failed,Terminated
    - I notice that this request URL doesn’t have the TaskHubName (TestHubName) or connection query parameters. Suggests that these aren’t passed into the RpcManagementOptions instance, which seems to be the case looking at the code.

I have some other durable functions which are triggered by orchestratorTriggers. I’m able to trigger them using Core Tools and they work without any issues, it’s only this particular timer-triggered function to purge the storage tables which isn’t working as expected. I noticed that when I trigger one of these other durable functions, the response contains a purgeHistoryDeleteUri which has both the taskHub query paramter (TestHubName) and connection query parameter (Storage). I’ve tried passing these parameters directly into the RpcManagementOptions but I still get a response status of 400.

I suspect the reason I’m getting a response status of 400 is due to something in the request url not being quite right, but I can’t put my finger on it at the moment. I’m open to hearing other thoughts though, let me know what you think.

Also just as an aside, I think the documentation (https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-instance-management?tabs=python#purge-instance-history) still needs to be amended, the `purge_instance_history_by()` example is HTTP-Triggered, not timer-triggered. I think this is captured in Issue #2406.

### Expected behavior
Purge function to return PurgeHistoryResult with number of instances deleted.

### Actual behavior
Purge function raises exception: 
> "The operation failed with an unexpected status code 400".

Worth noting that this in itself isn't a bug, this exception is raised if the list returned by `delete_async_request()` method doesn't contain a response status of 200 or 404. The issue i'm trying to resolve is why I get the 400 error. I haven't seen this particular issue on the github issues page although I've come across others regarding the purge functionality (#2406 mainly)

#### function.json file: (have tested various CRON timer schedules and they all trigger the function OK).
```
{
    "scriptFile": "main.py",
    "bindings": [
        {
            "name": "mytimer",
            "type": "timerTrigger",
            "direction": "in",
            "schedule": "0 0 20 * * *"
        },
        { 
            "name": "starter",
            "type": "orchestrationClient",
            "direction": "in"
        }
    ]
}
```
#### Timer-triggered function (main.py)
```
import azure.functions as func
from azure.durable_functions import DurableOrchestrationClient
from azure.durable_functions.models import PurgeHistoryResult
from azure.durable_functions.models.DurableOrchestrationStatus import OrchestrationRuntimeStatus
from datetime import datetime, timedelta
import logging


async def main(mytimer: func.TimerRequest, starter: str) -> None:
    
    utc_timestamp_start = datetime.now()
    logging.info(f"Purging old orchestrator function data from Azure storage tables at {utc_timestamp_start}")

    # Create durable orchestration client      
    client = DurableOrchestrationClient(context=starter)    

    # Specify timeperiod in which data should be purged (anything older than 14 days)
    created_time_from = datetime.min
    created_time_to = datetime.today() + timedelta(days=-14)

    # Set runtime status filter to use when purging orchestrator function storage tables
    runtime_status = [OrchestrationRuntimeStatus.Completed,
                    OrchestrationRuntimeStatus.Failed,
                    OrchestrationRuntimeStatus.Terminated]
    
    # Run purge function to purge storage tables
    # TODO: Tidy up code below once working as expected
    try:
        purge_result = await client.purge_instance_history_by(created_time_from, created_time_to, runtime_status)

        # Get and log purge result
        if isinstance(purge_result, PurgeHistoryResult):
            logging.info(f" Purge operation was successful. Total deleted count: {purge_result.instances_deleted}")
        else:
            logging.info(f"Purge operation failed: {purge_result}\n") 

    except Exception as e:
        logging.info("Purge operation unsuccessful. Exception raised during invokation of purge method.")
        logging.info(f"Exception raised: {e}\n")

    utc_timestamp_complete = datetime.now()
    logging.info(f"Purge operation completed at {utc_timestamp_complete}\n") 

    return None
```
### App Details

- **Core Tools version**: 4.0.5198
- **Azure Functions runtime version**: 4.21.1.20667
- **Programming language used**: Python

andystaples:
## Summary
- accept one- to three-digit years in Durable HTTP date filters
- preserve UTC semantics while normalizing Python/Linux short-year timestamps
- cover filtered purge requests using datetime.min-style values

Fixes #2531

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
+++ b/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
@@ -610,15 +610,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
             IDurableOrchestrationClient client = this.GetClient(request);
             var queryNameValuePairs = request.GetQueryNameValuePairs();
-            string createdTimeFromValue = queryNameValuePairs[CreatedTimeFromParameter];
-            bool hasBareCreatedTimeFrom = queryNameValuePairs.GetValues(string.Empty)?
-                .Any(value => string.Equals(value, CreatedTimeFromParameter, StringComparison.OrdinalIgnoreCase)) == true;
-            DateTime createdTimeFrom;
-            if (createdTimeFromValue == null && !hasBareCreatedTimeFrom)
-            {
-                createdTimeFrom = DateTime.MinValue;
-            }
-            else if (!TryGetDateTimeQueryParameterValue(queryNameValuePairs, CreatedTimeFromParameter, out createdTimeFrom))
+            if (!TryGetDateTimeQueryParameterValue(queryNameValuePairs, CreatedTimeFromParameter, out DateTime createdTimeFrom))
             {
                 var badRequestResponse = request.CreateResponse(
                     HttpStatusCode.BadRequest,
@@ -805,10 +797,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             {
                 // Python's strftime can omit leading zeroes from years before 1000 on Linux.
                 string paddedValue = value.PadLeft(value.Length + 4 - yearSeparatorIndex, '0');
-                if (DateTime.TryParse(
+                if (DateTime.TryParseExact(
                         paddedValue,
+                        "yyyy-MM-dd'T'HH:mm:ss.ffffff'Z'",
                         CultureInfo.InvariantCulture,
-                        DateTimeStyles.RoundtripKind,
+                        DateTimeStyles.AssumeUniversal | DateTimeStyles.AdjustToUniversal,
                         out dateTimeValue))
                 {
                     return true;
--- a/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
+++ b/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
@@ -789,6 +789,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         private static bool TryGetDateTimeQueryParameterValue(NameValueCollection queryStringNameValueCollection, string queryParameterName, out DateTime dateTimeValue)
         {
             string value = queryStringNameValueCollection[queryParameterName];
+            if (DateTime.TryParse(value, out dateTimeValue))
+            {
+                return true;
+            }
+
             int yearSeparatorIndex = value?.IndexOf('-') ?? -1;
             if (value != null &&
                 yearSeparatorIndex > 0 &&
@@ -799,16 +804,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 string paddedValue = value.PadLeft(value.Length + 4 - yearSeparatorIndex, '0');
                 if (DateTime.TryParseExact(
                         paddedValue,
-                        "yyyy-MM-dd'T'HH:mm:ss.ffffff'Z'",
+                        "yyyy-MM-dd'T'HH:mm:ss.ffffffK",
                         CultureInfo.InvariantCulture,
-                        DateTimeStyles.AssumeUniversal | DateTimeStyles.AdjustToUniversal,
+                        DateTimeStyles.RoundtripKind,
                         out dateTimeValue))
                 {
                     return true;
                 }
             }
 
-            return DateTime.TryParse(value, out dateTimeValue);
+            return false;
         }
 
         private static bool TryGetBooleanQueryParameterValue(NameValueCollection queryStringNameValueCollection, string queryParameterName, out bool boolValue)
--- a/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
+++ b/src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
@@ -610,7 +610,15 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
             IDurableOrchestrationClient client = this.GetClient(request);
             var queryNameValuePairs = request.GetQueryNameValuePairs();
-            if (!TryGetDateTimeQueryParameterValue(queryNameValuePairs, CreatedTimeFromParameter, out DateTime createdTimeFrom))
+            string createdTimeFromValue = queryNameValuePairs[CreatedTimeFromParameter];
+            bool hasBareCreatedTimeFrom = queryNameValuePairs.GetValues(string.Empty)?
+                .Any(value => string.Equals(value, CreatedTimeFromParameter, StringComparison.OrdinalIgnoreCase)) == true;
+            DateTime createdTimeFrom;
+            if (createdTimeFromValue == null && !hasBareCreatedTimeFrom)
+            {
+                createdTimeFrom = DateTime.MinValue;
+            }
+            else if (!TryGetDateTimeQueryParameterValue(queryNameValuePairs, CreatedTimeFromParameter, out createdTimeFrom))
             {
                 var badRequestResponse = request.CreateResponse(
                     HttpStatusCode.BadRequest,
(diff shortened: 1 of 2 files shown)


test/FunctionsV2/Tests/Unit/HttpApiHandlerTests.cs
```diff
@@ -851,6 +851,42 @@ public async Task HandleGetStatusRequestAsync_Correctly_Parses_InstanceId_With_S
...
+                .Callback((DateTime createdTimeFrom, DateTime? createdTimeTo, IEnumerable<OrchestrationStatus> runtimeStatus) =>
+                {
+                    actualCreatedTimeFrom = createdTimeFrom;
+                })
+                .Returns(Task.FromResult(new PurgeHistoryResult(1)));
+
+            var purgeRequestUriBuilder = new UriBuilder(TestConstants.NotificationUrl);
+            purgeRequestUriBuilder.Path += "/Instances/";
+            purgeRequestUriBuilder.Query = $"createdTimeFrom={createdTimeFrom}";
+
```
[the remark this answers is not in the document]
andystaples: Updated the regression test to URL-encode the `createdTimeFrom` value with `WebUtility.UrlEncode`, matching the request construction used elsewhere in this test suite.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -788,6 +789,20 @@ private static bool TryGetStringQueryParameterValue(NameValueCollection queryStr
...
+                // Python's strftime can omit leading zeroes from years before 1000 on Linux.
+                string paddedValue = value.PadLeft(value.Length + 4 - yearSeparatorIndex, '0');
+                return DateTime.TryParse(
+                    paddedValue,
+                    CultureInfo.InvariantCulture,
+                    DateTimeStyles.RoundtripKind,
+                    out dateTimeValue);
+            }
+
             return DateTime.TryParse(value, out dateTimeValue);
```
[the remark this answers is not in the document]
andystaples: Updated the short-year branch to fall back to the existing `DateTime.TryParse` behavior when normalized invariant parsing fails, preserving compatibility for other accepted date formats.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -788,6 +789,20 @@ private static bool TryGetStringQueryParameterValue(NameValueCollection queryStr
         private static bool TryGetDateTimeQueryParameterValue(NameValueCollection queryStringNameValueCollection, string queryParameterName, out DateTime dateTimeValue)
         {
             string value = queryStringNameValueCollection[queryParameterName];
+            int yearSeparatorIndex = value?.IndexOf('-') ?? -1;
+            if (yearSeparatorIndex > 0 &&
+                yearSeparatorIndex < 4 &&
+                value.Take(yearSeparatorIndex).All(char.IsDigit))
```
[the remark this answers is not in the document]
andystaples: Added an explicit `value != null` guard before calling `Take` and `PadLeft`, making the null-safety contract clear.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -609,7 +610,15 @@ private async Task<HttpResponseMessage> HandleDeleteHistoryWithFiltersRequestAsy
...
             IDurableOrchestrationClient client = this.GetClient(request);
             var queryNameValuePairs = request.GetQueryNameValuePairs();
-            if (!TryGetDateTimeQueryParameterValue(queryNameValuePairs, CreatedTimeFromParameter, out DateTime createdTimeFrom))
+            string createdTimeFromValue = queryNameValuePairs[CreatedTimeFromParameter];
+            bool hasBareCreatedTimeFrom = queryNameValuePairs.GetValues(string.Empty)?
+                .Any(value => string.Equals(value, CreatedTimeFromParameter, StringComparison.OrdinalIgnoreCase)) == true;
+            DateTime createdTimeFrom;
+            if (createdTimeFromValue == null && !hasBareCreatedTimeFrom)
+            {
+                createdTimeFrom = DateTime.MinValue;
```
berndverst: Please keep `createdTimeFrom` required. This turns a bare `DELETE .../instances/` into `PurgeInstanceHistoryAsync(DateTime.MinValue, DateTime.UtcNow, null)`. In AzureStorage 2.9.1, `MinValue` omits the lower-bound filter and a null status list adds no status filter, so this request deletes every instance created up to now, including running orchestrations. That is a destructive behavior change unrelated to parsing Python's explicit `datetime.min` value; the endpoint has returned 400 for an omitted lower bound since it was introduced. Remove this default and preserve the existing missing/invalid-parameter response.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
```diff
@@ -788,6 +797,24 @@ private static bool TryGetStringQueryParameterValue(NameValueCollection queryStr
         private static bool TryGetDateTimeQueryParameterValue(NameValueCollection queryStringNameValueCollection, string queryParameterName, out DateTime dateTimeValue)
         {
             string value = queryStringNameValueCollection[queryParameterName];
+            int yearSeparatorIndex = value?.IndexOf('-') ?? -1;
+            if (value != null &&
+                yearSeparatorIndex > 0 &&
+                yearSeparatorIndex < 4 &&
+                value.Take(yearSeparatorIndex).All(char.IsDigit))
```
berndverst, line 801: Please restrict this normalization to the Python ISO timestamp shape. As written, it reinterprets date strings that the existing fallback already accepts: under `en-US`, `DateTime.TryParse("12-11-10")` yields 2010-12-11, while this branch pads it to `0012-11-10` and returns year 12. Because this helper is also used by list, entity, and purge filters, that can drastically widen a purge. Requiring the `yyyy-MM-ddT...` structure (or using `TryParseExact` after padding) would accept the intended 1-3 digit ISO years without changing other legacy formats.

berndverst (changes requested):
Requesting changes for two behavior regressions: an unfiltered DELETE can now purge the entire task hub, and the broad short-year detection reinterprets previously accepted date strings. The explicit Python datetime.min case itself is covered and parses successfully.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
andystaples: Agreed. I restored the required `createdTimeFrom` behavior so a bare collection DELETE remains a 400 and cannot become an unfiltered purge. Supporting an all-time purge should be handled separately with an explicit, safe client contract.

src/WebJobs.Extensions.DurableTask/HttpApiHandler.cs
andystaples: Agreed. The normalized value now uses `TryParseExact` for Python's `%Y-%m-%dT%H:%M:%S.%fZ` shape, then falls back to the original parser. I also added coverage proving `12-11-10` retains its existing interpretation.

berndverst (approved):
Both requested changes are addressed: collection DELETE still requires createdTimeFrom, and short-year normalization is constrained to the Python timestamp format while preserving existing parser behavior.

PR #3505, opened 2026-07-29
Fire the Terminated Event Grid lifecycle notification when an orchestration is terminated

### LINKED ISSUE #286: Event Grid notification does not fire for orchestration termination
I noticed this while refactoring the event grid notification unit tests. It seems the `LifeCycleNotificationHelper.OrchestratorTerminatedAsync` method is never called anywhere in our codebase. The existing unit test code unfortunately did not catch this.

Unfortunately I don't see a simple way to fix this. The challenge is that termination doesn't require the orchestrator function to actually run, so there isn't any code inside of **Microsoft.Azure.WebJobs.Extensions.DurableTask** that we can hook into. It might require us to add hooks directly into **DurableTask.Core**.

YunchuWang:
# Summary
## What changed?
- Raise the `Terminated` Event Grid lifecycle notification from the orchestration dispatch middleware, which is the common ancestor of every orchestration work item.
- Add `DurableTaskExtension.GetTerminationEventOrNull`, which inspects only `OrchestrationRuntimeState.NewEvents` so the notification fires exactly once and is not re-raised on every subsequent replay.
- On the middleware-passthrough path (`OutOfProcMiddleware`), publish `Terminated` instead of the previously-published, incorrect `Completed` notification.
- Enable the existing E2E assertion that was pinned to the buggy behavior, and add unit coverage.
- Add a release note for the user-visible bug fix.

## Why is this change needed?
- A termination is applied by the DTFx orchestration executor itself. The instance is completed without the orchestrator function ever running to completion, so no code path inside the orchestrator shim was in a position to publish the notification.
- The result was that `LifeCycleNotificationHelper.OrchestratorTerminatedAsync` was effectively dead code: terminating an instance published no `Terminated` Event Grid notification at all on the in-process and legacy out-of-proc paths.
- On the middleware-passthrough path (.NET isolated), a terminated instance still produced a `CompleteOrchestrationAction` with status `Terminated`, which set `RemoteOrchestratorContext.OrchestratorCompleted` and caused a **`Completed`** notification to be published for an instance that was actually terminated.
- This was already a known and documented gap in the test suite. `DurableTaskLifeCycleNotificationTest` carried the assertion `Assert.Equal(1, callCount)` with the comment `// TODO: There should be two calls, but the termination notification is not being fired.` referencing this issue. That assertion is now `Assert.Equal(2, callCount)` and the TODO is removed.

## Issues / work items
- Resolves #286

---

# Project checklist
  - [x] Otherwise: Notes added to `release_notes.md`
  - [ ] Otherwise: Ensure EventIds are within the supported range in the existing Windows infrastructure (validate via deployed telemetry).
  - [x] Otherwise: This change applies exclusively to `WebJobs.Extensions.DurableTask` v3.x and will be retained only in the `dev` and `main` branches
- [x] Breaking change?
  - [x] If yes:
    - Impact: On the middleware-passthrough path (.NET isolated), terminating an orchestration previously published a `Completed` Event Grid notification. It now publishes `Terminated`. Any subscriber that is currently keying off the incorrect `Completed` event to detect terminated instances on isolated .NET will stop seeing it. On the in-process and legacy out-of-proc paths this is purely additive — a notification that was previously never sent is now sent.
    - Migration guidance: Subscribers should handle the `Terminated` runtime status, which is what the documented Event Grid contract already specifies. No configuration change is required. Flagged here for visibility rather than because the previous behavior was intentional — the old `Completed` event was incorrect for a terminated instance.

---

- [ ] No

If AI was used:
- Tool(s): GitHub Copilot CLI
- AI-assisted areas/files: the `GetTerminationEventOrNull` helper and dispatch-middleware hook in `DurableTaskExtension.cs`, the notification swap in `OutOfProcMiddleware.cs`, the `OutOfProcMiddlewareTests` theory, the E2E assertion update, and the release note.
- What you changed after AI output: The original brief only covered the in-process path. Reading the passthrough branch revealed that isolated .NET was not silent but actively publishing a **wrong** `Completed` notification, so the fix was widened from "add a missing notification" to "replace an incorrect one" and the breaking-change disclosure above was added. The helper was also narrowed from scanning the full event history to scanning only `NewEvents`, because matching against the full history would re-publish the notification on every subsequent replay of the instance.

---

# Testing
## Automated tests
- Result: Passed
  - `OutOfProcMiddlewareTests` + `DurableTaskLifeCycleNotificationTest`: 62/62 on .NET 8 and 62/62 on .NET 10
  - New unit theory `CallOrchestratorAsync_TerminatedInstance_RaisesTerminatedNotification`, covering both a supplied termination reason (`"sayōnara"`, which also exercises non-ASCII input) and a `null` reason. Both cases are genuinely terminated: an earlier revision of this test appended the `ExecutionTerminatedEvent` only when a reason was supplied, so the `null` case silently exercised the *Completed* path and passed with or without the fix. Caught in review and corrected.
  - New unit test `CallOrchestratorAsync_CompletedInstance_RaisesCompletedNotification` as a control, so the Completed path stays covered deliberately rather than as a side effect of the theory
  - Mutation-checked rather than assumed: forcing the production guard to `false` yields 2 failed / 1 passed, confirming both terminated cases genuinely depend on the fix and the control case does not
  - Existing E2E `DurableTaskLifeCycleNotificationTest` assertion moved from `1` to `2` lifecycle calls, removing the `// TODO` that documented this bug

- Environment (OS, .NET version, components): Windows, .NET 8 and .NET 10, local Functions V2 test host, Azurite
- Steps + observed results:
  1. Ran `DurableTaskLifeCycleNotificationTest` against the unmodified `dev` branch with the assertion raised to `2`. It failed with one observed lifecycle call, reproducing the missing notification.
  2. Applied the fix and reran. Two lifecycle calls were observed, the second being `Terminated`.
  3. Ran the combined unit + lifecycle E2E suite on both target frameworks with the results above.

---

# Notes for reviewers
- `AddDeferredTask` is used on the in-process path so the notification is dispatched consistently with the other lifecycle notifications raised from that middleware, rather than blocking the dispatch itself.
- Only `NewEvents` is inspected, never the full history. This is the load-bearing detail that keeps the notification single-shot; a history-wide scan would re-fire it on every replay of a terminated instance.
- The passthrough path **replaces** the notification rather than emitting both, so a terminated instance produces exactly one lifecycle notification, not two.
- Out of scope, flagged so it is not lost: `TagActivityWithOrchestrationStatus` still tags terminated instances as `Completed` for Application Insights in the passthrough branch. That is telemetry tagging rather than the Event Grid contract this issue is about, so it is intentionally not changed here. Happy to open a separate issue for it if you agree it is worth tracking.


YunchuWang:
Addressed both review comments in 40fb6abb.

**1. Test coverage gap — this was a real defect in my test, thanks for catching it.**

`SetupCompletedOrchestratorTest` only appended the `ExecutionTerminatedEvent` when `terminationReason != null`. So the `[InlineData(null)]` case was exercising the **Completed** path under a test named `..._TerminatedInstance_RaisesTerminatedNotification`. Termination with no reason — a supported input — had zero coverage, and the assertion passed for the wrong reason.

Fixed by separating "is terminated" from "what the reason was": the helper now takes an explicit `isTerminated` flag, and both theory cases are genuinely terminated. Added `CallOrchestratorAsync_CompletedInstance_RaisesCompletedNotification` as a control so the Completed path is still covered deliberately rather than accidentally.

Verified by mutation rather than by inspection — reverting the production guard to `if (false)` gives **2 failed / 1 passed**, i.e. both terminated cases now genuinely depend on the fix. Before this change the null-reason case passed either way.

**2. Nullability annotations — applied, with one wrinkle worth recording.**

`GetTerminationEventOrNull` already promised null in both its name and its doc comment, so the annotation was warranted. The catch: `DurableTaskExtension.cs` is nullable-*oblivious* apart from one small island at lines 523-565, so a bare `?` there emits **CS8632** ("nullable annotation used outside a #nullable context"). I wrapped the method in its own `#nullable enable` / `#nullable restore` island, matching the pattern already used in this file. `OutOfProcMiddleware.cs` (the only consumer) *is* nullable-enabled, so it now gets a checked contract instead of an oblivious one.

Builds with **0 warnings**. Tests: **62/62 on net8.0 and net10.0** (up from 61 — the split added one).

bachuv (approved):

PR #3506, opened 2026-07-29
Unwrap TargetInvocationException thrown by class-based entity operations

### LINKED ISSUE #1230: Unwrap TargetInvocationExceptions thrown by DispatchAsync
When using class-based entities, the exceptions thrown inside an operation are currently wrapped by a TargetInvocationException (because the entity operations are invoked using reflection).

This is inconvenient and makes not much sense from a developer experience perspective - the use of reflection is an implementation detail, not part of the application logic, and the TargetInvocationException carries no useful information.

I believe we should automatically unwrap the InnerException and return it instead of the TargetInvocationException. Then developers can more easily catch and handle application-specific exceptions thrown by entity operations.

YunchuWang:
# Summary
## What changed?
- Unwrap `TargetInvocationException` when a class-based entity operation throws, rethrowing the original application exception with its stack trace preserved via `ExceptionDispatchInfo.Capture(...).Throw()`.
- Update the existing `CallFaultyEntity` E2E orchestration to expect the application exception rather than the reflection wrapper.
- Add a release note under **Breaking Changes**.

## Why is this change needed?
- Class-based entity operations are invoked through reflection (`MethodInfo.Invoke`) inside `IDurableEntityContext.DispatchAsync<T>`, which wraps any exception the operation throws in a `TargetInvocationException`.
- Reflection is an implementation detail of how the extension dispatches operations, not part of application semantics. Leaking the wrapper forces callers to write `catch (TargetInvocationException e)` and then re-inspect `e.InnerException` to find out what actually went wrong.
- The wrapper carries no useful information of its own, and it makes class-based entities behave differently from function-based entities, which propagate the application exception directly because they never go through reflection.
- After this change, both entity styles surface the same exception type for the same failure, so `catch (MyDomainException)` works uniformly.

## Issues / work items
- Resolves #1230

---

# Project checklist
  - [x] Otherwise: Notes added to `release_notes.md`
  - [ ] Otherwise: Ensure EventIds are within the supported range in the existing Windows infrastructure (validate via deployed telemetry).
  - [x] Otherwise: This change applies exclusively to `WebJobs.Extensions.DurableTask` v3.x and will be retained only in the `dev` and `main` branches
- [x] Breaking change?
  - [x] If yes:
    - Impact: Application code that currently catches `TargetInvocationException` around a call to a **class-based** entity operation will no longer catch anything, because the original exception is now propagated in its place. The clearest evidence of this is inside this PR itself: the existing `CallFaultyEntity` E2E orchestration had to change from `catch (TargetInvocationException) when (isClassBased)` to `catch (FaultyEntity.UnserializableKaboom) when (isClassBased)`. Any user with the equivalent pattern is affected the same way. Function-based entities are unaffected, since they never went through reflection.
    - Migration guidance: Replace `catch (TargetInvocationException e)` + `e.InnerException` inspection with a direct `catch` on the application exception type. Callers that want to keep handling both shapes during a transition can catch the application exception type and `TargetInvocationException` separately. This is the behavior change requested in #1230 and it makes class-based entities consistent with function-based entities.

---

- [ ] No

If AI was used:
- Tool(s): GitHub Copilot CLI
- AI-assisted areas/files: the `DispatchAsync<T>` unwrap in `DurableEntityContext.cs`, the `CallFaultyEntity` expectation update in `TestOrchestrations.cs`, and the release note.
- What you changed after AI output: The narrowing of the E2E catch clause was audited specifically for false confidence. Changing `catch (TargetInvocationException)` — which matched *any* wrapped exception — to `catch (UnserializableKaboom)` silently reduces what the test tolerates, so every operation reachable through the shared `ExpectException` helper was traced to confirm the narrower type is actually correct (see Notes for reviewers). The release note was also moved from "Bug Fixes" to "Breaking Changes" once it was clear the test change was itself proof of a user-visible break.

---

# Testing
## Automated tests
- Result: Passed
  - Full unit suite: **311/311** on .NET 8, no regressions
  - `src/WebJobs.Extensions.DurableTask` clean rebuild: **0 warnings, 0 errors**
- **Honest gap, please read:** the regression that directly covers this change is the E2E `DurableEntity_CallFaultyEntity`, and it **could not be executed locally**. Every test deriving from `DurableTaskEndToEndTestBase` fails at construction on this machine with `UnauthorizedAccessException: Error Starting ETW: Access Denied (Administrator rights required)`, thrown from `LogEventTraceListener.CaptureLogs`. This is an environment limitation, not a defect in the change, but it does mean **CI is the first place this E2E will actually run.** Flagging it explicitly rather than implying coverage I did not obtain.

- Environment (OS, .NET version, components): Windows, .NET 8 / .NET 10, non-elevated shell (hence the ETW limitation above)
- Steps + observed results: In place of the unrunnable E2E, the correctness of the narrowed catch clause was established by static analysis of every reachable throw path — see Notes for reviewers. `ExceptionDispatchInfo.Capture(...).Throw()` is the standard rethrow-preserving-stack-trace idiom; the trailing `throw;` is unreachable and exists only to satisfy definite-assignment analysis on `result`.

---

# Notes for reviewers
- The `when (e.InnerException != null)` guard means a `TargetInvocationException` with no inner exception is left untouched rather than being replaced by something less informative.
- On the narrowed E2E catch clause, the thing worth double-checking: the class-based `FaultyEntity` can throw **three** different exception types — `Throw()` produces `SerializableKaboom`, `ThrowUnDeserializable()` produces `UnDeserializableKaboom`, and `ThrowUnserializable()`/`SetThenThrow()`/`SendThenThrow()`/`DeleteThenThrow()` all produce `UnserializableKaboom`. The old `catch (TargetInvocationException)` matched all three; the new clause matches only `UnserializableKaboom`. That is safe **only** because all three `ExpectException(...)` call sites in `CallFaultyEntity` use `SetThenThrow` or `DeleteThenThrow`, both of which route to `Throw(false, false)` → `UnserializableKaboom`. It is correct, but it is correct by a narrow margin — if a future test routes `Throw()` or `ThrowUnDeserializable()` through `ExpectException`, this clause will need widening.
- `System.Runtime.ExceptionServices` was already imported in `DurableEntityContext.cs`; `System.Reflection` was dropped from `TestOrchestrations.cs` since `TargetInvocationException` is no longer referenced there.


YunchuWang:
## Local end-to-end validation

I validated the exception behavior locally using a **real .NET 8 in-process Functions host**, **Azurite 3.36.0**, and a locally built Durable extension. The scenario performed a real HTTP orchestration start, orchestration-to-class-based-entity call, durable message/history round trip, and returned the exception type observed by the orchestrator.

### Results

| Build/configuration | Orchestrator output |
|---|---|
| Merge base `91eb6695` | `TYPE=System.Reflection.TargetInvocationException; INNER_TYPE=UnserializableKaboom` |
| PR HEAD `e9061c90` | `TYPE=UnserializableKaboom; INNER_TYPE=<null>` |
| PR HEAD plus the proposed local `UnwrapEntityOperationExceptions=false` compatibility patch | `TYPE=System.Reflection.TargetInvocationException; INNER_TYPE=UnserializableKaboom` |

This confirms that the PR unwraps the reflection-generated exception as intended, while the proposed opt-out restores the previous observable behavior.

### Environment and commands

```powershell
azurite --location $env:TEMP\durable-pr3506-azurite --silent --skipApiVersionCheck
func host start --port 7171 --verbose
```

The host logs confirmed the net8 in-process host and the expected locally built extension DLL were loaded. For the compatibility run, the logs also confirmed:

```text
unwrapEntityOperationExceptions:false
"UnwrapEntityOperationExceptions": false
```

Automated coverage for the local compatibility patch:

```text
Targeted option/dispatch tests: 11 passed, 0 failed
Non-E2E regression suite:       339 passed, 0 failed
```

The proposed compatibility-option changes are currently a **local working-tree patch on top of PR HEAD and have not been pushed**, so the third row documents validation of that proposal rather than code currently present in the remote PR.

YunchuWang:
Closing this draft PR. The affected code path is limited to **.NET in-process class-based Durable Entities**; .NET isolated and function-based entities are not affected. Support for the Azure Functions .NET in-process model ends on **November 10, 2026**, and this change would intentionally alter the observable exception type for existing applications.

Given the narrow, retiring surface area and the compatibility cost, we don't plan to ship this change. Customers should migrate to the .NET isolated worker model for continued support: https://learn.microsoft.com/azure/durable-task/durable-functions/durable-functions-migrate

The linked issue will be closed as not planned.

PR #3510, opened 2026-07-31
Forward Durable settings to gRPC workers

### LINKED ISSUE #3503: Serialize useForwardedHost in middleware durable-client binding

The middleware-passthrough durable-client binding serializes `httpBaseUrl`, `maxGrpcMessageSizeInBytes`, and `grpcHttpClientTimeout`, but it does not serialize the effective `extensions.durableTask.httpSettings.useForwardedHost` value.

Language SDKs that construct HTTP management payloads therefore cannot distinguish trusted proxy configurations from untrusted request headers. Trusting `Forwarded` or `X-Forwarded-*` unconditionally can expose the management system key through an attacker-controlled `Location` host; ignoring them unconditionally means apps that enable `useForwardedHost` cannot receive externally correct management URLs.

- microsoft/durabletask-python#233
- microsoft/durabletask-python#237

Add the effective `HttpOptions.UseForwardedHost` value to the middleware-passthrough durable-client binding JSON, for example as `useForwardedHost`.

Update the exact emitted-payload tests to verify both:

- the default configuration serializes `useForwardedHost: false`;
- `extensions.durableTask.httpSettings.useForwardedHost: true` serializes `useForwardedHost: true`.

This lets language SDKs safely gate adoption of `Forwarded` and `X-Forwarded-*` while remaining secure with older hosts and default configurations.

- Always trust forwarded headers: rejected because an untrusted client can redirect management URLs containing the system key.
- Always ignore forwarded headers: secure, but breaks the documented proxy behavior when `useForwardedHost` is enabled.
- Infer trust from header presence: rejected because header presence does not establish that the request passed through a trusted proxy.

**Additional context**

At the current `dev` implementation, `DurableTaskClientConverter.DurableClientInputData` consumes `httpBaseUrl` but has no forwarded-host opt-in field. The option currently appears in `HttpOptions` and `HttpApiHandler`, so it needs to be carried into the middleware-passthrough binding contract before downstream SDK support can become effective.

andystaples:
## Summary

- expose `httpSettings.useForwardedHost` in middleware-passthrough durable-client binding payloads while leaving legacy payloads unchanged
- expose `rollbackEntityOperationsOnExceptions` in the existing entity batch protobuf properties map
- preserve backward compatibility so older workers can ignore the new settings

## Testing

- `dotnet test test\FunctionsV2\WebJobs.Extensions.DurableTask.Tests.V2.csproj --framework net8.0 --no-restore --filter "FullyQualifiedName~ProtobufUtilsTests|FullyQualifiedName~TestBindingHelper_"`

Fixes #3503

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
@@ -52,6 +52,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     RpcBaseUrl = localRpcAddress,
                     RequiredQueryStringParameters = this.config.HttpApiHandler.GetUniversalQueryStrings(),
                     HttpBaseUrl = this.config.HttpApiHandler.GetBaseUrl(),
+                    UseForwardedHost = this.config.Options.HttpSettings.UseForwardedHost,
                     MaxGrpcMessageSizeInBytes = this.config.Options.MaxGrpcMessageSizeInBytes,
                     GrpcHttpClientTimeout = JsonConvert.SerializeObject(this.config.Options.GrpcHttpClientTimeout),
                 });
@@ -149,6 +150,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
             /// </summary>
             [JsonProperty("httpBaseUrl")]
             public string? HttpBaseUrl { get; set; }
+
+            /// <summary>
+            /// Gets or sets a value indicating whether workers should use forwarded headers for URL construction.
+            /// </summary>
+            [JsonProperty("useForwardedHost", NullValueHandling = NullValueHandling.Ignore)]
+            public bool? UseForwardedHost { get; set; }
 
             /// <summary>
             /// Optional setting that specifies the maximum gRPC receive message size (in bytes) for the DurableTaskClient.
--- a/src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
@@ -52,7 +52,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                     RpcBaseUrl = localRpcAddress,
                     RequiredQueryStringParameters = this.config.HttpApiHandler.GetUniversalQueryStrings(),
                     HttpBaseUrl = this.config.HttpApiHandler.GetBaseUrl(),
-                    UseForwardedHost = this.config.Options.HttpSettings.UseForwardedHost,
+                    UseForwardedHost = this.config.Options.HttpSettings?.UseForwardedHost ?? false,
                     MaxGrpcMessageSizeInBytes = this.config.Options.MaxGrpcMessageSizeInBytes,
                     GrpcHttpClientTimeout = JsonConvert.SerializeObject(this.config.Options.GrpcHttpClientTimeout),
                 });
--- a/src/WebJobs.Extensions.DurableTask/Bindings/EntityTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/EntityTriggerAttributeBindingProvider.cs
@@ -139,7 +139,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 {
                     // Generate a byte array which is the serialized protobuf payload
                     // https://developers.google.com/protocol-buffers/docs/csharptutorial#parsing_and_serialization
-                    var entityBatchRequest = remoteContext.Request.ToEntityBatchRequest(remoteContext.Configurations);
+                    var entityBatchRequest = remoteContext.Request.ToEntityBatchRequest(
+                        remoteContext.Configurations,
+                        remoteContext.RollbackEntityOperationsOnExceptions);
 
                     // We convert the binary payload into a base64 string because that seems to be the most commonly supported
                     // format for Azure Functions language workers. Attempts to send unencoded byte[] payloads were unsuccessful.
--- a/src/WebJobs.Extensions.DurableTask/Bindings/EntityTriggerAttributeBindingProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/Bindings/EntityTriggerAttributeBindingProvider.cs
@@ -139,9 +139,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
                 {
                     // Generate a byte array which is the serialized protobuf payload
                     // https://developers.google.com/protocol-buffers/docs/csharptutorial#parsing_and_serialization
-                    var entityBatchRequest = remoteContext.Request.ToEntityBatchRequest(
-                        remoteContext.Configurations,
-                        remoteContext.RollbackEntityOperationsOnExceptions);
+                    var entityBatchRequest = remoteContext.Request.ToEntityBatchRequest(remoteContext.Configurations);
 
                     // We convert the binary payload into a base64 string because that seems to be the most commonly supported
                     // format for Azure Functions language workers. Attempts to send unencoded byte[] payloads were unsuccessful.
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteEntityContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteEntityContext.cs
@@ -14,15 +14,16 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         public RemoteEntityContext(EntityBatchRequest batchRequest, DurableTaskOptions options, bool isExtendedSession, bool includeEntityState)
         {
             this.Request = batchRequest;
-            this.RollbackEntityOperationsOnExceptions = options.RollbackEntityOperationsOnExceptions;
+            this.Configurations = new RemoteInstanceConfiguration
+            {
+                RollbackEntityOperationsOnExceptions = options.RollbackEntityOperationsOnExceptions,
+            };
+
             if (options.ExtendedSessionsEnabled)
             {
-                this.Configurations = new RemoteInstanceConfiguration
-                {
-                    IsExtendedSession = isExtendedSession,
-                    IncludeState = includeEntityState,
-                    ExtendedSessionIdleTimeoutInSeconds = options.ExtendedSessionIdleTimeoutInSeconds,
-                };
+                this.Configurations.IsExtendedSession = isExtendedSession;
+                this.Configurations.IncludeState = includeEntityState;
+                this.Configurations.ExtendedSessionIdleTimeoutInSeconds = options.ExtendedSessionIdleTimeoutInSeconds;
             }
         }
 
@@ -30,10 +31,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         internal EntityBatchRequest Request { get; private set; }
 
         [JsonProperty("configurations")]
-        public RemoteInstanceConfiguration? Configurations { get; }
-
-        [JsonIgnore]
-        internal bool RollbackEntityOperationsOnExceptions { get; }
+        public RemoteInstanceConfiguration Configurations { get; }
 
         [JsonIgnore]
         internal EntityBatchResult? Result { get; set; }
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteEntityContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteEntityContext.cs
@@ -14,16 +14,15 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         public RemoteEntityContext(EntityBatchRequest batchRequest, DurableTaskOptions options, bool isExtendedSession, bool includeEntityState)
         {
             this.Request = batchRequest;
-            this.Configurations = new RemoteInstanceConfiguration
-            {
-                RollbackEntityOperationsOnExceptions = options.RollbackEntityOperationsOnExceptions,
-            };
-
+            this.RollbackEntityOperationsOnExceptions = options.RollbackEntityOperationsOnExceptions;
             if (options.ExtendedSessionsEnabled)
             {
-                this.Configurations.IsExtendedSession = isExtendedSession;
-                this.Configurations.IncludeState = includeEntityState;
-                this.Configurations.ExtendedSessionIdleTimeoutInSeconds = options.ExtendedSessionIdleTimeoutInSeconds;
+                this.Configurations = new RemoteInstanceConfiguration
+                {
+                    IsExtendedSession = isExtendedSession,
+                    IncludeState = includeEntityState,
+                    ExtendedSessionIdleTimeoutInSeconds = options.ExtendedSessionIdleTimeoutInSeconds,
+                };
             }
         }
 
@@ -31,7 +30,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         internal EntityBatchRequest Request { get; private set; }
 
         [JsonProperty("configurations")]
-        public RemoteInstanceConfiguration Configurations { get; }
+        public RemoteInstanceConfiguration? Configurations { get; }
+
+        [JsonIgnore]
+        internal bool RollbackEntityOperationsOnExceptions { get; }
 
         [JsonIgnore]
         internal EntityBatchResult? Result { get; set; }
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteEntityContext.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteEntityContext.cs
@@ -14,6 +14,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         public RemoteEntityContext(EntityBatchRequest batchRequest, DurableTaskOptions options, bool isExtendedSession, bool includeEntityState)
         {
             this.Request = batchRequest;
+            this.RollbackEntityOperationsOnExceptions = options.RollbackEntityOperationsOnExceptions;
             if (options.ExtendedSessionsEnabled)
             {
                 this.Configurations = new RemoteInstanceConfiguration
@@ -30,6 +31,9 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         [JsonProperty("configurations")]
         public RemoteInstanceConfiguration? Configurations { get; }
+
+        [JsonIgnore]
+        internal bool RollbackEntityOperationsOnExceptions { get; }
 
         [JsonIgnore]
         internal EntityBatchResult? Result { get; set; }
--- a/src/WebJobs.Extensions.DurableTask/RemoteInstanceConfiguration.cs
+++ b/src/WebJobs.Extensions.DurableTask/RemoteInstanceConfiguration.cs
@@ -30,10 +30,5 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// Gets or sets the amount of time in seconds before an idle extended session times out.
         /// </summary>
         internal int ExtendedSessionIdleTimeoutInSeconds { get; set; }
-
-        /// <summary>
-        /// Gets or sets whether an uncaught exception should roll back the effects of an entity operation.
-        /// </summary>
-        internal bool RollbackEntityOperationsOnExceptions { get; set; } = true;
     }
 }
--- a/src/WebJobs.Extensions.DurableTask/RemoteInstanceConfiguration.cs
+++ b/src/WebJobs.Extensions.DurableTask/RemoteInstanceConfiguration.cs
@@ -30,5 +30,10 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         /// Gets or sets the amount of time in seconds before an idle extended session times out.
         /// </summary>
         internal int ExtendedSessionIdleTimeoutInSeconds { get; set; }
+
+        /// <summary>
+        /// Gets or sets whether an uncaught exception should roll back the effects of an entity operation.
+        /// </summary>
+        internal bool RollbackEntityOperationsOnExceptions { get; set; } = true;
     }
 }
--- a/test/FunctionsV2/Tests/Unit/LocalGrpcListenerTests.cs
+++ b/test/FunctionsV2/Tests/Unit/LocalGrpcListenerTests.cs
@@ -19,6 +19,7 @@ using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Extensions.Options;
 using Moq;
+using Newtonsoft.Json.Linq;
 using Xunit;
 using Xunit.Abstractions;
 using P = Microsoft.DurableTask.Protobuf;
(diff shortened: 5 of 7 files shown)


src/WebJobs.Extensions.DurableTask/Bindings/EntityTriggerAttributeBindingProvider.cs
```diff
@@ -139,7 +139,9 @@ public Task<ITriggerData> BindAsync(object value, ValueBindingContext context)
                 {
                     // Generate a byte array which is the serialized protobuf payload
                     // https://developers.google.com/protocol-buffers/docs/csharptutorial#parsing_and_serialization
-                    var entityBatchRequest = remoteContext.Request.ToEntityBatchRequest(remoteContext.Configurations);
+                    var entityBatchRequest = remoteContext.Request.ToEntityBatchRequest(
+                        remoteContext.Configurations,
+                        remoteContext.RollbackEntityOperationsOnExceptions);
```
sophiatev: Can this new property be made a part of `Configurations`?

test/FunctionsV2/Tests/Unit/ProtobufUtilsTests.cs
```diff
@@ -108,6 +109,44 @@ public void ToEntityBatchResult_PassesDefaultVersionToActions()
             Assert.Equal(defaultVersion, action.Version);
         }
 
+        [Theory]
+        [InlineData(null, true)]
```
sophiatev: I'm surprised - if the value is explicitly unset the default is "true"?

src/WebJobs.Extensions.DurableTask/Bindings/EntityTriggerAttributeBindingProvider.cs
andystaples: Done — `RollbackEntityOperationsOnExceptions` is now part of `RemoteInstanceConfiguration`, and `RemoteEntityContext` always creates that configuration so the existing property-map conversion forwards it.

test/FunctionsV2/Tests/Unit/ProtobufUtilsTests.cs
andystaples: Yes — `DurableTaskOptions.RollbackEntityOperationsOnExceptions` defaults to `true`. I split this into explicitly named default-enabled and configured-disabled tests so that behavior is clear without nullable theory data.

src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteEntityContext.cs
```diff
@@ -13,23 +13,25 @@ internal class RemoteEntityContext
...
-            if (options.ExtendedSessionsEnabled)
-            {
-                this.Configurations = new RemoteInstanceConfiguration
-                {
-                    IsExtendedSession = isExtendedSession,
-                    IncludeState = includeEntityState,
-                    ExtendedSessionIdleTimeoutInSeconds = options.ExtendedSessionIdleTimeoutInSeconds,
-                };
+            this.Request = batchRequest;
+            this.Configurations = new RemoteInstanceConfiguration
```
YunchuWang: **F0 [High] - constructing `Configurations` unconditionally changes what every non-extended-session entity batch sends to the worker.**

On `dev` this object only existed when `ExtendedSessionsEnabled == true`, so `ToEntityBatchRequest` called `ConvertPocoToProtoMap(null)`, which returns an **empty map**. With `Configurations` now always non-null, `ConvertPocoToProtoMap` reflects over every non-public instance property of `RemoteInstanceConfiguration`.

I ran the reflection against a copy of the type to be sure rather than assuming. For the default configuration (`ExtendedSessionsEnabled = false`), `EntityBatchRequest.Properties` goes from empty to:

```
HttpDefaultAsyncRequestSleepTimeMilliseconds = 30000
IncludeState                                 = True
IsExtendedSession                            = False
ExtendedSessionIdleTimeoutInSeconds          = 0      <-- newly present, and 0
RollbackEntityOperationsOnExceptions         = True
```

That is four extra keys beyond the one this PR intends to add, on the **default** configuration, i.e. for essentially every entity app that does not opt into extended sessions.

`ExtendedSessionIdleTimeoutInSeconds = 0` is the one I would want checked: a worker doing `properties.get("ExtendedSessionIdleTimeoutInSeconds", <default>)` previously fell through to its own default because the key was absent, and now reads a literal `0`. `IncludeState = true` does preserve today's `EntityState` behaviour - `configurations?.IncludeState == false` evaluated to `false` both when `configurations` was null and now that it defaults to `true` - so that one is safe.

Worth noting the previous revision of this PR did not have this problem: threading the flag through `ToEntityBatchRequest` as its own argument kept the change to exactly one new key. The tidier-looking refactor is the riskier one here.

Suggestions, in order of preference:

1. Go back to passing the flag explicitly and leave `Configurations` null when extended sessions are off.
2. Have `ConvertPocoToProtoMap` skip properties still at their default value - but that also affects the orchestrator path, so it needs its own consideration.
3. Keep this shape, confirm with each SDK owner that the four extra keys are inert, and add a test pinning the **exact** key set for `ExtendedSessionsEnabled = false` so it cannot drift silently again.

In all three cases I would like a test that asserts the full `Properties` key set on both branches - that map is the wire contract, and nothing covers it today.

YunchuWang (commented):
Re-reviewed at head `1eb0f7c` (after "Address gRPC settings review feedback"). CI is green, and dropping the extra `ToEntityBatchRequest` parameter in favour of putting the flag on `RemoteInstanceConfiguration` does read more cleanly.

**However, I think that refactor introduced a wider behavioural change than intended, and it is not covered by any test. Details inline on `RemoteEntityContext.cs`; summarised below.**

### F0 [High] - `Configurations` is now always constructed, which expands the entity `Properties` map from 0 keys to 5 for every app that does not enable extended sessions

`RemoteInstanceConfiguration` used to be created *only* when `ExtendedSessionsEnabled == true`. It is now created unconditionally so it can carry the new flag. But `ProtobufUtils.ToEntityBatchRequest` does:

```csharp
batchRequest.Properties.Add(ProtobufUtils.ConvertPocoToProtoMap(configurations));
```

and `ConvertPocoToProtoMap` reflects over **every** non-public instance property, with `ConvertPocoToProtoMap(null)` returning an empty map. So for the default configuration (`ExtendedSessionsEnabled = false`):

| | `Properties` sent to the worker |
|---|---|
| `dev` | *(empty)* |
| this PR | `HttpDefaultAsyncRequestSleepTimeMilliseconds=30000`, `IncludeState=true`, `IsExtendedSession=false`, `ExtendedSessionIdleTimeoutInSeconds=0`, `RollbackEntityOperationsOnExceptions=true` |

I confirmed the reflection output empirically against a copy of the type rather than inferring it.

`ExtendedSessionIdleTimeoutInSeconds = 0` is the one that worries me: a worker doing `properties.get("ExtendedSessionIdleTimeoutInSeconds", <default>)` previously fell through to its default because the key was absent, and now reads a literal `0`. The remaining keys look benign, and `IncludeState = true` preserves the existing `EntityState` behaviour (`configurations?.IncludeState == false` evaluated to `false` in both the old null case and the new default case) - but "looks benign" is doing a lot of work here for a change that ships to every entity app on the default configuration.

Worth noting that the **previous** revision of this PR did not have this problem: threading the flag through as its own argument kept the blast radius to exactly one new key. This is a case where the tidier-looking refactor is the riskier one.

Options, roughly in order of preference:

1. Go back to passing the flag explicitly and leave `Configurations` null when extended sessions are off.
2. Keep the current shape, but have `ConvertPocoToProtoMap` skip properties still at their default value - though that changes behaviour for the orchestrator path too, so it needs its own care.
3. Keep the current shape and confirm with each SDK owner that the four newly-present keys are inert, then **add a test that pins the exact key set** for `ExtendedSessionsEnabled = false`, so this cannot drift again silently.

Either way I would like a test that asserts the full `Properties` key set on both branches, since that is the actual wire contract and nothing currently covers it.

### F1 [Medium] - the entity change is unscoped relative to the linked issue, and has no consumer

`Fixes #3503`, but #3503 is exclusively about `useForwardedHost`. `RollbackEntityOperationsOnExceptions` is a separate change with no tracking issue and no design note.

The description also undersells what it fixes. In middleware-passthrough mode the worker has never received this flag, so **`rollbackEntityOperationsOnExceptions` in host.json has effectively been a no-op for out-of-process entity apps**. That is a real user-facing bug fix and deserves to be stated.

I grepped `durabletask-python`, `durabletask-js`, `azure-functions-durable-js` and `durabletask-dotnet`: **neither key has a consumer today.** For `useForwardedHost` that is expected and intentional, since #3503 explicitly states the host side has to land first. For the entity flag there is no equivalent record.

Concretely: in `durabletask-dotnet`, `TaskEntityShim.ExecuteOperationBatchAsync` rolls back **unconditionally** - the `catch` block calls `state.Rollback()` and `context.Rollback()` with no flag check:

```csharp
catch (Exception applicationException)
{
    ...
    // the user code threw an unhandled exception, so we roll back the state and the actions.
    this.state.Rollback();
    this.context.Rollback();
}
```

So `rollbackEntityOperationsOnExceptions: false` is ignored by the flagship out-of-process SDK today, and will still be ignored after this merges, because nothing reads the new key. Fine as a first step, but it means this half of the PR ships a wire key with no behavioural effect - which is exactly why I would like the consuming work linked.

One design note for whoever implements the worker side: `StateShim.SetState` serializes eagerly on every call, so merely skipping `Commit()` / `Rollback()` will **not** deliver the serialization saving that the `RollbackEntityOperationsOnExceptions` doc comment advertises ("if false, the entity state is serialized only after an entire batch of operations completes"). Realizing that needs deferred serialization too. Worth deciding up front whether this flag is a correctness knob, a performance knob, or both.

Questions:

1. Has the wire key `"RollbackEntityOperationsOnExceptions"` been agreed with the SDK owners? Once released this is effectively permanent.
2. Which SDKs will honour it, and is there a PR or issue to link?
3. Should this be split from the `useForwardedHost` change? The two halves have different risk surfaces and different release-coordination needs - F0 is a good illustration of that.

### F6 [Low] - `release_notes.md` is not updated

Both changes are user-visible, and the entity one especially: a host.json switch that previously did nothing for out-of-process apps starts taking effect, which can change entity failure semantics for existing users on upgrade. Please add entries under `Microsoft.Azure.WebJobs.Extensions.DurableTask` -> `New Features`.

### What is done well

- `bool?` plus `NullValueHandling.Ignore` keeps the legacy client payload byte-identical, and `TestBindingHelper_OmitsUseForwardedHostFromLegacyPayload` is a genuinely useful guard rather than filler.
- Omitting `useForwardedHost` from the legacy payload is correct: legacy SDKs consume `managementUrls` / `baseUrl` that `HttpApiHandler` already produced host-side, so the setting has been applied by the time the worker sees it. **Please put that reasoning in the PR description** - as written it reads like a possible oversight.
- Reusing the existing `Properties` map rather than changing the `.proto` remains the right extension mechanism; F0 is about the map's *contents* growing, not about the mechanism.

### Suggested order

1. **Must**: resolve F0 - either narrow it back down or pin the key set with a test; update `release_notes.md` (F6).
2. **Should**: expand the PR description per F1; assert the literal wire key rather than `nameof` (inline on `RemoteInstanceConfiguration.cs`).
3. **Optional**: revert the whitespace-only churn (inline); null-safe `HttpSettings` (inline); collapse the duplicated `[Fact]`s and close the coverage gaps (inline).
4. **Needs an answer**: the three questions under F1.

src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteEntityContext.cs
```diff
@@ -13,23 +13,25 @@ internal class RemoteEntityContext
...
-            this.Request = batchRequest;
-            if (options.ExtendedSessionsEnabled)
-            {
-                this.Configurations = new RemoteInstanceConfiguration
-                {
-                    IsExtendedSession = isExtendedSession,
-                    IncludeState = includeEntityState,
-                    ExtendedSessionIdleTimeoutInSeconds = options.ExtendedSessionIdleTimeoutInSeconds,
-                };
+            this.Request = batchRequest;
```
YunchuWang: **[Low] - this line, and roughly 11 others in this file, changed only because of line endings.**

Comparing the blobs byte for byte:

```
RemoteEntityContext.cs         dev -> CRLF=11 LF=57   PR -> CRLF=0 LF=70
RemoteInstanceConfiguration.cs dev -> CRLF=10 LF=24   PR -> CRLF=0 LF=39
```

Normalizing to LF is fine in itself, but it is why `RemoteInstanceConfiguration.cs` shows `+15/-10` for what is really a five-line addition, and it makes the substantive change in this constructor harder to spot during review. Since the repo has no `.gitattributes`, the next Windows editor to touch these files will flip some of it back.

Suggest either reverting the whitespace-only lines, or landing a separate `.gitattributes` (`* text=auto`) that fixes this repo-wide.

src/WebJobs.Extensions.DurableTask/RemoteInstanceConfiguration.cs
```diff
@@ -14,21 +14,26 @@ internal class RemoteInstanceConfiguration
...
-        /// Gets or sets the amount of time in seconds before an idle extended session times out.
+        /// <summary>
+        /// Gets or sets the amount of time in seconds before an idle extended session times out.
         /// </summary>
         internal int ExtendedSessionIdleTimeoutInSeconds { get; set; }
+
+        /// <summary>
+        /// Gets or sets whether an uncaught exception should roll back the effects of an entity operation.
+        /// </summary>
+        internal bool RollbackEntityOperationsOnExceptions { get; set; } = true;
```
YunchuWang: **[Medium] - the wire key is generated by reflection, so this property name *is* the protocol; please pin it.**

`ConvertPocoToProtoMap` emits `property.Name` verbatim, so renaming this property silently changes the gRPC contract with every language worker. There is no compile-time or review-time signal that would catch it.

The test in `ProtobufUtilsTests` does not close the gap either - it asserts using `nameof(DurableTaskOptions.RollbackEntityOperationsOnExceptions)`, which is a *different* identifier that merely happens to spell the same thing today. Renaming this property would change the wire format, and renaming the `DurableTaskOptions` one would break the test while the wire stayed correct. Neither failure mode maps to what you actually care about.

Suggest declaring an `internal const string` for the key and asserting the **literal string** in the test, so the wire name is pinned independently of any C# identifier.

Also worth settling while you are here: the orchestrator path adds its extra key as a camelCase literal (`Properties.Add("defaultVersion", ...)` in `OrchestrationTriggerAttributeBindingProvider.cs`), so the same `Properties` concept now carries both `IncludeState` (Pascal) and `defaultVersion` (camel). SDK implementers will trip over that; the intended convention should be documented somewhere.

src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
```diff
@@ -52,6 +52,7 @@ public string DurableOrchestrationClientToString(IDurableOrchestrationClient cli
                     RpcBaseUrl = localRpcAddress,
                     RequiredQueryStringParameters = this.config.HttpApiHandler.GetUniversalQueryStrings(),
                     HttpBaseUrl = this.config.HttpApiHandler.GetBaseUrl(),
+                    UseForwardedHost = this.config.Options.HttpSettings.UseForwardedHost,
```
YunchuWang: **[Low] - `HttpSettings` is dereferenced without a null check, inconsistent with this type's own precedent.**

`DurableTaskOptions.HttpSettings` has a public setter, and `DurableTaskOptions` guards it defensively elsewhere in the same file:

```csharp
{ nameof(this.HttpSettings), this.HttpSettings != null ? JObject.FromObject(this.HttpSettings) : null },
```

`HttpApiHandler` also dereferences it unguarded, so the pattern is pre-existing - but that path only runs when building a management payload, whereas this one runs on **every** durable-client binding invocation. `?.UseForwardedHost ?? false` costs nothing and removes the asymmetry.

test/FunctionsV2/Tests/Unit/ProtobufUtilsTests.cs
```diff
@@ -108,6 +109,58 @@ public void ToEntityBatchResult_PassesDefaultVersionToActions()
...
+        public void ToEntityBatchRequest_IncludesDefaultRollbackEntityOperationsOnExceptions()
+        {
+            var request = new EntityBatchRequest
+            {
+                InstanceId = "@TestEntity@test-key",
+                Operations = new List<OperationRequest>(),
+            };
+            var options = new DurableTaskOptions
+            {
+                ExtendedSessionsEnabled = false,
```
YunchuWang: **[Nit] - both new tests hardcode `ExtendedSessionsEnabled = false`, so the branch that actually changed is untested.**

The constructor now branches on this flag, and the `true` path is where `IsExtendedSession`, `IncludeState` and `ExtendedSessionIdleTimeoutInSeconds` get overwritten after `Configurations` is built. Nothing exercises it.

More importantly, neither test asserts the **key set** of `Properties` - only that one key has the expected value. That is what let F0 through: the map silently went from empty to five keys on this exact code path and both tests still pass.

Suggest adding coverage for `ExtendedSessionsEnabled = true`, and asserting the exact expected key set on both branches.

test/FunctionsV2/Tests/Unit/ProtobufUtilsTests.cs
```diff
@@ -108,6 +109,58 @@ public void ToEntityBatchResult_PassesDefaultVersionToActions()
             Assert.Equal(defaultVersion, action.Version);
         }
 
+        [Fact]
```
YunchuWang: **[Nit]** These two `[Fact]`s are identical apart from one option value and one assertion. A single `[Theory]` with `[InlineData(true)]` / `[InlineData(false)]` would express the same coverage without the duplication - and would make it obvious at a glance that the `ExtendedSessionsEnabled = true` case is missing.

test/FunctionsV2/Tests/Unit/LocalGrpcListenerTests.cs
```diff
@@ -833,6 +834,59 @@ public void TestBindingHelper_ThrowsGrpcChannelUnavailableException_WhenGrpcAddr
             Assert.Contains("transient", ex.Message);
         }
 
+        [Theory]
+        [InlineData(null, false)]
```
YunchuWang: **[Nit] - coverage asymmetry between the two files.**

This theory covers `null -> false` and `true -> true`, but never an explicitly configured `false`. The tests in `ProtobufUtilsTests` have the mirror-image gap: default `true` and explicit `false`, but never an explicit `true`.

Behaviourally the missing cases match the defaults today, so this is low value on its own - but they pin intent and would catch an accidental default flip.

src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteEntityContext.cs
andystaples: Agreed. I restored `Configurations` to its extended-sessions-only behavior and pass rollback separately, so a non-extended entity batch now gains exactly one property; the tests pin the complete key set for both branches.

src/WebJobs.Extensions.DurableTask/ContextImplementations/RemoteEntityContext.cs
andystaples: Cleaned up. `RemoteInstanceConfiguration.cs` no longer appears in the PR diff, and `RemoteEntityContext.cs` retains its existing mixed endings so only the four substantive added lines remain.

src/WebJobs.Extensions.DurableTask/RemoteInstanceConfiguration.cs
andystaples: Pinned the wire key as the literal `RollbackEntityOperationsOnExceptions` behind an internal constant, and the test asserts that literal independently. I kept the existing PascalCase convention used by configuration properties rather than changing the unrelated, pre-existing `defaultVersion` contract.

src/WebJobs.Extensions.DurableTask/Bindings/BindingHelper.cs
andystaples: Updated this to `HttpSettings?.UseForwardedHost ?? false` and added coverage for a null `HttpSettings` object.

test/FunctionsV2/Tests/Unit/ProtobufUtilsTests.cs
andystaples: The consolidated theory now covers extended sessions enabled and disabled and asserts the complete, exact property-key set for each branch.

test/FunctionsV2/Tests/Unit/ProtobufUtilsTests.cs
andystaples: Consolidated the rollback cases into one theory, including both extended-session branches and all relevant configured values.

test/FunctionsV2/Tests/Unit/LocalGrpcListenerTests.cs
andystaples: Added explicit `true` and `false` cases for both settings, while retaining the default-value coverage.

YunchuWang (commented):
Re-reviewed at `d37f736e`. All seven items from my earlier review are addressed. I verified each against the diff rather than the replies -- here is what I confirmed:

**F0 (High) -- resolved, and the fix is better than what I suggested.** Instead of keeping the setting inside `Configurations`, you moved it to a separate `bool?` parameter on `ToEntityBatchRequest` that is only written when `HasValue`, and restored the `ExtendedSessionsEnabled` guard on `Configurations`. That keeps the value entirely out of the reflection-based `ConvertPocoToProtoMap` path. The net effect on the wire for a default-configuration app is exactly one new key instead of five, and `ExtendedSessionIdleTimeoutInSeconds = 0` is no longer leaked to workers that never asked for extended sessions.

**F1 (Medium) -- resolved.** `RollbackEntityOperationsOnExceptionsPropertyName` as an `internal const` means renaming the C# property can no longer silently change the wire contract.

**Tests -- this is the part worth highlighting.** `Assert.Equal(expectedKeys, result.Properties.Keys.OrderBy(...))` pins the *complete* key set for both branches. That is exactly the right guard: any future property added to `RemoteInstanceConfiguration` now fails a test instead of silently widening the payload. This is what would have caught F0.

**Line endings -- resolved.** `RemoteInstanceConfiguration.cs` is out of the diff entirely, and `RemoteEntityContext.cs` is now a clean `+4/-0`.

The source diff is down to `+26/-2` with no incidental churn, and CI is green across all language workers on both platforms.

On compatibility of the new key, I will retract my earlier caution: apps with extended sessions enabled have always received a populated `Properties` map on `dev`, so any worker that supports extended sessions already tolerates non-empty maps. This change only routes default-configuration apps down that same path; it does not introduce a new shape. Combined with the green node/python/java/powershell/dotnet-isolated runs, I consider that risk adequately covered.

---

**One forward-looking item I would like resolved before this merges**, because it is much cheaper to decide now than after the fact.

`durabletask-dotnet` currently rolls back **unconditionally** -- `TaskEntityShim.ExecuteOperationBatchAsync` calls `Rollback()` in its catch block with no configuration check. So an app that sets `RollbackEntityOperationsOnExceptions: false` in host.json today is, in practice, still getting rollback protection in the out-of-process worker.

Once a worker-side change starts honoring this key, those apps flip to non-rollback behavior on a package upgrade, with no code change and no warning on their side. That is the failure mode where a failed operation's partial mutations stay on the live state object and feed into the *next* operation in the same batch. Because batch size is 1 under low load, it tends to reproduce only under production traffic.

So the SDK-side follow-up deserves an explicit decision rather than an implicit one:

1. Honor the flag, and treat it as a documented behavior change with release notes; or
2. Ignore it, mark `DurableTaskOptions.RollbackEntityOperationsOnExceptions` `[Obsolete]`, and keep rollback unconditional everywhere.

I lean toward (2). The only documented benefit of `false` (from #1231) is skipping per-operation state serialization, and it carries a much larger blast radius on serialization failure: a batch-end writeback failure converts *every* successful operation's result into an error, versus failing a single operation under `true`. It is also undocumented on Microsoft Learn and was never carried forward into `durabletask-dotnet`.

Could you link the worker-side tracking issue(s) here so that decision has a home? Same for `useForwardedHost`, which is the half that actually closes #3503 -- this PR is inert until something consumes it.

Minor, non-blocking: `release_notes.md` still is not updated.

Code quality is approval-ready from my side. Leaving this as a comment rather than an approval only so the question above gets an answer first.

bachuv (approved):

YunchuWang (commented):
Sharpening my previous comment with code references, because I think it materially changes the calculus on the SDK-side question.

I framed the concern as a silent behavior flip for apps that already set `false`. That is the weaker argument -- honoring an explicit configuration is defensible on its own. Here is the stronger, purely technical point:

**In `durabletask-dotnet`, honoring `false` would save exactly zero serializations.**

The in-process and out-of-process shims have fundamentally different state architectures, and the performance rationale for this setting only holds for one of them.

*In-process* (`DurableEntityContext`): `CurrentState` is a live object and serialization happens inside `TryWriteback()`. With rollback enabled, `TryWriteback` runs after every operation (`Listener/TaskEntityShim.cs:508`); with it disabled, it runs once per batch (`Listener/TaskEntityShim.cs:255`). So `false` genuinely collapses N serializations into 1. The rationale from #1231 is real here.

*Out-of-process* (`durabletask-dotnet`, `src/Worker/Core/Shims/TaskEntityShim.cs`): `StateShim.SetState` serializes eagerly on every call:

```csharp
public override void SetState(object? state)
{
    this.value = this.dataConverter.Serialize(state);
    this.cachedValue = state;
}
```

Rollback support adds nothing on top of that. `Commit()` is `checkpointValue = value` (line 147) and `Rollback()` is `CurrentState = checkpointValue` (line 152) -- both plain string reference assignments. The serialization count is driven entirely by how many times user code calls `SetState`, and is identical whether rollback is on or off.

So for the out-of-process SDKs, implementing `false` would deliver none of its only documented benefit, while introducing the dirty-state-propagation failure mode (a failed operation's partial mutations remaining on the live state object and feeding into the next operation in the same batch) plus the batch-wide blast radius on writeback failure. Across five language workers that is a pure-loss change.

That is why I lean strongly toward option (2) from my previous comment: mark `DurableTaskOptions.RollbackEntityOperationsOnExceptions` `[Obsolete]` and keep rollback unconditional out-of-process, rather than asking five SDK teams to implement a mode that cannot deliver its stated benefit and that nobody should enable.

If that is the direction, the `RollbackEntityOperationsOnExceptions` plumbing here is arguably unnecessary. It is harmless while no worker consumes the key, and it does preserve the option, so I am not asking for its removal -- but it should be a deliberate choice rather than a default.

To be explicit about what I am *not* blocking on: `useForwardedHost` stands entirely on its own merits, and it is the half that actually addresses #3503. I would not want it held up by the rollback discussion. If the rollback question needs more time, splitting the two settings into separate PRs would let the security fix land immediately.

PR #3512, opened 2026-08-06
Add Entra authentication to Durable distributed tracing

### LINKED ISSUE #3497: Distributed Tracing V2 emits no telemetry when Application Insights local auth is disabled (regression of #2870, fix reverted by #3053)
### Description

Durable Functions Distributed Tracing V2 silently emits no telemetry when the Application Insights resource has local authentication disabled (`DisableLocalAuth: true`) and the Function App authenticates via Microsoft Entra using `APPLICATIONINSIGHTS_AUTHENTICATION_STRING`.

This is a regression of #2870. That issue was closed as completed by #3009 ("Replace ITelemetryActivator with ITelemetryModule"), which was then reverted three weeks later by #3053 ("Revert ITelemetryModule updates") because that change caused missing spans. #2870 was never reopened, so the original defect is still present on `dev` today.

New information versus #2870: that issue was labeled `dotnet-isolated`, but this reproduces on a Python app, which suggests the defect is language-independent and lives entirely in the host-side extension.

### Expected behavior

With `distributedTracingEnabled: true` and `version: "V2"` in host.json and a valid `APPLICATIONINSIGHTS_CONNECTION_STRING`, orchestration and activity spans appear in Application Insights as `Request` / `Dependency` rows prefixed with `orchestration:` / `activity:`, regardless of whether the component accepts local authentication or requires Entra authentication.

### Actual behavior

Zero `orchestration:` / `activity:`-prefixed rows are ever ingested. All other telemetry from the same app is unaffected: host traces, and request telemetry for HTTP, timer, queue and entity triggers all arrive normally, because the Functions host applies the Entra credential to its own telemetry pipeline.

The V2 code path is confirmed active: `TelemetryActivator.Initialize` only emits the "Distributed Tracing V2 is GA now!" announcement when tracing is disabled or set to V1, and that warning is absent from the workspace, so the host.json block binds correctly.

`src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs` builds its own `TelemetryConfiguration` and offers no way to supply a token credential:

```csharp
private TelemetryConfiguration SetupTelemetryConfiguration()
{
    TelemetryConfiguration config = TelemetryConfiguration.CreateDefault();
    // ...
    string resolvedInstrumentationKey = this.nameResolver.Resolve("APPINSIGHTS_INSTRUMENTATIONKEY");
    string resolvedConnectionString = this.nameResolver.Resolve("APPLICATIONINSIGHTS_CONNECTION_STRING");
    // ...
    config.ConnectionString = resolvedConnectionString;   // no SetAzureTokenCredential(...)
}
```

`APPLICATIONINSIGHTS_AUTHENTICATION_STRING` does not appear anywhere in this repository, so this configuration cannot authenticate against a component with `DisableLocalAuth: true`. The equivalent handling in the WebJobs SDK is at `ApplicationInsightsServiceCollectionExtensions.cs` (linked in #2870).

### Known workarounds

1. Re-enable local authentication on the Application Insights component. This defeats the security posture that motivated disabling it, so it is not viable where Entra-only ingestion is a compliance requirement.
2. Accept the loss of distributed tracing and derive orchestration and activity visibility from host traces instead, querying `AppTraces` for the `Executed 'Functions.<name>'` records rather than `AppRequests`.

Neither the docs page (`durable-functions-diagnostics#distributed-tracing`) nor `samples/distributed-tracing/v2/DistributedTracingSample/README.md` mentions this limitation; both document only connection string or instrumentation key. A note there would save others the investigation.

### App Details

- **Durable Functions extension version**: `Microsoft.Azure.WebJobs.Extensions.DurableTask` 3.14.0, via extension bundle 4.38.0 (host.json range `[4.30.0, 5.0.0)`)
- **Azure Functions runtime version**: 4.x
- **Programming language used**: Python 3.13
- **Hosting plan**: Flex Consumption
- **Storage provider**: Azure Storage (default)

host.json (relevant parts):

```json
{
  "version": "2.0",
  "extensionBundle": {
    "id": "Microsoft.Azure.Functions.ExtensionBundle",
    "version": "[4.30.0, 5.0.0)"
  },
  "extensions": {
    "durableTask": {
      "tracing": {
        "distributedTracingEnabled": true,
        "version": "V2"
      }
    }
  },
  "logging": {
    "logLevel": { "Host.Results": "Information" },
    "applicationInsights": {
      "samplingSettings": { "isEnabled": true, "excludedTypes": "Request; Exception" }
    }
  }
}
```

App settings:

```
APPLICATIONINSIGHTS_CONNECTION_STRING     = InstrumentationKey=...;IngestionEndpoint=...
APPLICATIONINSIGHTS_AUTHENTICATION_STRING = Authorization=AAD;ClientId=<user-assigned identity>
```

Application Insights component: workspace-based, `DisableLocalAuth: true`. `APPINSIGHTS_INSTRUMENTATIONKEY` is not set, so the extension's "both keys defined" and "neither key defined" warnings do not fire.

### Minimal repro

1. Create a Python Durable Functions app with one orchestrator and one activity.
2. Point it at a workspace-based Application Insights component and set `DisableLocalAuth: true` on that component.
3. Grant the app's identity **Monitoring Metrics Publisher** and set `APPLICATIONINSIGHTS_AUTHENTICATION_STRING` accordingly.
4. Enable DT V2 in host.json as above; run a few orchestrations.
5. Query: `union AppRequests, AppDependencies | where Name has ":"` - no `orchestration:` or `activity:` rows, while `AppTraces` shows the orchestrator and activity executing and non-durable triggers still produce `AppRequests`.

Setting `DisableLocalAuth: false` and restarting is the expected control case.

### If deployed to Azure

- **Timeframe issue observed**: continuously; verified over a 30-day window ending 2026-07-28
- **Azure region**: West Europe
- **Function name(s)**: `orchestrator_process_task` (orchestrator trigger), plus its activity triggers
- **Orchestration instance IDs** (both completed successfully):

  | Instance ID | Orchestrator | Created (UTC) |
  | --- | --- | --- |
  | `stage3-batch-0efdd02c-cb4c-40c7-8842-111349931e96` | `orchestrator_process_task` | 2026-07-27T11:03:39Z |
  | `03ea08e1effc459f963792e01efc3b92` | `orchestrator_process_task` | 2026-07-26T11:21:22Z |

- **Worked example** - instance `stage3-batch-0efdd02c-cb4c-40c7-8842-111349931e96`:

  ```kusto
  // 7 records: the orchestration demonstrably ran
  AppTraces
  | where TimeGenerated between (datetime(2026-07-27) .. datetime(2026-07-28))
  | where Message has "stage3-batch-0efdd02c-cb4c-40c7-8842-111349931e96"

  // 0 records: no Distributed Tracing V2 spans were ever ingested
  union AppRequests, AppDependencies
  | where TimeGenerated between (datetime(2026-07-27) .. datetime(2026-07-28))
  | where Name has "stage3-batch-0efdd02c-cb4c-40c7-8842-111349931e96"
     or OperationName has "stage3-batch-0efdd02c-cb4c-40c7-8842-111349931e96"
  ```

  The second instance shows the same shape: 8 `AppTraces` records between 2026-07-26T11:21:23Z and 23:21:57Z, and no `AppRequests` / `AppDependencies` rows at all.

  Across a 30-day window this holds for every orchestration and activity in three independent environments: `AppTraces` shows them executing, and `AppRequests` / `AppDependencies` contain no `orchestration:` or `activity:` prefixed rows at all. Entity, queue, timer and HTTP triggers in the same apps produce `AppRequests` normally throughout.

- **Function App name / storage account name**: not shared publicly. The orchestration instance IDs above should be resolvable on your side; happy to provide more detail through a private channel if needed.

YunchuWang:
## Summary

- apply `APPLICATIONINSIGHTS_AUTHENTICATION_STRING` to Durable's private Application Insights pipeline
- support both system-assigned and user-assigned managed identities
- forward Durable telemetry to the Functions host's Application Insights channel, which already carries the host's Entra credential, so Durable never creates or reads an `Azure.Core` type
- preserve Durable's processor-free telemetry configuration, V1/V2 behavior, test hooks, and lifecycle ownership
- log only the selected identity type; configured IDs and authentication-string contents are never logged

Fixes #3497

## Root cause

Durable distributed tracing creates a private `TelemetryConfiguration`. It copied the Application Insights instrumentation key or connection string, but not the Microsoft Entra credential configured by the Functions host. With `DisableLocalAuth=true`, host telemetry stayed authenticated while Durable Request/Dependency telemetry was rejected at ingestion.

The obvious fix — construct a `ManagedIdentityCredential` and call `SetAzureTokenCredential` on Durable's configuration — does not work inside the Functions host. Application Insights 2.22 resolves `Azure.Core.TokenCredential` through `ReflectionCredentialEnvelope`, which binds the type in the host/default load context. A credential constructed in the function app load context is a different type identity and is rejected with `ArgumentException`. This was confirmed on a live Function App, not inferred.

## Implementation

Durable does not obtain a credential at all. It wraps the host's `ITelemetryChannel` and forwards its own telemetry through it. The host channel already holds the credential envelope and the Entra ingestion endpoint, so the authentication problem disappears without Durable ever touching an `Azure.Core` type — no reflection, and no coupling to the host's load context or to Application Insights internals.

- parse the documented authentication-string forms using the WebJobs Application Insights parser:
  - system-assigned: `Authorization=AAD`
  - user-assigned: `Authorization=AAD;ClientId=<CLIENT_ID>`
- add a host-aware `TelemetryActivator` constructor while retaining the existing two-argument public constructor
- add `HostForwardingTelemetryChannel`, which forwards `Send`/`Flush` to the host channel
- retain direct `ManagedIdentityCredential` fallback behavior for standalone WebJobs hosts, where there is no host channel to reuse
- emit sanitized warnings for malformed settings or configurations that cannot be applied
- document the required settings and `Monitoring Metrics Publisher` role

Three properties of the Application Insights SDK make the wrapper safe, and they are structural rather than conventions this code has to uphold:

- `ISupportCredentialEnvelope` is `internal`, so an external wrapper cannot implement it. The `TelemetryChannel` setter's call to `SetTelemetryChannelCredentialEnvelope` is therefore a no-op on the wrapper, and Durable's credential-free configuration cannot clear the host's credential.
- `SetTelemetryChannelEndpoint` only writes `EndpointAddress` for `InMemoryChannel` and `ServerTelemetryChannel`. A custom channel is never written to, so configuring Durable's connection string cannot downgrade the host to the non-Entra endpoint.
- `Dispose` is intentionally empty. The host owns the channel's lifetime, and Durable's configuration must not be able to stop host telemetry.

The host `TelemetryConfiguration` itself is still intentionally not reused for telemetry. Host telemetry processors include operation filtering that is incompatible with Durable operations that do not carry the host `Category` property; retaining the private configuration avoids the missing-span behavior that led to the earlier revert in #3053. Only the channel — the transport — is shared.

## Real Azure-hosted validation

All runs used temporary Azure Functions v4 Windows Consumption apps, real Azure Storage, workspace-based Application Insights with `DisableLocalAuth=true`, and real managed identities. Exactly one identity at a time had `Monitoring Metrics Publisher` scoped to the Application Insights component. No application code used `AzureCliCredential`.

### Current head `7a435487` — channel forwarding, .NET 8 in-process

A diagnostic endpoint resolved the production-registered `ITelemetryActivator` from the live host and reported the wiring:

```json
{
  "activatorType": "...DurableTask.Correlation.TelemetryActivator",
  "durableChannelType": "...DurableTask.Correlation.HostForwardingTelemetryChannel",
  "hostChannelType": "Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel.ServerTelemetryChannel",
  "hostChannelAssembly": "Microsoft.AI.ServerTelemetryChannel",
  "wrappedIsHostChannel": true,
  "durableConfigurationIsHostConfiguration": false
}
```

The wrapper holds the host's real `ServerTelemetryChannel` instance, and Durable's configuration is still its own — so the isolated processor/module pipeline that #3053 depended on is intact.

| Identity | Orchestration | Durable Requests | Durable Dependencies | Result |
| --- | --- | ---: | ---: | --- |
| User-assigned | Completed | 2 | 2 | One operation ID; full parent chain |
| System-assigned | Completed | 2 | 2 | One operation ID; full parent chain |

Ingested span chain, identical in both runs (user-assigned shown):

```text
request     Issue3497Start                     <- host control telemetry
dependency  create_orchestration:Orchestrator  <- parent: Issue3497Start
request     orchestration:Orchestrator         <- parent: create_orchestration
dependency  activity:Activity                  <- parent: orchestration
request     activity:Activity                  <- parent: activity dependency
request     Issue3497Status                    <- host control telemetry
```

Initialization logged the expected identity type for each run, and the host-channel-unavailable warning never fired. Ordinary host telemetry continued to flow throughout, confirming the shared channel was not disturbed.

### Earlier heads

The same harness previously reproduced the bug and validated the credential-reuse approach. Those results still establish the failure mode and the breadth of the fix.

| Phase | Runtime | Identity | Durable Requests | Durable Dependencies | Result |
| --- | --- | --- | ---: | ---: | --- |
| Official 3.14.0 | .NET 8 in-process | UAMI | 0 | 0 | Reproduced #3497 |
| Official 3.14.0 | .NET 8 isolated | UAMI | 0 | 0 | Host/control telemetry and MI token acquisition succeeded |
| Patched | .NET 8 in-process | UAMI / System | 2 | 2 | Full parent chain |
| Patched host DLL | .NET 8 isolated | UAMI / System | 2 | 2 | Full parent chain |
| Patched host DLL | Node.js out-of-process | System | 2 | 2 | Full parent chain |

The isolated and Node.js runs confirm the fix applies to host-generated Durable spans independently of the language worker. Worker-direct Application Insights/OpenTelemetry remains a separate pipeline. Normal non-.NET rollout still requires a Durable extension release followed by an extension-bundle release.

## Automated validation

- `TelemetryActivatorTests`: **10/10** on net8.0 and net10.0
- `CorrelationEndToEndTests`: **37/37** on net8.0 and net10.0
- Release build/package: **0 warnings, 0 errors**
- deployed host DLL hashes matched their local publish payloads exactly

New coverage for this design:

- forwarding is selected when the host exposes a configuration, and the wrapper holds the host's own channel
- telemetry tracked on Durable's configuration physically arrives on the host channel
- the host's credential survives Durable's wiring, and Durable's configuration still has none
- configuring Durable's connection string does not overwrite the host channel's endpoint
- disposing Durable's configuration does not dispose the host channel, which keeps sending afterwards
- the `OnSend` test channel is preserved rather than replaced
- standalone hosts still fall back to a managed identity credential, for both identity forms
- end to end: with the activator built purely by production DI, all four Durable spans arrive through the forwarding channel with the correct parent chain

## Cleanup and security

- all temporary resource groups were deleted
- ARM/subscription checks confirmed no matching Function App, storage account, managed identity, workspace, Application Insights component, or scoped role assignment remained
- deployment archives, publishing material, binaries, and transient credentials were removed
- retained evidence and source artifacts passed secret/resource-identifier scans

## Configuration

The connection string remains required:

```text
APPLICATIONINSIGHTS_CONNECTION_STRING=<connection string>
```

Choose one authentication form:

```text
# System-assigned identity
APPLICATIONINSIGHTS_AUTHENTICATION_STRING=Authorization=AAD

# User-assigned identity
APPLICATIONINSIGHTS_AUTHENTICATION_STRING=Authorization=AAD;ClientId=<client ID>
```

The selected identity needs `Monitoring Metrics Publisher` on the Application Insights component.

Official references:

- [Configure Application Insights for Azure Functions](https://learn.microsoft.com/azure/azure-functions/configure-monitoring#require-microsoft-entra-authentication)
- [`APPLICATIONINSIGHTS_AUTHENTICATION_STRING`](https://learn.microsoft.com/azure/azure-functions/functions-app-settings#applicationinsights_authentication_string)
- [Microsoft Entra authentication for Application Insights](https://learn.microsoft.com/azure/azure-monitor/app/azure-ad-authentication)
- [Upgrade the Durable Functions extension](https://learn.microsoft.com/azure/durable-task/durable-functions/durable-functions-extension-upgrade)


Directory.Packages.props
```diff
@@ -25,6 +25,7 @@
     <PackageVersion Include="Microsoft.Azure.WebJobs" Version="3.0.45" />
     <PackageVersion Include="Microsoft.Azure.WebJobs.Extensions.DurableTask.Analyzers" Version="0.5.0" />
     <PackageVersion Include="Microsoft.Azure.WebJobs.Extensions.Rpc" Version="3.0.39" />
+    <PackageVersion Include="Microsoft.Azure.WebJobs.Logging.ApplicationInsights" Version="3.0.41" />
```
bachuv: Microsoft.Azure.WebJobs.Logging.ApplicationInsights 3.0.41 transitively resolves System.Drawing.Common 4.7.0, which is affected by Critical advisory GHSA-rxg9-xrhp-64gj on both target frameworks. Could we use 3.0.45 instead? It aligns with the existing WebJobs 3.0.45 dependency and uses the updated Application Insights dependency chain.

test/FunctionsV2/Tests/E2E/CorrelationEndToEndTests.cs
```diff
@@ -124,6 +124,130 @@ public async Task DistributedTracingV2_DefaultOperationNames_ArePopulated()
...
+            };
+
+            var result = await this.ExecuteOrchestrationWithExceptionAsync(
+                orchestrationFunctionNames,
+                "DistributedTracingV2EntraAuthentication",
+                "world",
+                extendedSessions: false,
+                protocol: "W3CTraceContext",
+                version: DurableDistributedTracingVersion.V2,
+                authenticationString: "Authorization=AAD");
```
bachuv: This test uses onSend, which explicitly constructs the two-argument TelemetryActivator and bypasses production DI. Could we adjust it to test the three-argument path and verify that Durable reuses the host credential?

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
```diff
@@ -238,7 +260,95 @@ private TelemetryConfiguration SetupTelemetryConfiguration()
...
+            return credentialProperty?.GetValue(credentialEnvelope);
+        }
+
+        private void LogInvalidAuthenticationString()
+        {
+            this.endToEndTraceHelper.ExtensionWarningEvent(
+                hubName: this.options.HubName,
+                functionName: string.Empty,
+                instanceId: string.Empty,
+                message: "APPLICATIONINSIGHTS_AUTHENTICATION_STRING is invalid and will not be used for Durable distributed tracing.");
```
bachuv:
```suggestion
                message: "APPLICATIONINSIGHTS_AUTHENTICATION_STRING is invalid and will not be used for Durable Functions distributed tracing.");
```

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
```diff
@@ -238,7 +260,95 @@ private TelemetryConfiguration SetupTelemetryConfiguration()
...
+        }
+
+        private void LogAuthenticationStringCouldNotBeApplied()
+        {
+            this.endToEndTraceHelper.ExtensionWarningEvent(
+                hubName: this.options.HubName,
+                functionName: string.Empty,
+                instanceId: string.Empty,
+                message: "APPLICATIONINSIGHTS_AUTHENTICATION_STRING could not be applied and will not be used for Durable distributed tracing.");
```
bachuv:
```suggestion
                message: "APPLICATIONINSIGHTS_AUTHENTICATION_STRING could not be applied and will not be used for Durable Functions distributed tracing.");
```

Directory.Packages.props
YunchuWang: Done - moved to 3.0.45 in 3d58aa68.

I verified the advisory in both directions on `src/WebJobs.Extensions.DurableTask` with `dotnet list package --vulnerable --include-transitive`:

- **3.0.41** -> `System.Drawing.Common 4.7.0`, Critical `GHSA-rxg9-xrhp-64gj`, on both `net8.0` and `net10.0`
- **3.0.45** -> `System.Drawing.Common 6.0.0`, "no vulnerable packages"

So the audit is demonstrably working and the upgrade is what clears it, which matches your read of the dependency chain.

One knock-on worth flagging: 3.0.45's newer Application Insights chain raises the transitive requirement to `System.Drawing.Common >= 6.0.0`, which collided with the 4.7.3 pin in `test/Directory.Packages.props` and failed restore with `NU1109`. I raised that one pin to 6.0.0. It is scoped to test projects, it is strictly higher than the value it replaces, and several test projects already override to 6.0.0/7.0.0. Both projects now audit clean and the full solution builds in Release.

test/FunctionsV2/Tests/E2E/CorrelationEndToEndTests.cs
YunchuWang: Good catch - fixed in 3d58aa68.

The root of it was the helper, so I fixed it there rather than working around it in the test. `PlatformSpecificHelpers.CreateJobHost` no longer hand-builds the activator; it now calls `ActivatorUtilities.CreateInstance<TelemetryActivator>(serviceProvider)` and only attaches the `OnSend` hook, so constructor selection comes from the container exactly like the production `AddSingleton<ITelemetryActivator, TelemetryActivator>()` registration. That applies to every `onSend`-based correlation test, not just this one.

The test now registers a host `TelemetryConfiguration` the way the Functions host does, resolves `ITelemetryActivator` out of the running host's service provider, and asserts:

- the host-aware constructor was selected (the activator's `hostTelemetryConfiguration` is the instance we registered)
- Durable's own telemetry configuration carries the **same credential instance** the host owns (`Assert.Same`) - that identity is the actual root cause of #3497, since an equivalent-but-separate credential is what gets rejected
- the four-span parent chain still holds

It is a `[Theory]` now so the fallback branch is covered as well: with no host credential, Durable builds its own `ManagedIdentityCredential` and emits the warning, and spans are still produced.

I also checked that it genuinely guards the regression rather than passing by construction - stubbing the reuse lookup to return `null` fails both the new `Assert.Same` and the unit-level `Assert.True(reusedHostCredential)`; restoring it returns to green.

Results: `CorrelationEndToEndTests` 37/37 and `TelemetryActivatorTests` 6/6 on both `net8.0` and `net10.0`, and the solution builds clean in Release.

Directory.Packages.props
YunchuWang: Follow-up on this thread: the 3.0.45 upgrade had a knock-on effect worth flagging, now resolved.

3.0.45 raises `System.Configuration.ConfigurationManager` to 6.0.0 in the transitive graph. Combined with `Microsoft.Data.SqlClient 3.1.7`, that makes `System.Security.Permissions -> System.Windows.Extensions -> System.Drawing.Common (>= 6.0.0)` binding, so every remaining `4.7.3` pin became a NU1605/NU1109 downgrade (warnings are errors here). Raised four pins to `6.0.0`, which is what the repo already uses elsewhere via `VersionOverride`, and is still above the vulnerable 4.7.0 those pins originally guarded against:

- `test/Directory.Packages.props`
- `samples/Directory.Packages.props`
- `test/SmokeTests/BackendSmokeTests/MSSQL/MSSQL.csproj`
- `test/e2e/Apps/BasicDotNetIsolated/Directory.Build.props` (the generated `WorkerExtensions.csproj`)

Verified by flipping the package version back and forth rather than assuming: at 3.0.41 all 25 projects that reference the Durable extension restore clean, at 3.0.45 they fail, and after the pin bumps all 25 restore clean again. `dotnet list package --vulnerable --include-transitive` is clean on both the extension and the V2 test project.

CI is now 49 pass / 1 fail. The remaining failure is `Run V4 .NET in-proc w/ MSSQL Smoke Test`, which is unrelated to this PR: it reads `secrets.SA_PASSWORD`, and fork PRs cannot access repository secrets, so the SQL Server container rejects the empty password ("The password must be at least 8 characters"). Every `fork=true` run of that workflow fails and every non-fork run passes, including on `dev`.

test/FunctionsV2/Tests/E2E/CorrelationEndToEndTests.cs
YunchuWang: Update on this thread: the design changed in 7a435487, and this test changed with it — it now uses no `onSend` hook at all.

The reflection that read the host's credential is gone. Instead of copying a credential onto Durable's private configuration, Durable now wraps the host's `ITelemetryChannel` and forwards through it. The host channel already carries the credential envelope and the Entra endpoint, so Durable never creates or reads an `Azure.Core` type.

That made a strictly better test possible. Since `onSend` was the thing that bypassed production DI, and it also happens to be the thing that would suppress forwarding, the new test drops it entirely:

- the host `TelemetryConfiguration` is registered with a capturing channel
- `onSend` is left unset, so `PlatformSpecificHelpers` registers no `ITelemetryActivator` override at all and the activator comes purely from the extension's own `AddSingleton<ITelemetryActivator, TelemetryActivator>()`
- the test asserts the resolved activator holds the injected host configuration, and that its channel is a `HostForwardingTelemetryChannel` wrapping the host's own channel instance
- the four Durable spans are then captured *on the host channel*, which proves the credential-bearing path is the one telemetry actually travels

The `onSend` hook itself is still covered — `ApplyEntraAuthentication_PreservesExistingTestChannel` asserts an existing test channel is never replaced by forwarding — and the standalone fallback keeps its own end-to-end test.

Verified on a live Functions v4 app with `DisableLocalAuth=true`. A probe resolving the production-registered `ITelemetryActivator` reported `wrappedIsHostChannel: true` against the host's real `ServerTelemetryChannel`, and both user-assigned and system-assigned identities ingested 2 Requests + 2 Dependencies with the full parent chain. Details are in the PR description.

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
YunchuWang: Updated the warning wording in commit `fffc980f7a5258bf75678a55c0203e9f16fbcead` and kept the corresponding assertion aligned. `TelemetryActivator_InvalidAuthenticationString_LogsWarning` passes all 3 theory cases on both net8.0 and net10.0.

src/WebJobs.Extensions.DurableTask/Correlation/TelemetryActivator.cs
```diff
@@ -238,7 +264,115 @@ private TelemetryConfiguration SetupTelemetryConfiguration()
...
+            }
+
+            ManagedIdentityId managedIdentityId = tokenCredentialOptions.ClientId != null
+                ? ManagedIdentityId.FromUserAssignedClientId(tokenCredentialOptions.ClientId)
+                : ManagedIdentityId.SystemAssigned;
+            durableConfiguration.SetAzureTokenCredential(new ManagedIdentityCredential(managedIdentityId));
+            return false;
+        }
+
+        private void LogInvalidAuthenticationString()
```
bachuv: The following 3 methods are nearly identical. Can we have a single `private void LogTracingWarning(string message)` helper to remove redundant code?

YunchuWang: Addressed in commit `c04600394b911a69537b10f9910fdff128d61136`: all three wrappers are removed, and the existing call sites now use one `LogTracingWarning(string message)` helper without changing any warning text or trigger conditions. Targeted validation passed 16/16 on net8.0 and 16/16 on net10.0 (full `TelemetryActivatorTests` plus all `TelemetryActivator_InvalidAuthenticationString_LogsWarning` theory cases); 0 failed and 0 skipped on each framework. `git diff --check` also passed.

YunchuWang:
[query_data.csv](https://github.com/user-attachments/files/31274147/query_data.csv)
cc @bachuv for local manual test traces captured from app insights

bachuv (approved):

PR #3513, opened 2026-08-06
Include sub-orchestration instance IDs in history

### LINKED ISSUE #854: Include SubOrchestration InstanceId in HTTP API InstanceHistory
For more complex workflows, there is a need to retrieve the outcome of SubOrchestrations that are executed from a ParentOrchestration. 

Currently the HTTP API will return the status and history of an instance, but it wont include a way to retrieve information about any SubOrchestration instance, well in cases where the ParentOrchestration is in the 'Completed' state I have noticed.

For occasions where the ParentOrchestration is still 'Running', you can see the `InstanceId` of SubOrchestration in the `HistoryEvents` array. My question is, is there a way to more easily retrieve SubOrchestration status with only the ParentOrchestration `InstanceId` or is this a feature that needs to be requested?

YunchuWang:
## Summary
- preserve the child orchestration `InstanceId` when compacting `SubOrchestrationInstanceCreated` into completed or failed history entries
- keep activity history unchanged by only projecting the property when the scheduling event contains an instance ID
- cover completed, failed, and in-flight sub-orchestration history across input/output projection options
- add a .NET isolated E2E test that queries the generated management status URI with `showHistory=true` and verifies the completed sub-orchestration ID

Fixes #854.

## Testing
- `DurableClientHistoryTests` on `net8.0` and `net10.0`
- `GetStatusHistory_CompletedSubOrchestrationIncludesInstanceId` on `net8.0` with Azure Storage/Azurite
- confirmed the E2E test fails on the parent commit with `Actual: null` and passes with this fix

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -1050,7 +1050,7 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private static void TrackNameAndScheduledTime(JObject historyItem, EventType eventType, int index, Dictionary<string, EventIndexDateMapping> eventMapper)
         {
-            eventMapper.Add($"{eventType}_{historyItem["EventId"]}", new EventIndexDateMapping { Index = index, Name = (string)historyItem["Name"], Date = (DateTime)historyItem["Timestamp"], Input = (string)historyItem["Input"] });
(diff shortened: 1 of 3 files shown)


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -1059,6 +1059,11 @@ private static void AddScheduledEventDataAndAggregate(ref Dictionary<string, Eve
             {
                 historyItem["ScheduledTime"] = taskScheduledData.Date;
                 historyItem["FunctionName"] = taskScheduledData.Name;
+                if (taskScheduledData.InstanceId != null)
```
bachuv: Can we change this to check for null or ""?

```suggestion
                if (!string.IsNullOrEmpty(taskScheduledData.InstanceId))
```

Can we also add a test for empty Instance Ids?

kaibocai (approved):

kaibocai (approved):

bachuv (approved):

PR #3514, opened 2026-08-06
Reject starts for disabled orchestrators

### LINKED ISSUE #1933: Unexpected Behavior Starting Disabled Orchestrator Function
Creating the issue on behalf of the Visual Studio customer: Details: https://developercommunity.visualstudio.com/t/Unexpected-Behavior-Starting-Disabled-Or/1416301

We are seeing unexpected behavior when we use IDurableOrchestrationClient.StartNewAsync to start a disabled Function:

Expected behavior: According to the documentation linked above, this should result in an ArgumentException.

Actual behavior: Method executes without error.  After a certain period of time, an exception is thrown by runtime.

We first observed the behavior when the Function was hosted in Azure, but we can reproduce it on our local machines as well.

The Function is using a service bus trigger with auto-complete disabled.

We expect this scenario to throw an error, so that we can dead letter the service bus message.  Since the code is executing successfully, and the status is returned as Pending, we are completing the service bus message.  This means we are unable to re-process that message after the exception is thrown, that data is gone, and we are unable to properly act on that message.

Code starting/checking Orchestrator Function:

`instanceId = await orchestrationClient.StartNewAsync(contextFunctionName, msgBody); // Does not throw exception as expected
DurableOrchestrationStatus status = await orchestrationClient.GetStatusAsync(instanceId);

if (
  status. RuntimeStatus == OrchestrationRuntimeStatus.Canceled 
  || status. RuntimeStatus == OrchestrationRuntimeStatus.Failed 
  || status. RuntimeStatus == OrchestrationRuntimeStatus.Terminated
{
  // Dead letter service bus message
}  else {
  // Complete service bus message
}
`

YunchuWang:
Fixes #1933

## Problem

A function disabled at host startup is still indexed by name, but its `RegisteredFunctionInfo` remains `null` because no listener is created. Starts could therefore appear successful and remain pending indefinitely.

## Change

- Reject new and restarted current-app instances only when the orchestrator is known to be startup-disabled.
- Apply the disabled-function check to direct clients, the HTTP management API, and the gRPC sidecar before provider scheduling.
- Return HTTP 400 and gRPC `InvalidArgument` while retaining `ArgumentException` compatibility for direct clients.
- Keep active and deregistered listener-drain behavior and deterministic replay unchanged.

## Compatibility intentionally preserved

Disabled validation deliberately reuses the existing `ClientReferencesCurrentApp` predicate used by direct missing-function validation. This PR does not change current-app classification:

- `ExternalClient=true` bypasses local validation.
- TaskHub comparison remains exact-case.
- Connection comparison retains its existing behavior, including same-hub/different-connection self-comparison.
- Missing orchestrator/entity behavior is unchanged.
- HTTP and gRPC still allow unknown function names.
- No provider-specific TaskHub casing or host-default-provider connection correction is included.

## Tests

Coverage includes direct, HTTP, and gRPC starts and restarts; exact and case-variant TaskHubs; same-hub/different-connection behavior; external clients; missing, disabled, active, and deregistered functions; provider call suppression; replay safety; and an E2E assertion that a rejected disabled start is never persisted.

- Focused current-app gate and disabled-path tests: 27 passed on net8.0; 27 passed on net10.0
- Full relevant unit classes (`DurableClientBaseTests`, `DisabledFunctionDispatchTests`, `HttpApiHandlerTests`, `LocalGrpcListenerTests`): 183 passed on net8.0; 183 passed on net10.0


src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -361,15 +362,21 @@ private bool ClientReferencesCurrentApp(DurableClient client)
...
         {
             var taskHubName = this.durableTaskOptions.HubName;
-            return client.TaskHubName.Equals(taskHubName);
+            return string.Equals(client.TaskHubName, taskHubName, StringComparison.OrdinalIgnoreCase);
         }
 
         private bool ConnectionNameMatchesCurrentApp(DurableClient client)
         {
-            return this.DurabilityProvider.ConnectionNameMatches(client.DurabilityProvider);
+            return this.config != null &&
```
YunchuWang: why here use config defaultprovider

YunchuWang: `config.DefaultDurabilityProvider` is the host/current-app side of this comparison, while `client.DurabilityProvider` is the target selected for this client.

This matters because the new `ReferencesCurrentApp()` calls `ClientReferencesCurrentApp(this)`. If we kept `this.DurabilityProvider.ConnectionNameMatches(client.DurabilityProvider)`, that path would compare the client provider with itself and always report a connection match, even when the client targets a non-default connection. We could then incorrectly validate a remote target against this host's local function registrations.

Using `config.DefaultDurabilityProvider` anchors the comparison to the provider actually owned by the current Functions app. The `config != null` guard also means a standalone/external client, which has no local host metadata, cannot be classified as referencing the current app. The remote hub/connection tests cover these bypass cases.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -361,15 +364,31 @@ private bool ClientReferencesCurrentApp(DurableClient client)
...
+                ? StringComparison.OrdinalIgnoreCase
+                : StringComparison.Ordinal;
+            return string.Equals(client.TaskHubName, taskHubName, comparison);
+        }
+
+        private static bool UsesCaseInsensitiveTaskHubNames(DurabilityProvider provider)
+        {
+            return provider != null &&
+                (string.Equals(provider.Name, AzureStorageProviderName, StringComparison.OrdinalIgnoreCase) ||
+                string.Equals(provider.Name, AzureManagedProviderName, StringComparison.OrdinalIgnoreCase));
```
bachuv: How come we're only checking Azure Storage and DTS? Are MSSQL and Netherite different when it comes to task hub name casing? Is this documented somewhere?

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
```diff
@@ -361,15 +364,31 @@ private bool ClientReferencesCurrentApp(DurableClient client)
...
         }
 
         private bool ConnectionNameMatchesCurrentApp(DurableClient client)
         {
-            return this.DurabilityProvider.ConnectionNameMatches(client.DurabilityProvider);
+            return this.config != null &&
+                this.config.DefaultDurabilityProvider.ConnectionNameMatches(client.DurabilityProvider);
```
bachuv: Why are we updating this from using `this.DurabilityProvider` to `this.config.DefaultDurabilityProvider`?

src/WebJobs.Extensions.DurableTask/DurableTaskExtension.cs
```diff
@@ -1525,6 +1525,18 @@ internal void ThrowIfFunctionDoesNotExist(string name, FunctionType functionType
             }
         }
 
+        internal void ThrowIfOrchestratorFunctionIsDisabled(string name)
+        {
+            // A null registration means the function was indexed but disabled, so its listener was
+            // never created. A missing registration and IsDeregistered are deliberately allowed:
+            // the target may exist in another app, or its listener may be restarting on this host.
```
bachuv, line 1532: How are we taking into account for `A missing registration and IsDeregistered are deliberately allowed:
            // the target may exist in another app, or its listener may be restarting on this host.`? Does `this.knownOrchestrators` already filter those functions out?

YunchuWang: No—`knownOrchestrators` does not filter these states out. Retaining them is what lets this check distinguish the listener lifecycle states:

- During indexing, the binding provider calls [`RegisterOrchestrator(name, null)`](https://github.com/Azure/azure-functions-durable-extension/blob/32a2a3f13aba4245905de486d4b3cf88fb5119c8/src/WebJobs.Extensions.DurableTask/Bindings/OrchestrationTriggerAttributeBindingProvider.cs#L72).
- If a listener is created, it replaces that value with a non-null [`RegisteredFunctionInfo`](https://github.com/Azure/azure-functions-durable-extension/blob/32a2a3f13aba4245905de486d4b3cf88fb5119c8/src/WebJobs.Extensions.DurableTask/Bindings/OrchestrationTriggerAttributeBindingProvider.cs#L225).
- Deregistration sets `IsDeregistered` on the existing info; it does not remove the dictionary entry.

Therefore, `TryGetValue(...) && info == null` is the only definitive startup-disabled state: this host indexed the function but never created its listener. A missing key is inconclusive because an app sharing the hub may own that function. A non-null `IsDeregistered` entry can be transient while a listener stops/restarts, so this helper deliberately does not reject it; [`StartNewAsync_DeregisteredOrchestrator_SchedulesInstance`](https://github.com/Azure/azure-functions-durable-extension/blob/32a2a3f13aba4245905de486d4b3cf88fb5119c8/test/FunctionsV2/Tests/Unit/DurableClientBaseTests.cs#L181-L216) covers that behavior.

For the direct in-app `DurableClient` path, the existing `ThrowIfFunctionDoesNotExist` check still runs before this helper, so its missing-function behavior is unchanged. HTTP/gRPC use this narrower helper so they do not introduce a new rejection for functions that may be owned by another app sharing the hub.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
YunchuWang: This changes the comparison from the target client's provider to the provider used by the current host worker.

`ReferencesCurrentApp()` calls `ClientReferencesCurrentApp(this)`. With the previous expression, `this.DurabilityProvider.ConnectionNameMatches(client.DurabilityProvider)`, both sides therefore refer to the same client provider, making the connection check effectively always true. A client using the same hub name against a different connection could then be misclassified as referencing this app and receive this host's local function-metadata validation.

`config.DefaultDurabilityProvider` is the backend used by this host's TaskHub worker/listeners, whose `knownOrchestrators` metadata is being consulted. `client.DurabilityProvider` is the backend targeted by this client. Comparing host-default versus client-target is what establishes whether the client actually references the current app. The HTTP and gRPC different-connection tests preserve the existing behavior that an external target skips local disabled-function validation.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
YunchuWang: This is provider-specific; the fallback is intentionally exact rather than assuming every backend normalizes task hub names.

- Azure Storage already compares its configured hub with the provider settings using [`OrdinalIgnoreCase`](https://github.com/Azure/azure-functions-durable-extension/blob/32a2a3f13aba4245905de486d4b3cf88fb5119c8/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProviderFactory.cs#L192), and the new [E2E](https://github.com/Azure/azure-functions-durable-extension/blob/32a2a3f13aba4245905de486d4b3cf88fb5119c8/test/FunctionsV2/Tests/E2E/StatusAndQueryTests.cs#L62-L103) exercises a case-variant Azure Storage hub and verifies that the rejected start is not persisted.
- For `azureManaged`, I also verified against a live Durable Task Scheduler that a mixed-case host hub and lower-case client override resolve to the same hub. I did not find a public DTS casing statement, so this part of the allowlist is based on that service-level A/B verification.
- MSSQL explicitly documents that task hub casing [depends on the database collation](https://github.com/microsoft/durabletask-mssql/blob/8d46f5fbe5ca38501a804640af41453c992d9a8a/docs/taskhubs.md#case-sensitivity). Its preferred collation is `Latin1_General_100_BIN2_UTF8`, and the provider [uses that binary collation when creating a database](https://github.com/microsoft/durabletask-mssql/blob/8d46f5fbe5ca38501a804640af41453c992d9a8a/src/DurableTask.SqlServer/SqlDbManager.cs#L204), so case-only names can be distinct on a supported configuration.
- Netherite retains the original task hub name and performs an [exact `TaskhubName != HubName` check](https://github.com/microsoft/durabletask-netherite/blob/d6e58072e40be6081729148e18a89eda246025ce/src/DurableTask.Netherite/TransportLayer/EventHubs/EventHubsTransport.cs#L93-L99).
- Custom providers have no common casing contract, so exact matching is the conservative fallback.

This is not claiming every MSSQL database is case-sensitive; it means we cannot safely make this comparison case-insensitive for all MSSQL, Netherite, or custom-provider configurations.

src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
bachuv: Is this a breaking change though? Just want to confirm there is not behavior change.

YunchuWang: I verified this more deeply. There **is an observable behavior change**, so describing this as “no behavior change” would be inaccurate. It changes an incorrectly rejected cross-app call into a successful dispatch; I did not find a previously successful valid path that becomes unsuccessful.

The old code is a self-comparison for `StartNewAsync`: `ReferencesCurrentApp()` calls `ClientReferencesCurrentApp(this)`, so

```csharp
this.DurabilityProvider.ConnectionNameMatches(client.DurabilityProvider)
```

compares the target client's provider to itself. With the same task-hub name, a client bound to App B's connection is therefore misclassified as App A, and App A's `knownOrchestrators` is consulted before anything is sent to B.

I reproduced this with two real `ITestHost`/`JobHost` instances and two independent Azurite backends. Both apps used the same task-hub name; App A invoked a genuine `[DurableClient(TaskHub = "%CrossAppTaskHub%", ConnectionName = "TargetStorage")]` binding through `JobHost.CallAsync`; the target orchestrator was absent from A and active in B.

- Base `53a64841`: App A threw `FunctionInvocationException` -> `ArgumentException` from its local missing-orchestrator check; App B had no instance record.
- PR head: no local exception; App A's backend remained untouched; App B persisted and completed the instance (`REMOTE_COMPLETED`).
- Controls: same-app active calls completed on both revisions; same-app missing calls remained rejected; disabled-in-A/active-in-B cross-app calls completed in B; no tested successful valid path regressed.
- The same shared predicate also fixes custom-bound entity signaling: base rejected it against A's local registry, while head delivered it to B.

This comparison is also required for the disabled-start change itself. Without host-default-vs-client-target comparison, a function disabled in A but enabled in B would be newly rejected by A when using a same-hub/different-connection client.

I added focused regression coverage in `990df360`:
- [`StartNewAsync_DifferentConnection_DoesNotValidateLocalOrchestrator`](https://github.com/Azure/azure-functions-durable-extension/blob/990df360bc1efb15188e5f581677af7302927adb/test/FunctionsV2/Tests/Unit/DurableClientBaseTests.cs#L83-L95)
- [`SignalEntityAsync_DifferentConnection_DoesNotValidateLocalEntity`](https://github.com/Azure/azure-functions-durable-extension/blob/990df360bc1efb15188e5f581677af7302927adb/test/FunctionsV2/Tests/Unit/DurableClientBaseTests.cs#L255-L270)
- [The setup explicitly uses the host hub with different default/client connections](https://github.com/Azure/azure-functions-durable-extension/blob/990df360bc1efb15188e5f581677af7302927adb/test/FunctionsV2/Tests/Unit/DurableClientBaseTests.cs#L521-L551).

With only the old self-comparison restored, both tests fail on net8.0 and net10.0 with the expected local missing-orchestrator/entity exceptions. On the PR head, both pass 2/2 on each TFM.

So I would classify this as a compatibility bug fix with intentional observable behavior, not a binary/public-API break and not a zero-behavior-change refactor.

bachuv: Since this is a behavior change, can we keep the existing behavior and just add this check separately? I would prefer to not change the behavior as part of this PR. I think the bug fix should be in a separate, isolated PR.

YunchuWang: sounds good, this pr updated to fix disabled orch start only

bachuv (approved):

PR #3517, opened 2026-08-14
Prevent Unsafe Type Materialization when Deserializing Entity Response Exceptions

nytian:
Validates that entity response exception types derive from Exception before deserialization and disables Json.NET type metadata handling. This prevents attacker-controlled types from being instantiated while preserving custom exception propagation. Adds regression tests for root and nested type metadata.


src/WebJobs.Extensions.DurableTask/EntityScheduler/ResponseMessage.cs
```diff
@@ -62,8 +62,17 @@ public T GetResult<T>(MessagePayloadDataConverter messageDataConverter, MessageP
                 // do a best-effort attempt at deserializing this exception
                 try
                 {
-                    var type = Type.GetType(this.ExceptionType, true);
-                    e = (Exception)errorDataConverter.Deserialize(this.Result, type);
+                    Type type = Type.GetType(this.ExceptionType, true);
+                    if (typeof(Exception).IsAssignableFrom(type))
+                    {
+                        JsonSerializer serializer = JsonSerializer.Create(errorDataConverter.JsonSettings);
+                        serializer.TypeNameHandling = TypeNameHandling.None;
```
bachuv, line 69: Should we document that we are deserializing using `TypeNameHandling.None` and serializing with `TypeNameHandling.Objects`?

test/FunctionsV2/Tests/Unit/ResponseMessageTests.cs
bachuv, line 1: Do we already have a happy path test somewhere? Seems like the three new tests cover blocked types, blocked $type metadata, and DateParseHandling.

src/WebJobs.Extensions.DurableTask/EntityScheduler/ResponseMessage.cs
```diff
@@ -62,8 +62,17 @@ public T GetResult<T>(MessagePayloadDataConverter messageDataConverter, MessageP
...
+                        using (var stringReader = new System.IO.StringReader(this.Result))
+                        using (var jsonReader = new JsonTextReader(stringReader) { DateParseHandling = serializer.DateParseHandling })
+                        {
+                            e = (Exception)JToken.Load(jsonReader).ToObject(type, serializer);
```
bachuv, line 73: Can we use `serializer.Deserialize(jsonReader, type)` here instead of `JToken.Load().ToObject()`?

test/FunctionsV2/Tests/Unit/ResponseMessageTests.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        public class DeserializationCallback
+        {
+            public DeserializationCallback()
+            {
+                WasCreated = true;
+            }
+
+            public static bool WasCreated { get; set; }
```
bachuv, line 94: `WasCreated` is a shared static field so if xUnit parallelism is ever enabled for this class (e.g., via `[assembly: CollectionBehavior(MaxParallelThreads = N)]`), concurrent tests will silently corrupt this assertion. Consider using an instance counter with `Interlocked.Increment` instead, or at minimum add a comment noting that these tests must run single-threaded.

nytian:
Close as not needed anymore

PR #3527, opened 2026-08-20
Fix Durable client current-app detection

### LINKED ISSUE #3523: Fix Durable client current-app detection across connections and TaskHub casing
## Problem

`DurableClient.ClientReferencesCurrentApp` controls whether local function metadata validation is applied to a durable client operation. Its current backend classification has two problems:

1. `ConnectionNameMatchesCurrentApp` compares the target client provider with `this.DurabilityProvider`. Existing callers pass `this` into `ClientReferencesCurrentApp`, so this becomes a self-comparison. A client using the same TaskHub name but a different connection is therefore classified as targeting the current app.
2. `TaskHubMatchesCurrentApp` always uses exact-case string equality, even though TaskHub identity/casing semantics can differ by durability provider.

As a result, local metadata can incorrectly block supported cross-App operations, while a case-only TaskHub difference can incorrectly bypass local validation for providers that consider the names equivalent.

PR #3514 deliberately preserves this legacy predicate so its disabled-orchestrator fix remains isolated. This issue tracks correcting the predicate separately as an explicit compatibility change.

## Current behavior

| Scenario | Current result |
|---|---|
| Same exact TaskHub, different connection, missing orchestrator | Local `ArgumentException`; target provider is not called |
| Same exact TaskHub, different connection, missing entity | Local `ArgumentException`; message is not sent |
| Case-only TaskHub difference | Treated as another app for every provider |
| `ExternalClient = true` | Local metadata validation is skipped |

After #3514, startup-disabled validation also consumes this same predicate, so any correction must keep missing-function, entity, and disabled-function behavior consistent.

## Expected behavior

- Determine whether the target client points to the host's current backend by comparing the host default durability provider with the target client's durability provider, rather than comparing the target provider with itself.
- Apply TaskHub identity/casing semantics appropriate to the selected durability provider. Explicitly define and test Azure Storage and Azure Managed behavior, while preserving exact-case behavior for providers where case is significant.
- Skip local missing/disabled metadata validation for a genuinely remote connection so the target provider receives the operation.
- Apply local validation when the client points to the same logical backend.
- Preserve the explicit `ExternalClient = true` opt-out behavior unless a separate API decision is made.
- Keep orchestration, entity, disabled-orchestrator, HTTP, and gRPC consumers of `ClientReferencesCurrentApp` consistent.

## Acceptance criteria

- Regression tests cover missing orchestrator start, restart where applicable, and missing entity signal for same-hub/different-connection clients.
- Tests cover case-variant TaskHub names for both case-insensitive and exact-case provider semantics.
- Tests cover startup-disabled orchestrators now that #3514 reuses this predicate.
- Tests verify whether provider calls/messages occur, not only exception types.
- The behavior change is documented separately from #3514.

## Related history

- #1281 — cross-Function-App orchestration starts
- #1248 — signaling an entity in another Function App
- #1306 — original cross-App `ThrowIfFunctionDoesNotExist` correction
- #3514 — disabled-orchestrator rejection, intentionally retaining legacy current-app classification

YunchuWang:
## Summary

Fixes #3523.

This compatibility correction builds on merged PR #3514 (`477a92915cbeb46423fe55ceb36c036b845b00b0`). It keeps the single shared `DurableClient.ClientReferencesCurrentApp` gate introduced there, but compares a target client with the host's actual default durability provider instead of comparing the target provider with itself.

## Before / after

| Scenario | Before | After |
|---|---|---|
| Same exact TaskHub, different connection, missing orchestrator | Rejected by local metadata; target provider not called | Local validation skipped; target provider receives the start |
| Same exact TaskHub, different connection, missing entity | Rejected by local metadata; message not sent | Local validation skipped; entity signal is sent |
| Same exact TaskHub, different connection, locally startup-disabled orchestrator start/restart | Rejected locally | Local validation skipped; target backend receives start/restart |
| Azure Storage case-only TaskHub difference | Classified as remote | Classified as current app because Azure Storage hub identity is case-insensitive |
| Azure Managed / exact-case provider case-only TaskHub difference | Classified as remote | Remains classified as remote |
| `ExternalClient = true` | Bypasses local validation | Unchanged |
| Unknown registry key / disabled null entry / active or deregistered listener | #3514 behavior | Unchanged; only current-app classification is corrected |

## Design rationale

- `ClientReferencesCurrentApp` remains the only predicate used by missing-orchestrator, missing-entity, and startup-disabled-orchestrator validation.
- Connection identity is evaluated by virtual dispatch on `DurableTaskExtension.DefaultDurabilityProvider`, passing the target provider. This preserves provider-specific connection matching direction.
- TaskHub identity is an internal durability-provider semantic. The base implementation uses ordinal matching (including Azure Managed's shipped exact-case semantics), while Azure Storage overrides it with ordinal-ignore-case matching to reflect its lowercased storage resources.
- No provider-name string switches, public API additions, entity dispatch changes, or HTTP/gRPC error-mapping changes are introduced.

## Compatibility

The intentional behavior change is limited to local metadata validation at corrected current/remote boundaries. Cross-connection operations now reach their configured backend; Azure Storage case variants now receive current-app validation. Exact-case providers, the explicit external-client bypass, replay, and deregistered listener drain behavior remain unchanged.

## Tests

TDD RED evidence on `net8.0` before production changes:
- `DurableClientBaseTests`: 8 expected failures covering cross-connection missing/disabled orchestrator start, restart, entity signal, and Azure Storage case-only matching.
- HTTP cross-connection disabled start: expected `Accepted`, observed `BadRequest`.
- gRPC cross-connection disabled start: observed the expected pre-fix `InvalidArgument` local rejection.

GREEN results:
- Focused `DurableClientBaseTests`: 50/50 passed on `net8.0`; 50/50 passed on `net10.0`.
- Focused HTTP current/remote/missing/disabled tests: 4/4 passed on each TFM.
- Focused gRPC current/remote/unknown/disabled/restart tests: 5/5 passed on each TFM.
- Complete relevant classes (`DurableClientBaseTests`, `DisabledFunctionDispatchTests`, `HttpApiHandlerTests`, `LocalGrpcListenerTests`): 187/187 passed on `net8.0`; 187/187 passed on `net10.0`.
- `git -c core.whitespace=cr-at-eol diff --check`: passed.

A principal-level review found no critical or important findings.

### DIFF
--- a/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/AzureStorageDurabilityProvider.cs
@@ -73,6 +73,11 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         public override string EventSourceName { get; set; } = "DurableTask-AzureStorage";
 
+        internal override bool TaskHubNameMatches(string taskHubName, string otherTaskHubName)
+        {
+            return string.Equals(taskHubName, otherTaskHubName, StringComparison.OrdinalIgnoreCase);
+        }
+
         /// <inheritdoc/>
         public async override Task<IList<OrchestrationState>> GetAllOrchestrationStates(CancellationToken cancellationToken)
         {
--- a/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
+++ b/src/WebJobs.Extensions.DurableTask/ContextImplementations/DurableClient.cs
@@ -358,7 +358,8 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
 
         private bool ClientReferencesCurrentApp(DurableClient client)
         {
-            return !client.attribute.ExternalClient &&
+            return this.config != null &&
+                !client.attribute.ExternalClient &&
                 this.TaskHubMatchesCurrentApp(client) &&
                 this.ConnectionNameMatchesCurrentApp(client);
         }
@@ -374,12 +375,12 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         private bool TaskHubMatchesCurrentApp(DurableClient client)
         {
             var taskHubName = this.durableTaskOptions.HubName;
-            return client.TaskHubName.Equals(taskHubName);
+            return this.config.DefaultDurabilityProvider.TaskHubNameMatches(client.TaskHubName, taskHubName);
         }
 
         private bool ConnectionNameMatchesCurrentApp(DurableClient client)
         {
-            return this.DurabilityProvider.ConnectionNameMatches(client.DurabilityProvider);
+            return this.config.DefaultDurabilityProvider.ConnectionNameMatches(client.DurabilityProvider);
         }
 
         /// <inheritdoc />
--- a/src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
+++ b/src/WebJobs.Extensions.DurableTask/DurabilityProvider.cs
@@ -617,6 +617,17 @@ namespace Microsoft.Azure.WebJobs.Extensions.DurableTask
         }
 
         /// <summary>
+        /// Returns true if two task hub names identify the same task hub for this durability provider.
+        /// </summary>
+        /// <param name="taskHubName">The first task hub name.</param>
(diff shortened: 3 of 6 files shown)


YunchuWang:
Superseded by #3528. The replacement PR uses a branch in Azure/azure-functions-durable-extension so secret-dependent CI workflows can run.

